PR #81, opened 2020-01-19
More Clippy warnings

jsulmont:


src/parser/ast.rs
```diff
@@ -95,6 +95,7 @@ pub struct VariableDeclaration {
 }
 
 #[derive(Debug, PartialEq)]
+#[allow(clippy::vec_box)]
```
seanyoung: Actually I think clippy is right here. This should be fixed (and I think I already fixed this)

seanyoung:
Thanks for the PR, looks great!

Just one small comment.

Hyperledger needs a DCO, please add a Signed-off-by: to your commit. Thanks again!

seanyoung:
Looks great, thanks. Just needs rebasing.

jsulmont:
@seanyoung not sure where you need this rebased

PR #110, opened 2020-02-15
Remove burrow target now that burrow uses ewasm

seanyoung:


seanyoung:
See: https://github.com/hyperledger/burrow/pull/1338

PR #139, opened 2020-03-25
fix #5

### LINKED ISSUE #5: File not found show ugly panic
If the solidity source is not found, the error is very ugly:

```
$ solang foo
thread 'main' panicked at 'file not found: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

knarfytrebil:
Fixes the ugly panic message and gracefully quit when file does not exist.

### DIFF
--- a/src/bin/solang.rs
+++ b/src/bin/solang.rs
@@ -130,7 +130,17 @@ fn process_filename(
 
     let mut json_contracts = HashMap::new();
 
-    let mut f = File::open(&filename).expect("file not found");
+    let mut f = match File::open(&filename) {
+        Err(err_info) => {
+            eprintln!(
+                "error: cannot open {:?}: {}",
+                &filename,
+                err_info.to_string()
+            );
+            std::process::exit(1);
+        }
+        Ok(file) => file,
+    };
 
     let mut contents = String::new();
     f.read_to_string(&mut contents)


src/bin/solang.rs
```diff
@@ -130,7 +130,13 @@ fn process_filename(
 
     let mut json_contracts = HashMap::new();
 
-    let mut f = File::open(&filename).expect("file not found");
+    let mut f = match File::open(&filename) {
+        Err(err_info) => {
+            eprintln!("{}, {:?}", err_info.to_string(), &filename);
```
seanyoung: The error should look similar to other errors and warnings. So, it should be something like: 
error: cannot open ‘foo.sol’: No such file or directory

Also your commit message should say what your are changing rather than just referencing an issue. Something like "improve error handling on file system errors"

Per hyperledger dco rules you need to have your real name in the commit message. You also have two signed-off-by lines in the commit message.

Looks very good thank you very much!

knarfytrebil: Hi Sean, I have done the following as requested.
- Changed output message
- Changed commit message
- Signed with my real name
Have a nice day, cheers !

seanyoung:
That looks perfect thank you very much!

PR #148, opened 2020-04-02
Substrate function selectors should be string of bytes, not u32

seanyoung:


Stefie:
Amazing, that was quick!

PR #204, opened 2020-05-31
Substrate contracts tester

seanyoung:


seanyoung:
Changes are in https://github.com/paritytech/pallet-contracts-waterfall/ instead.

PR #241, opened 2020-08-01
Implement dynamic array push and pop

### LINKED ISSUE #177: memory arrays should have push() and pop()
Dynamic memory arrays currently cannot be resized. This is true for Solang and solc solidity. However, Solang uses a heap, and has a realloc() function. Thus it is is possible for Solang to support pop() and push() on memory arrays.

Note that push and pop already supported on contract storage arrays. Also, arrays are stored in struct vectors in Solang, which are allocated with malloc.

Steps:
- In [method call](https://github.com/hyperledger-labs/solang/blob/master/src/resolver/expression.rs#L4124) add some code to detect "push" and "pop" call on memory arrays. This should result in a new Expression::DynamicArrayPush() and Expression::DynamicArrayPop().
Note that push() with no arguments should return a reference to the new element.
- In the [code emitter](https://github.com/hyperledger-labs/solang/blob/master/src/emit/mod.rs#L1473), some code is needed to emit code for Expression::DynamicArrayPush() and  Expression::DynamicArrayPop(). Both of these will new to call realloc() to resize the array, and of course pop needs to return the last element and push the new element.
- The call return value of realloc() might be a new pointer if the memory was be moved. The array variable will need to be updated.
- An alternative implementation would generate a Expression::DynamicArrayResize() for the code emitter.
- Arrays can be the `bytes` type, or every type followed by `[]`, so `int64[]`,  `address[]`, `struct-foo[]`, `enum-bar[]`. All types will need test cases like the one below.

```
contract c {
        function test() public {
                bytes foo = new bytes(1);

                foo[0] = 128;
                foo.push(64);

                assert(foo.length == 2);

                assert(foo[1] == 64);

                assert(64 == foo.pop());

                assert(foo.length == 1);
        }
}
```

KiChjang:
Fixes #177.

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -31,6 +31,11 @@ pub enum Instr {
         array: usize,
         value: Box<Expression>,
     },
+    PopMemory {
+        res: usize,
+        ty: Type,
+        array: usize,
+    },
     Set {
         res: usize,
         expr: Expression,
@@ -621,6 +626,12 @@ impl ControlFlowGraph {
                 self.vars[*array].id.name,
                 ty.to_string(ns),
                 self.expr_to_string(contract, ns, value),
+            ),
+            Instr::PopMemory { res, ty, array } => format!(
+                "%{}, %{} = pop array ty:{}",
+                self.vars[*res].id.name,
+                self.vars[*array].id.name,
+                ty.to_string(ns),
             ),
             Instr::AssertFailure { expr: None } => "assert-failure".to_string(),
             Instr::AssertFailure { expr: Some(expr) } => {
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -25,6 +25,12 @@ pub enum Instr {
         storage: Box<Expression>,
         offset: Box<Expression>,
     },
+    PushMemory {
+        res: usize,
+        ty: Type,
+        array: usize,
+        value: Box<Expression>,
+    },
     Set {
         res: usize,
         expr: Expression,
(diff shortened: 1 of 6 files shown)


src/codegen/cfg.rs
```diff
@@ -25,6 +25,17 @@ pub enum Instr {
         storage: Box<Expression>,
         offset: Box<Expression>,
     },
+    PushMemory {
+        res: usize,
+        ty: Type,
+        array_addr: usize,
```
seanyoung, line 31: This should be called `array` not `array_addr`.

src/codegen/cfg.rs
```diff
@@ -25,6 +25,17 @@ pub enum Instr {
...
+        value: Box<Expression>,
+    },
+    PopMemory {
+        res: usize,
+        ty: Type,
+        array_addr: usize,
```
seanyoung, line 37: This should be called `array` not `array_addr`.

src/codegen/expression.rs
```diff
@@ -389,6 +389,61 @@ pub fn expression(
             *loc,
             Box::new(expression(expr, cfg, contract_no, ns, vartab)),
         ),
+        Expression::DynamicArrayPush(loc, array, ty, value) => {
+            let elem_ty = match ty {
+                Type::Array(..) => match ty.array_elem() {
```
seanyoung, line 394: Use `array_deref()` and you should not need to do the match.

src/codegen/expression.rs
```diff
@@ -389,6 +389,61 @@ pub fn expression(
...
+                    array_addr,
+                    value: value.clone(),
+                },
+            );
+
+            Expression::Variable(*loc, elem_ty, address_res)
+        }
+        Expression::DynamicArrayPop(loc, array, ty) => {
+            let elem_ty = match ty {
+                Type::Array(..) => match ty.array_elem() {
```
seanyoung, line 422: Use `array_deref()`

src/codegen/expression.rs
KiChjang: Looks like `array_deref()` doesn't work here, because it always wraps a layer of `Ref` before it gets to the actual array element type.

seanyoung:
LGTM

Thank you!

PR #272, opened 2020-09-11
Allow implicit casting between bytesN to bytes and vice versa

### LINKED ISSUE #158: casting from bytesN to bytes and vice versa should be possible
Note that bytesN is represented as llvm int, so it needs byte order swapping.

KiChjang:
Fixes #158.


seanyoung:
Perfect, thank you very much!

PR #287, opened 2020-09-29
Produce warning if tx.gasprice is used on Substrate

seanyoung:
tx.gasprice may round to zero giving the impression gas is free.

### DIFF
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -2560,7 +2560,21 @@ uint128 ``tx.gasprice(uint64 gas)``
     if the gas price is less than 1 it may round down to 0, giving the incorrect appearance gas is free.
     Therefore, avoid the ``tx.gasprice`` member in favour of the function ``tx.gasprice(uint64 gas)``.
 
-    This function is not available on the Ethereum Foundation Solidity compiler.
+    To avoid rounding errors, pass the total amount of gas into ``tx.gasprice(uint64 gas)`` rather than
+    doing arithmetic on the result. As an example, **replace** this bad example:
+
+    .. code-block::
+
+        // BAD example
+        uint128 cost = num_items * tx.gasprice(gas_per_item);
+
+    with:
+
+    .. code-block::
+
+        uint128 cost = tx.gasprice(num_items * gas_per_item);
+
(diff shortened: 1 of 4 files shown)


seanyoung:
@athei what do you think?

docs/language.rst
```diff
@@ -2546,17 +2546,21 @@ address ``msg.sender``
...
-    On Parity Substrate, the cost of one gas unit may not be an exact whole round value. So, to
-    avoid rounding errors, use the ``tx.gasprice()`` function to get the correct pricing
-    for a total units of gas. This function is not available on the Ethereum Foundation Solidity
-    compiler.
+.. warning::
+    On Parity Substrate, the cost of one gas unit may not be an exact whole round value. In fact,
+    if the gas price is less than 1 it may round down to 0, giving the incorrect appearance gas is free.
+    Therefore, avoid the ``tx.gasprice`` member in favour of the function ``tx.gasprice(uint64 gas)``.
+
+    This function is not available on the Ethereum Foundation Solidity compiler.
```
athei, line 2563: I think an example goes a long way here. You probably need to modify/correct my suggestion before comitting because I messed up the language and syntax of that comment:

```suggestion
.. warning::
    On Parity Substrate, the cost of one gas unit may not be an exact whole round value. In fact,
    if the gas price is less than 1 it may round down to 0, giving the incorrect appearance gas is free.
    Therefore, avoid the ``tx.gasprice`` member in favour of the function ``tx.gasprice(uint64 gas)``.
    
    One should always calculate the complete gas amount and pass that into ``tx.gasprice(uint64 gas)`` rather
    than using the result of the function in a larger function where possible.
    
    **Don't** do
    ```
    uint128 cost = num_items * tx.gasprice(gas_per_item);
    ```

    Rather:
    ```
    uint128 cost = tx.gasprice(num_items * gas_per_item);
    ```

    This function is not available on the Ethereum Foundation Solidity compiler.
```

athei (approved):
LGTM

docs/language.rst
seanyoung: That example is very useful. Thanks, I've added it.

PR #296, opened 2020-10-22
Add experimental solana target support

seanyoung:


jackcmay:
Looks like llvm10 is making now making use of one of the `TargetInfo` members.  Try applying the following changes:
```
diff --git a/lld/ELF/Arch/BPF.cpp b/lld/ELF/Arch/BPF.cpp
index f70cc4e8b15..e12baaf9da5 100644
--- a/lld/ELF/Arch/BPF.cpp
+++ b/lld/ELF/Arch/BPF.cpp
@@ -25,6 +25,7 @@ namespace {
 class BPF final : public TargetInfo {
 public:
   BPF();
+  RelType getDynRel(RelType type) const override;
   RelExpr getRelExpr(RelType Type, const Symbol &S,
                      const uint8_t *Loc) const override;
   void relocateOne(uint8_t *Loc, RelType Type, uint64_t Val) const override;
@@ -36,6 +37,13 @@ BPF::BPF() {
   relativeRel = R_BPF_64_RELATIVE;
 }
 
+RelType BPF::getDynRel(RelType type) const {
+  if (type == R_BPF_64_32) {
+    return type;
+  }
+  return R_ARM_NONE;
+}
+
 RelExpr BPF::getRelExpr(RelType Type, const Symbol &S,
                         const uint8_t *Loc) const {
   switch (Type) {
```

Issue #319, opened 2020-11-13
Conflicting use of 'view' in Solang as compared to Remix IDE


JeremyConnor:
This is my code AssetTransfer.sol, with a little modification. I have added two new events which I am using in the ```viewBalance``` and ```findOwner``` functions. When I deploy the contract on Remix IDE, I am prompted to remove the ```view``` from these function types as I am not just viewing the data but also emitting it. But, when I go on to create WASM and ABI file in Solang, it doesn't work and it hints me to add view to the functions. These both are conflicting. I think this might be the reason I am not able to deploy my modified contract on Polkadot UI.
```
pragma solidity ^0.6.0;

contract AssetTransfer {
    
    /*
    *    SendTo is a structure which will store the basic information for every receiver.
    *    SendTo.sender is the person sending the Asset.
    *    SendTo.hexOfAsset is the hash of the Asset being sent.
    *    SendTo.paid will initially be false, and will be updated to true when the reciver has paid for the asset.
    */
    struct SendTo{
        address sender;
        string hexOfAsset;
        bool paid;
    }
    
    //    registered is a map of every Registered user.
    mapping(address => uint) registered;
    
    /*
    *    balance is a map of every registered user.
    *    By default, the balance would be 1000.
    */
    mapping(address => uint) balance;
    
    //    assetToOwner maps hash of every Asset to it's current owner.
    mapping(string => address) assetToOwner;
    
    /*   send maps the receiver's address to the SendTo structure.
    *    where SendTo is the information for every corresponding receiver.
    */
    mapping(address => SendTo) send;
    
    event successfulRegistration(address _newUser);
    event assetSent(address sender, address receiver, string hexOfAsset);
    event addedAsset(address sender, string _hexAsset);
    event viewingBalance(address sender, uint balance);
    event foundOwner(address owner, string _hashOfAsset);
    event assetReceived(address owner, string _hashOfAsset);
    
    /*    
    *    onlyRegistered is a modifier which checks that the person who 
    *    has called a particular function must be a registered user.
    */
    modifier onlyRegistered() {
        require(registered[msg.sender] > 0, "You are not a registered user!");
        _;
    }
    
    //    viewBalance is used to view balance of the person who has called the function.
    function viewBalance(address _user) public returns(uint) {
        emit viewingBalance(_user, balance[_user]);
        return balance[_user];
    }
    
    /*
    *    Every user must be registered before participating in a transaction.
    *    For this purpose, a person must make a function call to registre function.
[...]

PR #321, opened 2020-11-15
Fix event generation in substrate

seanyoung:


seanyoung:
See #319 
See https://github.com/polkadot-js/api/issues/2850

PR #358, opened 2021-01-05
Use blake2 to generate selector for substrate

tylerztl:
Adaptive substrate ink! contract metadate

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ serde_json = "1.0"
 serde = "1.0"
 serde_derive = { version = "1.0" }
 inkwell = { version = "0.1.0-llvm10sample", features = ["target-webassembly", "target-bpf", "llvm10-0"] }
+blake2 = "0.9"
 blake2-rfc = "0.2.18"
 phf = { version = "0.8", features = ["macros"] }
 unicode-xid = "0.2.0"
--- a/src/sema/ast.rs
+++ b/src/sema/ast.rs
@@ -133,6 +133,7 @@ pub struct Function {
     pub symtable: Symtable,
     // What events are emitted by the body of this function
     pub emits_events: Vec<usize>,
+    pub target: Target,
 }
 
 impl Function {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -29,6 +29,7 @@ serde_json = "1.0"
 serde = "1.0"
 serde_derive = { version = "1.0" }
 inkwell = { version = "0.1.0-llvm10sample", features = ["target-webassembly", "target-bpf", "llvm10-0"] }
+blake2 = "0.9"
```
seanyoung, line 32: Is there a reason for using blake2 crate rather than blake2-rfc?

src/sema/ast.rs
```diff
@@ -133,6 +133,7 @@ pub struct Function {
     pub symtable: Symtable,
     // What events are emitted by the body of this function
     pub emits_events: Vec<usize>,
+    pub target: Target,
```
seanyoung, line 136: Presumably you're copying in the compilation target into the Function struct since you don't have it available at the various selector() call sites.

It might be better to calculate the selector in Function::new() and then keep the selector as a member. This prevents the selector being recomputed every time (for metadata generation and code generation for example), and also there is no need to keep a copy of the target around.

src/sema/ast.rs
```diff
@@ -187,6 +197,17 @@ impl Function {
         u32::from_le_bytes([res[0], res[1], res[2], res[3]])
     }
 
+    /// Generate selector with this function name for substrate
+    pub fn selector_blake2(&self) -> u32 {
+        let name= if self.name.is_empty() {
```
seanyoung, line 202: `cargo fmt ` does not like this.

There can also be multiple constructors and they must have different selectors.
Solidity permits functions to be overloaded, so functions with the same name must have different selectors.

This is presumably why the tests are failing.

seanyoung (changes requested):
The code change looks good in general, just some minor points. Thank you!

src/sema/ast.rs
tylerztl: Yes，i agree with you.

src/sema/ast.rs
tylerztl: Hi, @seanyoung , is there any way to adapt the selector of substrate ink! ? Because if we want to use the ink! contract on the substrate to call solang contract, the selector must be consistent.

seanyoung:
Using blake2 for substrate to generate selectors might be a good idea, so that the function selectors match what Ink! expects. However, this will not work in practise since the functions may overloaded. So, the selector is based on a hash of the function name and signature.

One idea would be to generate a different function selector based on whether a function is overloaded or not. However, this means that if a function was not overloaded, adding and overloaded function with the same name suddenly changes the function selector. I don't think this is a good idea.

Nevertheless, thank you for trying to fix this -- it would have been great if there was a good solution for this.

PR #359, opened 2021-01-07
Add constant folding optimization pass after codegen

### LINKED ISSUE #289: You are using the wrong sha3 function
Should be keccak256

seanyoung:
This implements reaching definitions and constant folding.

The folded constants are visible in the vs code extension hover. Here all code paths result in x being 50.
image

PR also  fixes #289


Hyperion101010:
Looks good, it works on variables as contants and also on enums!

Hyperion101010 (commented):
It works with contants and enums. A good addition to the language server and the compiler overall :+1:

PR #360, opened 2021-01-09
Math overflow checking

### LINKED ISSUE #282: Implement math overflow checking
This should be an optional feature which can be enabled on the command line, with something like `solang --math-overflow`. 

The following code:

```
contract overflow {
        function foo(uint64 x) internal {
                uint64 y = x - 1;
        }

        function bar() public {
                foo(0);
        }
}
```

Calling `bar()` should fail with an assertion failure if `--math-overflow` has been specified. Overflow checking should be done for:

- Addition (both signed and unsigned)
- Subtraction (both signed and unsigned)
- Division/Multiplication/Modulo (signed and unsigned, up-to 64 bit types)
- Power does not need an implementation yet

Tests are required for all types.

This issue is purely for llvm overflow checking, see http://llvm.org/docs/LangRef.html#arithmetic-with-overflow-intrinsics

Multiplication, division, modulo larger than 64 bit and power have their own implementations which can be implemented in a separate issue.

KiChjang:
Fixes #282.

### DIFF
--- a/src/parser/solidity.rs
+++ b/src/parser/solidity.rs
@@ -1,5 +1,5 @@
-// auto-generated: "lalrpop 0.19.1"
-// sha256: c1a758427bd0bd88a580a61f4a19cb7d57b966a8fb335e8e9b2170878113d21d
+// auto-generated: "lalrpop 0.19.4"
+// sha3: 527dcc8958c490bbe826ae4acc91ed3eb8d7a5ce1f8bc36b9eb237f9988bb2dc
 use std::str::FromStr;
 use num_bigint::BigInt;
 use num_bigint::BigUint;
@@ -14,6 +14,8 @@ use super::doc::tags;
 extern crate lalrpop_util as __lalrpop_util;
 #[allow(unused_imports)]
 use self::__lalrpop_util::state_machine as __state_machine;
+extern crate core;
+extern crate alloc;
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 mod __parse__SourceUnit {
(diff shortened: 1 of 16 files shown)


src/parser/solidity.rs
```diff
@@ -1,4 +1,4 @@
-// auto-generated: "lalrpop 0.19.1"
+// auto-generated: "lalrpop 0.19.0"
```
seanyoung, line 1: Your parser was generated using an older version of lalrpop. Please remove this change (you can remove `Cargo.lock` and rebuild).

src/sema/contracts.rs
```diff
@@ -15,11 +15,18 @@ use crate::emit;
 
 impl ast::Contract {
     /// Create a new contract, abstract contract, interface or library
-    pub fn new(name: &str, ty: pt::ContractTy, tags: Vec<ast::Tag>, loc: pt::Loc) -> Self {
+    pub fn new(
+        name: &str,
+        ty: pt::ContractTy,
+        tags: Vec<ast::Tag>,
+        loc: pt::Loc,
+        math_overflow_check: bool,
```
seanyoung, line 23: I don't think this new parameter is necessary. The optimization is also not specified, which is also a codegen issue (and not sema).

seanyoung (changes requested):
Looks great, just two minor details.

Thank you very much!

src/parser/solidity.rs
KiChjang: Haha, I ended up generating another file using a newer version of lalrpop. Is this ok?

seanyoung: of course

seanyoung (approved):

seanyoung:
Very nice, thank you very much!

PR #367, opened 2021-01-18
Expand constant folding unit tests

mmagician:
Note: fail tests & pass tests must be in separate `.sol` files, else the code won't compile at all and so you won't be able to test the passing tests.


seanyoung (approved):
Perfect, thank you very much!

PR #384, opened 2021-02-12
Support rational constants

### LINKED ISSUE #284: Support rational constants
In Solidity, rationals are supported as long as they are compile-time constants. e.g. this is supported:
```
uint x = .5 * 8;
```
Also, fractions can be written with scientific notation: `1e-2` or `0.001e2`.

Fractions can be added/subtracted at compiler time, but fractions cannot be added to integers, like:
```
uint x = 1;
uint y = x + 0.1;
```

This change requires:
- Changes to the lexer to support lexing fractions
- Add new type to AST Type::Rational
- Add funtion to evaluate expression of rationals at compile time
- Implicit casting should allow round rationals convert to integers
- Tests for this

janus:
Support rational constants

close #284


Cargo.toml
```diff
@@ -17,7 +17,7 @@ cc = "1.0"
 [dependencies]
 lalrpop-util = "0.19"
 regex = "1"
-num-bigint = "0.4"
+num-bigint = "0.4.0"
```
seanyoung, line 20: Why is this changed?

seanyoung (changes requested):
Code so far looks fine.

Next step would be to make a contract like:

```
contract c {
    function f() returns (int) {
        return 1.4 + 1.6;
    }
```
Compile (and return 3).

There are changes needed in sema and codegen.
sema should check that the code is valid.
So, when sema::expression::expression encounters and add/subtract/multiply/divide/modulo, if the two arguments are rational, that is ok. If one of the arguments is a rational, then the rational value must be a whole number, so can be replaced with Expression::NumberLiteral() in codegen state.
In codegen::expression::expression all the rationals must be gone and replace with Expression::NumberLiteral.
Please let me know if you need any help

janus:
> Code so far looks fine.
> 
> Next step would be to make a contract like:
> 
> ```
> contract c {
>     function f() returns (int) {
>         return 1.4 + 1.6;
>     }
> ```
> 
> Compile (and return 3).
> 
> There are changes needed in sema and codegen.
> sema should check that the code is valid.
> So, when sema::expression::expression encounters and add/subtract/multiply/divide/modulo, if the two arguments are rational, that is ok. If one of the arguments is a rational, then the rational value must be a whole number, so can be replaced with Expression::NumberLiteral() in codegen state.
> In codegen::expression::expression all the rationals must be gone and replace with Expression::NumberLiteral.
> Please let me know if you need any help

I would follow these steps. I would report back soon.

Cargo.toml
janus: I wanted to have something that would match precisely two versions of num_bigint I have. But I would try out returning it to previous value to see if it works well.

janus:
@seanyoung  Please review my code before. I changed `coerce_integer` to `coerce_number`.

seanyoung:
Please make the CI tests pass (fix the clippy lints) and rebase this into one commit. Your commit needs your real name in the `Signed-off-by` line. Then it is done, thank you!

src/sema/format.rs
```diff
@@ -100,6 +100,7 @@ pub fn string_format(
                     arg_ty,
                     Type::Uint(_)
                         | Type::Int(_)
+                        | Type::Rational
```
seanyoung, line 103: `.format()` does not support rationals and this will break. Please remove.

src/sema/format.rs
```diff
@@ -111,7 +112,7 @@ pub fn string_format(
                     diagnostics.push(Diagnostic::error(
                         arg.loc(),
                         String::from(
-                            "argument must be a bool, enum, address, contract, string, or bytes",
+                            "argument must be a bool, enum, address, contract, string, or bytes, int, uint, rational",
```
seanyoung, line 115: `.format()` does not support rationals

src/sema/types.rs
```diff
@@ -761,6 +764,7 @@ impl Type {
             Type::Contract(_) | Type::Address(_) => BigInt::from(ns.address_length),
             Type::Bytes(n) => BigInt::from(*n),
             Type::Uint(n) | Type::Int(n) => BigInt::from(n / 8),
+            Type::Rational => BigInt::from(8), //(64 / 8) ToDO .. figure out the size of Rational
```
seanyoung, line 767: replace with `unreachable!()`

src/sema/types.rs
```diff
@@ -817,6 +821,7 @@ impl Type {
             Type::Bool => 1,
             Type::Int(n) => *n,
             Type::Uint(n) => *n,
+            Type::Rational => 64, //ToDo figure out the size
```
seanyoung, line 824: replace with `unreachable!()`

src/sema/types.rs
```diff
@@ -845,6 +850,17 @@ impl Type {
...
     }
 
+    pub fn is_rational(&self) -> bool {
+        match self {
+            Type::Rational => true,
+            Type::Ref(r) => r.is_rational(),
+            Type::StorageRef(_, r) => r.is_rational(),
+            Type::Int(_) => true,
+            Type::Uint(_) => true,
+            Type::Value => true,
```
seanyoung, line 860: Int/Uint/Value is not rational.

seanyoung (changes requested):
Some minor nits

src/sema/types.rs
janus: I get error message if I remove Int/Uint/Value.

```
test.sol:4:26-28: error: expected ‘uint256’, found rational
Line 4:
	uint x = .5 * 8;
	---------^^
thread 'solana_tests::simple::returns' panicked at 'called `Option::unwrap()` on a `None` value', src/emit/mod.rs:4348:54
```
There may be something wrong with my cast then.

seanyoung:
@janus as you are clearly struggling please see https://github.com/janus/solang/pull/1

janus:
@seanyoung  I have not being paid.  It looks like the sponsor is not aware that the task is closed.

seanyoung:
@janus I was not aware of this. We should raise this in the issue, not the PR.

janus:
> @janus I was not aware of this. We should raise this in the issue, not the PR.

Ok

Issue #396, opened 2021-02-28
Some errors when running llvm-opt with the bitcode generated by Solang


taquangtrung:
Hi,

I tested the following simple contract with Solang to generate LLVM bitcode.

```
pragma solidity ^0.4.0;

contract SimpleStorage {
  uint storedData;

  function set(uint x) public {
    storedData = x;
  }

  function get() public view returns (uint) {
    return storedData;
  }
}
```

When running llvm-opt with the generated bitcode, there was an error like below. 

```
opt -mem2reg SimpleStorage.bc

Global is external, but doesn't have external or weak linkage!
void (i64, i8*, i32*)* @seal_weight_to_fee
/usr/bin/opt: SimpleStorage.bc: error: input module is broken!
```

The IR corresponding to the above error is:

```
declare internal void @seal_weight_to_fee(i64, i8*, i32*)
```

Is it the intended behavior of Solang to not introduce keywords mentioned by llvm-opt?

Btw, thank you for the great tool. So far, it's the best compiler from Solidity to LLVM, AFAIK.

seanyoung:
This is obviously broken, I'll see if I can fix it.

In the mean time, you can use `--emit llvm-ir` rather than `--emit llvm-bc` and simply remove the offending line in the text format.

seanyoung:
Should be fixed now. Note that it takes a few hours for the docker images to update

seanyoung:
Is it ok to close this issue or is there anything else?

taquangtrung:
Yes, it worked perfectly! Thank you very much!

seanyoung:
I'm not sure exactly what you're doing, but just in case: actually I haven't done a particularly good job of selecting the passes that solang runs when compiling a contract. If you find any more that should be added, please let me know or open a PR. (see pass manager in src/emit/mod.rs)

taquangtrung:
Thanks for the kind notice! I will definitely discuss with you when I face any difficulties with Solang

PR #397, opened 2021-02-28
Make sure generated llvm IR is valid

seanyoung:
This ensures that the llvm IR is valid for llc or llvm-opt. Steps to
reproduce:

  solang --emit llvm-ir -v storage.sol
  llc storage.ll


seanyoung:
fixes #396

PR #404, opened 2021-03-23
Make sure to include tests for --no-dead-storage flag on

mmagician:


seanyoung:
The issue with the repolinter test failing is unrelated to your work. That  issue will be fixed in https://github.com/hyperledger-labs/solang/pull/405

Thank you @mmagician

PR #407, opened 2021-04-13
Implement solana cross contract calls without return values

seanyoung:
This modifies the instruction so that the public key for the called account is included.

Note this does not implement return values yet.


chuckczhao (approved):

chuckczhao (approved):

integration/solana/index.ts
```diff
@@ -157,6 +162,10 @@ class Program {
         let abi: AbiItem = JSON.parse(this.abi).find((e: AbiItem) => e.name == name);
 
         const input: string = Web3EthAbi.encodeFunctionCall(abi, params);
+        const data = Buffer.concat([
+            this.contractStorageAccount.publicKey.toBuffer(),
+            Buffer.from(input.substr(2), 'hex')
```
chuckczhao, line 167: seems it is moved from #141. Wonder what 2 means here.

seanyoung: Yes, you are right -- this could be a lot clearer. The problem is that web3-eth-abi returns the encoded data as a hex string starting with "0x". So, this strips the "0x" and then node's buffer can hex-decode it. I'll see how can I make this clearer.

PR #429, opened 2021-07-02
Unused variable detection

LucasSte:
This pull request implements unused variable detection for the Solang compiler.
It generates warnings for unused state variables, global constants and contract storage.
As a bonus, I implemented the detection of events that have never been emitted.

This PR also implements tests for the new code.

This is the first milestone for the Linux Foundation mentorship on the Solang compiler.
For more information, please refer to the [project plan](https://wiki.hyperledger.org/display/INTERN/Project+plan+-+Solang+compiler+passes+2021).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 Cargo.lock
 /target
 **/*.rs.bk
+.DS_Store
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@
 Cargo.lock
 /target
 **/*.rs.bk
-.DS_Store
-.idea
+
--- a/src/sema/mod.rs
+++ b/src/sema/mod.rs
@@ -41,16 +41,18 @@ pub const SOLANA_BUCKET_SIZE: u64 = 251;
 pub const SOLANA_FIRST_OFFSET: u64 = 16;
 pub const SOLANA_SPARSE_ARRAY_SIZE: u64 = 1024;
 
-/// Performs semantic analysis and checks for unused variables
+/// Load a file file from the cache, parse and resolve it. The file must be present in
+/// the cache.
 pub fn sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
-    perform_sema(file, cache, ns);
+    sema_file(file, cache, ns);
+
+    // Checks for unused variables
     check_unused_namespace_variables(ns);
     check_unused_events(ns);
 }
 
-/// Load a file file from the cache, parse and resolve it. The file must be present in
-/// the cache. This function is recursive for imports.
-pub fn perform_sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
+/// Parse and resolve a file and its imports in a recursive manner.
+fn sema_file(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
     let file_no = ns.files.len();
 
     let source_code = cache.get_file_contents(&file.full_path);
@@ -177,7 +179,7 @@ fn resolve_import(
         }
         Ok(file) => {
             if !ns.files.iter().any(|f| *f == file.full_path) {
-                perform_sema(file.clone(), cache, ns);
+                sema_file(file.clone(), cache, ns);
 
                 // give up if we failed
                 if diagnostics::any_errors(&ns.diagnostics) {
--- a/src/sema/mod.rs
+++ b/src/sema/mod.rs
@@ -21,6 +21,7 @@ mod statements;
 pub mod symtable;
 pub mod tags;
 mod types;
+mod unused_variable;
 mod variables;
 
 use self::contracts::visit_bases;
@@ -30,6 +31,7 @@ use self::functions::{resolve_params, resolve_returns};
 use self::symtable::Symtable;
 use self::variables::var_decl;
 use crate::file_cache::{FileCache, ResolvedFile};
+use crate::sema::unused_variable::{check_unused_events, check_unused_namespace_variables};
 
 pub type ArrayDimension = Option<(pt::Loc, BigInt)>;
 
@@ -39,9 +41,16 @@ pub const SOLANA_BUCKET_SIZE: u64 = 251;
 pub const SOLANA_FIRST_OFFSET: u64 = 16;
 pub const SOLANA_SPARSE_ARRAY_SIZE: u64 = 1024;
 
+/// Performs semantic analysis and checks for unused variables
+pub fn sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
+    perform_sema(file, cache, ns);
+    check_unused_namespace_variables(ns);
+    check_unused_events(ns);
+}
+
 /// Load a file file from the cache, parse and resolve it. The file must be present in
 /// the cache. This function is recursive for imports.
-pub fn sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
+pub fn perform_sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
     let file_no = ns.files.len();
 
     let source_code = cache.get_file_contents(&file.full_path);
@@ -168,7 +177,7 @@ fn resolve_import(
         }
         Ok(file) => {
             if !ns.files.iter().any(|f| *f == file.full_path) {
-                sema(file.clone(), cache, ns);
+                perform_sema(file.clone(), cache, ns);
 
                 // give up if we failed
                 if diagnostics::any_errors(&ns.diagnostics) {
@@ -1469,14 +1478,14 @@ impl ast::Namespace {
         expr: &pt::Expression,
         diagnostics: &mut Vec<ast::Diagnostic>,
     ) -> Result<ArrayDimension, ()> {
-        let symtable = Symtable::new();
+        let mut symtable = Symtable::new();
 
         let size_expr = expression(
             &expr,
             file_no,
             contract_no,
             self,
-            &symtable,
+            &mut symtable,
             true,
             diagnostics,
             Some(&ast::Type::Uint(256)),
--- a/src/sema/statements.rs
+++ b/src/sema/statements.rs
@@ -319,9 +319,7 @@ fn statement(
                 diagnostics,
             )?;
 
-            let mut initialized = false;
             let initializer = if let Some(init) = initializer {
-                initialized = true;
                 let expr = expression(
                     init,
                     file_no,
@@ -332,6 +330,7 @@ fn statement(
                     diagnostics,
                     Some(&var_ty),
                 )?;
+
                 used_variable(ns, &expr, symtable);
 
                 Some(cast(&expr.loc(), expr, &var_ty, true, ns, diagnostics)?)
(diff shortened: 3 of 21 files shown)


src/lib.rs
```diff
@@ -140,6 +140,8 @@ pub fn parse_and_resolve(filename: &str, cache: &mut FileCache, target: Target)
         }
         Ok(file) => {
             sema::sema(file, cache, &mut ns);
+            sema::unused_variable::check_unused_namespace_variables(&mut ns);
+            sema::unused_variable::check_unused_events(&mut ns);
         }
```
seanyoung, line 145: It is better if `sema::sema()` calls the unused variable checks. I realise that `sema::sema()` is recursive, and you don't want this to be called for every recursive call. So maybe rename `sema::sema_file()` or something similar and create a new function `sema::sema()` which calls the unused variable checks.

src/sema/unused_variable.rs
```diff
@@ -0,0 +1,314 @@
...
+        level: Level::Warning,
+        ty: ErrorType::Warning,
+        pos: Some(loc),
+        message: text.clone(),
+        notes
+    })
+}
+
+/// Emit different warning types according to the state variable usage
+pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Box<Diagnostic>> {
```
seanyoung, line 178: Why is the diagnositic in a box?

src/sema/unused_variable.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// Marks as used variables that appear in an expression with right and left hand side.
+pub fn check_var_usage_expression(ns: &mut Namespace, left: &Expression, right: &Expression,
+                                  symtable: &mut Symtable) {
+    used_variable(ns, left, symtable);
+    used_variable(ns, right, symtable);
+}
+
+/// Generate warnings for unused varibles
+fn generate_unused_warning(loc: Loc, text: &String, notes: Vec<Note>) -> Box<Diagnostic> {
```
seanyoung, line 167: Why in a box?

seanyoung (changes requested):
Looks great! Just some minor comments, I think other than some details it looks god.

I would say it is good practise to write tests as are you coding, so you have all the edge cases in mind and don't forget about them when writing tests.

seanyoung:
Also please squash all your work into a single commit once you are done - or a few commits if each commit makes logical sense.

src/sema/unused_variable.rs
LucasSte: I create the Diagnostic struct in function `generate_unused_warning` and it is returned to `emit_warning_local_variable`. There I return the value again to the original caller. I wanted to avoid memory copy between these three returns. I thought that using a smart pointer would be good.

src/lib.rs
LucasSte: Fixed

src/sema/unused_variable.rs
seanyoung: The only caller of `emit_warning_local_variable()` derefs the box with `*`, which effectively means unboxing it. Also using box for efficiency isn't the right thing to do: rust does a lot of inlining and tries to solve these issues for you.

LucasSte: Fixed

src/sema/unused_variable.rs
```diff
@@ -0,0 +1,409 @@
...
+        | Expression::StorageArrayLength {
+            loc: _,
+            ty: _,
+            array,
+            ..
+        } => {
+            used_variable(ns, array, symtable);
+        }
+
+        Expression::StorageLoad(_, _, expr)
```
seanyoung, line 91: This function only recurses for limited set of expressions. For example, a simple `Expression::Add` is not mentioned. If you want to proper recursion there is a `recurse()` method on `Expression`.

However, as I read the code, I don't think this function needs to be recurse at all. sema resursively resolves expressions already.

src/sema/unused_variable.rs
```diff
@@ -0,0 +1,409 @@
...
+
+/// Emit different warning types according to the state variable usage
+pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Diagnostic> {
+    match &variable.usage_type {
+        VariableUsage::Parameter => {
+            if !variable.read {
+                return Some(generate_unused_warning(
+                    variable.id.loc,
+                    &format!(
+                        "Function parameter '{}' has never been read.",
```
seanyoung, line 234: compiler diagnostics do not need a full stop or start with a capital letter

src/sema/unused_variable.rs
```diff
@@ -0,0 +1,409 @@
...
+            }
+            None
+        }
+
+        VariableUsage::ReturnVariable => {
+            if !variable.assigned {
+                return Some(generate_unused_warning(
+                    variable.id.loc,
+                    &format!(
+                        "Return variable '{}' has never been assigned",
```
seanyoung, line 248: Same here and the others

src/sema/mod.rs
```diff
@@ -39,9 +41,16 @@ pub const SOLANA_BUCKET_SIZE: u64 = 251;
...
+pub fn sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
+    perform_sema(file, cache, ns);
+    check_unused_namespace_variables(ns);
+    check_unused_events(ns);
+}
+
 /// Load a file file from the cache, parse and resolve it. The file must be present in
 /// the cache. This function is recursive for imports.
-pub fn sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
+pub fn perform_sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
```
seanyoung, line 53: `sema_file()` might be a better name.

src/sema/statements.rs
```diff
@@ -309,12 +333,20 @@ fn statement(
...
             } else {
                 None
             };
 
-            if let Some(pos) = symtable.add(&decl.name, var_ty.clone(), ns) {
+            if let Some(pos) = symtable.add(
+                &decl.name,
+                var_ty.clone(),
+                ns,
+                initialized,
```
seanyoung, line 347: You could replace this with `initializer.is_some()`

seanyoung (changes requested):
Fantastic work!  I have some very minor comments.

Please also note that the clippy warnings need to be fixed. You can probably enable clippy in your ide.

.gitignore
```diff
@@ -2,3 +2,5 @@
 Cargo.lock
 /target
 **/*.rs.bk
+.DS_Store
+.idea
```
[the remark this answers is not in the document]
LucasSte: `.idea` contains configuration files from my IDE. `.DS_Store` is a Mac OS file that holds meta-data about the folder it resides in. They are not supposed to be committed so I added them to gitignore. I'll remove them from this file and add to my global .gitignore then.

src/sema/unused_variable.rs
LucasSte: Sema does resolve expressions recursively, as you mentioned. The problem is that using only `Expression::Variable` (and its variants) does not provide me the context in which the variable has been used, i.e., I cannot differentiate between an assignment and a read. Once I target that, I'll need to recurse other expressions that encapsulate a variable, such as `StorageLoad` and `ZeroExt` (but not limited to these two cases).

For `Expression::Add` and other expressions that have a left and a right side, I use the function `check_var_usage_expression`.

I agree that this function is worth some refactoring. I am happy to discuss more during our meeting tomorrow.

src/sema/unused_variable.rs
```diff
@@ -0,0 +1,409 @@
...
+                        variable.id.name
+                    ),
+                    vec![],
+                ));
+            }
+            None
+        }
+
+        VariableUsage::StateVariable => {
```
LucasSte, line 257: State variables are not local variables. The name is misleading. Please, fix this @LucasSte.

LucasSte: Fixed

src/sema/statements.rs
LucasSte: Done

src/sema/mod.rs
LucasSte: Renamed!

src/sema/statements.rs
```diff
@@ -685,6 +718,11 @@ fn statement(
                 return Err(());
             }
 
+            // for offset in symtable.returns.iter() {
```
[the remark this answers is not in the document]
LucasSte: Done

src/sema/unused_variable.rs
LucasSte: I simplified the function. Please, take a look again. I tried to recurse only for expressions that directly wrap a variable.

src/sema/mod.rs
```diff
@@ -39,9 +41,16 @@ pub const SOLANA_BUCKET_SIZE: u64 = 251;
...
+pub fn sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
+    sema_file(file, cache, ns);
+    check_unused_namespace_variables(ns);
+    check_unused_events(ns);
+}
+
 /// Load a file file from the cache, parse and resolve it. The file must be present in
 /// the cache. This function is recursive for imports.
-pub fn sema(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
+pub fn sema_file(file: ResolvedFile, cache: &mut FileCache, ns: &mut ast::Namespace) {
```
seanyoung, line 53: sema_fiile should no longer be `pub`

src/sema/mod.rs
```diff
@@ -39,9 +41,16 @@ pub const SOLANA_BUCKET_SIZE: u64 = 251;
 pub const SOLANA_FIRST_OFFSET: u64 = 16;
 pub const SOLANA_SPARSE_ARRAY_SIZE: u64 = 1024;
 
+/// Performs semantic analysis and checks for unused variables
```
seanyoung, line 44: The comment is not really correct. This function is the entry point for sema. Maybe move the comment from sema_file to here and put this comment inside the function.

src/sema/statements.rs
```diff
@@ -1236,7 +1295,9 @@ fn destructure_values(
                 diagnostics,
                 Some(&Type::Bool),
             )?;
-
+            std::println!("Aqui!!");
+            std::println!("{:?}", cond);
```
seanyoung, line 1299: Please remove your debugging :smiley:

src/sema/variables.rs
```diff
@@ -250,6 +252,8 @@ pub fn var_decl(
         ty: ty.clone(),
         constant: is_constant,
         initializer,
+        read: false,
+        assigned: initialized,
```
seanyoung, line 256: `initializer.is_some()` ?

seanyoung (changes requested):
Looks good, just a few very minor comments.

src/sema/mod.rs
LucasSte: Done

src/sema/mod.rs
LucasSte: I moved the comments and added a new one to `sema_file`

src/sema/statements.rs
LucasSte: As a curiosity, `aqui` means `here` in Portuguese :)

src/sema/variables.rs
LucasSte: I changed `initializer` to `initializer: initializer.clone()` as well.

seanyoung:
Would you mind rebasing against latest main, so it can be merged and also squashing the commits into one. Thanks.

LucasSte:
> Would you mind rebasing against latest main, so it can be merged and also squashing the commits into one. Thanks.

I left two commits: one for the code itself and another for the tests.

seanyoung (approved):

PR #435, opened 2021-07-14
Do not warning about unused public storage variable or shadowed event declarations

seanyoung:
@LucasSte please review if you have the time

### DIFF
--- a/src/sema/unused_variable.rs
+++ b/src/sema/unused_variable.rs
@@ -382,6 +382,7 @@ fn emit_warning_contract_variables(variable: &ast::Variable) -> Option<Diagnosti
             vec![],
         ));
     }
+
     //Solidity attributes zero value to contract values that have never been assigned
     //There is no need to raise warning if we use them, as they have a valid value.
 
--- a/src/sema/unused_variable.rs
+++ b/src/sema/unused_variable.rs
@@ -3,7 +3,7 @@ use crate::sema::ast::{
     Builtin, Diagnostic, ErrorType, Expression, Level, Namespace, Note, Statement,
 };
 use crate::sema::symtable::{Symtable, VariableUsage};
-use crate::sema::{ast, symtable};
+use crate::sema::{ast, contracts::visit_bases, symtable};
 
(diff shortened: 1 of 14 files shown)


src/sema/unused_variable.rs
```diff
@@ -362,21 +362,26 @@ pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Diag
...
-        ));
-    } else if !variable.assigned && !variable.read {
-        return Some(generate_unused_warning(
-            variable.loc,
-            &format!("storage variable '{}' has never been used", variable.name),
-            vec![],
-        ));
+    if !matches!(
+        variable.visibility,
+        Visibility::Public(_) | Visibility::External(_)
```
LucasSte, line 367: Why are you restricting the warning? Public variables can also be unused.

seanyoung: Any public storage variable becomes an accessor function. So if you have this code:
```
contract c {
      int public foo;
      
      function bar() public returns (int) {
           return 1;
      }
}
```
Then when you compile this, you'll get a contract which you can deploy, and you can call `foo()` on it which returns 0, and you can call `bar()` on it which returns 1.

LucasSte: I am detecting the usage of variables accessed by accessor functions. Please check the test `function_call` at `unused_variable_detection.rs`. Even though `foo()` returns zero, it might be never used in any other contract.

LucasSte: Please, check also my implementation at `src/unused_variable.rs` line 179.

seanyoung: The accessor function can be called in a transaction. You can submit a transaction to the blockchain which calls the foo accessor function, just like any public function (like bar).

LucasSte: Makes sense. If that is the case, I cannot remove public storage variables at all, right?

seanyoung: Unfortunately not no.

LucasSte (approved):

PR #437, opened 2021-07-15
Check used variables in Load expression and length call

LucasSte:
This PR implements the used variable detection for the `Load` expression and marks arrays as assigned when there is a `length` call. The latter is necessary to prevent array being removed from the intermediary representation.


seanyoung:
Looks good, I agree this is the right thing to do.

PR #448, opened 2021-07-23
Unused variable elimination

LucasSte:
This PR implements unused variable elimination for the Solang compiler and improves the unused variable detection.

For unused variable detection, this PR improves the following cases:

- Variables inside `format()`
- Functions assigned to local variables
- Storage load
- Usage of methods `address` and `selector`

For the unused variable elimination, the PR allows Solang to remove from the intermediate representation the following cases:

- Variables that have never been read nor assigned.
- Variables that have been assigned but never read.
- Assignments of unused variables, including destructures.
- Assignments of unused variables inside expressions (`a = b + (c = 1)`)

This PR is part of the Linux Foundation mentorship for the Solang compiler. For more information, please refer to the [project plan](https://wiki.hyperledger.org/display/INTERN/Project+plan+-+Solang+compiler+passes+2021).

### DIFF
--- a/src/codegen/statements.rs
+++ b/src/codegen/statements.rs
@@ -36,7 +36,8 @@ pub fn statement(
                     vartab,
                 };
 
-                check_side_effects_expressions(init, &mut params);
+                //If we remove the assignment, we must keep expressions that have side effects
+                init.recurse(&mut params, process_side_effects_expressions);
                 return;
             }
 
@@ -90,7 +91,7 @@ pub fn statement(
                         ns,
                         vartab,
                     };
-                    right.recurse(&mut params, check_side_effects_expressions);
+                    right.recurse(&mut params, process_side_effects_expressions);
 
                     if !reachable {
                         cfg.add(vartab, Instr::Unreachable);
@@ -1230,7 +1231,10 @@ impl Type {
     }
 }
 
-pub fn check_side_effects_expressions(
+/// This function looks for expressions that have side effects during code execution and
+/// processes them.
+/// They must be added to the cfg event if we remove the assignment
+pub fn process_side_effects_expressions(
     exp: &Expression,
     ctx: &mut SideEffectsCheckParameters,
 ) -> bool {
@@ -1239,10 +1243,29 @@ pub fn check_side_effects_expressions(
         | Expression::ExternalFunctionCall { .. }
         | Expression::ExternalFunctionCallRaw { .. }
         | Expression::Constructor { .. }
-        | Expression::Builtin(..) => {
+        | Expression::Assign(..)
+        | Expression::DynamicArrayPop(..)
+        | Expression::DynamicArrayPush(..) => {
             let _ = expression(exp, ctx.cfg, ctx.contract_no, ctx.func, ctx.ns, ctx.vartab);
             false
         }
+
+        Expression::Builtin(_, _, builtin_type, _) => match &builtin_type {
+            Builtin::PayableSend
+            | Builtin::ArrayPush
+            | Builtin::ArrayPop
+            // PayableTransfer, Revert, Require and SelfDestruct do not occur inside an expression
+            // for they return no value. They should not bother the unused variable elimination.
+            | Builtin::PayableTransfer
+            | Builtin::Revert
+            | Builtin::Require
+            | Builtin::SelfDestruct => {
+                let _ = expression(exp, ctx.cfg, ctx.contract_no, ctx.func, ctx.ns, ctx.vartab);
+                false
+            }
+
+            _ => true,
+        },
 
         _ => true,
     }
--- a/src/codegen/statements.rs
+++ b/src/codegen/statements.rs
@@ -3,6 +3,9 @@ use std::collections::LinkedList;
 
 use super::cfg::{ControlFlowGraph, Instr, Vartable};
 use super::expression::{assign_single, emit_function_call, expression};
+use crate::codegen::unused_variable::{
+    should_remove_assignment, should_remove_variable, SideEffectsCheckParameters,
+};
 use crate::parser::pt;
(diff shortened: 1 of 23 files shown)


src/sema/unused_variable.rs
```diff
@@ -8,6 +8,7 @@ use crate::sema::{ast, contracts::visit_bases, symtable};
 /// Mark variables as assigned, either in the symbol table (for local variables) or in the
 /// Namespace (for storage variables)
 pub fn assigned_variable(ns: &mut Namespace, exp: &Expression, symtable: &mut Symtable) {
+    //std::println!("{:?}", exp);
```
seanyoung, line 11: Please remove

src/codegen/unused_variable.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub vartab: &'a mut Vartable,
+}
+
+/// Check if we should remove an assignment. The expression in the argument is the left-hand side
+/// of the assignment
+pub fn should_remove_assignment(ns: &Namespace, exp: &Expression, func: &Function) -> bool {
+    match &exp {
+        Expression::StorageVariable(_, _, contract_no, offset) => {
+            let var = &ns.contracts[*contract_no].variables[*offset];
+            if matches!(
```
seanyoung, line 22: var.read is set to true for public/external variables already; this check should be redundant (i.e. remove line 22 to 27)

src/codegen/statements.rs
```diff
@@ -1181,3 +1230,25 @@ impl Type {
...
+    ctx: &mut SideEffectsCheckParameters,
+) -> bool {
+    match &exp {
+        Expression::InternalFunctionCall { .. }
+        | Expression::ExternalFunctionCall { .. }
+        | Expression::ExternalFunctionCallRaw { .. }
+        | Expression::Constructor { .. }
+        | Expression::Assign(..)
+        | Expression::DynamicArrayPush(..)
+        | Expression::Builtin(..) => {
```
seanyoung, line 1247: Expression::DynamicArrayPop is missing
The Expression::Builtin can be restricted further.
You only need the builtins: PayableSend, PayableTransfer, ArrayPush, ArrayPop, Revert, Require, selfdestruct

seanyoung (changes requested):
Looks great! Some minor review points. Please also squash the commits.

I guess the windows tests just randomly failed.

tests/codegen_testcases/slice1.sol
```diff
@@ -4,14 +4,15 @@ contract c {
 	function test1() public pure {
 		bytes x = "foo1";
 		// x is not being used, so it can be a slice
-// CHECK: alloc slice uint32 4 "foo1"
+// CHECK:
```
LucasSte, line 7: @seanyoung Do you have any suggestion on how I should fix this test? I tried returning bytes, but it didn't work.

src/codegen/statements.rs
LucasSte: PayableTransfer, ArrayPush, Revert, Require and selfdestruct do not return any value, so they are not allowed inside an expression. Do I still need to take care of these cases?

LucasSte:
> Looks great! Some minor review points. Please also squash the commits.
> 
> I guess the windows tests just randomly failed.

tests/codegen_testcases/unused_variable_elimination.sol
```diff
@@ -0,0 +1,198 @@
+// RUN: --emit cfg
+
+contract c2 {
+    int public cd;
+    function sum(int32 a, int32 b) public pure returns (int32) {
+        return a+b;
+    }
+// BEGIN-CHECK: c2::function::doSomething
```
LucasSte, line 8: Windows tests failed at this line. I am getting an intermittent error here. Do you have any clue on that?

seanyoung: From the output it looks like the test did not produce output. Not sure what happened there, but it was probably spurious. Let's see how is goes next time.

src/codegen/statements.rs
seanyoung: Good point, that is totally true. However the function tells you if the expression has side effects. Might as well include them with a comment that this should not occur for unused variable elimination.

LucasSte: The function also adds the expressions to the CFG. Renaming it may be a better option.

tests/codegen_testcases/unused_variable_elimination.sol
LucasSte: The problem appears to be the leading white spaces.

tests/codegen_testcases/slice1.sol
seanyoung: Yes, returning bytes will make it into a slice (this needs fixing, but needs an analysis pass across functions).

You can return a subscript:
```
        function test1() public returns (byte) {
                bytes x = "foo1";

                return x[0];
        }
```

seanyoung (approved):

src/codegen/unused_variable.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+/// Checks if we should remove a variable
+pub fn should_remove_variable(
+    pos: &usize,
+    func: &Function,
+    initializer: Option<&Expression>,
+) -> bool {
+    let var = func.symtable.vars.get(pos).unwrap();
```
seanyoung, line 56: This is an elaborate way of saying `let var = &func.symtable.vars[pos];`

seanyoung:
The CI build has an issue:

> No files were found with the provided path: ./target/debug/solang.exe. No artifacts will be uploaded.

This line `path: ./target/debug/solang.exe` needs to be update to `path: ./target/releasesolang.exe`

PR #468, opened 2021-08-16
Implement undefined varaible detection and prettify warning messages

LucasSte:
Signed-off-by: Lucas Steuernagel <lucas.tnagel@gmail.com>


src/file_cache.rs
```diff
@@ -196,4 +198,88 @@ impl FileCache {
 
         Err(format!("file not found ‘{}’", filename))
     }
+
```
seanyoung, line 201: This should not longer be necessary, this is now in https://github.com/hyperledger-labs/solang/blob/main/src/sema/file.rs

`formated_message()` no longer needs a file cache argument in the latest version because of it.

src/sema/expression.rs
```diff
@@ -30,83 +30,6 @@ use crate::Target;
 use base58::{FromBase58, FromBase58Error};
 
 impl Expression {
-    /// Return the location for this expression
```
seanyoung, line 33: Why was this moved?

seanyoung (changes requested):
Looks great so far!

src/file_cache.rs
LucasSte: I print the complete line when I am printing `Diagnotics`. I needed to fetch the line from somewhere, so I added `FileCache` again to formatted message. The function in line 203 retrieves the line and the offsets so that I can print an indicator (^) to the point in the line where there is an error. Can I do the same using only `File`?

src/sema/expression.rs
LucasSte: `fn loc` is a very essential method for `Expression`. I thought it would be nicer to move it closer to the `struct Expression` declaration to ease development and organize the code better.

src/codegen/cfg.rs
```diff
@@ -649,6 +760,7 @@ impl ControlFlowGraph {
                     .collect::<Vec<String>>()
                     .join(", ")
             ),
+            Expression::Undefined(_) => "".to_string(),
```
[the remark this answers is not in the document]
LucasSte: Fixed

src/file_cache.rs
seanyoung: `get_line_and_offset_from_loc()` belongs in `File` and `struct FileOffsets` should be removed.

src/sema/expression.rs
seanyoung: Good point

src/file_cache.rs
LucasSte: I updated the comment for this function. Please, take a look now. I can also change the name to make it clearer that the function retrieves a line from the source code.

seanyoung (approved):

PR #476, opened 2021-08-23
Move to llvm 12

seanyoung:


seanyoung:
This needs to be merged first so the CI image can be rebuilt.

PR #503, opened 2021-09-30
Implement available expression analysis

LucasSte:
This PR implements an available expression analysis for the control flow graph of Solang. This analysis is going to be necessary for the common subexpression elimination pass to be implemented.

This PR is part of the Linux mentorship. For more information, please refer to the [project's wiki](https://wiki.hyperledger.org/display/INTERN/Project+plan+-+Solang+compiler+passes+2021).

### DIFF
--- a/src/codegen/available_expressions.rs
+++ b/src/codegen/available_expressions.rs
@@ -1,0 +1,807 @@
+use crate::codegen::cfg::Instr;
+use crate::sema::ast::{Expression, StringLocation};
+use num_bigint::BigInt;
+use std::cell::RefCell;
+use std::collections::{HashMap, HashSet};
+use std::rc::Rc;
+
+/*
+The available expression analysis implemented here build a graph to track expressions. Each
+operand and each operation represents a vertex. Edges are directed from operands to an operation.
+Let's say we have a+b. 'a', 'b' and 'e1=a+b' are vertexes. Edges are directed from 'a' to 'e1=a+b'
+and from 'b' to 'a+b'. If we add now 'a+b-c', we will have two new nodes: 'c' and 'e2=e1-c'.
+Edges will connect 'c' to 'e2=e1-c' and 'e1=a+b' to 'e2=e1-c'. Whenever a variable becomes
+unavailable (i.e. we kill its definition), we recursively remove the operand node and all its
+children operations from the graph.
+ */
+
+/// This enum defines operator types for the graph
+#[derive(PartialEq, Eq, Hash, Copy, Clone)]
+enum Operator {
+    Add,
+    Subtract,
+    Multiply,
+    Divide,
+    Modulo,
+    Power,
+    BitwiseOr,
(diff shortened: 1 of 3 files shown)


src/codegen/available_expressions.rs
```diff
@@ -0,0 +1,866 @@
...
+        );
+
+        expr_set.expr_map.insert(expr_type, self.global_id_counter);
+        self.global_id_counter += 1;
+
+        self.global_id_counter - 1
+    }
+
+    /// Add a node to represent a binary expression
+    pub fn add_expression_node(
```
seanyoung, line 199: This should be `fn add_binary_node()`

src/codegen/available_expressions.rs
```diff
@@ -0,0 +1,866 @@
...
+
+        self.global_id_counter += 1;
+
+        self.global_id_counter - 1
+    }
+}
+
+impl AvailableExpressionSet {
+    /// Check if a commutative expression exists in the set
+    fn check_commutative(
```
seanyoung, line 278: Rather than return `(bool, NodeId)` I think it would be more rust-like to return `Option<NodeId>`

src/codegen/available_expressions.rs
```diff
@@ -0,0 +1,866 @@
...
+            _ => unreachable!("Not a constant"),
+        }
+    }
+
+    fn add_variable_or_arg(
+        &mut self,
+        exp: &Expression,
+        expr_type: &ExpressionType,
+        ave: &mut AvailableExpression,
+    ) -> (bool, NodeId) {
```
seanyoung, line 358: The `bool` return is always true, not needed I guess

src/codegen/available_expressions.rs
```diff
@@ -0,0 +1,866 @@
...
+            right_id, left_id, operator,
+        )) {
+            (true, *exp_id)
+        } else {
+            (false, 0)
+        }
+    }
+
+    /// Add a commutative expression to the set if it is not there yet
+    fn process_commutative(
```
seanyoung, line 307: Return `Option<NodeId>`

seanyoung (changes requested):
This could do with some documentation. I think this is original work, after all. Maybe something in `docs/optimizer.rst` and/or a big comment at the top of the file.

src/codegen/available_expressions.rs
LucasSte: Done

src/codegen/available_expressions.rs
LucasSte: Done

src/codegen/available_expressions.rs
LucasSte: Done

src/codegen/available_expressions.rs
LucasSte: Done

PR #505, opened 2021-10-01
Resolve BytesLiteral to String when hinted

sushi-shi:


seanyoung:
Thank you!

PR #507, opened 2021-10-04
Resolve constants from libraries

sushi-shi:


seanyoung (approved):
Looks very good, thanks!

PR #508, opened 2021-10-04
Allow calling constants from contracts as well

sushi-shi:


src/sema/expression.rs
```diff
@@ -4907,8 +4907,8 @@ fn member_access(
         return Ok(expr);
     }
 
-    // is it a constant in a library (unless basecontract is a local variable)
-    match library_constant(e, id, file_no, ns, symtable) {
+    // is it a constant (unless basecontract is a local variable)
+    match constant(e, id, file_no, ns, symtable) {
         None => (),
         Some(expr) => return Ok(expr),
```
seanyoung, line 4913: This seems like an baroque way of saying:

```
if let Some(expr) =  constant(e, id, file_no, ns, symtable) {
    return Ok(expr);
}
```

What do yo think?

src/sema/expression.rs
```diff
@@ -5251,7 +5251,7 @@ fn member_access(
     Err(())
 }
 
-fn library_constant(
+fn constant(
```
seanyoung, line 5254: `fn constant()` is not a great name. How about `fn contract_constant()`?

src/sema/expression.rs
```diff
@@ -5269,16 +5269,8 @@ fn library_constant(
 
     let library_no = ns.resolve_contract(file_no, namespace)?;
```
seanyoung, line 5270: contract_no?

src/sema/expression.rs
sushi-shi: Yes, of course!

seanyoung:
Very nice :+1:

PR #518, opened 2021-10-08
Cast ternary expressions

sushi-shi:


seanyoung:
Looks great, thank you for fixing that

PR #520, opened 2021-10-08
Fix StorageRefs in ternary expressions

sushi-shi:


seanyoung:
Again thanks for the fix! Looks great. The solana CI failure is due to edge docker image being broken, I'm looking at that now. No reason not to merge this.

PR #550, opened 2021-11-04
Implement common subexpression elimination

LucasSte:
This PR implements common subexpression elimination for the Solang compiler. It runs an available expression analysis during the first pass. Then, during a second pass we exchange common expressions by a temporaries.

This PR is the last milestone for the Linux Mentorship for the Hyperledger foundation. For more information about the project, please check the [wiki](https://wiki.hyperledger.org/display/INTERN/Project+plan+-+Solang+compiler+passes+2021).


examples/test.sol
```diff
@@ -0,0 +1,42 @@
+contract foo {
```
LucasSte, line 1: I will remove this file in the future.

seanyoung:
The codegen test fails randomly, that's why the mac/windows tests have failed.

seanyoung:
Please can you merge your commits into one commit, or more than one if it makes sense. It would be nice if commits are a set of logical changes.

Also please ensure your commits have the correct `Signed-off-by:` line, this is what the DCO failure is about.

src/bin/solang.rs
```diff
@@ -156,6 +156,12 @@ fn main() {
                 .long("math-overflow")
                 .display_order(5),
         )
+        .arg(
+            Arg::with_name("COMMONEXPRESSIONELIMINATION")
+                .help("Disable common subexpression elimination")
+                .long("no-cse")
+                .display_order(6),
+        )
```
seanyoung, line 164: --no-cse should be ordered before --math-overflow, please set the display order to 5 (and move math-overflow below this)

src/codegen/cfg.rs
```diff
@@ -1254,6 +1257,10 @@ pub fn optimize_and_check_cfg(
     if opt.dead_storage {
         dead_storage::dead_storage(cfg, ns);
     }
+
+    if opt.common_subexpression_elimination && func_no.is_some() {
```
seanyoung, line 1261: Any reason for `func_no.some()`? 

A default constructor has no function number, so it is `None` in that case.

seanyoung (changes requested):
Please add some words to `docs/optimizer.rst`. The documentation should at least explain that there is a CSE pass, and what it does (not necessarily how it works)

src/codegen/cfg.rs
LucasSte: A default constructor has no sub expression to be optimized.

seanyoung: True, although that is worth a comment.

LucasSte: Done

src/bin/solang.rs
LucasSte: Done

seanyoung (approved):

PR #565, opened 2021-11-14
Divide optimizer into sections

LucasSte:
Currently, the 'Unused Variable Elimination' and 'Common Subexpression Eimination' are sub sections under 'Vector to Slice Pass' in [Solang docs](https://solang.readthedocs.io/en/latest/optimizer.html). They should be a new section under Optimization Passes, instead of a subsection. This PR fix this issue.

Screen Shot 2021-11-14 at 09 53 10

### DIFF
--- a/docs/optimizer.rst
+++ b/docs/optimizer.rst
@@ -125,8 +125,11 @@ This optimization pass can be disabled by running `solang --no-vector-to-slice`.
 having this optimization pass on by comparing the output of `solang --no-vector-to-slice --emit cfg foo.sol` with
 `solang --emit cfg foo.sol`.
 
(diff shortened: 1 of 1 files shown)


seanyoung (changes requested):
This won't fix the issue. The line under "Common Subexpression Elimination" should be made using `-` rather than `_`. 

See:
https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings
(Note that we use different characters)

seanyoung:
You can test it out by running `make html` in the `docs` directory.

seanyoung (approved):

Issue #558, opened 2021-11-11
Modifiers of functions with return values containing multiple `_` fail LLVM assertions


sushi-shi:
This contract fails with
```
solang: /home/runner/work/solang/solang/llvm-project/llvm/lib/Transforms/Utils/ValueMapper.cpp:861: 
  void {anonymous}::Mapper::remapInstruction(llvm::Instruction*): 
    Assertion `(Flags & RF_IgnoreMissingLocals) && "Referenced value not in value map!"' failed.
```
```solidity
contract c {
    int foo;
    bool bar;

    function func() external mod() returns (int) {
        return foo;
    }

    modifier mod() {
        if (bar) {
            _;
        } else {
            _;
        }
    }
}
```

seanyoung:
This problem should be fixed now.

PR #579, opened 2021-11-25
Fix modifier with return values needing phis

seanyoung:
When the return value is stored in a local variable, then phis might be
needed.


seanyoung:
Fixes #558

Issue #597, opened 2021-12-10
Feature gate LLVM dependency


shekhirin:
## Problem

Currently there's no way to use `parser` without compiling the `inkwell` dependency which in turn requires the LLVM to be installed.

I encountered that issue while working on Solidity formatter in Foundry (https://github.com/gakonst/foundry/pull/201) where we use only `parser::parse` function to construct the AST. Without the hard LLVM dependency it would be much easier for users to install our CLI since they won't need to have LLVM installed on their machine.

## Solution

I tried to add an `llvm` feature to `Cargo.toml` and use `#[cfg(feature = "llvm")]` in `src/lib.rs` but that didn't work straight away since `parser` is writing lots of diagnostic messages while parsing and these messages are written from the package that’s currently depends on LLVM. So simple `[features]` in `Cargo.toml` won’t solve the issue for now.

The proposed solution is to separate the diagnostics-related types & functions into the separate module which can be used without LLVM dependency.

seanyoung:
This is totally a good idea. The solang vscode plugin also does not need llvm, and if did not need llvm, it could be built as a onode wasm module.

The llvm dependencies *should* be in `src/emit/`. If they're not, they should be abstracted/moved. 

An Solidity formatter is a great.I'd love to see that happen.

PR #615, opened 2021-12-30
Make solidity parser a standalone crate

mattsse:
This is essentially a followup on #597 and makes the solidity parser a standalone crate (`solang-parser`)

## Changes
* move `src/parser/*` to `solang-parser`, including lalrpop build
* move `src/sema/ast/Diagnostic` related structs to `solang-parser/src/diagnostics/`. This was necessary because the parser's error is the Diagnostic, but has the downside that some `Diagnostic` implementation is now in two places:
  * made `src/sema/ast/Diagnostic::formatted_message` a function that accepts a `&Diagnostic` instead
* rm unused dependencies in solang 
* reexport the parser, so it doesn't break


gakonst (approved):
lgtm

seanyoung:
Good idea. Just reviewing now.

seanyoung (approved):
LGTM

seanyoung:
I've uploaded https://crates.io/crates/solang-parser

mattsse:
Amazing, thank you!

PR #630, opened 2022-01-08
Add location to solidity import directive

rkrasiuk:
Pertaining to the discussion in [ethers-rs/feat(solc): flatten](https://github.com/gakonst/ethers-rs/pull/774) PR, add location to the import directives to be able to parse them out.

cc: @mattsse


seanyoung (approved):
Changes look good, as long as CI passes

PR #634, opened 2022-01-09
feat(parser): try statements without low-level catch clause

shekhirin:
Resolves https://github.com/hyperledger-labs/solang/issues/633

### DIFF
--- a/solang-parser/src/lib.rs
+++ b/solang-parser/src/lib.rs
@@ -90,6 +90,18 @@ mod test {
                     }
                     string __abba_$;
                     int64 $thing_102;
+                }
+                
+                function bar() {
+                    try sum(1, 1) returns (uint sum) {
+                        assert(sum == 2);
+                    } catch (bytes memory b) {
+                        revert('meh');
+                    } catch Error(string memory error) {
+                        revert(error);
+                    } catch Panic(uint x) {
+                        revert('feh');
+                    }
                 }";
 
         let lex = lexer::Lexer::new(src);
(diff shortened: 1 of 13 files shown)


shekhirin:
@seanyoung does such fix looks alright? I'll write some tests if so and make it ready for review then.

seanyoung (commented):
I think you're on the right track, however I think the situation is more complicated than that.
This is also valid solidity (news to me!):

```
pragma solidity >=0.7.0 <0.9.0;

interface Summer {
    function sum(uint a, uint b) external returns (uint);
}

contract Contract {
    constructor(Summer summer) {
        try summer.sum(1, 1) returns (uint sum) {
            assert(sum == 2);
	} catch (bytes memory b) {
	   revert('meh');
        } catch Error(string memory error) {
            revert(error);
        } catch Panic(uint x) {
            revert('feh');
        }
    }
}
```
However we can stick with the change you have so far, just needs a test case.

shekhirin:
@seanyoung you're right, I see it was [added](https://github.com/ethereum/solidity/releases/tag/v0.8.1) in Solidity 0.8.1.

I supported both `Error` and `Panic` identifiers in the PR. It also required the support of multiple named catches (when both `Error` and `Panic` are present).

shekhirin:
@seanyoung I think it would be useful to have Solidity -> Parse Tree tests similar to Substrate or Solana `.dot` files. E.g. Solana doesn't support converting `assembly` statements, so we can't test them that way.

PR #636, opened 2022-01-11
feat(parser): missing assembly statements

shekhirin:
Resolves https://github.com/hyperledger-labs/solang/issues/631

### DIFF
--- a/solang-parser/src/lexer.rs
+++ b/solang-parser/src/lexer.rs
@@ -66,7 +66,6 @@ pub enum Token<'input> {
     Throw,
     Emit,
     Return,
-    Leave,
     Returns,
 
     Uint(u16),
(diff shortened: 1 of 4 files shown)


seanyoung:
`leave` etc are not keywords in solidity unfortunately. I think they are in yul though.

seanyoung (approved):
I think the changes are fine.

There is an issue that now `leave`, `switch`, `case`, and `default` are now keywords in solidity. They are not keywords in solidity when using the ethereum solidity compiler.

shekhirin:
> `leave` etc are not keywords in solidity unfortunately. I think they are in yul though.

oh I see. Yeah, I'll figure that out and also rebase onto fresh master branch. Thanks for the review!

PR #637, opened 2022-01-12
Support skiping implicit type cast checking by using flag --no-implicit-type-cast-check

taquangtrung:
Hi,

I created this PR to resolve the issue reported in https://github.com/hyperledger-labs/solang/issues/632 where Solang reports errors on implicit type truncation of integers (which Solc can compile well).

In particular, I added a field `pub implicit_type_cast_check: bool` into the structure `Options` to capture this flag.

I also move the structure `Options` from `codegen/mod.rs` into `lib.rs` so that the compiler options can be accessed by both `sema` and `codegen` modules.

Can you review this PR and advise if my implementation can be merged?

Thank you!

### DIFF
--- a/src/bin/languageserver/mod.rs
+++ b/src/bin/languageserver/mod.rs
@@ -57,10 +57,12 @@ impl SolangServer {
 
             let os_str = path.file_name().unwrap();
 
-            let mut ns = parse_and_resolve(os_str, &mut resolver, self.target);
+            let opt = &Default::default();
+
+            let mut ns = parse_and_resolve(os_str, &mut resolver, self.target, opt);
 
             // codegen all the contracts; some additional errors/warnings will be detected here
-            codegen(&mut ns, &Default::default());
+            codegen(&mut ns, opt);
 
             let diags = ns
                 .diagnostics
--- a/src/bin/solang.rs
+++ b/src/bin/solang.rs
@@ -9,10 +9,13 @@ use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 
 use solang::abi;
-use solang::codegen::{codegen, OptimizationLevel, Options};
+use solang::codegen::codegen;
 use solang::emit::Generate;
 use solang::file_resolver::FileResolver;
 use solang::sema::{ast::Namespace, diagnostics};
+
+use solang::OptimizationLevel;
+use solang::Options;
 
 mod doc;
 mod languageserver;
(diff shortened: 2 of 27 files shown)


seanyoung (changes requested):
Sorry for taking time to review this. I've been considering your change and comparing the behaviour of solc.

I think the idea is right, but I'm afraid it will need some changes.

The issue is that anywhere where value is used, the type is different in solc/evm. In ewasm and substrate, value is represented as `uint128` (see https://github.com/ewasm/design/blob/master/eth_interface.md) and as `uint64` in solana. This is the nature of their VM.

However, lots of solidity code assumes that value is `uint256` and this is causing a lot of errors when compiling solidity code. We do need a fix for this.

However your change suppresses *all* implicit conversion which is not correct. Take this solidity code for example:

```
contract foo {
	function f(uint256 x) public returns (uint128) {
		return x;
	}
}
```
This does not compile with solc either, however your PR suppresses the error which is just not correct.

The only change that is needed is that when `uint256` is converted to `Type::Value`, then it should be permitted with a warning and a conversion check. `Expression::Trunc` throws away the higher bits. Optionally there could be an option to suppress the warning output.

Perhaps `Expression::Trunc` could have an extra boolean field which signifies that the codegen should check no value is actually being truncated (if the value fits into `uint256` and not into `uint128`).

Please let me know if this makes sense, and let me know what you think

taquangtrung:
@seanyoung: thank you for the feedback.

> The only change that is needed is that when `uint256` is converted to `Type::Value`, then it should be permitted with a warning and a conversion check. `Expression::Trunc` throws away the higher bits. Optionally there could be an option to suppress the warning output.
> 

Yes, I agree that for the current state of Solang, maybe we only need to check the conversion of `uint256` and `Expression::Trunc`. 

Currently, I move the `Options` data structure to `lib.rs` so that it can be used by `sema` (previously it's in `codegen` modules). Are you OK with this change?


> Perhaps `Expression::Trunc` could have an extra boolean field which signifies that the codegen should check no value is actually being truncated (if the value fits into `uint256` and not into `uint128`).
> 

Do you think this check is dependent on target, e.g., `ewasm`, `solana`, ...? If so, we can add the target to `Options` and check it too.

seanyoung:
Issue resolved in #650

taquangtrung:
Thanks for the update! Yes, your solution is more systematic!

Issue #639, opened 2022-01-16
Reserved keywords collision in Yul assembly


shekhirin:
Such snippet
```solidity
contract Contract {
    constructor() {
        assembly {
            return(0, 0)
        }
    }
}
```
can't be parsed due to 
```
 [Diagnostic { level: Error, ty: ParserError, pos: Some(Loc(0, 71, 77)), message: "unrecognised token `return', expected \"address\", \"bool\", \"break\", \"bytes\", \"continue\", \"false\", \"for\", \"if\", \"leave\", \"let\", \"mapping\", \"payable\", \"string\", \"switch\", \"true\", \"}\", Bytes, Int, Uint, hexnumber, identifier, number, string", notes: [] }]
```
error. The reason for this is that `return` is a reserved keyword in Solidity but just a regular function in Yul.

This is an issue similar to https://github.com/hyperledger-labs/solang/pull/636#pullrequestreview-853224787, i.e. requires separation of Yul and Solidity keywords to prevent such collisions.

seanyoung:
So I have been thinking about this problem for some time. I would like it to be possible to specify amounts like `1 sol` or `50000 lamports` for Solana, without introducing new keywords. 

The lalrpop parser has a separate lexer, so this is hard to work around. One way is not make `return` a keyword, and then parse `return ...` as `Identifier ...` and then check if the identifier is `return`. This is how you traditionally work around these issues, but it's a workaround.

Another issue is #625. I don't think this is possible to fix without some changes to lalrpop itself.

So I have been thinking about this problem for a while, and I've been reading a lot about parser generators. My first idea was to use peg grammars, which allow for left recursion. This is why I started https://github.com/seanyoung/lrpeg This would solve the keyword issue.

However, peg grammars don't have great performance when it comes to back-tracking. Rat-pack parsing helps a little, but LR type grammars are much faster, as much less backtracking is done.

There is another issue with peg grammars and lalrpop. You want the parser to parse as much as possible when syntax errors are present. In other words, if someone types "for let ()" in one function, parsing should not fail completely. This is important for code editors / language servers, like the solang vscode extension. Code is mostly broken when the developer is typing.

So I think we want GLR parsing, with error recovery and context-sensitive keyword parsing. This has lead me to consider tree sitter, which has all these properties. There is also a grammar for solidity https://crates.io/crates/tree-sitter-solidity

[...]

PR #652, opened 2022-02-01
Permit assembly keywords in solidity and vice versa

seanyoung:
This ensures that `switch`, `default`, `case`, and `leave` are permitted in solidity, and that `return` is permitted in assembly.


seanyoung:
Fixes #639

PR #656, opened 2022-02-02
The parser should record all comments

seanyoung:


seanyoung:
See #652

seanyoung:
@shekhirin what do you think? Could this work?

gakonst:
@seanyoung following up on this - https://github.com/gakonst/foundry/issues/780#issuecomment-1047064967

> comments are not part of the AST right? this makes it significantly harder to traverse the tree and simply to node -> text.
> perhaps we can solve this with a macro that we add in every visit function, that checks if the next comment's location is before the location of current node, if so visit_comment?

What would be the recommended way for us to figure out where the comments are and restore them in the source code?

seanyoung:
So the way I think this could work is like so. You generate the source from the parse tree. So you walk the parse tree, and make the source code.

```
fn print_source(tree: &pt::SourceUnit) {
    for unit in tree.0 {
         match unit {
              SourceUnitPart::EnumDefinition { name, values, .. } => {
                  println!("enum {} = {{ {} }}", name.name);
                  println!("{}", values.map(|id| id.name.as_str()).join(", "));
                  println!("}}");
              }
              _ => todo!(),
         }
    }
}
```

Now, we add in the comments. Each time we print *anything* we should check if there is a comment before it. This will litter the code with tons of calls. 

```
struct Formatter {
      last: usize,
     comments: Vec<pt::Comments>,
}

impl Formatter {
   fn print_comment(&mut self, loc: Loc) {
       if self.comments.is_empty() {
             return;
       } 

      while self.comments[0].2 < loc.1 {
            match &self.comments[0] {
                  Comment::Line(_, s) => println!("// {}",s);
                  Comment::Block_, s) => println!("/* {} */",s);
            }

            // this will cause a memcpy of all elements! Maybe keep track of "comments to be processed"
            self.comments.remove(0);
      }
   }   

   fn print_source(&self, tree: &pt::SourceUnit) {
    for unit in tree.0 {
         match unit {
              SourceUnitPart::EnumDefinition { loc, name, values, .. } => {
                  println!("enum ");
                  self.print_comments(&name.loc);
                  println!("{} = {{", name.name);
                  for (i, value) in values.iter().enumerate() {
                       if i == 0  {
                           println!(", ");
                       }
                       self.print_comments(&value.loc);
                       println!("{}", value.name);
                  }
                  self.print_comments(loc);
                  println!("}}");
              }
              _ => todo!(),
         }
    }
}
```

I have not compiled this, it's more pseudo-code.

Does this make sense at all?

PR #662, opened 2022-02-10
feat(parser): add support for custom error types

mattsse:
Add
* ErrorDefinition
* ErrorParameter

support in lalrpop and ast types (based on existing EventDef/EventParam), following https://docs.soliditylang.org/en/v0.8.11/grammar.html#a4.SolidityParser.errorDefinition

### DIFF
--- a/solang-parser/src/lexer.rs
+++ b/solang-parser/src/lexer.rs
@@ -39,6 +39,7 @@ pub enum Token<'input> {
 
     Struct,
     Event,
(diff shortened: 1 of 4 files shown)


seanyoung:
Unfortunately this does make `error` a keyword.

seanyoung:
Let's solve the `error` keyword in another commit, this is a bit of an edge case.

LGTM

PR #676, opened 2022-02-21
fix: inkwell cc link bug

dannysun85:
inkwell cc link bug


seanyoung (changes requested):
- Why is `no-libffi-linking` removed? 
- What is the inkwell cc link bug?
- Please your `Signed-off-by` to the commit, see https://wiki.hyperledger.org/display/TSC/DCO+Validation+During+Contribution+Review

PR #677, opened 2022-02-23
Support abi.encodeCall() syntax (introduced in solc 0.8.12)

seanyoung:
This is nice syntax from solc-0.8.12 which type-checks the arguments.


src/sema/builtin.rs
```diff
@@ -250,6 +250,18 @@ static BUILTIN_FUNCTIONS: Lazy<[Prototype; 24]> = Lazy::new(|| {
...
+        Prototype {
+            builtin: Builtin::AbiEncodeCall,
+            namespace: Some("abi"),
+            method: None,
+            name: "encodeCall",
+            args: vec![],
+            ret: vec![],
+            target: vec![],
+            doc: "Abi encode given arguments with function signature",
+            // it should be allowed in constant context, but we don't supported that yet
```
LucasSte, line 262: There is a type at the comment.

src/sema/builtin.rs
```diff
@@ -1122,6 +1135,85 @@ pub fn resolve_namespace_call(
...
+                    | Type::InternalFunction { params, .. } => {
+                        resolved_args.push(function);
+
+                        if args.len() - 1 != params.len() {
+                            diagnostics.push(Diagnostic::error(
+                                *loc,
+                                format!(
+                                    "function takes {} arguments, {} provided",
+                                    params.len(),
+                                    args.len() - 1
```
LucasSte, line 1161: `abi.encodeCall` receives a function as the first argument. Does `params.len()` takes that into account? I think it would be confusing if you don't count the function as an argument in the error message.

src/sema/statements.rs
```diff
@@ -293,6 +293,7 @@ pub fn resolve_function_body(
 }
 
 /// Resolve a statement
+#[allow(clippy::ptr_arg)]
```
LucasSte, line 296: What did add this?

LucasSte (changes requested):

src/sema/builtin.rs
seanyoung: Good catch, fixed now.

src/sema/builtin.rs
seanyoung: `params` is a reference to the function prototype parameters, and `args` is the arguments provided to `abi.encodeCall()`. I think this is right.

src/sema/statements.rs
seanyoung: The commit description could be better. 

I've been testing rust beta channel, so that things work when rust 1.59.0 is released later today. This release includes a bizarre clippy warning that says that `& mut Vec<_>` parameter can be replaced with `&mut [_]` which is not the case because we want to push onto the vector.

So we silence the clippy warning.

seanyoung: I've improved the commit description

src/sema/builtin.rs
LucasSte: When the function is properly invoked `args.len() == params.len()` should be true, right? You compare `args.len()-1` with `params.len()`.

seanyoung: Take this example:
```
    contract c {
        function f1() public {
            bytes foo = abi.encodeCall(c.bar, 102, true);
        }
        function bar(int a, bool b) public {}
    }
```

The call to `abi.encodeCall()` has three arguments: the function, and the two arguments for the function.

LucasSte (approved):

seanyoung:
Thanks!

Issue #678, opened 2022-02-23
LLVM ERROR: Unsupported dynamic stack allocation on "transfer"


NicolasWent:
Hello,

I have an issue on transferring solana. I have the latest version of Solang.

Here is the smart contract:
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.7.0;

contract Shares {
    event Transfer(address to, uint amount, uint balance);

    struct Share {
        address payable shareOwner;
        uint amount;
    }

    Share[] private _shares;

    /// @notice Create the shares object that gives the shares to every body
    constructor() {
        _shares.push(Share(msg.sender, 1000));
    }

    function getShares() external view returns(address[] memory, uint[] memory) {
        address[] memory retAddress = new address[](uint32(_shares.length));
        uint[] memory retShare = new uint[](uint32(_shares.length));
        for (uint i = 0; i < _shares.length; i++) {
            retAddress[i] = _shares[i].shareOwner;
            retShare[i] = _shares[i].amount;
        }
        return (retAddress, retShare);
    }

    function _senderIsAShareOwner() private view returns (bool) {
        for (uint i = 0; i < _shares.length; i++) {
            if (_shares[i].shareOwner == msg.sender) {
                return true;
            }
        }
        return false;
    }

    /**
    ** @dev Allow a share owner to retrieve his money. It empty the money contained inside of the smart contract to give it to owners.
     */
    function withdraw() external {
        require(_senderIsAShareOwner(), "You can't withdraw if you are not a share owner");
        uint curr_balance = address(this).balance;
        require(curr_balance > 0, "There is nothing to withdraw");
        for (uint i = 0; i < _shares.length; i++) {
            uint to_transfer = curr_balance * _shares[i].amount / 1000;
            _shares[i].shareOwner.transfer(uint64(to_transfer));
            emit Transfer(_shares[i].shareOwner, to_transfer, curr_balance);
        }
        if (address(this).balance > 0) {
            // Send the remaining money to the one who withdraw so there is nothing left on
            // the contract
            payable(msg.sender).transfer(address(this).balance);
        }
    }
}
```

Then I run
> ./solang --target solana Shares.sol

It outputs:
> LLVM ERROR: Unsupported dynamic stack allocation
> Aborted (core dumped)

Can someone help me with that please? There was also a related issue on stack overflow: https://stackoverflow.com/questions/71174709/solang-compiler-fails-when-sending-lamports-within-a-loop-targeting-solana

[...]

PR #680, opened 2022-02-24
Add missing cast which causes pointer to address not to be loaded

seanyoung:


seanyoung:
Make solidity in issue #678 compile

src/sema/expression.rs
```diff
@@ -5995,21 +5995,30 @@ fn method_call_pos_args(
                 ResolveTo::Type(&Type::Value),
             )?;
 
+            let address = cast(
+                &var_expr.loc(),
+                var_expr,
```
LucasSte, line 6000: Is there any case in which `loc` is not the same `loc` saved in an expression for the `cast` function? It seems redundant to have these two arguments as the former is an attribute of the latter.

seanyoung: That is a good point, I had not considered that. 

I think you are totally right and this is remnant of the way the code used to be structured. 

Let me see if I remove the argument.

seanyoung: I started this work and now I found a case where we use a different location for the first argument and the expression. You can see that here.

image

LucasSte (approved):

PR #681, opened 2022-02-24
Enable assembly compatibility on Solang parser

LucasSte:
This PR updates the Solang-parser to be compatible with inline assembly and does some refactoring.


solang-parser/src/pt.rs
```diff
@@ -597,7 +597,8 @@ pub enum Statement {
     },
     Assembly {
         loc: Loc,
-        assembly: AssemblyBlock,
+        assembly: Vec<AssemblyStatement>,
+        dialect: Option<String>,
```
seanyoung, line 601: Very much a nitpick, but in the parse tree I would place fields in the order they appear in the original source, so `dialect` first. Also I think `statements` would be a better name than `assembly`.

solang-parser/src/pt.rs
```diff
@@ -653,28 +658,52 @@ pub enum AssemblyStatement {
...
+    FunctionCall(Box<AssemblyFunctionCall>),
     Member(Loc, Box<AssemblyExpression>, Identifier),
     Subscript(Loc, Box<AssemblyExpression>, Box<AssemblyExpression>),
 }
 
-pub type AssemblyBlock = Vec<AssemblyStatement>;
+#[derive(Debug, PartialEq, Clone)]
+pub struct AssemblyTypedIdentifier {
+    pub loc: Loc,
+    pub identifier_name: Identifier,
```
seanyoung, line 683: When naming a field in a struct, repeating the name of the struct in fields in unnecessary. These fields can be called `name` and  `type`.

solang-parser/src/solidity.lalrpop
```diff
@@ -752,8 +752,10 @@ NonIfStatement: Statement = {
     <l:@L> "unchecked" "{" <statements:Statement*> "}" <r:@R> => {
         Statement::Block { loc: Loc::File(file_no, l, r), unchecked: true, statements }
     },
-    <l:@L> "assembly" <assembly:AssemblyBlock> <r:@R> => {
-        Statement::Assembly { loc: Loc::File(file_no, l, r), assembly }
+    <l:@L> "assembly" <dialect:string?> <assembly:AssemblyBlock> <r:@R> => {
+        Statement::Assembly {
+         loc: Loc::File(file_no, l, r), assembly,
+          dialect: if let Some(dialect_name) = dialect {Some(dialect_name.to_owned())} else {None}}
```
seanyoung, line 758: You can write the if statement as `dialect.map(|dialect| dialect.to_owned())`, much shorter.

The indentation is askew.

solang-parser/src/solidity.lalrpop
```diff
@@ -779,30 +781,21 @@ NonIfStatement: Statement = {
...
-    },
+     <l:@L> "{" <statement:AssemblyStatement> "}" <r:@R> => { AssemblyStatement::Block(Loc::File(file_no, l, r),
+ Box::new(statement))},
+ <AssemblyVariableDeclaration> => <>,
+ <AssemblyAssignment> => <>,
+ <AssemblyFunctionCall> => AssemblyStatement::FunctionCall(Box::new(<>)),
+ <AssemblyIf> => <>,
+ <AssemblyFor> => <>,
+ <AssemblySwitch> => <>,
+ <AssemblyFunctionDefinition> => <>,
```
seanyoung, line 798: Indentation

solang-parser/src/solidity.lalrpop
```diff
@@ -866,23 +827,107 @@ AssemblyExpression0: AssemblyExpression = {
...
+       };
+     AssemblyStatement::FunctionDefinition(Box::new(AssemblyFunctionDefinition{
+    loc: Loc::File(file_no, l, r),
+    name,
+    params,
+    returns,
+    body
+ }))
+
+ }
```
seanyoung, line 873: indentation

solang-parser/src/solidity.lalrpop
```diff
@@ -866,23 +827,107 @@ AssemblyExpression0: AssemblyExpression = {
...
+AssemblyTypedIdentifierList: Vec<AssemblyTypedIdentifier> = {
+ CommaOne<<AssemblyType>> => <>
+}
+
+AssemblyType: AssemblyTypedIdentifier = {
+    <l:@L> <var_name:AssemblyIdentifier> ":" <var_type:AssemblyIdentifier> <r:@R> =>
+ AssemblyTypedIdentifier{loc: Loc::File(file_no, l, r),
+  identifier_name: var_name, identifier_type: Some(var_type)},
+ <l:@L> <var_name:AssemblyIdentifier> <r:@R> =>
+ AssemblyTypedIdentifier{loc: Loc::File(file_no, l, r), identifier_name: var_name, identifier_type: None},
```
seanyoung, line 886: indentation

seanyoung (changes requested):
Looks great just some minor issues.I hope they make sense.

solang-parser/src/pt.rs
LucasSte: `type` is a reserved word in Rust. I can't name the field as `type`, so I am changing to `assembly_type`.

solang-parser/src/solidity.lalrpop
```diff
@@ -752,8 +752,12 @@ NonIfStatement: Statement = {
     <l:@L> "unchecked" "{" <statements:Statement*> "}" <r:@R> => {
         Statement::Block { loc: Loc::File(file_no, l, r), unchecked: true, statements }
     },
-    <l:@L> "assembly" <assembly:AssemblyBlock> <r:@R> => {
-        Statement::Assembly { loc: Loc::File(file_no, l, r), assembly }
+    <l:@L> "assembly" <dialect:string?> <statements:AssemblyBlock> <r:@R> => {
```
seanyoung, line 755: Here you us the string token, which will give the string, but not its location. If the contents is something we do not expect, then in sema we would like to give a warning with the correct location. Therefore it is better to `StringLiteral`.

solang-parser/src/solidity.lalrpop
```diff
@@ -812,48 +807,16 @@ AssemblyStatement: AssemblyStatement = {
...
-    },
-    AssemblyExpression0
-}
-
-AssemblyExpression0: AssemblyExpression = {
-    <a:@L> "true" <b:@R> => AssemblyExpression::BoolLiteral(Loc::File(file_no, a, b), true),
-    <a:@L> "false" <b:@R> => AssemblyExpression::BoolLiteral(Loc::File(file_no, a, b), false),
-    <l:@L> <n:number> <r:@R> => {
+AssemblyLiteral: AssemblyExpression = {
+<a:@L> "true" <t_type:(":" <AssemblyIdentifier>)?> <b:@R> => AssemblyExpression::BoolLiteral(Loc::File(file_no, a, b), true, t_type),
```
seanyoung, line 817: Indentation is a bit messy here.

solang-parser/src/solidity.lalrpop
```diff
@@ -866,23 +829,110 @@ AssemblyExpression0: AssemblyExpression = {
...
+            body
+            }
+          )
+        )
+ }
+
+}
+
+AssemblyTypedIdentifierList: Vec<AssemblyTypedIdentifier> = {
+ CommaOne<<AssemblyType>> => <>
```
seanyoung, line 881: Indentation is a bit messy.

solang-parser/src/pt.rs
seanyoung: This issue comes up in other parts of the parse tree. In other parts we have used `ty` instead.

PR #685, opened 2022-03-04
array of structs should not be array of pointers to struct

seanyoung:
Fixed sized arrays and structs should be stored inline rather than
storing a pointer to them. Any many cases this reduces the number of
allocation required, the amount of pointer chasing we need to do and
most importantly, makes it possible to use the same data layout for
solana system calls.

This has the following effects on the type system:

 - a variable declared as `int[2] a` is a pointer to an array of two
   elements with type Type::Array(Type::Int(256), Some(2)).
 - a structure declared as `struct S { bool f1; int[2] f2; }` is now
   a single structure and f2 is no longer a pointer to another object.
 - when `f2` of structure `S` is accessed, sema gives it the type:
   Type::Ref(Type::Array(Type::Int(256), Some(2))). This is still a
   simple pointer (as would have been without the Type::Ref. The
   Type::Ref() means is can be used as an l-value, else it would be
   un-assignable.
 - however, assigning to this type does mean copying the entire member
   rather than just a pointer.
 - in many instances, pointer chasing is no longer necessary

This is a minimal implementation which needs further optimization. For
example, this code:

        struct S { int[2] a; bool b; }

        S s = S([1,2], true);

This will allocate an [1,2] array and then copy it into the larger
struct. This allocate/copy should be avoided. The same issue appears in
many other situations:

        struct S { int[2] a; bool b; }

        function lit() returns (int[2]) {
                return [1, 2]
        }

        S s = S(lit(), true);

and

        contract C {
                struct S { int[2] a; bool b; }
                int[2] b;

                function foo() public {
                        S s = S(b, true);
                }

In each instance, we should allocate a single struct and then pass the
pinter for the a member to the struct literal generation, function call,
or the codegen for reading from storage. This would avoid the unnecessary
copy.


LucasSte (commented):
I know all tests have passed, but do you think you can add a test that explicitly verifies the behavior you implemented?
Usually, we should try to test all branches of our code.

seanyoung:
> I know all tests have passed, but do you think you can add a test that explicitly verifies the behavior you implemented? Usually, we should try to test all branches of our code.

This is  implicitly tested by the existing tests. However, maybe it's time for some tests that check llvm IR. Let me have a look.

seanyoung:
> > I know all tests have passed, but do you think you can add a test that explicitly verifies the behavior you implemented? Usually, we should try to test all branches of our code.
> 
> This is implicitly tested by the existing tests. However, maybe it's time for some tests that check llvm IR. Let me have a look.

I've added an llvm test which checks the type of the structs in structs, arrays of structs etc. Ready for review again.

LucasSte (approved):

PR #687, opened 2022-03-04
Resolve yul literals

LucasSte:
This PR implements the semantic analysis of YUL literals. The code is not yet fully integrated into Solang's sema, so there are workarounds to avoid build warnings.

### DIFF
--- a/solang-parser/src/solidity.lalrpop
+++ b/solang-parser/src/solidity.lalrpop
@@ -834,6 +834,12 @@ AssemblyLiteral: AssemblyExpression = {
     <l:@L> <n:hexnumber> <t_type:(":" <AssemblyIdentifier>)?> <r:@R> => {
         AssemblyExpression::HexNumberLiteral(Loc::File(file_no, l, r), n.to_owned(), t_type)
     },
+    <l:@L> <s:hexstring> <t_type:(":" <AssemblyIdentifier>)?> <r:@R> => {
+        let v = s.to_string();
+        let hex_len = v.len() - 5;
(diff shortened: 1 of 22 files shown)


solang-parser/src/solidity.lalrpop
```diff
@@ -885,10 +891,10 @@ AssemblyType: AssemblyTypedIdentifier = {
     <l:@L> <var_name:AssemblyIdentifier> ":" <var_type:AssemblyIdentifier> <r:@R> =>
        AssemblyTypedIdentifier{
             loc: Loc::File(file_no, l, r),
-            name: var_name, ty: Some(var_type)
+            id: var_name, ty: Some(var_type)
        },
     <l:@L> <var_name:AssemblyIdentifier> <r:@R> =>
-       AssemblyTypedIdentifier{loc: Loc::File(file_no, l, r), name: var_name, ty: None},
+       AssemblyTypedIdentifier{loc: Loc::File(file_no, l, r), id: var_name, ty: None},
```
seanyoung, line 897: total nitpick: you could do:
```
    <l:@L> <name:AssemblyIdentifier> <r:@R> =>
       AssemblyTypedIdentifier{loc: Loc::File(file_no, l, r), name, ty: None},
```
Slightly shorter

seanyoung (approved):

src/lib.rs
```diff
@@ -89,7 +89,7 @@ impl Target {
     }
 
     /// Size of a pointer in bytes
```
LucasSte, line 91: @seanyoung This should be "Size of a pointer in bits", shouldn't it?

seanyoung: Good catch, yet

Issue #679, opened 2022-02-23
msg.sender provides wrong AccountId on substrate


h4nsu:
(seems to me like this could be related to #666)

I'm using solang 0.1.9 installed via cargo from crates.io and a custom substrate node (https://github.com/Cardinal-Cryptography/aleph-node/tree/smartnet) which is based on substrate 4.0.0-dev. It looks like `msg.sender` is providing an incorrect value. This is my minimal example:
```
contract mytoken {
    function test(address account, bool sender) public view returns (address) {
        if (sender) {
            return msg.sender;
        }
        return account;
    }
}
```
When I call that function from some address and provide the same address as the `account` parameter, I'm getting two different values when switching the bool parameter:

![image](https://user-images.githubusercontent.com/9969515/155400749-133514bb-e67b-42e0-bf4d-d58e62092687.png)

LucasSte:
Hi @h4nsu ,
Please check [this](https://github.com/hyperledger-labs/solang/pull/682) PR. I added your example contract to our mock VM for Substrate test and our integration tests with Substrate. Everything seems to be working as expected.

I also tested your contract on Substrate's interface and see that the returned addresses aren't equal. Perhaps, this is a issue with Substrate's UI interface. Did you experience `msg.sender` returning the incorrect address in any other circumstances?

LucasSte:
Hello @h4nsu,
I added the tests you suggested [here](https://github.com/hyperledger-labs/solang/pull/690). Our CI tests passes as well.
Do you have any suggestion of integration tests we should be testing to recreate your issue?
Integration tests deploy a Solang compiled contract into Substrate and allow us to call its methods. Have a look at the test I added for your issue on my last PR:
https://github.com/hyperledger-labs/solang/blob/main/integration/substrate/msg_sender.spec.ts#L7

extraymond:
@LucasSte 
I think it might possible pallet-contract now takes a salt argument. And the integration test uses old pallet-contract code, which didn't have that as a option, also the polkadot-js api used in integration tests didn't take that salt argument either.

This is my related substrate so issue.
https://substrate.stackexchange.com/questions/714/cross-contract-call-failure-when-using-different-salt-to-instantiate-contract

As you can see, the address generated with different salt argument ends up having different behavior.

seanyoung:
@extraymond this issue here does not use the instantiate contract api call.

extraymond:
[...]

PR #690, opened 2022-03-08
Improve msg.sender test

LucasSte:
This PR updates the test of `msg.sender`, according to the suggestions at issue #679.

### DIFF
--- a/tests/substrate_tests/builtins.rs
+++ b/tests/substrate_tests/builtins.rs
@@ -705,4 +705,16 @@ fn my_token() {
 
     runtime.function("test", TokenTest(addr, false).encode());
     assert_eq!(&runtime.vm.output[..], &addr[..]);
+
+    runtime.function(
+        "test",
+        TokenTest(<[u8; 32]>::try_from(&runtime.vm.caller[..]).unwrap(), true).encode(),
+    );
+    assert_eq!(&runtime.vm.caller[..], &runtime.vm.output[..]);
+
+    runtime.function(
+        "test",
+        TokenTest(<[u8; 32]>::try_from(&runtime.vm.caller[..]).unwrap(), false).encode(),
+    );
+    assert_eq!(&runtime.vm.caller[..], &runtime.vm.output[..]);
 }


tests/substrate_tests/builtins.rs
```diff
@@ -705,4 +705,16 @@ fn my_token() {
 
     runtime.function("test", TokenTest(addr, false).encode());
     assert_eq!(&runtime.vm.output[..], &addr[..]);
+
+    runtime.function(
+        "test",
+        TokenTest(<[u8; 32]>::try_from(&runtime.vm.caller[..]).unwrap(), true).encode(),
+    );
+    assert_eq!(&runtime.vm.caller[..], &runtime.vm.output[..]);
```
seanyoung, line 713: `runtime.vm.caller` is of type `Account`, which is defined as `type Account = [u8; 32];`. This `try_from..unwrap()` is not really adding anything. In fact I suspect your could do:
` TokenTest(<[u8; 32]>::from(&runtime.vm.caller[..]), true)`

This (and below) is not testing anything different from what we've tested above.

seanyoung (approved):

tests/substrate_tests/builtins.rs
LucasSte: This test was a suggestion of [this](https://github.com/hyperledger-labs/solang/issues/679#issuecomment-1060802286) comment, so I thought it would be interesting to add.

LucasSte: `<[u8; 32]>::from(&runtime.vm.caller[..])` doesn't work because `the trait `From<&[u8]>` is not implemented for `[u8; 32]``

seanyoung: I guess that should be `TokenTest(<[u8; 32]>::from(&runtime.vm.caller), true)` then

seanyoung: either way, nothing wrong with adding extra tests :smiley:

LucasSte: > I guess that should be `TokenTest(<[u8; 32]>::from(&runtime.vm.caller), true)` then

That doesn't work either because ``` the trait `From<&[u8; 32]>` is not implemented for `[u8; 32]` ```

PR #691, opened 2022-03-09
Resolve assembly expressions

LucasSte:
This PR implements the semantic analysis of YUL expressions. As we don't analyze YUL statements yet, there are workarounds to avoid warnings during build. There are also some `TODOs` in the code to indicate implementation details I must not forget when creating the analysis of statements.

I added most unity tests, however, a few were too difficult to create, because they required building a complex parse tree. I am going to test those functions as soon as the semantic analysis is complete for YUL.


src/sema/assembly/builtin.rs
```diff
@@ -0,0 +1,654 @@
...
+        let index = self as usize;
+        &ASSEMBLY_BUILTIN[index]
+    }
+}
+
+// Assembly built-in functions.
+// Descriptions copied and slightly modified from: https://docs.soliditylang.org/en/v0.8.12/yul.html
+static ASSEMBLY_BUILTIN: Lazy<[AssemblyBuiltinPrototype; 76]> = Lazy::new(|| {
+    [
+        AssemblyBuiltinPrototype{
```
seanyoung, line 197: So the `AssemblyBuiltInFunction` is an index into this array. If there is anything misplaced, then we will get strange results. I would suggest adding a field to `AssemblyBuiltinPrototype` of type `AssemblyBuiltInFunction`. That way you can do an `assert_eq!()` or `debug_assert_eq!()` to make sure there nothing off in the array. It does cost a tiny amount of memory of course.
What do you think?

src/sema/assembly/expression.rs
```diff
@@ -207,3 +272,427 @@ fn resolve_string_literal(
...
+    symtable: &Symtable,
+    ns: &mut Namespace,
+) -> Result<AssemblyExpression, ()> {
+    if func_call.id.name.starts_with("verbatim") {
+        ns.diagnostics.push(Diagnostic::error(
+            func_call.id.loc,
+            "verbatim functions are not yet supported in Solang".to_string(),
+        ));
+        return Err(());
+    } else if assembly_unsupported_builtin(&func_call.id.name[..]) {
```
seanyoung, line 364: This could be written as `assembly_unsupported_builtin(func_call.id.name.as_str())`. I guess this a question of taste though.

src/sema/assembly/expression.rs
```diff
@@ -207,3 +272,427 @@ fn resolve_string_literal(
...
+        ns.diagnostics.push(Diagnostic::error(
+            func_call.id.loc,
+            format!(
+                "'{}' is an internal EVM built-in not yet supported",
```
seanyoung, line 368: I would rephrase this `"the internal EVM built-in ‘{}’ is not yet supported"`

src/sema/assembly/expression.rs
```diff
@@ -207,3 +272,427 @@ fn resolve_string_literal(
...
+                    prototype.no_args,
+                    func_call.arguments.len()
+                ),
+                notes: vec![],
+            });
+            return Err(());
+        }
+
+        let default_builtin_parameter = AssemblyFunctionParameter {
+            loc: Loc::File(0, 0, 0),
```
seanyoung, line 407: Loc::Builtin

src/sema/assembly/expression.rs
```diff
@@ -207,3 +272,427 @@ fn resolve_string_literal(
...
+            name: Identifier {
+                loc: Loc::File(0, 0, 0),
```
seanyoung, line 409: Loc::Builtin

seanyoung (changes requested):
Looks great! Just some minor things.

src/sema/assembly/builtin.rs
```diff
@@ -0,0 +1,654 @@
...
+impl AssemblyBuiltInFunction {
+    pub(crate) fn get_prototype_info(self) -> &'static AssemblyBuiltinPrototype {
+        let index = self as usize;
+        &ASSEMBLY_BUILTIN[index]
+    }
+}
+
+// Assembly built-in functions.
+// Descriptions copied and slightly modified from: https://docs.soliditylang.org/en/v0.8.12/yul.html
+static ASSEMBLY_BUILTIN: Lazy<[AssemblyBuiltinPrototype; 76]> = Lazy::new(|| {
```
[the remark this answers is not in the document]
LucasSte: I wasn't sure if that is really necessary. I saw that you used it [here](https://github.com/hyperledger-labs/solang/blob/6f56cd058b3596a7dd9879864c6569b0e544e4ee/src/sema/builtin.rs#L27) and wanted to follow your pattern.

src/sema/assembly/builtin.rs
LucasSte: That is a good idea. Another way of solving this would be creating a direct mapping between the keyword and the built-in prototype. This way, I can carry only static references to such item throughout the code. That will save me memory and I can eliminate the enum. What do you think of this approach?

src/sema/assembly/builtin.rs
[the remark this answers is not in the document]
seanyoung: Here https://github.com/hyperledger-labs/solang/blob/6f56cd058b3596a7dd9879864c6569b0e544e4ee/src/sema/builtin.rs#L477 allocations happen. So, that code is not possible without lazy (or rather, I couldn't figure out how to do this without lazy).

This case should be possible.

src/sema/assembly/builtin.rs
seanyoung: This is a nice solution. You may still want to have the extra field, so we can do codegen dependant on the builtin type.

seanyoung: Only if needed of course.

LucasSte: Either way, we need check to verify that. So I'll follow your suggestion.

seanyoung (approved):

PR #692, opened 2022-03-10
Pushing onto two different arrays causes many reallocations

seanyoung:
This fixes an issue raised on discord:

https://discord.com/channels/905194001349627914/905834552965103638/949343674272219226

On Solana, contract storage in storage in account data which is a
continious piece of memory. In order to support growable objects like
dynamic arrays, we have a mini heap implementation. This fixes a problem
in the realloc implementation.


LucasSte (approved):
Please, provide a little description on the PR's summary next time, so that we can have more context on what you're trying to achieve here.

seanyoung:
Yes, you are right. I have added an explanation to the commit and the pr description.

PR #694, opened 2022-03-10
Include trait for loc() functions

LucasSte:
This PR adds traits for `fn loc(&self) -> Loc` and `fn loc(&self) -> Option<Loc>`, so that we can maintain more uniformity throughout the code, while creating the loc() functions for many structs.

### DIFF
--- a/solang-parser/src/lexer.rs
+++ b/solang-parser/src/lexer.rs
@@ -9,7 +9,7 @@ use std::iter::Peekable;
 use std::str::CharIndices;
 use unicode_xid::UnicodeXID;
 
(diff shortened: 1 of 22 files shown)


seanyoung (changes requested):
What about `Function`, `Contract`, `Parameter`, etc?

LucasSte:
> What about `Function`, `Contract`, `Parameter`, etc?

Good point. They have a public `loc` attribute, so we actually don't use its accessor corresponding function. Do you think it is a good practice to implement the trait for them as well?

PR #695, opened 2022-03-11
Fix issue with dead storage eliminating the wrong storage load

seanyoung:
When the same variable loaded from storage twice in a loop, then first
may be eliminated rather than the second. The issue is that the reaching
definition for the second load reaches the first; the possible
undefined value from the entry point of the function is not reaching.

To fix this, for block 0, populate the variable table all undefined
values.

### DIFF
--- a/src/codegen/dead_storage.rs
+++ b/src/codegen/dead_storage.rs
@@ -5,15 +5,29 @@ use std::collections::{HashMap, HashSet};
 use std::fmt;
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
-struct InstrDef {
-    pub block_no: usize,
-    pub instr_no: usize,
+enum Definition {
+    Undefined,
+    Instr { block_no: usize, instr_no: usize },
 }
+
+impl fmt::Display for Definition {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Definition::Undefined => {
+                write!(f, "undef")
+            }
+            Definition::Instr { block_no, instr_no } => {
+                write!(f, "({}, {})", block_no, instr_no)
+            }
+        }
+    }
+}
+
 #[allow(clippy::large_enum_variant)]
 #[derive(Clone, PartialEq)]
 enum Transfer {
     Gen {
-        def: InstrDef,
+        def: Definition,
         var_no: usize,
     },
     Copy {
@@ -24,7 +38,7 @@ enum Transfer {
         var_no: usize,
     },
     Store {
-        def: InstrDef,
+        def: Definition,
         expr: Option<Expression>,
     },
 }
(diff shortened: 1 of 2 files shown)


src/codegen/dead_storage.rs
```diff
@@ -92,8 +102,20 @@ fn reaching_definitions(cfg: &mut ControlFlowGraph) -> (Vec<Vec<Vec<Transfer>>>,
...
+            assert_eq!(block_no, 0);
+
+            let mut vars: HashMap<usize, HashMap<InstrDef, Option<Expression>>> = HashMap::new();
+
+            for var_no in cfg.vars.keys() {
+                let mut defs = HashMap::new();
+
+                defs.insert(InstrDef::Undefined, None);
+
+                vars.insert(*var_no, defs);
```
LucasSte, line 114: Why don't `vars.insert(*var_no, HashMap::new())`?

src/codegen/dead_storage.rs
```diff
@@ -580,31 +604,43 @@ fn get_storage_definition<'a>(
...
+            ReachingDefs {
+                vars: HashMap::new(),
+                stores: Vec::new(),
+            }
+        };
 
-    vars
+        vars
+    } else {
+        unreachable!();
```
LucasSte, line 642: Please, leave a message inside `unreachable`, explaining why no flow should read this line.

tests/solana_tests/simple.rs
```diff
@@ -280,3 +280,29 @@ fn two_arrays() {
...
+
+            constructor() {
+                for(uint64 i = 0; i < 10; i++) {
+                    print("bitches0");
+                    uint x = i*10e34;
+                    print("bitches1");
+                    v = x%maxlen;
+                    print("bitches2");
+                    z =((0)%maxlen);
+                    print("bitches3");
```
LucasSte, line 301: I see that dogs inspired you today 🐕‍🦺🦮🐶🐩

LucasSte (changes requested):

src/codegen/dead_storage.rs
seanyoung: That won't work because we want a HashMap with a single `Undefined` entry. I don't think there is a macro for hashmap like there is for `Vec`: `vec![1]`.

src/codegen/dead_storage.rs
seanyoung: That is a good point. I've added some words,

tests/solana_tests/simple.rs
seanyoung: ah that was not supposed to be in the PR 😳 

I've improved the test somewhat, now it tests more than just success.

src/codegen/dead_storage.rs
LucasSte: Why not `unreachable!("description")`?

seanyoung: Another good point. Updated.

LucasSte (approved):

PR #700, opened 2022-03-14
feat(parser): support revert statement

### LINKED ISSUE #696: custom error reverts unsupported by solang-parser
Stumbled upon this when looking at this for example https://github.com/withtally/rollcall/blob/main/src/Executor.sol 

which solang-parser fails to parse at the parentheses 

https://github.com/withtally/rollcall/blob/6ffdcffc41ba163a83a93f1b03c97815b6e49fa4/src/Executor.sol#L28

I couldn't find where "revert" is implemented so would need some pointers where to modify lalrpop rules for this before I can try fixing this

mattsse:
Closes #696

Followed these instructions https://github.com/hyperledger-labs/solang/issues/696#issuecomment-1066666159

revert statement remains `unimplemented!` in `sema`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ itertools = "0.10"
 num-rational = "0.4"
 indexmap = "1.8"
 once_cell = "1.9"
-solang-parser = { path = "solang-parser", version = "0.1.2" }
+solang-parser = { path = "solang-parser", version = "0.1.11"}
 codespan-reporting = "0.11"
 phf = "0.10.1"
 
--- a/solang-parser/Cargo.toml
+++ b/solang-parser/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "solang-parser"
-version = "0.1.10"
+version = "0.1.11"
 authors = ["Sean Young <sean@mess.org>"]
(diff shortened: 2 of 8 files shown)


seanyoung:
My suggested parser changes did not work. I'm trying to see what changes are needed, let me get back to you

seanyoung:
@mattsse I don't think I can add commits to your PR, but here is a commit which fixes the grammar: https://github.com/seanyoung/solang/commit/16ee98629f67c94adfab6ba5970832f65f3d415c

mattsse:
Thanks so much!
https://github.com/seanyoung/solang/commit/16ee98629f67c94adfab6ba5970832f65f3d415c appended

seanyoung:
There is also an issue in the solang-parser test, I had not tested that, this should fix that https://github.com/seanyoung/solang/commit/787515bf9d3f4d1a3f5fed3d863f6e784a00f932

seanyoung:
The vscode extension needs a small fix for the test to pass, in vscode/src/test/suite/extension.test.ts

seanyoung (approved):

seanyoung:
Looks great, I'll fixup the vscode test in a follow up PR

PR #702, opened 2022-03-17
Yul semantic analysis

LucasSte:
This PR implements the semantic analysis for YUL statements.

Two contract test cases have been removed, because the unused variable detection and the graphviz for YUL are not implemented yet, so the tests weren't passing. I'll bring back these tests in another PR when I implement the missing features.


seanyoung (commented):
I've read all the way through, looks great!

function could do with some doc comments.

src/sema/assembly/functions.rs
```diff
@@ -1,23 +1,279 @@
...
+                ns,
+                VariableInitializer::Assembly(true),
+                VariableUsage::AssemblyLocalVariable,
+                None,
+            )
+            .is_none()
+        {
+            ns.diagnostics.push(Diagnostic::error(
+                item.id.loc,
+                format!("variable name '{}' already taken", item.id.name),
```
seanyoung, line 232: This is not a good error message. Also what is the location of the previous definition?

src/sema/assembly/functions.rs
```diff
@@ -1,23 +1,279 @@
...
+                ns,
+                VariableInitializer::Assembly(false),
+                VariableUsage::AssemblyLocalVariable,
+                None,
+            )
+            .is_none()
+        {
+            ns.diagnostics.push(Diagnostic::error(
+                item.id.loc,
+                format!("variable name '{}' already taken", item.id.name),
```
seanyoung, line 252: This is not a good error message. Also what is the location of the previous definition?

seanyoung (changes requested):
Some of error messages for expressions are not great.

```
 "singed value cannot fit in unsigned type"
```
This should repeat the value and the type. Also note the "singed".

`suppoerted` should be `supported `

`"{}-bit` should be `"{} bit` (to be consistent with the other error messages).

src/sema/assembly/expression.rs
```diff
@@ -173,7 +174,7 @@ fn resolve_number_literal(
                     pos: *loc,
                     level: Level::Error,
                     ty: ErrorType::TypeError,
-                    message: "singed value cannot fit in unsigned type".to_string(),
+                    message: "singed integer cannot fit in unsigned integer".to_string(),
```
seanyoung, line 177: signed

seanyoung (approved):
Please fix typo before merging

PR #708, opened 2022-03-23
Add yul to unused variable detection

LucasSte:
This PR integrates the YUL semantic analysis to the unused variable detection algorithm. I also remove repeated and unused code from the `sema/unsused_variable.rs` file.


src/sema/assembly/expression.rs
```diff
@@ -640,7 +639,101 @@ fn resolve_member_access(
...
+            AssemblyExpression::SolidityLocalVariable(
+                _,
+                _,
+                Some(StorageLocation::Storage(_)),
+                ..,
+            )
+            | AssemblyExpression::StorageVariable(..) => {
+                return Some(Diagnostic::error(
+                    expr.loc(),
+                    "storage variables cannot be assigned any value in assembly. You may use \"sstore()\"".to_string()
```
seanyoung, line 659: We have inconsistent quoting in solang error messages. I think we should always use ‘foo’. However some messages use 'foo' (which should be fixed) and now this introduces "foo".

src/sema/assembly/expression.rs
```diff
@@ -640,7 +639,101 @@ fn resolve_member_access(
...
+        AssemblyExpression::SolidityLocalVariable(
+            _,
+            Type::Array(_, ref dims),
+            Some(StorageLocation::Calldata(_)),
+            ..,
+        ) => {
+            if dims[0].is_none() {
+                return Some(Diagnostic::error(
+                    expr.loc(),
+                    "Calldata arrays must be accessed with \".offset\", \".length\" and the \"calldatacopy\" function".to_string()
```
seanyoung, line 729: quotes

src/sema/unused_variable.rs
```diff
@@ -347,20 +330,31 @@ pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Diag
...
 
             None
         }
         VariableUsage::AssemblyLocalVariable => {
-            //unimplemented!("Variable usage not implemented for YUL")
+            let has_value = variable.assigned || variable.initializer.has_initializer();
+            if !variable.read && !has_value {
+                return Some(Diagnostic::warning(
+                    variable.id.loc,
+                    "assembly variable has never been read or assigned".to_string(),
```
seanyoung, line 349: Please do include the variable name. This is consistent with the other unused messages, and also I think explicitly repeating the variable name is useful. In an ide for example you see the diagnostics in a separate window. Calling the variable out there is good.

Also note this is what gcc/clang do as well.

src/sema/unused_variable.rs
```diff
@@ -347,20 +330,31 @@ pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Diag
...
+                ));
+            } else if !variable.read {
+                return Some(Diagnostic::warning(
+                    variable.id.loc,
+                    "assembly variable has never been read".to_string(),
```
seanyoung, line 354: Same as above

src/sema/unused_variable.rs
```diff
@@ -370,19 +364,17 @@ pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Diag
...
             ),
-            vec![],
         ));
     } else if !variable.assigned && !variable.read {
-        return Some(generate_unused_warning(
+        return Some(Diagnostic::warning(
             variable.loc,
-            &format!("storage variable '{}' has never been used", variable.name),
-            vec![],
+            format!("storage variable '{}' has never been used", variable.name),
```
seanyoung, line 377: Ideally quoting here should be fixed as well

src/sema/unused_variable.rs
```diff
@@ -370,19 +364,17 @@ pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Diag
 /// Emit warnings depending on the storage variable usage
 fn emit_warning_contract_variables(variable: &ast::Variable) -> Option<Diagnostic> {
     if variable.assigned && !variable.read {
-        return Some(generate_unused_warning(
+        return Some(Diagnostic::warning(
             variable.loc,
-            &format!(
+            format!(
                 "storage variable '{}' has been assigned, but never read",
```
seanyoung, line 370: ideally quoting here should be fixed as well

src/sema/unused_variable.rs
```diff
@@ -402,14 +394,29 @@ pub fn check_unused_namespace_variables(ns: &mut Namespace) {
...
+    // Global constants should have been initialized during declaration
     for constant in &ns.constants {
         if !constant.read {
-            ns.diagnostics.push(generate_unused_warning(
+            ns.diagnostics.push(Diagnostic::warning(
                 constant.loc,
-                &format!("global constant '{}' has never been used", constant.name),
-                vec![],
-            ))
+                format!("global constant '{}' has never been used", constant.name),
```
seanyoung, line 402: ideally quoting here should be fixed as well

src/sema/unused_variable.rs
```diff
@@ -481,10 +475,9 @@ pub fn check_unused_events(ns: &mut Namespace) {
                 }
             }
 
-            ns.diagnostics.push(generate_unused_warning(
+            ns.diagnostics.push(Diagnostic::warning(
                 event.loc,
-                &format!("event '{}' has never been emitted", event.name),
-                vec![],
+                format!("event '{}' has never been emitted", event.name),
```
seanyoung, line 480: ideally quoting here should be fixed as well

seanyoung (changes requested):
Sorry about being pedantic

src/sema/assembly/expression.rs
```diff
@@ -640,7 +639,101 @@ fn resolve_member_access(
...
+    } else {
+        used_variable(ns, expr, symtable);
+    }
+
+    match expr {
+        AssemblyExpression::SolidityLocalVariable(_, _, Some(StorageLocation::Storage(_)), ..)
+        | AssemblyExpression::StorageVariable(..) => {
+            return Some(Diagnostic::error(
+                expr.loc(),
+                "Storage variables must be accessed with \".slot\" or \".offset\"".to_string(),
```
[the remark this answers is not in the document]
LucasSte: Thanks for noticing. I'll fix the quotes. Actually this code appeared by mistake during rebase. It it exactly
the same thing you see a few lines above.

src/sema/unused_variable.rs
LucasSte: I've changed all the occurrences of `'{}'` in the file to ``` ´{}´ ```

seanyoung (changes requested):
I meant curly quotes like `‘foo’`, see https://github.com/hyperledger-labs/solang/blob/main/src/sema/contracts.rs#L114 for example.

PR #710, opened 2022-03-24
Detect unused functions in inline assembly

LucasSte:
This PR implements an algorithm to detect unused functions in YUL. I've included tests as well.

There was some refactoring due to bugs I found along the way.

### DIFF
--- a/src/sema/assembly/block.rs
+++ b/src/sema/assembly/block.rs
@@ -1,5 +1,7 @@
 use crate::ast::Namespace;
-use crate::sema::assembly::functions::{process_function_header, FunctionsTable};
+use crate::sema::assembly::functions::{
+    process_function_header, resolve_function_definition, FunctionsTable,
+};
 use crate::sema::assembly::statements::{resolve_assembly_statement, AssemblyStatement};
 use crate::sema::expression::ExprContext;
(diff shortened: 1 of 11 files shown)


src/sema/assembly/block.rs
```diff
@@ -79,7 +91,11 @@ pub(crate) fn process_statements(
             ns,
         ) {
             Ok(can_reach_next_statement) => {
-                if !reachable && !has_unreachable {
+                // There shouldn't be warnings of unreachable statements for function definitions
```
seanyoung, line 94: I don't understand this comment. Could you elaborate the comment please?

seanyoung (approved):

src/sema/assembly/block.rs
LucasSte: Take this example:
```
let x := foo(1, 2)
return(x, 2)
function foo(a, b) -> ret {
   ret := add(a, b)
}
```
Every statement after `return` is unreachable. Although the function definition is also a statement, it cannot be unreachable because it is not a statement that executes something. In addition, the function was invoked, so it is not dead code.

seanyoung: Ok, that makes sense. I had not thought of that. Can you you add that to the comment in the code please?

PR #716, opened 2022-03-28
Integrate inline assembly to graphviz

LucasSte:
This PR integrates the YUL AST to the graphviz visualization tool. I also fixed a bug in which Solang was panicking when we used `--emit ast-dot` with an invalid contact.

Please, feel free to plot the dot files I added to the tests to check they are OK.


src/bin/solang.rs
```diff
@@ -470,6 +470,11 @@ fn process_file(
             eprintln!("{}: error: {}", dot_filename.display(), err);
             std::process::exit(1);
         }
+
+        if ns.contracts.is_empty() || diagnostics::any_errors(&ns.diagnostics) {
+            return Err(());
+        }
+
```
seanyoung, line 477: That is not the right solution, as if when you run `solang --target solana --emit ast-dot foo.sol` and foo.sol is valid, a binary file is still created (`bundle.so` and abi files)

src/sema/assembly/builtin.rs
```diff
@@ -194,6 +194,13 @@ impl AssemblyBuiltInFunction {
     }
 }
 
+impl ToString for AssemblyBuiltInFunction {
+    fn to_string(&self) -> String {
+        let prototype = self.get_prototype_info();
+        prototype.name.parse().unwrap()
+    }
+}
+
```
seanyoung, line 203: Why are you using `.parse().unwrap()` rather than `.to_owned()`?
Why implement `ToString` rather than `Display`? The former does a copy/malloc.

src/sema/assembly/mod.rs
```diff
@@ -6,12 +6,12 @@ use crate::sema::expression::ExprContext;
...
-mod builtin;
-mod expression;
+pub mod block;
+pub mod builtin;
+pub mod expression;
 mod for_loop;
-mod functions;
-mod statements;
+pub mod functions;
+pub mod statements;
```
seanyoung, line 14: You're having to make these modules `pub` because the ast definition is scattered over sema now. The original ast had a definition in a single file, `ast.rs`.

seanyoung (changes requested):
Support for assembly seems to be called `assembly` in some places and `yul` in other places. We should probably settle on one name.

src/bin/solang.rs
LucasSte: Does this problem only happen with the Solana target?

seanyoung: I think so, you could try it out.

src/sema/assembly/builtin.rs
LucasSte: To maintain coherence with other constructs in Solang, I implemented the `to_string` functions. It is also implemented for `Type` and `Loc`.
Returning a String as in `fn to_string(&self) -> String` doesn't do a malloc. The compiler either moves the result or builds the value in place at the line of calling.
What advantages do you see for `Display` over `ToString`?

src/sema/assembly/mod.rs
LucasSte: The AST for YUL is scattered across many files in `sema/assembly`. I can either create a new file `sema/assembly/ast.rs` to concentrate every AST structure or I can move them all to `sema/ast.rs`. Do you have any preference?

src/bin/solang.rs
seanyoung: Sorry, I think I am mistaken. This does not happen on solana either.

However, when you run `solang --target solana --emit ast-doc foo.sol` and there are compiler errors, then you both get a dot file and the message `error: not all contracts are valid` and non-zero exit code. This suggests that the command did not work, but it did and you do have a .dot file

tests/contract_testcases/solana/assembly/function_cal_cond.dot
```diff
@@ -0,0 +1,152 @@
...
+	yul_function_call_66 [label="yul function call ‘foo‘\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:20:36-39"]
+	yul_number_literal_67 [label="uint256 literal: 3\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:20:40-41"]
+	yul_number_literal_68 [label="uint256 literal: 2\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:20:44-45"]
+	yul_number_literal_69 [label="uint256 literal: 4\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:20:48-49"]
+	yul_var_decl [label="yul variable declaration\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:31:17-27"]
+	var_decl_item [label="yul variable declaration uint256 x\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:31:21-22"]
+	yul_number_literal_72 [label="uint256 literal: 5\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:31:26-27"]
+	diagnostic [label="found contract ‘testTypes’\nlevel Debug\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:1:1-20"]
+	diagnostic_75 [label="evm assembly not supported on target solana\nlevel Error\ntests/contract_testcases/solana/assembly/function_cal_cond.sol:3:9-33:10"]
```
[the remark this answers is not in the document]
LucasSte: Even though the sema for inline assembly is mostly complete, we still don't support the code generation, so I left there the same error message you wrote. When I start working on code generation, I'll move that error to another place.

src/bin/solang.rs
seanyoung: I think the issue is that `solang::compile_many()` is called when there are compiler errors for the solana target

tests/contract_testcases/solana/assembly/function_cal_cond.dot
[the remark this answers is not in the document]
seanyoung: ok, makes sense!

src/bin/solang.rs
LucasSte: I think that the issue is that we don't return errors from the function `process_file` when we emit the `ast-dot`.
Have a look here:
https://github.com/hyperledger-labs/solang/blob/a524ff58f5cc033b849195c36cb0e07e06fe05fb/src/bin/solang.rs#L473
As we always return `Ok`, we will never exit the process on the following condition:
https://github.com/hyperledger-labs/solang/blob/a524ff58f5cc033b849195c36cb0e07e06fe05fb/src/bin/solang.rs#L336

LucasSte: When I add the checks above, I return an error from `process_file` and the `solang::compile_many()` won't be called.

seanyoung: The problem with that is that the behaviour of the solang command line is a bit in-consistent with your change:

```
[sean@xywoleh solang]$ cargo run foo.sol --target solana --emit ast-dot -v
   Compiling solang v0.1.10 (/home/sean/git/solang)
    Finished dev [unoptimized + debuginfo] target(s) in 26.59s
     Running `target/debug/solang foo.sol --target solana --emit ast-dot -v`
info: Solang version v0.1.10-20-g51904a31
help: found contract ‘c’
  ┌─ /home/sean/git/solang/foo.sol:1:1
  │
1 │ contract c {
  │ ^^^^^^^^^^^

error: function cannot have same name as the contract
  ┌─ /home/sean/git/solang/foo.sol:2:2
  │
2 │     function c() public {}
  │     ^^^^^^^^^^^^^^^^^^^

info: Saving graphviz dot ./foo.dot
[sean@xywoleh solang]$ echo $?
1
```

So now it returns an error exit code but it did actually save the dot file. That is not behaviour I would expect. Without the `-v` option it is even worse: you have no idea that it succeeded in writing the dot file.

LucasSte: What do you suggest then?

seanyoung: Change line 346 to:
```
if target == solang::Target::Solana && !diagnostics::any_errors(&ns.diagnostics) {
```

seanyoung (approved):

LucasSte:
> Support for assembly seems to be called `assembly` in some places and `yul` in other places. We should probably settle on one name.

I'll change names in another PR.

PR #719, opened 2022-03-30
Rename assembly to yul in Solang

LucasSte:
This PR refactors Solang Parser. We decided to adopt the name "YUL" for everything that refers to Solidity assembly, so we are changing the nomenclature everywhere.


src/sema/assembly/block.rs
```diff
@@ -17,7 +17,7 @@ pub struct AssemblyBlock {
 /// Returns the resolved block and a boolean that tells us if the next statement is reachable.
```
seanyoung, line 17: The module is still called `assembly` (the one in sema)

seanyoung (changes requested):

src/sema/assembly/block.rs
LucasSte: >  Changes to the sema will be made in another PR to avoid development conflicts.

@seanyoung

seanyoung: Ah. Why multiple PRs? Because you're still working on sema in other PRs?

Ideally PRs should be self-contained, complete pieces of work.

LucasSte: Ok, I'll change that after merging the other PRs.

LucasSte: Ready for review now!

LucasSte: Ready for review now!

seanyoung (approved):

PR #722, opened 2022-04-01
Implement mutability check for yul

LucasSte:
This PR implements mutability checks for YUL. We check if a pure function reads or writes the state and if a view function writes the state. I've provided tests as well.


src/codegen/mod.rs
```diff
@@ -87,8 +87,8 @@ impl Default for Options {
     }
 }
 
-/// The contracts are fully resolved but they do not have any a CFG which is needed for
-/// the llvm code emitter. This will also do addition code checks.
+/// The contracts are fully resolved but they do not have a CFG which is needed for
```
seanyoung, line 90: Each contract has multilple CFGs, one per function, constructor and one for the storage initializer. So "a CFG" is not correct, it should be "any CFGs"/"the CFGs" or something like that.

seanyoung (changes requested):
Tiny nitpick, sorry

seanyoung (approved):

PR #725, opened 2022-04-04
chore: add missing clone derives

mattsse:


mattsse:
I'm using some types for `solc` preprocessing in [ethers-solc](https://github.com/gakonst/ethers-rs/tree/master/ethers-solc), especially `FunctionDefinition` which was lacking `Clone` and didn't play nice with existing types that are `Clone`.

I can totally live without this PR, but thought the missing derive was missing rather than deliberately omitted.
feel free to close this if that's the case.

seanyoung:
I agree they were missed rather than omitted on purpose. When structures are `pub` in a crate, they should really have `Clone` derived. I think this was even mentioned in [rust for rustaceans](https://rust-for-rustaceans.com/).

seanyoung (approved):

PR #726, opened 2022-04-06
Refactoring in sema for yul and solidity compatibility

LucasSte:
This PR implements the following refactoring to improve compatibility between Solidity Functions and Yul functions so that we can share the same CFG structures for them both.

1. Parameters are now saved in an `Arc<Vec<Parameter>>>`, because they are shared between `Function`/`YulFunction`, `FunctionsTable` and `ControlFlowGraph`. This way we avoid unnecessary clones from all the elements of a vector.
2. `llvm_symbol` is now under a trait, so that we can implement the same interface for both `Function` and `YulFunction`.
3. Yul Functions are processed in `external_functions`, so that we know which functions should be laid out at each contract.
4. `ty_loc` is now `Option<Loc>` in `Parameter`, because Yul Function parameters might not have a specified type.
5. Yul functions now use `Parameter` instead of `YulFunctionParameter` for saving parameters.
6. The function number in `ControlFlowGraph` is now an enum to distinguish between Yul Functions and Solidity Functions.
7. All Yul functions are centralized in `Namespace`. A Yul function number is its respective index in `Namespace::yul_functions`. Likewise, `FunctionsTable` in `sema/yul/functions.rs` saves such an index and works with an offset for its lookup vector.


src/codegen/cfg.rs
```diff
@@ -336,13 +337,20 @@ pub struct ControlFlowGraph {
     current: usize,
 }
 
+#[derive(Clone)]
+pub enum FunctionNumber {
+    SolidityFunction(usize),
+    YulFunction(usize),
+    None,
+}
```
seanyoung, line 345: The `FunctionNumber` is a nice idea, although I am not keen on the name. This is a reference to the AST, so how about `ASTFunction`?

src/sema/yul/ast.rs
```diff
@@ -10,7 +10,8 @@ use std::sync::Arc;
 pub struct InlineAssembly {
     pub loc: pt::Loc,
     pub body: Vec<YulStatement>,
-    pub functions: Vec<YulFunction>,
+    // (begin, end) offset for Namespace::yul_functions
+    pub functions: (usize, usize),
```
seanyoung, line 14: This looks like a open-coded version of `std::ops::Range<usize>`.

src/codegen/cfg.rs
```diff
@@ -324,9 +325,9 @@ impl BasicBlock {
 #[derive(Clone)]
 pub struct ControlFlowGraph {
     pub name: String,
-    pub function_no: Option<usize>,
-    pub params: Vec<Parameter>,
-    pub returns: Vec<Parameter>,
+    pub function_no: FunctionNumber,
+    pub params: Arc<Vec<Parameter>>,
+    pub returns: Arc<Vec<Parameter>>,
```
seanyoung, line 330: You may avoid the cost of the malloc/memcpy, but are introducing atomic instructions which come with memory barriers for each iteration. I don't know whether this is faster or `clone()` is faster. Up to you

seanyoung (changes requested):

src/codegen/cfg.rs
seanyoung: I do wonder if we need a `DefaultConstructor`, `StorageInitializer`, `Modifier` field. Not sure what use that would be though

src/codegen/cfg.rs
seanyoung: Another way of doing this is by keeping a reference to the parameters in the AST, if possible. We do want to add diagnostics in codegen so we can't make the namespace read-only by keeping a reference. Having said that, we would never modify a functions parameters/returns in codegen.

src/codegen/cfg.rs
LucasSte: > The `FunctionNumber` is a nice idea, although I am not keen on the name. This is a reference to the AST, so how about `ASTFunction`?

Currently, there is an `Option<usize>` at the CFG funtion_no, so we don't distinguish between Solidity function and AST Function. As Yul function also have a number, we must differentiate yul function numbers and Solidity function numbers, as they represent indexes to different vectors in NS.

LucasSte: Any suggestions for the name?

src/codegen/cfg.rs
LucasSte: I'd rather use a `Rc`, however, the language server is asynchronous, so it requires atomic operations. Keep in mind that the content of an `Arc` is immutable, so the only atomic operations that happen here are incrementing or decrementing the reference counter. Hopefully, they only happen when we call `clone()` and when an `Arc` goes out of scope. There is no race condition from multiple threads reading the elements of an `Arc`. In addition, `Arc` may use atomic operations from the hardware. x86, for instance, [implements some](https://stackoverflow.com/questions/38447226/atomicity-on-x86/38465341#38465341), so I don't expect `Arc` to be slow.

LucasSte: > Another way of doing this is by keeping a reference to the parameters in the AST, if possible. We do want to add diagnostics in codegen so we can't make the namespace read-only by keeping a reference. Having said that, we would never modify a functions parameters/returns in codegen.

That's is an option too, but the reference keeper has the same parent as the object the reference points to. I think this is unsafe, due to this answer, you can find on [this thread](https://www.reddit.com/r/rust/comments/dymc8f/selfreference_struct_how_to/):

> Self referential structs are inherently unsafe since they are implicitly invalidated if they are ever moved. This is why the Rust type system doesn't allow this in safe Rust. I will check on the code and get back to you.

seanyoung: > I'd rather use a `Rc`, however, the language server is asynchronous, so it requires atomic operations. Keep in mind that the content of an `Arc` is immutable, so the only atomic operations that happen here are incrementing or decrementing the reference counter. Hopefully, they only happen when we call `clone()` and when an `Arc` goes out of scope. There is no race condition arising from multiple threads reading the elements of an `Arc`. In addition, `Arc` may use atomic operations from the hardware. x86, for instance, [implements some](https://stackoverflow.com/questions/38447226/atomicity-on-x86/38465341#38465341), so I don't expect `Arc` to be slow.

atomic operations have to implemented in hardware, else they cannot be atomic. atomic instructions are quite expensive due inter cpu locking protocols.

seanyoung: > > Another way of doing this is by keeping a reference to the parameters in the AST, if possible. We do want to add diagnostics in codegen so we can't make the namespace read-only by keeping a reference. Having said that, we would never modify a functions parameters/returns in codegen.
> 
> That's is an option too, but the reference keeper has the same parent as the object the reference points to. I think this is unsafe, due to this answer you can find on [this thread](https://www.reddit.com/r/rust/comments/dymc8f/selfreference_struct_how_to/):

I thought the original is in `Namespace` and the reference would be in `ControlFlowGraph`. That's not the same object.

src/codegen/cfg.rs
LucasSte: Sounds good.

PR #727, opened 2022-04-08
Remove double call to expression in destructure codegen

LucasSte:
This PR is a quick fix for calling `expression` twice when doing code generation for restructure statements.
Instead of calling `expression`, I call `storage_load` to create the `LoadStorage` instruction in the CFG.

### DIFF
--- a/src/codegen/statements.rs
+++ b/src/codegen/statements.rs
@@ -7,6 +7,7 @@ use super::{
     cfg::{ControlFlowGraph, Instr},
     vartable::Vartable,
 };
+use crate::codegen::expression::load_storage;
 use crate::codegen::unused_variable::{
     should_remove_assignment, should_remove_variable, SideEffectsCheckParameters,
 };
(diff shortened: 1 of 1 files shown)


seanyoung (approved):
The change works, but it's just as ugly as the original (albeit less broken).

LucasSte:
> The change works, but it's just as ugly as the original (albeit less broken).

Do you have any ideas on possible changes that are less ugly?

seanyoung:
> > The change works, but it's just as ugly as the original (albeit less broken).
> 
> Do you have any ideas on possible changes that are less ugly?

I've been thinking about this and nothing has come to mind. Do you think that your new version of `fn cast()` for codegen could deal with this?

LucasSte:
> > > The change works, but it's just as ugly as the original (albeit less broken).
> > 
> > 
> > Do you have any ideas on possible changes that are less ugly?
> 
> I've been thinking about this and nothing has come to mind. Do you think that your new version of `fn cast()` for codegen could deal with this?

My new version of Cast works basically the same way the old one, expect that now it is a method of `cfg::Expression`. If I were to solve the `LoadStorage` issue within `cast`, `cfg` should be passed as an argument to cast. I do believe that `cast` should be self-contained, in a way that it does not depend on the CFG.

seanyoung:
The same issue exists for `return foo()` on line https://github.com/hyperledger-labs/solang/blob/5da5377b6fce0d1aec98fa58041a4126540a226d/src/codegen/statements.rs#L842-L843

seanyoung:
I think we should have:
- Separate this functionality out into a function with a big fat comment explaining why it is needed (cast not done in sema)
- Reuse this function for all three cases.

PR #735, opened 2022-04-12
Update links to discord to the correct channel

seanyoung:
The links to discord chat is to the discord server, but we can do better
than that and link to the correct channel on the server.


LucasSte (approved):
Are these links permanent?
Do you know if the links that point directly to the channel expire?

seanyoung:
I've updated them to channel invites that don't expire.

PR #736, opened 2022-04-12
More info for overload

guanqun:
fix https://github.com/hyperledger-labs/solang/issues/707

### DIFF
--- a/src/sema/expression.rs
+++ b/src/sema/expression.rs
@@ -31,6 +31,7 @@ use crate::sema::unused_variable::{
 use crate::Target;
 use base58::{FromBase58, FromBase58Error};
 use num_rational::BigRational;
+use solang_parser::diagnostics::Note;
 
 impl Expression {
     /// Return the type for this expression. This assumes the expression has a single value,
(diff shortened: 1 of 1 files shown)


guanqun:
question: how to add a unit test for UI?

seanyoung (changes requested):
That error message looks great! :1st_place_medal: My first observation is the same change needs to be repeated for method calls, method calls with named arguments, constructor calls, constructor calls with named arguments. 

> question: how to add a unit test for UI?

I think the best way to add test cases for this is to do the following:
- Add some solidity files with bad overloaded calls to the `tests/contract_tests/solana/` directory.
- Re-generate the test data by uncommenting these three lines: https://github.com/hyperledger-labs/solang/blob/main/tests/contract.rs#L80-L82
- Run `cargo test --test contract` 
- Putting comments back on the three lines https://github.com/hyperledger-labs/solang/blob/main/tests/contract.rs#L80-L82
- Inspect the newly created dot files. Do they have the correct error message in them with the notes?

guanqun:
Thanks for the comments. Will add the overloaded info for other cases.

guanqun:
Sorry I don't have enough time on this. Closing this PR.

Issue #731, opened 2022-04-11
Error in sema parsing contract


ntrippar:
I found that parsing this contract ends in a stack overflow.
```solidity
pragma solidity ^0.8.7;

contract TestContract {
    struct Data {
        bytes32 a1;
        bytes32 a2;
    }

    mapping(uint176 => bytes32) Hashes;

    constructor() {}

    function getdata(uint176 id, bool needHash) public view returns (Data memory g) {
        (g.a1, g.a2) = (Hashes[id+1], Hashes[id+2]);
    }
}
```

this seems to be when its casting the `StorageRef(false, Bytes(32))` to  `Ref(Bytes(32))`

```js
from cast: StorageRef(false, Bytes(32))
to cast: Ref(Bytes(32))
from cast: Ref(StorageRef(false, Bytes(32)))
to cast: Ref(Bytes(32))
```
based on the result that I was getting with some debug prints in the function `cast` I noticed that the function checks if the `to` needs a `Ref`, and try creates a `Ref` from the `from`.  this cause an infinite loop.

https://github.com/hyperledger-labs/solang/blob/5da5377b6fce0d1aec98fa58041a4126540a226d/src/sema/expression.rs#L540-L550

one solution that seems to be working is not dereference when the `to` object type is a `Type::StorageRef`, but as I don't have experience with this codebase I didn't create a pull request.

```rust
  // First of all, if we need a ref then derefence it
  if matches!(to, Type::Ref(..)) && !matches!(from, Type::Ref(..)) && !matches!(from, Type::StorageRef(..)) {
      return cast(
          loc,
          Expression::GetRef(*loc, Type::Ref(Box::new(from)), Box::new(expr)),
          to,
          implicit,
          ns,
          diagnostics,
      );
  }
```

ntrippar:
Also do the project has a "repo" with a lot of testcases?, if not I would recommend that the project should have one with a lot of sol files named `{github_issue_id}.sol` an a `{github_issue_id}.json` with some information that can do continuous test to check if there is any regression of a bug.  or something in broken

seanyoung:
> Also do the project has a "repo" with a lot of testcases?, if not I would recommend that the project should have one with a lot of sol files named `{github_issue_id}.sol` an a `{github_issue_id}.json` with some information that can do continuous test to check if there is any regression of a bug. or something in broken

There are a whole bunch of testcases in the tests/ directories, including over 500 sol files.

seanyoung:
[...]

PR #734, opened 2022-04-12
chore: typo fixes

guanqun:

PR #738, opened 2022-04-13
Sema test for casting destructure values is broken

seanyoung:
When the left hand side of a destructure statements is a structure member or array element, the test incorrectly assumed the value was not assignable.
    
Fixes issue #731.


tests/contract_testcases/solana/address_cast.sol
```diff
@@ -0,0 +1,8 @@
+contract c {
+	function foo() public {
+		// We have code that cast address type to ref address
+		// in fn sema::cast(). Ensure that this does not cause
+		// address values to be assignable.
```
LucasSte, line 5: >  the test incorrectly assumed the value was not assignable.

This left me confused. If the test incorrectly assumed the value was not assignable, why are you testing here that the values should not be assignable?

LucasSte:
> When the left hand side of a destructure statements is a structure member or array element, the test incorrectly assumed the value was not assignable.
> 
> Fixes issue #734.

You linked a PR here. I think you meant an issue, didn't you? Please, make sure to change that before merging the code.

tests/contract_testcases/solana/address_cast.sol
seanyoung: This PR has two commits. This is related to the first commit

seanyoung:
> You linked a PR here. I think you meant an issue, didn't you? Please, make sure to change that before merging the code.

Thanks for spotting that. Fixed.

tests/contract_testcases/solana/address_cast.sol
LucasSte: I don't really pay attention to commit messages because they are usually hidden in GitHub. Thanks for signaling that!

tests/contract_testcases/solana/issue731-2.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity 0.6.12;
```
LucasSte, line 1: How about naming the file as `casting_desctructure.sol`. It is not easy to look at your file and guess what you're testing. I think you can mention the issue in the file itself.

seanyoung: I've given them both the best names I could come with and they're still not very descriptive. The name `issue731.sol` is very clear: have a look at issue 731.

Also I don't think the file name itself is also not terribly important. Only if the test fails do you need to figure out what it is testing; and even then, the test could fail for any reason if there is a code change.

LucasSte (approved):

Issue #732, opened 2022-04-11
Error in sema type when getting bits of Ref(Uint(N))


ntrippar:
When parsing the code fails with `type not allowed Ref(Uint(N))` this is due that the function `pub fn bits(&self, ns: &Namespace) -> u16` is not handling `Ref`

https://github.com/hyperledger-labs/solang/blob/4a463652125127f42f88b151c4f28bcd8d86eaef/src/sema/types.rs#L910-L923

## Testcase

```solidity
library FixedPoint {
   
    struct uq144x112 {
        uint _x;
    }

    struct uq112x112 {
        uint224 _x;
    }

    // this errors with `type not allowed Ref(Uint(224))`
    function decode(uq112x112 memory self) internal pure returns (uint112) {
        return uint112(self._x >> 112);
    }

    // this error with `type not allowed Ref(Uint(256))`
    function decode144(uq144x112 memory self) internal pure returns (uint144) {
        return uint144(self._x >> 112);
    }
}
```

seanyoung:
Thanks for a good quality bug report!

PR #739, opened 2022-04-13
Shift on integer struct member causes panic

seanyoung:
The syntax `self._x >> 112` causes a panic because retrieving the bit
width of a referene to an integer is broken.

Fixes issue #732.


tests/contract_testcases/solana/issue732.sol
```diff
@@ -0,0 +1,20 @@
+library FixedPoint {
```
LucasSte, line 1: How about specifying what you are trying to test in the file's name?
Nobody will know what `issue732.sol` tests just by looking at the code. You can mention in the file that it is related to the issue.

LucasSte: @seanyoung

seanyoung: I've renamed the file

LucasSte (approved):

PR #741, opened 2022-04-13
Create codegen::Expression and refactor codegen

LucasSte:
This PR refactor code generation. The main addition was the `codegen::Expression`. This means we no longer depend on `ast::Expression` to build the CFG. I've managed to improve some other stuff:

1. Now, we have a `cast` function for codegen. 
2. Sema's `cast` function is now a method of `ast::Expression`.
3. I created traits for common functions to facilitate implementing methods with generic arguments.
4. I removed from the AST expressions that do not belong there, except `FunctionArg`, to which I couldn't find an easy solution.
5. Some AST expressions do not exist is the CFG, so I removed references to them in codegen.
6. As we are branching at every AND and OR expression, they do not exist in `codegen::Expression`. However, I left some block comments containing the code to handle those cases, in case we add them in nearby future.

PS: This PR modifies 56 files, so please, @seanyoung, be pedantic!

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -13,14 +13,14 @@ use super::{
     vector_to_slice, Options,
 };
 use crate::codegen::subexpression_elimination::common_sub_expression_elimination;
-use crate::codegen::{undefined_variable, LLVMName};
+use crate::codegen::{undefined_variable, Expression, LLVMName};
 use crate::parser::pt;
 use crate::parser::pt::CodeLocation;
-use crate::sema::ast::{
-    CallTy, Contract, Expression, Function, Namespace, Parameter, StringLocation, Type,
-};
+use crate::sema::ast::RetrieveType;
+use crate::sema::ast::{CallTy, Contract, Function, Namespace, Parameter, StringLocation, Type};
 use crate::sema::contracts::{collect_base_args, visit_bases};
-use crate::Target;
+use crate::sema::Recurse;
+use crate::{ast, Target};
 
 #[derive(Clone)]
 #[allow(clippy::large_enum_variant)]
@@ -518,23 +518,9 @@ impl ControlFlowGraph {
                 self.expr_to_string(contract, ns, r)
             ),
             Expression::Variable(_, _, res) => format!("%{}", self.vars[res].id.name),
-            Expression::ConstantVariable(_, _, Some(var_contract_no), var_no)
-            | Expression::StorageVariable(_, _, var_contract_no, var_no) => format!(
-                "${}.{}",
-                ns.contracts[*var_contract_no].name,
-                ns.contracts[*var_contract_no].variables[*var_no].name
-            ),
-            Expression::ConstantVariable(_, _, None, var_no) => {
-                format!("${}", ns.constants[*var_no].name)
-            }
             Expression::Load(_, _, expr) => {
                 format!("(load {})", self.expr_to_string(contract, ns, expr))
             }
-            Expression::StorageLoad(_, ty, expr) => format!(
-                "(loadstorage ty:{} {})",
-                ty.to_string(ns),
-                self.expr_to_string(contract, ns, expr)
-            ),
             Expression::ZeroExt(_, ty, e) => format!(
                 "(zext {} {})",
                 ty.to_string(ns),
@@ -596,22 +582,6 @@ impl ControlFlowGraph {
                 self.expr_to_string(contract, ns, a),
                 f
             ),
-            Expression::Or(_, l, r) => format!(
-                "({} || {})",
-                self.expr_to_string(contract, ns, l),
-                self.expr_to_string(contract, ns, r)
-            ),
-            Expression::And(_, l, r) => format!(
-                "({} && {})",
-                self.expr_to_string(contract, ns, l),
-                self.expr_to_string(contract, ns, r)
-            ),
-            Expression::Ternary(_, _, c, l, r) => format!(
-                "({} ? {} : {})",
-                self.expr_to_string(contract, ns, c),
-                self.expr_to_string(contract, ns, l),
-                self.expr_to_string(contract, ns, r)
-            ),
             Expression::Not(_, e) => format!("!{}", self.expr_to_string(contract, ns, e)),
             Expression::Complement(_, _, e) => format!("~{}", self.expr_to_string(contract, ns, e)),
             Expression::UnaryMinus(_, _, e) => format!("-{}", self.expr_to_string(contract, ns, e)),
@@ -648,19 +618,6 @@ impl ControlFlowGraph {
                     .collect::<Vec<String>>()
                     .join(", ")
             ),
-            Expression::InternalFunction {
-                function_no,
-                signature,
-                ..
-            } => {
-                let function_no = if let Some(signature) = signature {
-                    contract.virtual_functions[signature]
-                } else {
-                    *function_no
-                };
-
-                ns.functions[function_no].print_name(ns)
-            }
             Expression::ExternalFunction {
                 address,
                 function_no,
@@ -673,41 +630,6 @@ impl ControlFlowGraph {
             Expression::InternalFunctionCfg(cfg_no) => {
                 format!("function {}", contract.cfg[*cfg_no].name)
             }
-            Expression::InternalFunctionCall { function, args, .. } => format!(
-                "(call {} ({})",
-                self.expr_to_string(contract, ns, function),
-                args.iter()
-                    .map(|a| self.expr_to_string(contract, ns, a))
-                    .collect::<Vec<String>>()
-                    .join(", ")
-            ),
-            Expression::Constructor {
-                contract_no,
-                constructor_no: Some(constructor_no),
-                args,
-                ..
-            } => format!(
-                "(constructor:{} ({}) ({})",
-                ns.contracts[*contract_no].name,
-                ns.functions[*constructor_no].signature,
-                args.iter()
-                    .map(|a| self.expr_to_string(contract, ns, a))
-                    .collect::<Vec<String>>()
-                    .join(", ")
-            ),
-            Expression::Constructor {
-                contract_no,
-                constructor_no: None,
-                args,
-                ..
-            } => format!(
-                "(constructor:{} ({})",
-                ns.contracts[*contract_no].name,
-                args.iter()
-                    .map(|a| self.expr_to_string(contract, ns, a))
-                    .collect::<Vec<String>>()
-                    .join(", ")
-            ),
             Expression::CodeLiteral(_, contract_no, runtime) => format!(
                 "({} code contract {})",
                 if *runtime {
@@ -717,32 +639,7 @@ impl ControlFlowGraph {
                 },
                 ns.contracts[*contract_no].name,
             ),
-            Expression::ExternalFunctionCall { function, args, .. } => format!(
-                "(external call {} ({})",
-                self.expr_to_string(contract, ns, function),
-                args.iter()
-                    .map(|a| self.expr_to_string(contract, ns, a))
-                    .collect::<Vec<String>>()
-                    .join(", ")
-            ),
             Expression::ReturnData(_) => "(external call return data)".to_string(),
-            Expression::Assign(_, _, l, r) => format!(
-                "{} = {}",
-                self.expr_to_string(contract, ns, l),
-                self.expr_to_string(contract, ns, r)
-            ),
-            Expression::PostDecrement(_, _, _, e) => {
-                format!("{}--", self.expr_to_string(contract, ns, e),)
-            }
-            Expression::PostIncrement(_, _, _, e) => {
-                format!("{}++", self.expr_to_string(contract, ns, e),)
-            }
-            Expression::PreDecrement(_, _, _, e) => {
-                format!("--{}", self.expr_to_string(contract, ns, e),)
-            }
-            Expression::PreIncrement(_, _, _, e) => {
-                format!("++{}", self.expr_to_string(contract, ns, e),)
-            }
             Expression::Cast(_, ty, e) => format!(
                 "{}({})",
                 ty.to_string(ns),
@@ -793,7 +690,7 @@ impl ControlFlowGraph {
         &self,
         contract: &Contract,
         ns: &Namespace,
-        l: &StringLocation,
+        l: &StringLocation<Expression>,
     ) -> String {
         match l {
             StringLocation::RunTime(e) => self.expr_to_string(contract, ns, e),
@@ -1254,11 +1151,11 @@ pub fn generate_cfg(
 
 /// resolve modifier call
 fn resolve_modifier_call<'a>(
-    call: &'a Expression,
+    call: &'a ast::Expression,
     contract: &Contract,
-) -> (usize, &'a Vec<Expression>) {
-    if let Expression::InternalFunctionCall { function, args, .. } = call {
-        if let Expression::InternalFunction {
+) -> (usize, &'a Vec<ast::Expression>) {
+    if let ast::Expression::InternalFunctionCall { function, args, .. } = call {
+        if let ast::Expression::InternalFunction {
             function_no,
             signature,
             ..
@@ -1596,7 +1493,7 @@ pub fn generate_modifier_dispatch(
     modifier: &Function,
     cfg_no: usize,
     chain_no: usize,
-    args: &[Expression],
+    args: &[ast::Expression],
     ns: &Namespace,
     opt: &Options,
 ) -> (ControlFlowGraph, usize) {
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -7,20 +7,23 @@ use super::{
     vartable::Vartable,
 };
 use crate::codegen::unused_variable::should_remove_assignment;
+use crate::codegen::Expression;
 use crate::parser::pt;
 use crate::parser::pt::CodeLocation;
+use crate::sema::ast;
+use crate::sema::ast::RetrieveType;
 use crate::sema::ast::{
-    Builtin, CallTy, Expression, FormatArg, Function, Namespace, Parameter, StringLocation, Type,
+    Builtin, CallTy, FormatArg, Function, Namespace, Parameter, StringLocation, Type,
 };
 use crate::sema::eval::{eval_const_number, eval_const_rational};
-use crate::sema::expression::{bigint_to_expression, cast, cast_shift_arg, ResolveTo};
+use crate::sema::expression::{bigint_to_expression, ResolveTo};
 use crate::Target;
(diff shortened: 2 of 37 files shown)


seanyoung:
Wonderful @LucasSte this cleans up a lot of ugliness, I'll review tomorrow morning when I've had some rest.

src/sema/expression.rs
```diff
@@ -6187,14 +6204,8 @@ fn resolve_using(
...
-                &libfunc.params[0].ty,
-                true,
-                ns,
-                &mut errors,
-            ) {
+            let param_ty = ns.functions[function_no].params[0].ty.clone();
+            match self_expr.cast(&self_expr.loc(), &param_ty, true, ns) {
                 Ok(e) => cast_args.push(e),
                 Err(()) => continue,
             }
```
seanyoung, line 6211: Here we try to convert the self expression to the first argument of the library function, and if this fails, we should try the next one without failing. Now any mismatch becomes a compiler error, which is broken. I've written an example here:

This example fails to compile on your branch but works fine on the main branch:
```
library lib1 {
	function fmuladd(int a, int b, int c) internal pure returns (int) {
		return a * b + c;
	}

	function fmuladd(int a, int b, bool c) internal pure returns (int) {
		return a * b;
	}

}

library lib2 {
	function fmuladd(int a, int b, bool c) internal pure returns (int) {
		return a * b;
	}

	function fmuladd(uint a, uint b, uint c) internal pure returns (uint) {
		return a * b + c;
	}
}

contract bar {
	using lib1 for *;
	using lib2 for *;

	function f1(int a, int b) public pure returns (int) {
		return a.fmuladd(b, 2);	
	}

	function f2(uint a, uint b) public pure returns (uint) {
		return a.fmuladd(b, 2);	
	}

}
```

src/sema/expression.rs
```diff
@@ -6218,7 +6229,7 @@ fn resolve_using(
                     }
                 };
 
-                match cast(&arg.loc(), arg.clone(), &ty, true, ns, &mut errors) {
+                match arg.cast(&arg.loc(), &ty, true, ns) {
                     Ok(expr) => cast_args.push(expr),
```
seanyoung, line 6233: again here the cast error should go to `errors`, not `ns.diagnostics`. If the argument cannot be casted, then try the next library function with same name.

src/sema/expression.rs
seanyoung: Please add this example to test cases

src/sema/expression.rs
```diff
@@ -6466,7 +6477,7 @@ fn method_call_named_args(
                     }
                 };
 
-                match cast(&arg.loc(), arg.clone(), &param.ty, true, ns, diagnostics) {
+                match arg.cast(&arg.loc(), &param.ty, true, ns) {
```
seanyoung, line 6480: again here the diagnostic should go to `diagnostics` not `ns.diagnostics`, if this particular overloaded function does not match then the loop should try the next one without failing.

This is an example of some solidity which fails on your branch, not on main:
```
contract c {
	function foo(int, bool) public { }
	function foo(bool, int) public { }
	function bar() public {
		foo(1, true);
		foo(true, 1);
	}
}
```

seanyoung: The same overloaded function match exists for named method call, function call, named function call, constructor call, named constructor call.

seanyoung (changes requested):
Your changes look very good and a huge leap forward, unfortunately the removal of diagnostics from `cast` is not correct, and the tests did not covert his.

Otherwise looks good!

src/sema/expression.rs
seanyoung: @LucasSte Please add this example to test cases

src/sema/expression.rs
seanyoung: @LucasSte Please add this example to test cases

src/sema/tests/mod.rs
```diff
@@ -125,3 +145,81 @@ fn test_statement_reachable() {
...
+		foo(1, true);
+		foo(true, 1);
+	}
+}
+    "#;
+
+    let ns = parse(file);
+    assert_eq!(ns.diagnostics.len(), 1);
+    assert_message_in_diagnostics(&ns.diagnostics, "found contract ‘c’");
+}
```
seanyoung, line 225: These tests are better in `tests/contract_tests/` because then the entire ast is tested.

src/codegen/cfg.rs
```diff
@@ -596,22 +591,22 @@ impl ControlFlowGraph {
...
+            //     "({} && {})",
+            //     self.expr_to_string(contract, ns, l),
+            //     self.expr_to_string(contract, ns, r)
+            // ),
+            // Expression::Ternary(_, _, c, l, r) => format!(
+            //     "({} ? {} : {})",
+            //     self.expr_to_string(contract, ns, c),
+            //     self.expr_to_string(contract, ns, l),
+            //     self.expr_to_string(contract, ns, r)
+            // ),
```
seanyoung, line 609: remove

src/sema/expression.rs
LucasSte: Have a look at `sema/tests/mod.rs`

src/codegen/statements.rs
```diff
@@ -946,31 +950,53 @@ fn destructure(
...
                 }
 
-                assign_single(left, &expr, cfg, contract_no, Some(func), ns, vartab, opt);
+                assign_single(left, expr, cfg, contract_no, Some(func), ns, vartab, opt);
             }
         }
     }
 }
 
+/// If the destructure statement returns a storage reference, we must load the value from storage
```
seanyoung, line 970: This needs explaining. Why are the casts not in the ast for example.

src/codegen/strength_reduce.rs
```diff
@@ -1617,57 +1497,57 @@ fn expression_values(expr: &Expression, vars: &Variables, ns: &Namespace) -> Has
...
+        //             }
+        //
+        //             Value {
+        //                 value,
+        //                 known_bits,
+        //                 bits: 1,
+        //             }
+        //         })
+        //         .collect()
+        // }
```
seanyoung, line 1550: remove

src/codegen/strength_reduce.rs
```diff
@@ -2632,74 +2512,74 @@ fn expresson_known_bits() {
...
+    //         Box::new(Expression::FunctionArg(loc, Type::Bool, 0)),
+    //     ),
+    //     &vars,
+    //     &ns,
+    // );
+    //
+    // assert_eq!(res.len(), 1);
+    // let v = res.iter().next().unwrap();
+    //
+    // assert!(!v.known_bits[0]);
```
seanyoung, line 2548: remove

src/codegen/strength_reduce.rs
```diff
@@ -2632,74 +2512,74 @@ fn expresson_known_bits() {
...
+    //         Box::new(Expression::FunctionArg(loc, Type::Bool, 0)),
+    //     ),
+    //     &vars,
+    //     &ns,
+    // );
+    //
+    // assert_eq!(res.len(), 1);
+    // let v = res.iter().next().unwrap();
+    // assert!(v.known_bits[0]);
+    // assert!(!v.value[0]);
```
seanyoung, line 2584: remove

src/codegen/subexpression_elimination/expression.rs
```diff
@@ -43,14 +44,13 @@ impl Expression {
...
-                Expression::And(*loc, Box::new(left.clone()), Box::new(right.clone()))
-            }
-
+            // Expression::Or(loc, ..) => {
+            //     Expression::Or(*loc, Box::new(left.clone()), Box::new(right.clone()))
+            // }
+            //
+            // Expression::And(loc, ..) => {
+            //     Expression::And(*loc, Box::new(left.clone()), Box::new(right.clone()))
+            // }
```
seanyoung, line 53: remove

src/codegen/subexpression_elimination/expression.rs
```diff
@@ -206,8 +206,8 @@ impl Expression {
             | Expression::BitwiseOr(_, _, left, right)
             | Expression::BitwiseAnd(_, _, left, right)
             | Expression::BitwiseXor(_, _, left, right)
-            | Expression::Or(_, left, right)
-            | Expression::And(_, left, right)
+            // | Expression::Or(_, left, right)
+            // | Expression::And(_, left, right)
```
seanyoung, line 210: remove

src/codegen/subexpression_elimination/expression.rs
```diff
@@ -255,10 +255,9 @@ impl Expression {
             Expression::BoolLiteral(_, value) => ConstantType::Bool(*value),
             Expression::NumberLiteral(_, _, value) => ConstantType::Number(value.clone()),
             Expression::BytesLiteral(_, _, value) => ConstantType::Bytes(value.clone()),
-            Expression::ConstantVariable(_, _, contract_no, var_no) => {
-                ConstantType::ConstantVariable(*contract_no, *var_no)
-            }
-
+            // Expression::ConstantVariable(_, _, contract_no, var_no) => {
+            //     ConstantType::ConstantVariable(*contract_no, *var_no)
+            // }
```
seanyoung, line 260: remove

src/codegen/subexpression_elimination/mod.rs
```diff
@@ -62,7 +62,7 @@ pub enum ConstantType {
     Bool(bool),
     Bytes(Vec<u8>),
     Number(BigInt),
-    ConstantVariable(Option<usize>, usize),
+    //ConstantVariable(Option<usize>, usize),
```
seanyoung, line 65: remove

src/codegen/subexpression_elimination/tests.rs
```diff
@@ -35,44 +36,56 @@ fn add_variable_function_arg() {
...
+//     let mut ave = AvailableExpression::default();
+//     let mut set = AvailableExpressionSet::default();
+//     let mut cst = CommonSubExpressionTracker::default();
+//
+//     set.process_instruction(&instr, &mut ave, &mut cst);
+//
+//     assert!(set.find_expression(&var).is_some());
+//     assert!(set.find_expression(&num).is_some());
+// }
+
```
seanyoung, line 63: remove

src/codegen/statements.rs
LucasSte: There is no cast in this function.

src/sema/tests/mod.rs
seanyoung: Whenever there is a change in a dot file in a PR then this needs reviewing. I certainly review it.

You could keep these tests if you want and also add them as dot files. The more tests the better.

src/codegen/statements.rs
LucasSte: I updated the comment. What is basically happening at this function is that we are adding a Load instruction to the CFG, given that we only have a reference to variable and not its value (this is not a cast).

src/codegen/mod.rs
```diff
@@ -327,3 +333,800 @@ impl LLVMName for Function {
...
+                // Expression::Or(loc, left, right) => Expression::Or(
+                //     *loc,
+                //     Box::new(filter(left, ctx)),
+                //     Box::new(filter(right, ctx)),
+                // ),
+                // Expression::And(loc, left, right) => Expression::And(
+                //     *loc,
+                //     Box::new(filter(left, ctx)),
+                //     Box::new(filter(right, ctx)),
+                // ),
```
seanyoung, line 1105: Remove this

src/codegen/statements.rs
```diff
@@ -946,31 +950,54 @@ fn destructure(
...
 
-                assign_single(left, &expr, cfg, contract_no, Some(func), ns, vartab, opt);
+                assign_single(left, expr, cfg, contract_no, Some(func), ns, vartab, opt);
             }
         }
     }
 }
 
+/// If the destructure statement returns a storage reference,
+/// we must add a load instruction in the CFG.
```
seanyoung, line 971: This needs an explanation. For normal assignments, sema will have cast the value to the correct type. However we do not for destructure statements.
```
contract c {
   int[2] x;
    function x() public returns (int[2] storage, int8) { return (x, 1); }
   function y()  public {
       (int[2] a, int16 b) =  x();
   }
}
```
The cast of `int8` to `int16` and from `int[2] storage` to `int[2]` is not stored in the ast. sema only checks that the cast can be done.

The cast in codegen does not insert `Instr::LoadStorage`.

seanyoung, line 971: This needs an explanation. For normal assignments, sema will have cast the value to the correct type. However we do not for destructure statements.
```
contract c {
   int[2] x;
    function x() public returns (int[2] storage, int8) { return (x, 1); }
   function y()  public {
       (int[2] a, int16 b) =  x();
   }
}
```
The cast of `int8` to `int16` and from `int[2] storage` to `int[2]` is not stored in the ast. sema only checks that the cast can be done.

The cast in codegen does not insert `Instr::LoadStorage`.

src/codegen/expression.rs
```diff
@@ -1043,38 +1087,8 @@ fn require(
...
-    func: Option<&Function>,
-    ns: &Namespace,
-    vartab: &mut Vartable,
-    opt: &Options,
-) -> Expression {
-    let recipient = expression(&args[0], cfg, contract_no, func, ns, vartab, opt);
-    cfg.add(vartab, Instr::SelfDestruct { recipient });
-    Expression::Poison
-}
-
```
seanyoung, line 1075: We used to have an issue where some contracts would overflow the stack when solang is built in debug mode. The problem with the debug compilation path is that each variable will use up more stack space even if the liveness does not overlap with other variables.
The result of this is that `codegen::expression::expression` had a huge stack frame and we got stack overflows with large expression depths. That's why all these small functions exists: `codegen:::expression::expression` must be kept as short as possible.

seanyoung, line 1075: We used to have an issue where some contracts would overflow the stack when solang is built in debug mode. The problem with the debug compilation path is that each variable will use up more stack space even if the liveness does not overlap with other variables.
The result of this is that `codegen::expression::expression` had a huge stack frame and we got stack overflows with simple contracts. That's why all these small functions exists: `codegen:::expression::expression` must be kept as short as possible.

src/codegen/expression.rs
```diff
@@ -1748,53 +1746,9 @@ fn interfaceid(ns: &Namespace, contract_no: &usize, loc: &pt::Loc) -> Expression
...
-) -> Expression {
-    Expression::Subtract(
-        *loc,
-        ty.clone(),
-        *unchecked,
-        Box::new(expression(left, cfg, contract_no, func, ns, vartab, opt)),
-        Box::new(expression(right, cfg, contract_no, func, ns, vartab, opt)),
-    )
-}
-
```
seanyoung, line 1794: Do not inline these functions

seanyoung, line 1794: Do not inline these functions

src/codegen/expression.rs
```diff
@@ -1643,33 +1657,17 @@ fn format_string(
...
-    vartab: &mut Vartable,
-    loc: &pt::Loc,
-    ty: &Type,
-    init: &Option<Vec<u8>>,
-    opt: &Options,
-) -> Expression {
-    let size = expression(size, cfg, contract_no, func, ns, vartab, opt);
-    Expression::AllocDynamicArray(*loc, ty.clone(), Box::new(size), init.clone())
-}
-
```
seanyoung, line 1661: Do not inline these functions

seanyoung, line 1661: Do not inline these functions

seanyoung (changes requested):

src/codegen/expression.rs
```diff
@@ -519,26 +537,30 @@ pub fn expression(
...
                 if ns.target == Target::Solana || args[0].ty().is_storage_bytes() {
                     array_push(loc, args, cfg, contract_no, func, ns, vartab, opt)
                 } else {
                     storage_slots_array_push(loc, args, cfg, contract_no, func, ns, vartab, opt)
                 }
             } else {
+                let mut second_arg = ty[0].default(ns).unwrap();
+                if args.len() > 1 {
+                    second_arg = expression(&args[1], cfg, contract_no, func, ns, vartab, opt);
+                };
```
seanyoung, line 563: Pointless call to default() for most cases, you could have:
```
                let value = if args.len() > 1 {
                    expression(&args[1], cfg, contract_no, func, ns, vartab, opt);
                } else {
                    ty[0].default(ns).unwrap();
                };
```

tests/contract_testcases/substrate/value/this_address.dot
```diff
@@ -3,7 +3,7 @@ strict digraph "tests/contract_testcases/substrate/value/this_address.sol" {
 	step1 [label="function step1\ncontract: b\ntests/contract_testcases/substrate/value/this_address.sol:3:13-62\nsignature step1()\nvisibility public\nmutability nonpayable"]
 	returns [label="returns\naddress payable "]
 	return [label="return\ntests/contract_testcases/substrate/value/this_address.sol:4:17-37"]
-	cast [label="cast address payable\ntests/contract_testcases/substrate/value/this_address.sol:4:24-37"]
+	cast [label="cast address payable\ntests/contract_testcases/substrate/value/this_address.sol:4:32-36"]
```
seanyoung, line 6: The location of the cast is wrong now. If I compile this contract on your branch:
```
        contract b {
            function step1() public returns (address) {
                return int(this);
            }
        }
```
I get the following output:

```
$ cargo run tests/contract_testcases/substrate/value/this_address.sol --target substrate
   Compiling solang v0.1.10 (/home/sean/git/solang)
    Finished dev [unoptimized + debuginfo] target(s) in 12.82s
     Running `target/debug/solang tests/contract_testcases/substrate/value/this_address.sol --target substrate`
error: conversion from contract b to int256 not possible
  ┌─ /home/sean/git/solang/tests/contract_testcases/substrate/value/this_address.sol:4:28
  │
4 │                 return int(this);
  │                            ^^^^

```
The location should be `int(this)`, not `this`. If I try the same on the main branch, I get:

```
$ cargo run tests/contract_testcases/substrate/value/this_address.sol --target substrate
   Compiling solang v0.1.10 (/home/sean/git/solang)
    Finished dev [unoptimized + debuginfo] target(s) in 28.16s
     Running `target/debug/solang tests/contract_testcases/substrate/value/this_address.sol --target substrate`
error: conversion from contract b to int256 not possible
  ┌─ /home/sean/git/solang/tests/contract_testcases/substrate/value/this_address.sol:4:24
  │
4 │                 return int(this);
  │                        ^^^^^^^^^

seanyoung (changes requested):
Sorry for not spotting this the first time round.

seanyoung (approved):
Looks great, thanks!

PR #742, opened 2022-04-13
feat: support byte as yulidentifier

mattsse:
prior to this `byte` was tokenized as `Token::Bytes(1)` as prior to 0.8.0 `byte` used to be an alias for `bytes1`

however this breaks the assembly `byte` function

```solidity
            assembly {
                v := byte(0, mload(add(signature, 0x60)))
            }
```

This PR

* adds a `"bytes" => Byte` token, that will be `Bytes(1)` if used as type or a yulidentifier


LucasSte (changes requested):
The code changes look great! That was a nice catch!

Please, make sure to fix the failing CI tests. As you added a new keyword, the error message from the parser changed, so tests have failed.

seanyoung:
I tried doing your change a different way and then realised your change is exactly right, it's perfect!

The error in the vscode extension is on line https://github.com/hyperledger-labs/solang/blob/main/vscode/src/test/suite/extension.test.ts#L32 you can simply copy and paste the new error from CI in there and it will be fine (adding `, byte` somewhere).

seanyoung (approved):

LucasSte (approved):

PR #743, opened 2022-04-13
feat: support projections in revert statement

mattsse:
I tried to fix this but I believe Hit some lalrpop issues, so this won't build I'm afraid. 
But not sure what the proper fix for this would be, so I'd need some help

Background

this is a valid revert:

```solidity
        function doRevert2() internal pure {
             revert LibName.ErrorName();
        }
```

if an error is defined inside a library `LibName` 

the solidity grammar states the revert stmt is `"revert" + expr + call-arg-list`
But I think only this is valid (Id)? ("." Id)*

### DIFF
--- a/solang-parser/src/solidity.lalrpop
+++ b/solang-parser/src/solidity.lalrpop
@@ -125,7 +125,7 @@ IdentifierProject: Vec<Identifier> = {
     => vec![],
     <i:SolIdentifier> <v:("." SolIdentifier)*> => {
             let mut i = vec![i];
-            i.extend(v);
+            i.extend(v.into_iter().map(|(_, id)|id));
             i
     }
 }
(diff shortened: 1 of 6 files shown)


solang-parser/src/solidity.lalrpop
```diff
@@ -121,6 +121,15 @@ StorageLocation: StorageLocation = {
     <l:@L> "calldata" <r:@R> => StorageLocation::Calldata(Loc::File(file_no, l, r)),
 }
 
+IdentifierProject: Vec<Identifier> = {
+    <i:SolIdentifier?> => vec![],
+    <i:SolIdentifier> <v:("." SolIdentifier)*> => {
```
seanyoung, line 126: This is ambiguous: both productions above can produce a single `SolIdentifier`. To fix this:

```
IdentifierProject: Vec<Identifier> = {
    => vec![],
    <i:SolIdentifier> <v:("." SolIdentifier)*> => {
            let mut i = vec![i];
            i.extend(v);
            i
    }
}
```

seanyoung (changes requested):
I think you have the right idea.

solang-parser/src/solidity.lalrpop
mattsse: great, that worked, updated.

seanyoung:
Solved by https://github.com/hyperledger-labs/solang/pull/806

PR #747, opened 2022-04-15
feat(parser): user defined value types

shekhirin:
Resolves https://github.com/hyperledger-labs/solang/issues/746

Better suggestions on resolving ambiguity with `type()` builtin are welcome 😄

### DIFF
--- a/solang-parser/src/solidity.lalrpop
+++ b/solang-parser/src/solidity.lalrpop
@@ -72,22 +72,19 @@ Type: Type = {
     FunctionType
 }
 
-ElementaryValueType: Type = {
+NoFunctionType: Type = {
     "bool" => Type::Bool,
     "address" => Type::Address,
     "address" "payable" => Type::AddressPayable,
-    Uint => Type::Uint(<>),
-    Int => Type::Int(<>),
-    Bytes => Type::Bytes(<>),
-    "byte" => Type::Bytes(1),
-}
-
-NoFunctionType: Type = {
-    ElementaryValueType,
     // payable is only used as a cast in solc
     "payable" => Type::Payable,
     "string" => Type::String,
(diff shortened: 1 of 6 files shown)


solang-parser/src/solidity.lalrpop
```diff
@@ -249,6 +254,14 @@ VariableDefinition: Box<VariableDefinition> = {
     }
 }
 
+TypeDefinition: Box<TypeDefinition> = {
+    <doc:DocComments> <l:@L> "type" <name:SolIdentifier> "is" <ty:ElementaryValueType> <r:@R> ";" => {
```
seanyoung, line 258: It would be nice if the parser could parse:
```
contract c {
	type foo is int[2];
}
```
Then we could produce a nice error message say "int[2]" is not an elementary value type (why it is not called a primitive type like every other language, I don't know!).

```
<doc:DocComments> <l:@L> "type" <name:SolIdentifier> "is" <ty:Precedence0> <r:@R> ";" => {
```
Will do the trick (I think).

seanyoung (changes requested):
Looks great, thanks!

Please can `contract { type x as bytes; }` and such be parsed so we can give nice error messages. If not we will have to rework this when we add the plumbing into sema for solang.

Thanks!

solang-parser/src/solidity.lalrpop
shekhirin: yeah totally agree, fixed

seanyoung (approved):
Perfect, thank you very much!

shekhirin:
> Perfect, thank you very much!

can we make a release after it's merged please? I want to incorporate these changes into Foundry's formatter 🙂

seanyoung:
solang-parser 0.1.12 published, thank you!

PR #749, opened 2022-04-18
Create codegen::Builtin

LucasSte:
This PR creates a `codegen::Builtin` so that we can have a single interface to manage both Yul and Solidity Builtins, some of which are common.


src/codegen/mod.rs
```diff
@@ -1120,3 +1121,125 @@ impl Expression {
...
+    WriteInt256LE,
+    WriteUint16LE,
+    WriteUint32LE,
+    WriteUint64LE,
+    WriteUint128LE,
+    WriteUint256LE,
+}
+
+impl Builtin {
+    pub fn from_ast(ast_builtin: &ast::Builtin) -> Builtin {
```
seanyoung, line 1185: Would it possible to implement the `From` trait here, like:

```
impl From<&ast::Builtin> for Builtin {
...
}
```
So you can use `.into()` to covert into `codegen::Builtin`

seanyoung (changes requested):

seanyoung (approved):

Issue #744, opened 2022-04-14
Files with \r\n can be incorrectly tokenized


alcuadrado:
Hi,

I just noticed that this file is tokenized incorrectly, and I have a strong suspicion that it's because it has `\r\n` end lines. In particular, it has some inside comments.

[9d694731f6291f4553fad6b8174e7c915e6219ec_TigerBNB.sol](https://github.com/hyperledger-labs/solang/files/8492764/9d694731f6291f4553fad6b8174e7c915e6219ec_TigerBNB.sol.txt)

If you compile it with `solc` the `pragma solidity ^0.6.12;` statement is correctly parsed. In `solang` it's part of a comment.

alcuadrado:
@seanyoung, I found more files that break like this. In case they are helpful, I'm posting them here:


[54bef7b75e1bec4594ecbc8caad54328d653f5c1_Spark.sol.txt](https://github.com/hyperledger-labs/solang/files/8496215/54bef7b75e1bec4594ecbc8caad54328d653f5c1_Spark.sol.txt)
[71d0e03abd2b4b4fb9ef73d4c6f35a650124791e_BabyRock.sol.txt](https://github.com/hyperledger-labs/solang/files/8496216/71d0e03abd2b4b4fb9ef73d4c6f35a650124791e_BabyRock.sol.txt)
[85d1f31777c5c9ca544a1a73b83341872bdca45e_PUSHNOU.sol.txt](https://github.com/hyperledger-labs/solang/files/8496217/85d1f31777c5c9ca544a1a73b83341872bdca45e_PUSHNOU.sol.txt)
[97f4706e90cbd5d882088a0a245c63cb6b5f27cf_Vitalic.sol.txt](https://github.com/hyperledger-labs/solang/files/8496218/97f4706e90cbd5d882088a0a245c63cb6b5f27cf_Vitalic.sol.txt)
[719de440d870b18de38782ca191ed6a651a2e67a_TotoToken.sol.txt](https://github.com/hyperledger-labs/solang/files/8496219/719de440d870b18de38782ca191ed6a651a2e67a_TotoToken.sol.txt)
[909b4e9a086ba9feb85b3244fd83b84cf8b2ab83_Saddam.sol.txt](https://github.com/hyperledger-labs/solang/files/8496220/909b4e9a086ba9feb85b3244fd83b84cf8b2ab83_Saddam.sol.txt)
[...]

PR #751, opened 2022-04-19
Fix interface inheritance bug

LucasSte:
This PR fixes a bug, in which we couldn't not call a function from an interfaced that was inherited from another one.

This is one of the bugs found in issue #744.


seanyoung (changes requested):
Looks like this routine does not check if the function is declared `public` or `external`. 

Also the same problem exists for external function calls with named arguments, e.g. `foo.bar({a: 1, b: 2});`

seanyoung:
So the code should check if the function is declared `public` or `external` and if not, give an error. Non-public functions are not callable so this would result in a broken transaction.

src/sema/expression.rs
```diff
@@ -5883,7 +5885,7 @@ fn method_call_pos_args(
             if matches {
                 diagnostics.truncate(marker);
 
-                if !ns.functions[function_no].is_public() {
+                if !ns.functions[*function_no].is_public() {
```
LucasSte, line 5888: @seanyoung We check here if the function is public or external.

LucasSte:
> Looks like this routine does not check if the function is declared `public` or `external`.
> 
> Also the same problem exists for external function calls with named arguments, e.g. `foo.bar({a: 1, b: 2});` (see `method_call_named_args`)

Can you provide an example of the `foo.bar({a: 1, b: 2})` case?

src/sema/expression.rs
seanyoung: > > Looks like this routine does not check if the function is declared `public` or `external`.
> > Also the same problem exists for external function calls with named arguments, e.g. `foo.bar({a: 1, b: 2});` (see `method_call_named_args`)
> 
> Can you provide an example of the `foo.bar({a: 1, b: 2})` case?

In the babylink example replace the call to `createPair` with:
```
        uniswapV2Pair = IUniswapV2Factory(_uniswapV2Router.factory())
            .createPair({tokenA: address(this), tokenB:_uniswapV2Router.WETH()});
```
does that work?

seanyoung:
> > Looks like this routine does not check if the function is declared `public` or `external`.
> > Also the same problem exists for external function calls with named arguments, e.g. `foo.bar({a: 1, b: 2});` (see `method_call_named_args`)
> 
> Can you provide an example of the `foo.bar({a: 1, b: 2})` case?

In the babylink example replace the call to `createPair` with:
```
        uniswapV2Pair = IUniswapV2Factory(_uniswapV2Router.factory())
            .createPair({tokenA: address(this), tokenB:_uniswapV2Router.WETH()});
```
does that work?

src/sema/expression.rs
seanyoung: Indeed, you are right.

Ideally we would give an error if a function is not public, rather than pretending it does not exist. However, that might be a bit much for this PR, going off-topic here

src/sema/expression.rs
```diff
@@ -5883,7 +5885,7 @@ fn method_call_pos_args(
...
                     diagnostics.push(Diagnostic::error(
                         *loc,
                         format!("function ‘{}’ is not ‘public’ or ‘external’", func.name),
```
LucasSte, line 5891: @seanyoung Who says we are not raising an error?

src/sema/expression.rs
LucasSte: We do raise an error here. Take a look at the lines following this comment.

seanyoung: Yes. Sorry, I should have paid more attention.

LucasSte:
> > > Looks like this routine does not check if the function is declared `public` or `external`.
> > > Also the same problem exists for external function calls with named arguments, e.g. `foo.bar({a: 1, b: 2});` (see `method_call_named_args`)
> > 
> > 
> > Can you provide an example of the `foo.bar({a: 1, b: 2})` case?
> 
> In the babylink example replace the call to `createPair` with:
> 
> ```
>         uniswapV2Pair = IUniswapV2Factory(_uniswapV2Router.factory())
>             .createPair({tokenA: address(this), tokenB:_uniswapV2Router.WETH()});
> ```
> 
> does that work?

I added the lines you suggested, but surprisingly the compilation does not fail.

seanyoung:
> I added the lines you suggested, but surprisingly the compilation does not fail.

Sorry that was not an inherited function. I wrote this.

```
interface A {
	function foo(int a, int b) external;
}

interface B is A {
	function bar(int a, int b) external;
}

contract C {
	function test(B b) public {
		b.foo({a: 1, b: 2});
	}
}
```
You should be able to do this yourself without asking me.

LucasSte:
> You should be able to do this yourself without asking me.

I deleted my comment, but you were faster to answer that 😅

seanyoung (approved):

PR #756, opened 2022-04-20
@return doccomments on unnamed returns permitted

seanyoung:
This doc comment is not permitted:
```
// @return feh
// @return foo
function f3() pure returns (int, int) { return (1, 2); }
```
Allow unnamed returned value to have doc comments, while ensuring that
there is on per return.

Also fix the location of doc-comments in diagnostics.

Found:
https://stackoverflow.com/questions/71789212/solang-panicked-at-type-not-allowed

The issue described on stackoverflow is not fixed yet.


src/sema/tags.rs
```diff
@@ -13,7 +13,7 @@ pub fn resolve_tags(
 ) -> Vec<Tag> {
     let mut res: Vec<Tag> = Vec::new();
 
-    for c in doc.iter().flat_map(pt::DocComment::comments) {
+    'next_doc: for c in doc.iter().flat_map(pt::DocComment::comments) {
```
LucasSte, line 16: This label makes the code very unreadable, because you continue the loop from down below. Is there any way to make it cleaner?

seanyoung: You're right, this can be written without labels. I've re-written without a for loop (find on iterator)

LucasSte (approved):

PR #757, opened 2022-04-20
Fix standalone function call bug

LucasSte:
This PR fixes a bug found in issue #744. We found out that when a function that returns multiple values was invoke as a top level statement, the compiler emitted an error. In sum, the following constructions should be allowed:

```solidity
function multipleReturns() returns (int, int) {
    return (1, 2)
}

function test() returns int {
    multipleReturns();
    return 3;
}
```


seanyoung (changes requested):
We should really be avoiding abi decoding for external function calls if there return values are not used, we could have a codegen test for this.

src/sema/expression.rs
```diff
@@ -7342,19 +7488,19 @@ pub fn named_function_call_expr(
 }
 
 /// Get the return values for a function call
-fn function_returns(ftype: &Function) -> Vec<Type> {
-    if !ftype.returns.is_empty() {
+fn function_returns(ftype: &Function, resolve_to: &ResolveTo) -> Vec<Type> {
```
LucasSte, line 7491: @seanyoung This is where I change the return type of the function. When I checked the CFG everything was correct, however the problem of `ResolveTo::Discard` arose when I set `Type::Void` to single return functions.

src/sema/expression.rs
```diff
@@ -1663,6 +1663,12 @@ pub fn expression(
         )),
         // unary expressions
         pt::Expression::Not(loc, e) => {
+            // The result should not be discarded
+            let resolve_to = if matches!(resolve_to, ResolveTo::Discard) {
+                ResolveTo::Unknown
+            } else {
+                resolve_to
+            };
```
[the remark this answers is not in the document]
LucasSte: I've changed my approach. I now keep track of the expression tree's depth. If we are deeper than the first level, `ResolveTo` will never be `Discard`.

tests/contract_testcases/ewasm/standalone_call.sol
```diff
@@ -0,0 +1,33 @@
...
+        1 | singleReturn();
+        1 & singleReturn();
+        1 << singleReturn();
+        1 >> singleReturn();
+        !returnBool();
+        ~singleReturn();
+        +singleReturn();
+        -singleReturn();
+        return b;
+    }
```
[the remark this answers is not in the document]
LucasSte: The named function call was already there. I added a call from an external contract's function.

Forgive my naiveness, but I don't know exactly what you meant by external call and method call. I tried Googling  these terms, but I found no reliable result.

seanyoung (changes requested):
I was hoping a different solution would come to mind, but nothing has occurred so far.

The following solidity inserts abi decoding for the return values even though they are discarded:
```
contract C {
        function foo() public returns (int, int) { return (1, 2); }
        function bar() public { this.foo(); }
}
```
You also mentioned this problem:
```
contract c {
function test() public returns (int) {
  1 + 1.5 + 1.3;
  return 5;
}
}
```
Anything that can be done about that? Or should we create a new issue?

seanyoung:
How about: resolving a `pt::Statement::Expression`, if the expression is a `pt::Expression::NamedFunctionCall` or `pt::Expression::FunctionCall`, then attempt to resolve without return values? You can call `named_call_expr` or `call_expr` with `is_destructible` set to true (may need renaming to something better).

LucasSte:
> How about: resolving a `pt::Statement::Expression`, if the expression is a `pt::Expression::NamedFunctionCall` or `pt::Expression::FunctionCall`, then attempt to resolve without return values? You can call `named_call_expr` or `call_expr` with `is_destructible` set to true (may need renaming to something better).
> 
> This may need some special handling for standalone casts/struct literals (even though they don't make any sense).

I accepted your suggestions 🙂.

src/sema/builtin.rs
```diff
@@ -868,7 +868,7 @@ pub fn resolve_call(
     namespace: Option<&str>,
     id: &str,
     args: &[pt::Expression],
-    context: &ExprContext,
+    context: &mut ExprContext,
```
LucasSte, line 871: Context does not need to be mutable. I'll change this tomorrow.

tests/ewasm.rs
```diff
@@ -2294,7 +2294,7 @@ fn selfdestruct() {
     runtime.function("step1", &[]);
     runtime.accounts.get_mut(&runtime.vm.cur).unwrap().1 = 0;
 
-    runtime.function_revert("step2", &[]);
+    runtime.function("step2", &[]);
```
seanyoung, line 2297: Do you know why this changed?

seanyoung (changes requested):

src/sema/expression.rs
```diff
@@ -1713,6 +1713,7 @@ pub fn expression(
             }
         },
         pt::Expression::UnaryPlus(loc, e) => {
+            // The results of the expression should not be discarded
```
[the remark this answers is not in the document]
LucasSte: I forgot to remove these comments.

src/sema/expression.rs
```diff
@@ -1723,6 +1724,7 @@ pub fn expression(
         }
 
         pt::Expression::Ternary(loc, c, l, r) => {
+            // The results of ternary expressions should not be discarded
```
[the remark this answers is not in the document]
LucasSte: I forgot to remove this comment.

tests/ewasm.rs
LucasSte: We should not remove the `abiDecode` instruction if we don't use the results of a function call, then right?

seanyoung: The test could be fixed by changing the to `bool foo = goaway();`.

As far eliding abi decoding, I still think it's a good idea. There are corner cases like this of course.  I am not sure what solc does (and if we care)

LucasSte: Fixed

seanyoung (approved):

PR #761, opened 2022-04-22
Add spl-token integration

### LINKED ISSUE #607: Receive and transfer Solana SPL token into smart contract
I'm building a smart contract that can receive and send Solana SPL tokens, but I don't see any document mention about that or any example from doing that.

Have Solang supported that yet? If it is, please help me find it out. Thank you.

seanyoung:
Provide spl-token solidity library, documentation and tests.

Fixes #607


examples/spl_token.sol
```diff
@@ -0,0 +1,135 @@
+import 'solana';
```
LucasSte, line 1: This file should also be on codegen/contract tests, as it should never be broken.

integration/solana/token.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { getOrCreateAssociatedTokenAccount, createMint, TOKEN_PROGRAM_ID } from '@solana/spl-token';
```
LucasSte, line 1: How about asking someone from Solana who knows typescript to review this file?

src/emit/mod.rs
```diff
@@ -4077,7 +4077,8 @@ pub trait TargetRuntime<'a> {
                         let address = if let Some(address) = address {
                             let address = self.expression(bin, address, &w.vars, function, ns);
 
-                            let addr = bin.builder.build_array_alloca(
+                            let addr = bin.build_array_alloca(
+                                function,
```
LucasSte, line 4081: What is the implication of this?

LucasSte (changes requested):

docs/targets/solana.rst
```diff
@@ -76,49 +73,93 @@ package has `documentation <https://solana-labs.github.io/solana-solidity.js/>`_
...
+---------------
 
-ewasm has been tested with `Hyperledger Burrow <https://github.com/hyperledger/burrow>`_.
-Please use the latest master version of burrow, as ewasm support is still maturing in Burrow.
+`spl-token <https://spl.solana.com/token>`_ is the solana native way of creating tokens, minting, burning, transfering
+and swapping tokens. We have created a library to use spl-token from Solidity. This file ``spl_token.sol`` should be
+imported in your solidity, and then you can use the library.
 
-Some language features have not been fully implemented yet on ewasm:
+.. include:: ../../examples/spl_token.sol
```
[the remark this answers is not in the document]
seanyoung: I was wondering about that too. We should definitely link to the file.

examples/spl_token.sol
seanyoung: We have tests for that.

integration/solana/token.spec.ts
seanyoung: This is just a test file, I don't it matters enough to bother someone else about it.

src/emit/mod.rs
seanyoung: On Solana, llvm requires that all stack allocations happen in the first block of a function. This is a limitation of the BPF backend in llvm.

The stack allocator in the llvm backend is very limited. Stack variables with non-overlapping life times do not reuse stack space, so if you have too many local variables then you overflow the 4KiB stack size (per-frame).

seanyoung: Without this change, compiling spl_token.sol crashes llvm with an assert

integration/solana/token.spec.ts
seanyoung: I think everyone in solana is a far too busy to bother about this. However, if we have typescript will be used by users, then a review by a type-script type person would be very helpful.

docs/targets/burrow.rst
```diff
@@ -0,0 +1,18 @@
+Hyperledger Burrow (ewasm)
+__________________________
```
LucasSte, line 2:
```suggestion
=========================
```

docs/targets/substrate.rst
```diff
@@ -0,0 +1,26 @@
+Parity Substrate
+________________
```
LucasSte, line 2:
```suggestion
================
```
I'm changing to be consistent with the other rst files.

integration/solana/token.sol
```diff
@@ -0,0 +1,29 @@
+import '../../examples/spl_token.sol';
```
LucasSte, line 1: Mos people only download Solang pre-compiled binary and will not have the repository on their system. Having to download an example is too burdensome. If you want to use x86 or arm intrinsics in C/C++, you just need to include a file, without any download.
One option to solve this would be to create an installation bundle that saves this includes in an installation folder and automatically configures the environment variables, so that we can just type `solang` on our terminal to use it. What do you think of this?

docs/targets/burrow.rst
LucasSte: Changing to be consistent with other rst files.

LucasSte (changes requested):

integration/solana/token.sol
seanyoung: We have seen various issues/questions being raised about installing solang. It would be nice to have a better way of dealing with this, but I don't know what the answer is. Something like rustup, solana-installer, or avm (anchor version manager) might be the right answer.

Hopefully we will get to a point where there will be many files that help users write solidity on solana (or other chains), including a port of openzeppelin. How that should be distributed, I don't know. Maybe something on solana-program-library.

I don't think that asking users to copy a file into their tree is too difficult, it's just not the most elegant solution.

However, the answer to all of this is not going to be solved in this PR.

LucasSte (approved):

PR #763, opened 2022-04-25
Remove Expression::FunctionArg from ast

seanyoung:
This expression type does not belong in the ast. It is used in the
implicit accessor functions of public storage variables. Rather than
using a codegen-like expression, create a proper symbol table entries
for the arguments and retrieve them.


src/sema/variables.rs
```diff
@@ -427,66 +430,101 @@ pub fn var_decl(
...
+            collect_parameters(value, symtable, params, expr, ns)
         }
         Type::Array(elem_ty, dims) => {
             let mut ty = Type::StorageRef(false, Box::new(ty.clone()));
             for _ in 0..dims.len() {
                 let map = (*expr).clone();
 
+                let arg_name = pt::Identifier {
+                    loc: pt::Loc::Implicit,
+                    name: format!("index{}", params.len()),
```
LucasSte, line 486: How do you make sure this name is not already used in the Symbol table? I believe that `index0` is a valid symbol for variables in Solidity.

LucasSte (changes requested):

src/sema/variables.rs
seanyoung: The only symbols in the symbol table are the parameters to the function, so there can't be a conflict.

LucasSte: Don't we use the symbol table to add all variables of a function?

seanyoung: Yes, we do.

This code creates an accessor function.

seanyoung: That is an entirely synthetic function.

seanyoung: I've changed it again. Now the parameters have no names, which probably makes more sense than coming up with synthetic useless names.

PR #764, opened 2022-04-26
Delay resolving on contract variable initializers

### LINKED ISSUE #762: Cannot find local function when creating state variable.
When creating a function pointer variable as an attribute of a contract, I can assign it a function of the contract itself. For instance, the following contract does so and works in solc.

```solidity
contract testing {
    
    function testPtr(int a) public pure returns (int) {
        return a/2;
    }

    function(int) external pure returns (int) sfPtr = this.testPtr;

    function doTest() public pure {
        assembly {
            let x := 1
        }
    }
}
```

However, Solang raises the following error:
```
contract ‘testing’ has no public function ‘testPtr’
```

seanyoung:
The initializer to contract variable may only be done once all the base
contract functions are resolved, as the initializer might be reference to
a function in base contract, for example:

    contract b {
        function testPtr(int a) public pure returns (int) {
            return a/2;
        }
    }

    contract testing is b {
        function(int) external pure returns (int) sfPtr = this.testPtr;
    }

Note this also fixes code like:

    contract test {
        uint x = y + 102;
        uint y = 102;
    }

Fixes #762


src/sema/contracts.rs
```diff
@@ -733,24 +736,34 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
     }
 }
 
+#[derive(Default)]
+
+struct ResolveLater<'a> {
+    functions: Vec<(usize, usize, &'a pt::FunctionDefinition)>,
```
LucasSte, line 742: Instead of this tuple, you could use a struct. This way we know what the first and the second `usize` mean.

LucasSte (changes requested):

src/sema/contracts.rs
seanyoung: Yes, you are right. I've added comments to this structure

seanyoung: Sorry, I'm introducing structs now

LucasSte (approved):
LGTM

seanyoung:
@LucasSte some errors have changed, please review the dot files too

LucasSte:
> @LucasSte some errors have changed, please review the dot files too

Thanks for bringing this up. I reviewed the dot files and compared the output with what we expect as errors from their respective `.sol` files. The updated error messages are correct. The variables have been assigned, but never read.

PR #766, opened 2022-04-26
Refactor diagnostics

LucasSte:
Signed-off-by: Lucas Steuernagel <lucas.tnagel@gmail.com>


src/sema/diagnostics.rs
```diff
@@ -5,48 +5,43 @@ use codespan_reporting::{diagnostic, files, term};
...
-        }
-
-        let diagnostic = convert_diagnostic(msg, &file_id);
-
-        term::emit(&mut writer.lock(), &config, &files, &diagnostic).unwrap();
-    }
+#[derive(Default)]
+pub struct Diagnostics {
+    pub contents: Vec<Diagnostic>,
+    has_error: bool,
```
seanyoung, line 11: one is pub and the other not?

src/sema/diagnostics.rs
```diff
@@ -5,48 +5,43 @@ use codespan_reporting::{diagnostic, files, term};
...
+    /// Borrow the contents vector
+    /// We should not modify the contents without this data strcuture
+    pub fn get_contents(&self) -> &Vec<Diagnostic> {
+        &self.contents
+    }
 
-    for msg in &ns.diagnostics {
-        if msg.level == Level::Debug && !debug {
-            continue;
+    pub fn add_diagnostic(&mut self, diagnostic: Diagnostic) {
```
seanyoung, line 25: you're repeating diagnostic in the function name. We already know it's a diagnostic. Please just `fn add()`

seanyoung, line 25: you're repeating diagnostic in the function name. We already know it's a diagnostic. Please just `fn add()`

src/sema/diagnostics.rs
```diff
@@ -5,48 +5,43 @@ use codespan_reporting::{diagnostic, files, term};
...
+    }
 
-        let diagnostic = convert_diagnostic(msg, &file_id);
-
-        let mut buffer = RawBuffer::new();
-
-        term::emit(&mut buffer, &config, &files, &diagnostic).unwrap();
-
-        println!("{}", buffer.into_string());
+    pub fn extend_diagnostic(&mut self, diagnostics: Vec<Diagnostic>) {
```
seanyoung, line 32: `fn extend()`

seanyoung, line 32: `fn extend()`

src/sema/diagnostics.rs
```diff
@@ -5,48 +5,43 @@ use codespan_reporting::{diagnostic, files, term};
...
-    let (files, file_id) = convert_files(ns, cache);
+impl Diagnostics {
+    pub fn any_errors(&self) -> bool {
+        self.has_error
+    }
 
-    let config = term::Config::default();
+    /// Borrow the contents vector
+    /// We should not modify the contents without this data strcuture
+    pub fn get_contents(&self) -> &Vec<Diagnostic> {
```
seanyoung, line 21: `fn get_contents()`is not a great name. How about just providing an iterator?

seanyoung, line 21: `fn get_contents()`is not a great name. How about just providing an iterator?

src/sema/diagnostics.rs
```diff
@@ -5,48 +5,43 @@ use codespan_reporting::{diagnostic, files, term};
...
+            self.has_error = diagnostics.iter().any(|m| m.level == Level::Error);
+        }
+        self.contents.extend(diagnostics);
     }
-}
 
-/// Do we have any errors
-pub fn any_errors(diagnotic: &[Diagnostic]) -> bool {
-    diagnotic.iter().any(|m| m.level == Level::Error)
+    pub fn append_diagnostic(&mut self, diagnostics: &mut Vec<Diagnostic>) {
```
seanyoung, line 39: `fn append()`

seanyoung, line 39: `fn append()`

src/sema/diagnostics.rs
seanyoung: maybe `fn push()` would be even better, keeps the diff small

seanyoung: maybe `fn push()` would be even better, keeps the diff small

seanyoung (changes requested):
You could provide `fn first_error()`, `fn count_warnings()`, `fn get_warnings()` as methods on `Diagnostics`

src/sema/diagnostics.rs
LucasSte: Both should be private. I forgot to remove the first `pub`. That was a lapse.

src/sema/diagnostics.rs
LucasSte: I'll rename it to push.

src/sema/diagnostics.rs
LucasSte: Some places require the vector, especially in tests. I do `contents.len()`, `contents[0]` and I pass a slice of `contents` to other functions. Are those things I can do with an iterator?

seanyoung: Yes you get can the length of an iterator or get the nth element

For tests, I think it's better to provide `impl Diagnostics` functions rather than implementing them in helper functions in the tests.

seanyoung (changes requested):

src/sema/diagnostics.rs
```diff
@@ -2,51 +2,119 @@ use super::ast::{Diagnostic, Level, Namespace};
...
-    for msg in &ns.diagnostics {
-        if msg.level == Level::Debug && !debug {
-            continue;
+    /// Borrow the contents vector
+    /// We should not modify the contents without this data structure
+    pub fn get_contents(&self) -> &Vec<Diagnostic> {
+        &self.contents
+    }
+
+    pub fn iter(&self) -> Iter<Diagnostic> {
```
seanyoung, line 27: Is this possible with `impl Iterator for Diagnostics`

LucasSte: I don't think that is necessary. Implementing the `Iterator` trait for diagnostics would require some extra code to handle functions like `next()`, which keeps track of the last returned element.

seanyoung (changes requested):

src/sema/diagnostics.rs
seanyoung: I think you're right, I can't find a way of doing this with `impl Iterator` either.

src/sema/diagnostics.rs
```diff
@@ -2,51 +2,117 @@ use super::ast::{Diagnostic, Level, Namespace};
...
+        for diag in &self.contents {
+            if matches!(diag.level, Level::Error) {
+                vec.push(diag);
+            }
+        }
+        vec
+    }
 
-        term::emit(&mut buffer, &config, &files, &diagnostic).unwrap();
+    pub fn assert_message_in_warnings(&self, message: &str) -> bool {
```
seanyoung, line 96: This function does not assert, it should be called `fn contains_message_in_warnings()` or `fn warning_contains()`

seanyoung (changes requested):
Looks good, just one comment

seanyoung (approved):

PR #767, opened 2022-04-27
Fix require() not found and staticcall in view function

seanyoung:
More fixes for #744. 

Two commits:

address.call() is permitted in view functions on ewasm and solana

and:

Error of garbage argument to builtin function not surfaced
   
A function call like:
    
            require("a" > 2);
    
Will just give the diagnostic:
    
             error: cannot find overloaded function which matches signature


tests/contract_testcases/ewasm/call/call_02.dot
```diff
@@ -13,6 +13,7 @@ strict digraph "tests/contract_testcases/ewasm/call/call_02.sol" {
 	diagnostic [label="found contract ‘x’\nlevel Debug\ntests/contract_testcases/ewasm/call/call_02.sol:2:9-20"]
 	diagnostic_14 [label="destructure variable ‘s‘ has never been used\nlevel Warning\ntests/contract_testcases/ewasm/call/call_02.sol:4:23-24"]
 	diagnostic_15 [label="destructure variable ‘bs‘ has never been used\nlevel Warning\ntests/contract_testcases/ewasm/call/call_02.sol:4:39-41"]
+	diagnostic_16 [label="function can be declared ‘view’\nlevel Warning\ntests/contract_testcases/ewasm/call/call_02.sol:3:13-49"]
```
LucasSte, line 16: Is this right? The function makes an external call.

tests/contract_testcases/ewasm/comment_tests.dot
```diff
@@ -826,12 +826,14 @@ strict digraph "tests/contract_testcases/ewasm/comment_tests.sol" {
...
-	diagnostic_831 [label="storage variable ‘_isExcluded‘ has never been used\nlevel Warning\ntests/contract_testcases/ewasm/comment_tests.sol:386:9-61"]
-	diagnostic_832 [label="storage variable ‘MAX‘ has been assigned, but never read\nlevel Warning\ntests/contract_testcases/ewasm/comment_tests.sol:390:9-51"]
-	diagnostic_833 [label="storage variable ‘burnrepeat‘ has been assigned, but never read\nlevel Warning\ntests/contract_testcases/ewasm/comment_tests.sol:408:5-44"]
-	diagnostic_834 [label="storage variable ‘_tFeeTotal‘ has never been used\nlevel Warning\ntests/contract_testcases/ewasm/comment_tests.sol:410:5-31"]
-	diagnostic_835 [label="storage variable ‘damada‘ has been assigned, but never read\nlevel Warning\ntests/contract_testcases/ewasm/comment_tests.sol:522:5-27"]
+	diagnostic_830 [label="function can be declared ‘view’\nlevel Warning\ntests/contract_testcases/ewasm/comment_tests.sol:188:5-75"]
+	diagnostic_831 [label="inline assembly is not yet supported\nlevel Error\ntests/contract_testcases/ewasm/comment_tests.sol:287:17-290:18"]
+	diagnostic_832 [label="function can be declared ‘view’\nlevel Warning\ntests/contract_testcases/ewasm/comment_tests.sol:264:5-270:37"]
```
LucasSte, line 831: This function utilizes `mload` in assembly. I don't think it can be declared `view`.

tests/contract_testcases/ewasm/call/call_02.dot
seanyoung: This is what the PR is trying to fix. solc says this is permitted.

tests/contract_testcases/ewasm/comment_tests.dot
seanyoung: `mload` loads from memory, not state

tests/contract_testcases/ewasm/call/call_02.dot
seanyoung: The question is whether the called contract modifies state. We don't know that, because `.call()` is without the abi.

solc assumes it views state, but does not modify it. This is useful because then call be used in rpc/readonly only mode. That means no transaction is required: this means it is free and fast.

Making users pay for transactions because you might modify state might not be appreciated.

On Substrate call is explicitly disallowed for rpc calls.

LucasSte (approved):

PR #769, opened 2022-04-27
Give a nice error message when 'now' is used

seanyoung:


tests/contract_testcases/solana/now.sol
```diff
@@ -0,0 +1,12 @@
...
+	struct S { int f1; }
+	function f(S s) public returns (int) {
+		return now.s;
+	}
+	function g() public {
+		now();
+	}
+	function h() public returns (uint) {
+		return now;
+	}
```
LucasSte, line 11: Am I allowed to declare a variable named `now`?
```solidity
function j() public returns (int) {
    int now = 5;
    return now;
}
```

seanyoung: Good point. I've added this test

PR #771, opened 2022-04-28
Improve compiler diagnostics

seanyoung:
Ensure that `now > startTime` gives error about block.timestamp
Improve error for old `function()` syntax
Ensure bad expression errors are surfaced for overloaded function calls

See #744


src/sema/expression.rs
```diff
@@ -1467,7 +1467,8 @@ pub fn compatible_mutability(left: &Mutability, right: &Mutability) -> bool {
 #[derive(PartialEq, Clone, Copy, Debug)]
 pub enum ResolveTo<'a> {
     Unknown,        // We don't know what we're looking for, best effort
-    Discard,        // We won't be using the result. For example, an expression as a statement/
+    Integer,        // Try to resolve to an integer type value
```
LucasSte, line 1470: What is the difference between this and `ResolveTo::Type(Type::Int(32))`? 
I guess you didn't want to specify the number of bits, right?

src/sema/expression.rs
```diff
@@ -5943,6 +5969,18 @@ fn method_call_pos_args(
...
+
+        for arg in args {
+            errors |=
+                expression(arg, context, ns, symtable, diagnostics, ResolveTo::Unknown).is_err();
+        }
+
+        if errors {
+            return Err(());
+        }
+
```
LucasSte, line 5983: This block of code looks familiar. 🤔

src/sema/builtin.rs
```diff
@@ -1305,6 +1305,17 @@ pub fn resolve_method_call(
...
+    let mut errors = false;
+
+    for arg in args {
+        errors |= expression(arg, context, ns, symtable, diagnostics, ResolveTo::Unknown).is_err();
+    }
+
+    if errors {
+        return Err(());
+    }
+
```
LucasSte, line 1318: This is exactly the same code from here:
https://github.com/hyperledger-labs/solang/blob/6a2ba29d45b808c6a526d7a9be743e8337f5727c/src/sema/builtin.rs#L881-L891
How about creating a function?

LucasSte (changes requested):

src/sema/expression.rs
seanyoung: Unsigned/signed any bit depth. I've add  that to the comment

src/sema/builtin.rs
seanyoung: I've introduce `fn args_sanity_check()`

LucasSte:
Does this PR also solve issue #779?

seanyoung:
> Does this PR also solve issue #779?

No

LucasSte (approved):

PR #772, opened 2022-04-28
Fix associativity of power

seanyoung:
1 ** 2 ** 3 should be parsed like 1 ** (2 ** 3), not (1 ** 2) ** 3.

See https://github.com/foundry-rs/foundry/issues/781#issuecomment-1112109532


LucasSte (approved):

jpopesculian:
amazing :heart_eyes: thanks for the clarification

PR #778, opened 2022-04-29
feat(parser): event on source unit level isn't allowed

shekhirin:
It's not allowed to have an `EventDefinition` on source unit level according to the [doc]( https://docs.soliditylang.org/en/v0.8.13/grammar.html#a4.SolidityParser.sourceUnit) and `solc` output:
```bash
echo 'event A();' | solc -

Error: Expected pragma, import directive or contract/interface/library/struct/enum/constant/function definition.
 --> <stdin>:1:1:
  |
1 | event A();
  | ^^^^^
```

@seanyoung I'm not sure it's the best to disallow it because of backward compatibility. We can add a diagnostic message instead. If you're ok with that change, I'll fix the tests.

### DIFF
--- a/solang-parser/src/pt.rs
+++ b/solang-parser/src/pt.rs
@@ -127,7 +127,6 @@ pub enum SourceUnitPart {
     ImportDirective(Vec<DocComment>, Import),
     EnumDefinition(Box<EnumDefinition>),
     StructDefinition(Box<StructDefinition>),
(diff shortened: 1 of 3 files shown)


shekhirin:
@seanyoung oh, I didn't see that issue and it's indeed pretty cool that solang already has a support for this! No problem from our side, I'm closing the PR, thanks!

Issue #780, opened 2022-04-29
Cannot find overloaded function


LucasSte:
The following contract works on solc, however solang cannot find an overloaded function for
```
error: cannot find overloaded function which matches signature
    ┌─ /home/lucas/Documents/solang/examples/test.sol:496:5
    │  
496 │ ╭     ApproveAndCallFallBack(spender).receiveApproval(msg.sender, tokens, address(
497 │ │ this
498 │ │ ), data);
    │ ╰────────^

```

```solidity
/**
 *Submitted for verification at BscScan.com on 2021-08-12
*/

pragma solidity >=
0.5
.17
;
library SafeMath {
  
function
 
add
(
uint a, uint b
) 
internal
 
pure
 
returns
 (
uint c
) 
{
    c = a + b;
    
require
(c >= a);
  }
  
function
 
sub
(
uint a, uint b
) 
internal
 
pure
 
returns
 (
uint c
) 
{
    
require
(b <= a);
    c = a - b;
  }
  
function
 
mul
(
uint a, uint b
) 
internal
 
pure
 
returns
 (
uint c
) 
{
    c = a * b;
    
require
(a == 
0
 || c / a == b);
  }
  
function
 
div
(
uint a, uint b
) 
internal
 
pure
 
returns
 (
uint c
) 
{
    
require
(b > 
0
);
    c = a / b;
  }
}
abstract contract BEP20Interface {
  
function
 
totalSupply
(
) 
public virtual
 
view
 
returns
 (
uint
)
;
  
function
 
balanceOf
(
address tokenOwner
) 
public virtual
 
view
 
returns
 (
uint balance
)
;
  
function
 
allowance
(
address tokenOwner, address spender
) 
public virtual
 
view
 
returns
 (
uint remaining
)
;
  
function
 
transfer
(
address to, uint tokens
) 
public virtual
 
returns
 (
bool success
)
;
  
function
 
approve
(
address spender, uint tokens
) 
public virtual
 
returns
 (
bool success
)
;
  
function
 
transferFrom
(
address 
from
, address to, uint tokens
) 
public virtual
 
returns
 (
bool success
)
;
  event Transfer(address indexed 
from
, address indexed to, uint tokens);
  event Approval(address indexed tokenOwner, address indexed spender, uint tokens);
}
abstract contract ApproveAndCallFallBack {
  
function
 
receiveApproval
(
address 
from
, uint256 tokens, address token, bytes memory data
) 
public virtual
;
}
[...]

PR #781, opened 2022-04-29
Ensure that virtual functions in abstract contracts can be called

seanyoung:
See https://github.com/hyperledger-labs/solang/issues/780.


LucasSte (approved):
Your modifications work for the provided example and fix the inheritance issue. However, the contract I posted in issue #780 still doesn't work (constant folding enters unreachable code).

PR #787, opened 2022-05-02
test(parser): semantic & syntax tests from ethereum/solidity

shekhirin:
Official Solidity repo contains [semantic tests](https://github.com/ethereum/solidity/tree/develop/test/libsolidity/semanticTests) and [syntax tests](https://github.com/ethereum/solidity/tree/develop/test/libsolidity/syntaxTests) that we can use to test the support for all languages features i.e. whether Solang can parse any valid Solidity code.

UPD: parser now passes all tests!


seanyoung:
Great idea. Let's work through the issues together.

shekhirin:
> Great idea. Let's work through the issues together.

I think these errors can be summed up into separate GitHub issues to make it clear what to fix.

seanyoung:
5 files with errors now

shekhirin:
@seanyoung I also added support for syntax tests from `libsolidity`, there are some interesting cases when parsing halts. I updated my first message with new errors.

seanyoung:
> @seanyoung I also added support for syntax tests from `libsolidity`, there are some interesting cases when parsing gets stuck. I updated my first message with new errors.

Great, thanks!

I've merged a fix for a negative exponent like `2e-2`, which is conflicting with a change you made.

solang-parser/src/test.rs
```diff
@@ -1003,3 +1007,113 @@ contract C {
...
+        });
+
+        match done_rx.recv_timeout(d) {
+            Ok(_) => Ok(handle.join().expect("Thread panicked")),
+            Err(_) => Err(format!("Thread timeout-ed after {d:?}")),
+        }
+    }
+
+    let source_delimiter = regex::Regex::new(r"====.*====").unwrap();
+    let error_matcher = regex::Regex::new(r"// ----\n// \w+( \d+)?:").unwrap();
```
seanyoung, line 1033: On Windows, this regex fails because the files are checked out with dos end of lines (`\r\n`).

The regex should it should be: `r"// ----\r?\n// \w+( \d+)?:"` and then the tests pass (on my windows machine at least)

shekhirin: fixed

LucasSte (approved):
Looks good!

PR #788, opened 2022-05-02
feat(parser): Gwei unit

shekhirin:
Support for `gwei` unit was added in Solidity 0.6.11: https://github.com/ethereum/solidity/blob/develop/Changelog.md#0611-2020-07-07


seanyoung (approved):
LGTM

PR #793, opened 2022-05-04
Add support for functions without parameters

LucasSte:
While working on code generation for Yul statements, I found that the parser did not support yul functions declared without parameters. This PR updates the parser to support this.

### DIFF
--- a/solang-parser/src/solidity.lalrpop
+++ b/solang-parser/src/solidity.lalrpop
@@ -890,17 +890,13 @@ YulExpression: YulExpression = {
 }
 
 YulFunctionDefinition: YulStatement = {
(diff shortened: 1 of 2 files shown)


solang-parser/src/solidity.lalrpop
```diff
@@ -890,11 +890,16 @@ YulExpression: YulExpression = {
...
     <body: YulBlock> <r:@R> => {
         let returns = match returns {
             Some(content) => content,
-            None => vec![]
+            None => vec![],
+        };
+
+        let params = match params {
+            Some(content) => content,
+            None => vec![],
```
seanyoung, line 902: `params.unwrap_or_default()` (same for returns)

seanyoung (changes requested):

seanyoung (approved):

PR #794, opened 2022-05-04
infrastructure we need for spl-token and calls to rust contracts

seanyoung:
We need to be able to:
 - Specify the AccountMetas for calls
 - Encode the arguments using a struct


docs/language.rst
```diff
@@ -1757,6 +1757,60 @@ The syntax for calling external call is the same as the external call, except fo
...
 
+Internal calls and externals calls
+___________________________________
+
+An internal function call is one that is executed by the current contract. This
+is much more efficient than an external call. An external call requires the
+address of the contract to call, the arguments must be *abi encoded* (also known
+as serialization), then the runtime must set up the VM for the called contract
+(the callee). It must decode the arguments, and then encode return values. Lastly
+the first contract (the caller) must decode return values.
```
LucasSte, line 1768:
```suggestion
An internal function call is executed by the current contract. This
is much more efficient than an external call, which requires the
address of the contract to call, whose arguments must be *abi encoded* (also known
as serialization). Then, the runtime must set up the VM for the called contract
(the callee), decode the arguments, and encode return values. Lastly,
the first contract (the caller) must decode return values.
```

docs/language.rst
```diff
@@ -1757,6 +1757,60 @@ The syntax for calling external call is the same as the external call, except fo
...
+
+Whenever a call is done on a contract type, this will intiate an external call.
```
LucasSte, line 1770:
```suggestion
A call done on a contract type will always initiate an external call.
```

docs/language.rst
```diff
@@ -1757,6 +1757,60 @@ The syntax for calling external call is the same as the external call, except fo
...
+Note that ``this`` returns the current contract, so ``this.foo()`` will do an
+external call, which is much more expensive than ``foo()``.
+
+Passing accounts on Solana on external calls
```
LucasSte, line 1774:
```suggestion
Passing accounts on Solana during external calls
```
I just wanted to avoid the repetition of `on`. Feel free to change the preposition, if `during` does not sound good.

docs/language.rst
```diff
@@ -1757,6 +1757,60 @@ The syntax for calling external call is the same as the external call, except fo
...
+____________________________________________
+
+The Solana runtime allows you the specify the accounts to be passed for an
+external call. This is specified in an array of the struct ``AccountMeta``. This
```
LucasSte, line 1778: Is `AccountMeta` going to be a reserved word?

LucasSte: We could do something like `SolanaSdk.AccountMeta`.

src/emit/solana.rs
```diff
@@ -2955,39 +2955,172 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
...
+                    "payload_len",
+                ),
+            );
+
+            let parameters = self.sol_parameters(binary);
+
+            let account_infos = binary.builder.build_pointer_cast(
+                binary
+                    .builder
+                    .build_struct_gep(parameters, 0, "ka")
```
LucasSte, line 3054: What does `ka` mean?

LucasSte (changes requested):

src/emit/solana.rs
seanyoung: The `ka` field of `SolParameters`

docs/language.rst
seanyoung: Now you've made `SolanaSdk` a reserved word (and it's really ugly).

docs/language.rst
seanyoung: changed it to with

docs/language.rst
seanyoung: All good, I've amened the commit with your suggestion

docs/language.rst
seanyoung: All good, I've amened the commit with your suggestion

docs/language.rst
seanyoung: So I've decided to bite the bullet and go ahead with a change I had planned for some time.

If we want to introduce new types, or builtin functions, we should hide them using solidity imports. The import system used by solidity mimics what javascript does. In javascript/node, builtins (apart from ancient onces) are hidden behind imports.

I've change the code and now there is an import called:

```
import 'solana';
```
Which imports `AccountMeta` and `AccountInfo`. You can do all the renaming tricks that import permits.

This keeps the namespace clean. 

On Solana we also have a `verifySignature()` call that should go into this import (something for another PR)

docs/language.rst
```diff
@@ -1757,6 +1757,64 @@ The syntax for calling external call is the same as the external call, except fo
...
+Note that ``this`` returns the current contract, so ``this.foo()`` will do an
+external call, which is much more expensive than ``foo()``.
+
+Passing accounts with external calls on Solana
+______________________________________________
+
+The Solana runtime allows you the specify the accounts to be passed for an
+external call. This is specified in an array of the struct ``AccountMeta``, which
+has to be imported from the builtin file ``'solana'``. This import is builtin when
+you build your Solidity for Solana.
```
LucasSte, line 1780: It is a nice idea do add a section in the docs to talk about the contents of this 'solana' file.

LucasSte: Specify the structs available, their attributes and the latter's types.

LucasSte (changes requested):

docs/language.rst
seanyoung: I've introduced a builtin imports section.

docs/language.rst
```diff
@@ -136,6 +136,79 @@ This also has a slightly more baroque syntax, which does exactly the same.
 
     import * as defs from "defines.sol";
 
+Builtin Imports
+_______________
+
+Some builtin functionality only available after importing. The following structs
```
LucasSte, line 142:
```suggestion
Some builtin functionality is only available after importing. The following structs
```

LucasSte (approved):
Looks good. I found one minor nit. You can just click `accept suggestion` to change the file without needing to commit anything.

docs/language.rst
seanyoung: Thanks for spotting that. Good reviews, I appreciate it!

PR #806, opened 2022-05-06
allow base contracts and libraries to be specified via import object

seanyoung:
A contract can be imported via an import object. For example, say you have
a file called a.sol:

	contract A {}
	library L {}

Now, another file can import this via an import object:

	import 'a.sol' as IMP;

	contract C is IMP.A {
		using IMP.L for *;
		constructor() IMP.A() {}
	}


src/sema/mod.rs
```diff
@@ -614,6 +614,47 @@ impl ast::Namespace {
         None
     }
 
+    /// Resolve a contract name with namespace
+    pub fn resolve_contract_with_namespace(
```
LucasSte, line 618: We shouldn't be adding all the functions in mod.rs, because it is getting too big. Perhaps we can create a file `namespace.rs` to contain the `impl` for `Namespace`. What do you think?

seanyoung: I think it's a great idea. I've done this change now.

Now `src/sema/mod.rs` is nice and short.

src/sema/variables.rs
```diff
@@ -303,11 +303,11 @@ pub fn variable_decl<'a>(
         None
     };
 
-    let bases: Vec<&str> = if let Some(contract) = contract {
+    let bases = if let Some(contract) = contract {
         contract
             .base
             .iter()
-            .map(|base| -> &str { &base.name.name })
```
LucasSte, line 310: `name.name` was so ugly. I'm happy to see it is gone!

LucasSte (approved):

Issue #802, opened 2022-05-06
Allow `using` for list of file-level or library functions


shekhirin:
```solidity
function id(uint x) pure returns (uint) {
    return x;
}

function zero(uint) pure returns (uint) {
    return 0;
}

contract C {
    function f(uint z) pure external returns(uint) {
        return z.id();
    }

    function g(uint z) pure external returns (uint) {
        return z.zero();
    }

    using {id, zero} for uint;
}
```

Currently, parser cannot deal with it:
```console
Diagnostic { level: Error, ty: ParserError, pos: File(0, 310, 311), message: "unrecognised token '{', expected \"case\", \"default\", \"error\", \"leave\", \"revert\", \"switch\", identifier", notes: [] }
```

found by https://github.com/hyperledger-labs/solang/pull/787

PR #814, opened 2022-05-07
Permit function list in using syntax

seanyoung:
This is new syntax introduced with solc 0.8.13.

	function foo(int) {};
	function foo2(int) {};

	contract c {
		using {foo,foo2} for int;

		function test(int c) public {
			c.foo();
			c.foo2();
		}
	}


seanyoung:
Fixes #802

docs/language.rst
```diff
@@ -2201,6 +2201,110 @@ not allowed to call functions or read variables in the initializer:
         uint constant byzantium_block = 4_370_000;
     }
 
+
+Binding methods to types with ``using``
+---------------------------------------
+
+The builtin types and the any user-defined types like structs, can be extended
+with methods using the ``using`` syntax. This can be done either using libraries
```
LucasSte, line 2209:
```suggestion
We can bind methods to builtin types and any user-defined types like structs using the ``using`` syntax. This can be done either using libraries
```
Please, note that if you decide not to accept this suggestion, the comma placed between "structs" and "can be" should be removed, because we must not split the subject and the verb with a comma.

docs/language.rst
```diff
@@ -2201,6 +2201,110 @@ not allowed to call functions or read variables in the initializer:
...
+or free standing functions.
+
+``using`` with free standing functions
+______________________________________
+
+First, declare a function with 1 or more arguments. The first argument must
```
LucasSte, line 2215:
```suggestion
First, declare a function with one or more arguments. The first argument must
```
This is just a matter of styling.

docs/language.rst
```diff
@@ -2201,6 +2201,110 @@ not allowed to call functions or read variables in the initializer:
...
+
+    using UserLibrary for User global;
+
+Scope for ``using``
+___________________
+
+The ``using`` statement may be scoped in various ways:
+
+  - Globally by adding the ``global`` keyword
+  - Per file, by omitteding the ``global`` keyword
```
LucasSte, line 2302:
```suggestion
  - Per file, by omitting the ``global`` keyword
```

docs/language.rst
```diff
@@ -2201,6 +2201,110 @@ not allowed to call functions or read variables in the initializer:
...
+  - Per contract, by putting the ``using`` declaration in a contract definition
+
+If the scope is per contract, then the type maybe be replaced with ``*`` and
+the type from the first argument of the function will be used.
```
LucasSte, line 2306: Please, provide an example for this case.

src/sema/expression.rs
```diff
@@ -6326,118 +6323,119 @@ fn resolve_using(
     ns: &mut Namespace,
     resolve_to: ResolveTo,
 ) -> Result<Option<Expression>, ()> {
-    // first collect all possible libraries that match the using directive type
+    // first collect all possible functions that could be used for using
     // Use HashSet for deduplication.
```
LucasSte, line 6327: `expression.rs` has more than 7,000 lines. If you can, please place the `fn resolve_using` in `using.rs`

src/sema/using.rs
```diff
@@ -0,0 +1,200 @@
+use super::ast;
+use crate::parser::pt;
+use crate::parser::pt::CodeLocation;
+use std::collections::HashSet;
+
+/// Resolve a using declaration in either file scope or contract scope
+pub fn using_decl(
```
LucasSte, line 7: This function is not used outside sema, please make it `pub(crate)`.

src/sema/using.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    ns.diagnostics.extend(diagnostics);
+
+    Ok(ast::Using { list, ty, file_no })
+}
+
+/// Given the using declarations, find all the possible functions that could be called via using
+/// for the given name, type and file scope
+pub fn possible_functions(
```
LucasSte, line 166: This function is not used outside sema, please make it pub(crate).

LucasSte (changes requested):

docs/language.rst
seanyoung: I think in technical writing, it is best to avoid "we" or "you". I've specifically avoided this in the documentation so far.

docs/language.rst
seanyoung: done

docs/language.rst
LucasSte: I've seen many technical publications using "we", including research papers. I don't think it is that bad. In addition, I've always heard that we should avoid using passive voice in English. If you want to avoid we, maybe this is a better way to write the phrase:
```suggestion
Custom methods can be bound to builtin types and any user-defined types like structs using the ``using`` syntax. This can be done either using libraries
```

seanyoung: In papers "we" means the authors of the paper did something. Here it is unclear who "we" refers to.

seanyoung: I've adopted your change without the "custom", thanks

LucasSte: Feel free to remove `custom`. This is just a suggestion.

PR #815, opened 2022-05-09
doccomment are allowed anywhere

seanyoung:
contract c {}
	/// foo

The comment foo is not attached to anything.

This fixes https://github.com/hyperledger-labs/solang/issues/812


src/sema/tags.rs
```diff
@@ -1,19 +1,166 @@
...
+                comments: single_tags,
+            }),
+            _ => {}
+        }
+    }
+
+    tags
+}
+
+/// Convert the comment to lines, stripping
```
LucasSte, line 112: Is this comment complete?

PR #823, opened 2022-05-10
Permit negative components in scientific notation

### LINKED ISSUE #819: Negative power cause solang to panic
```
contract c {
    function g() public returns(uint) {
        return 200e-2 wei;
    }
}
```

Found in #787

seanyoung:
200e-2 is a NumberLiteral (2)
2e-2 is a RationalNumber(0.02)

Fixes #819


solang-parser/src/lexer.rs
```diff
@@ -353,6 +353,7 @@ pub enum LexicalError {
     DoublePoints(Loc),
     UnrecognisedDecimal(Loc),
     ExpectedFrom(Loc, String),
+    ExponantTooLarge(Loc, String),
```
LucasSte, line 356: Isn't it `Exponent` instead of `Exponant`?

seanyoung: Yes, thanks for catching that

PR #824, opened 2022-05-10
Partial support for Yul

LucasSte:
This PR enables partial support for Yul in code generation. Many builtins are still missing, like `store` and `load`, and will be added on future PRs. In addition, the Yul `switch` statement requires the LLVM switch instruction, which will also be implemented in a future PR.

The Yul codegen still does not work with the undefined variable detection and with unused variable elimination. This feature is also left for another PR.

In addition to partial Yul support, this PR fixes a bug in common subexpression elimination, in which it was using a variable that was unavailable in a branch.

Documentation for Yul is to follow in another PR.


src/codegen/cfg.rs
```diff
@@ -503,13 +504,23 @@ impl ControlFlowGraph {
                 self.expr_to_string(contract, ns, l),
                 self.expr_to_string(contract, ns, r)
             ),
-            Expression::Divide(_, _, l, r) => format!(
-                "({} / {})",
+            Expression::SignedDivide(_, _, l, r) => format!(
+                "signed({} / {})",
```
seanyoung, line 508: This reads like "cast the result of divide to unsigned". That is not very clear. How about "(unsigned divide {} / {})"?

src/codegen/statements.rs
```diff
@@ -1362,22 +1363,22 @@ impl LoopScopes {
         LoopScopes(LinkedList::new())
     }
 
-    fn new_scope(&mut self, break_bb: usize, continue_bb: usize) {
+    pub(crate) fn new_scope(&mut self, break_bb: usize, continue_bb: usize) {
```
seanyoung, line 1366: `pub(super)` ?

src/codegen/statements.rs
```diff
@@ -1362,22 +1363,22 @@ impl LoopScopes {
...
         self.0.pop_front().expect("should be in loop scope")
     }
 
-    fn do_break(&mut self) -> usize {
+    pub(crate) fn do_break(&mut self) -> usize {
         self.0.front().unwrap().break_bb
     }
 
-    fn do_continue(&mut self) -> usize {
+    pub(crate) fn do_continue(&mut self) -> usize {
```
seanyoung, line 1381: `pub(super)` (and others)

src/sema/yul/expression.rs
```diff
@@ -386,6 +386,13 @@ pub(crate) fn resolve_function_call(
 
     if let Some(built_in) = parse_builtin_keyword(func_call.id.name.as_str()) {
         let prototype = &built_in.get_prototype_info();
+        if !prototype.is_available(&ns.target) {
+            ns.diagnostics.push(Diagnostic::error(
+                func_call.loc,
+                format!("builtin '{}' is not available for target {}. Please, open a GitHub issue if there is need to support this function.", prototype.name, ns.target)
```
seanyoung, line 392: What do you think of including a link to our issue page?

src/codegen/statements.rs
LucasSte: What's the advantage of `pub(super)` here? I understand that this function need to be visible in the current create, so that I can use it dealing with yul statements.

src/sema/yul/expression.rs
LucasSte: That sounds cools. I avoided the link to make the message shorter, but I can add it anyways.

src/codegen/statements.rs
LucasSte: @seanyoung

src/codegen/yul/builtin.rs
```diff
@@ -0,0 +1,513 @@
...
+        | YulBuiltInFunction::Log0
+        | YulBuiltInFunction::Log1
+        | YulBuiltInFunction::Log2
+        | YulBuiltInFunction::Log3
+        | YulBuiltInFunction::Log4
+        // origin is the same as tx.origin and is not implemented
+        | YulBuiltInFunction::Origin
+        => {
+            let function_ty = builtin_ty.get_prototype_info();
+            unimplemented!("{} yul builtin not implemented", function_ty.name);
```
seanyoung, line 145: please don't make solang crash

src/codegen/yul/statements.rs
```diff
@@ -0,0 +1,443 @@
...
+            ns,
+            cfg,
+            vartab,
+            early_return,
+            opt,
+        ),
+
+        YulStatement::Switch { .. } => {
+            // Switch statements should use LLVM switch instruction, which requires changes in emit.
+            unimplemented!("Switch statements for yul are not implemented yet");
```
seanyoung, line 70: please don't make solang crash

seanyoung (changes requested):

src/codegen/yul/builtin.rs
LucasSte: Solang will never crash here, because there is an error in sema.

src/codegen/yul/statements.rs
LucasSte: Solang will never crash here, because there is an error in sema.

src/codegen/statements.rs
LucasSte: `pub(crate)` means that the function is only available in the `codegen` crate.

src/codegen/subexpression_elimination/common_subexpression_tracker.rs
```diff
@@ -174,4 +204,113 @@ impl CommonSubExpressionTracker {
...
+    fn find_ancestor(&self, start_block: usize, colors: &[u8]) -> usize {
+        let mut candidate = start_block;
+        let mut queue: VecDeque<usize> = VecDeque::new();
+        let mut visited: HashSet<usize> = HashSet::new();
+
+        visited.insert(start_block);
+        queue.push_back(start_block);
+
+        let mut counter: usize;
+        let mut six_child: usize = 0;
```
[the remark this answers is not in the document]
LucasSte: I think this is a matter of style. As I change them every loop, I prefer to leave them outside it.

src/codegen/subexpression_elimination/common_subexpression_tracker.rs
```diff
@@ -174,4 +204,113 @@ impl CommonSubExpressionTracker {
...
+
+        We can't add colors to code. Here, adding 2 to a block's color mean painting with yellow.
+        Likewise, adding 4 means painting with blue. Green blocks have 6 (2+4) as their color
+        number.
+
+         */
+
+        self.coloring_dfs(block_1, 0, 2, &mut colors, &mut visited);
+        visited.fill(false);
+        self.coloring_dfs(block_2, 0, 4, &mut colors, &mut visited);
```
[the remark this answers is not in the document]
LucasSte: This is a great idea!

src/sema/yul/expression.rs
LucasSte: Due to this error, Solang will never reach the `unimplemented!` function in codegen.

src/sema/yul/statements.rs
```diff
@@ -108,6 +108,13 @@ pub(crate) fn resolve_yul_statement(
                 ns,
             )?;
             resolved_statements.push(resolved_switch.0);
+            ns.diagnostics.push(
+                Diagnostic::error(
+                    switch_statement.loc,
+                    "switch statements have no implementation in code generation yet. Please, file a GitHub issue \
+                    if there is urgent need for such a feature".to_string()
+                )
+            );
```
LucasSte, line 117: @seanyoung Due to this error, Solang will never reach the `unimplemented!` function in `codegen/yul/statement.rs`

src/codegen/yul/builtin.rs
seanyoung: should it be `unreachable!(..)` in that case?

I read `unimplemented!()` as "some code should be inserted here, developer has not finished their job"

src/codegen/statements.rs
seanyoung: `codegen` isn't a crate, it's a module.

`solang` is the crate.

seanyoung (changes requested):

src/codegen/yul/builtin.rs
```diff
@@ -0,0 +1,514 @@
...
+
+    match builtin_ty {
+        YulBuiltInFunction::Add => {
+            Expression::Add(*loc, left.ty(), true, Box::new(left), Box::new(right))
+        }
+        YulBuiltInFunction::Sub => {
+            Expression::Subtract(*loc, left.ty(), true, Box::new(left), Box::new(right))
+        }
+        YulBuiltInFunction::Mul => {
+            Expression::Multiply(*loc, left.ty(), true, Box::new(left), Box::new(right))
```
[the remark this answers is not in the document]
LucasSte: The overflow checking is always off. I thought that `true` would indicate that because of this line:

https://github.com/hyperledger-labs/solang/blob/7f113eba256dbd38948c9d4cd41db8e034621758/src/emit/mod.rs#L5313

If unchecked is true, then we build an operation with overflow detection.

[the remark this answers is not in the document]
seanyoung: Yes, you are right. Somehow I thought that the `bool` field meant "checked" and not "unchecked"

seanyoung (approved):
Looks great!

PR #825, opened 2022-05-10
Do not load from storage when reference is needed

### LINKED ISSUE #789: LLVM crashes when building for Solana
The following contract crashes LLVM when targeting Solana. It works, however, if we target Substrate.

```solidity
contract foo {
    struct S { int f1; }
        S[] arr;

    function g() private returns (S storage, S storage) {
        return (arr[0], arr[1]);
    }

    function f() public returns (S, S) {
        S[] storage ptrArr = arr;
        ptrArr.push(S({f1: 1}));
        ptrArr.push(S({f1: 2}));
        return g();
    }
}
```

Error:
```
Assertion failed: (cast<PointerType>(getOperand(1)->getType()) ->isOpaqueOrPointeeTypeMatches(getOperand(0)->getType()) && "Ptr must be a pointer to Val type!"), function AssertOK, file Instructions.cpp, line 1482.
```

LucasSte:
This PR fixes #789. We were loading a value from storage whenever we encountered a storage reference. However, we should not load if we want the reference, not the value.

To solve this, modified `fn destructure_load` and renamed it to `fn cast_and_try_load`. Another option would be merging such a function with `codegen::Expression::cast`, however I am not sure if the cast function should be allowed to manage the CFG.


src/codegen/statements.rs
```diff
@@ -970,33 +966,42 @@ fn destructure(
 /// During a destructure statement, sema only checks if the cast is possible. During codegen, we
 /// perform the cast to a StorageRef if that is the case. The cast function, however, does not add
 /// instructions to the CFG. Here, we add a load instruction to fetch the variables from storage.
-fn destructure_load(
+fn cast_and_try_load(
```
seanyoung, line 969: The comment doesn't say that a cast also from e.g. `uint32` to `uint64` is done.

seanyoung (changes requested):

src/codegen/statements.rs
```diff
@@ -950,12 +950,8 @@ fn destructure(
...
             }
             DestructureField::Expression(left) => {
-                let expr = destructure_load(
-                    &left.loc(),
-                    &right.cast(left.ty().deref_any(), ns),
-                    cfg,
-                    vartab,
-                );
+                let expr =
+                    cast_and_try_load(&left.loc(), &right, left.ty().deref_any(), ns, cfg, vartab);
```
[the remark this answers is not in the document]
LucasSte: Thanks for spotting that! I've fixed it.

src/codegen/statements.rs
LucasSte: Fixed

seanyoung (approved):

PR #832, opened 2022-05-12
v0.1.11: Nuremberg release

seanyoung:
### Added
- Added support for Solidity user types
- Support `using` syntax on file scope
- Support binding functions with `using`
- Implemented parsing and semantic analysis of yul (code generation is to
  follow)
- The language server uses the `--import` and `--importmap` arguments
- On Solana, it is possible to set the accounts during CPI using the
  `accounts:` call argument.

### Fixed
- Fixed associativity of the power operator
- A huge amount of fixes improving compatibility with solc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 All notable changes to [Solang](https://github.com/hyperledger-labs/solang/)
 will be documented here.
 
+## [0.1.11]
+
+### Added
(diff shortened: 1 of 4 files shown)


seanyoung:
Please check we mention every interesting change in the changelog @LucasSte

seanyoung:
ci *should * resolve itself :crossed_fingers:

PR #833, opened 2022-05-12
Use IdentifierPath instead of Expression in solang-parser

jpopesculian:
Using expression provides an unnecessary overhead to matching against fairly simple identifier paths, and this change seeks to simplify that

### DIFF
--- a/solang-parser/src/pt.rs
+++ b/solang-parser/src/pt.rs
@@ -83,6 +83,33 @@ pub struct Identifier {
     pub name: String,
 }
 
+impl Display for Identifier {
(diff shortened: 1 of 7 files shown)


seanyoung (changes requested):
That's a nice change! The code change looks good.

However hyperledger requires a `Signed-off-by: real name <email@address>` in the commit message to sign it off.

jpopesculian:
Ah gotcha! I'll amend it! I see all kinds of issues, I gotta fix up first too haha...

seanyoung:
make sure you test with `cargo test --workspace` locally

jpopesculian:
I see the DCO doesn't like the force push... I'll open a new PR

PR #834, opened 2022-05-12
Use IdentifierPath instead of Expression in solang-parser

jpopesculian:
[Using expression provides an unnecessary overhead to matching against fairly simple identifier paths, and this change seeks to simplify that](https://github.com/hyperledger-labs/solang/pull/833)

### DIFF
--- a/src/sema/namespace.rs
+++ b/src/sema/namespace.rs
@@ -309,36 +309,21 @@ impl Namespace {
     pub fn resolve_contract_with_namespace(
         &mut self,
         file_no: usize,
-        expr: &pt::Expression,
+        name: &pt::IdentifierPath,
         diagnostics: &mut Vec<Diagnostic>,
     ) -> Result<usize, ()> {
-        let (namespace, id, dimensions) = self.expr_to_type(file_no, None, expr, diagnostics)?;
+        let (id, namespace) = name
+            .identifiers
+            .split_last()
+            .map(|(id, namespace)| (id, namespace.iter().collect()))
+            .unwrap();
 
-        if !dimensions.is_empty() {
-            diagnostics.push(Diagnostic::decl_error(
-                expr.loc(),
-                "array type found where contract type expected".to_string(),
-            ));
-            return Err(());
-        }
-
-        let id = match id {
-            pt::Expression::Variable(id) => id,
-            _ => {
-                diagnostics.push(Diagnostic::decl_error(
-                    expr.loc(),
-                    "expression found where contract type expected".to_string(),
-                ));
-                return Err(());
-            }
-        };
-
-        let s = self.resolve_namespace(namespace, file_no, None, &id, diagnostics)?;
+        let s = self.resolve_namespace(namespace, file_no, None, id, diagnostics)?;
 
         if let Some(Symbol::Contract(_, contract_no)) = s {
             Ok(*contract_no)
         } else {
-            let error = Namespace::wrong_symbol(s, &id);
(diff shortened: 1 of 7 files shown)


src/sema/namespace.rs
```diff
@@ -309,36 +309,25 @@ impl Namespace {
...
-        expr: &pt::Expression,
+        name: &pt::IdentifierPath,
         diagnostics: &mut Vec<Diagnostic>,
     ) -> Result<usize, ()> {
-        let (namespace, id, dimensions) = self.expr_to_type(file_no, None, expr, diagnostics)?;
-
-        if !dimensions.is_empty() {
+        let (id, namespace) = if let Some((id, namespace)) = name.identifiers.split_last() {
+            (id, namespace.iter().collect())
+        } else {
```
seanyoung, line 317: The else should never occur. The `name.identifiers` should have at least one element, or there is a bug in the parser.

This might was well be an `.unwrap()` and get rid of the diagnostic.

src/sema/namespace.rs
```diff
@@ -309,36 +309,25 @@ impl Namespace {
...
         file_no: usize,
-        expr: &pt::Expression,
+        name: &pt::IdentifierPath,
         diagnostics: &mut Vec<Diagnostic>,
     ) -> Result<usize, ()> {
-        let (namespace, id, dimensions) = self.expr_to_type(file_no, None, expr, diagnostics)?;
-
-        if !dimensions.is_empty() {
+        let (id, namespace) = if let Some((id, namespace)) = name.identifiers.split_last() {
+            (id, namespace.iter().collect())
```
seanyoung, line 316: `.iter().collect()` can be `.to_vec()`, I think (not terribly important)

src/sema/namespace.rs
```diff
@@ -350,36 +339,25 @@ impl Namespace {
...
+        } else {
             diagnostics.push(Diagnostic::decl_error(
-                expr.loc(),
-                "array type found where function expected".to_string(),
+                name.loc,
+                "empty identifier path".to_string(),
```
seanyoung, line 350: Same as above, no need for diagnostic. The compiler should panic.

seanyoung (changes requested):
Your PR is very nice, this makes the code shorter and nicer.

The DCO bot is being difficult. This is the error:

Commit sha: [cbaba2f](https://github.com/hyperledger-labs/solang/pull/834/commits/cbaba2fe40a41d546772dc0ac699c9af84fbd7c0), Author: Julian Popescu, Committer: Julian Popescu; Expected "Julian Popescu [jpopesculian@gmail.com](mailto:jpopesculian@gmail.com)", but got "Julian Popescu [hi@julian.dev](mailto:hi@julian.dev)".

Looks like the problem is that you used a different email address in the signed-off-by than the commit author.

src/sema/namespace.rs
jpopesculian: I thought about this too, but I don't if this is possible here as `split_last` returns `(&T, &[T])` and I don't know if its possible to easily convert a `&[T]` to a `[&T]` to convert directly to a `Vec<&T>` like the call signature expects

jpopesculian:
Ahh I fixed my gitconfig. For some reason, I was expecting it to look at the GPG signature...

src/sema/namespace.rs
seanyoung: Right, I had not thought of that. Looks great!

seanyoung (approved):

seanyoung:
LGTM

PR #836, opened 2022-05-13
Parse `(new ty){value: 1}(args)`

seanyoung:
Three different parse tree for callargs:

	(new D{value: 1})();
	(new D){value: 1}();
	new D{value: 1}();

solc accepts this syntax, so should we.

Fixes https://github.com/hyperledger-labs/solang/issues/818


tests/substrate_tests/arrays.rs
```diff
@@ -743,7 +743,7 @@ fn dynamic_array_push() {
 
         contract foo {
             function test() public {
-                int[] bar = new int[](1);
+                int[] bar = (new int[])(1);
```
LucasSte, line 746: Why are you adding parenthesis here? Shouldn't we keep testing the old syntax too?

seanyoung: There are many tests for the existing syntax already. I wanted to add tests for the other syntax too.

There are a few in the same file, here for example.

https://github.com/hyperledger-labs/solang/blob/1c81845dda633e7e749978c57bdc99b9b105d65d/tests/substrate_tests/arrays.rs#L556

LucasSte (approved):

Issue #807, opened 2022-05-06
LLVM Crashes when building contract


LucasSte:
The following contract crashes LLVM, when building for target ewasm.

```solidity
/**
 *Submitted for verification at Etherscan.io on 2022-03-14
*/

//SPDX-License-Identifier: Unlicensed

pragma  solidity  ^0.8.5;
abstract  contract  Context  {
        function  _msgSender()  internal  view  virtual  returns  (address  payable)  {
                return  payable(msg.sender);
        }
        function  _msgData()  internal  view  virtual  returns  (bytes  memory)  {
                this;
                return  msg.data;
        }
}

contract  Ownable  is  Context  {
        address  private  _owner;
      
        event  OwnershipTransferred(address  indexed  previousOwner,  address  indexed  newOwner);
        constructor  ()  {
                address  msgSender  =  _msgSender();
                _owner  =  msgSender;
                emit  OwnershipTransferred(address(0),  msgSender);
        }
        function  owner()  public  view  returns  (address)  {
                return  _owner;
        }

        modifier  onlyOwner()  {
                require(_owner  ==  _msgSender(),  "Ownable:  caller  is  not  the  owner");
                _;
        }

        function  waiveOwnership()  public  virtual  onlyOwner  {
                emit  OwnershipTransferred(_owner,  address(0));
                _owner  =  address(0);
        }

        function  transferOwnership(address  newOwner)  public  virtual  onlyOwner  {
                require(newOwner  !=  address(0),  "Ownable:  new  owner  is  the  zero  address");
                emit  OwnershipTransferred(_owner,  newOwner);
                _owner  =  newOwner;
        }
        
        function  getTime()  public  view  returns  (uint256)  {
                return  block.timestamp;
        }
}

```

The error is the following:
```
llvm-project/llvm/lib/IR/Instructions.cpp:1480: void llvm::StoreInst::AssertOK(): Assertion `cast<PointerType>(getOperand(1)->getType()) ->isOpaqueOrPointeeTypeMatches(getOperand(0)->getType()) && "Ptr must be a pointer to Val type!"' failed.
```

PR #837, opened 2022-05-17
Refactor emit

LucasSte:
In order to solve #807, I need to implement changes in Binary. This PR removes Binary from `emit/mod.rs`, so that I can work on it without modifying the `TargetRuntime`.

The issue of LLVM renaming our types happens because we load the stdlib multiple times. Instead, we must load it once, build an LLVM module and link this same module to all other ones we create. Once this PR is merged, I'll submit my changes.

In addition, I split the contract test into three functions, so that they run in parallel and make our testing faster.


seanyoung (approved):
Looks good!

PR #838, opened 2022-05-18
Load stdlib only once

### LINKED ISSUE #807: LLVM Crashes when building contract

LucasSte:
This PR fixes #807. We are now only loading the stdlib module only once and linking it against any other modules we create. Although LLVM types are global, I believe LLVM adds a `.0` to the end of the type name for disambiguation, because it does not check if two types of equal names represent the same structure in the memory. In this sense, whenever we load stdlib again, LLVM understands that is a totally new module and tries to disambiguate type names.

Please, @seanyoung check if I did not remove any necessary pointer cast.


seanyoung (changes requested):
Looks good, however I found one pointer_cast() which can now be removed, I think.
binary.rs:926

seanyoung:
The burrow test is failing because of https://www.npmjs.com/package/solc/v/0.8.14 is broken (I think).

LucasSte:
> Looks good, however I found one pointer_cast() which can now be removed, I think. binary.rs:926

Thanks for catching that. I've removed that cast too.

LucasSte:
> The burrow test is failing because of https://www.npmjs.com/package/solc/v/0.8.14 is broken (I think).

A fix has already been released: https://www.npmjs.com/package/solc

seanyoung (approved):

PR #839, opened 2022-05-19
Fix parser locations of expressions

seanyoung:
The parser locations for unary and binary expression are inconsistent.
Often only the operator itself is included, and not operand(s). This is
causes incorrect diagnostics.
```
/home/sean/git/solang/tests/contract_testcases/substrate/arrays/array_dimensions.sol:3:21-22: error: zero size array not permitted
Line 3:
	bool[10 - 10] x;
	--------^
```
This is now fixed:
```
error: zero size array not permitted
  ┌─ /home/sean/git/solang/tests/contract_testcases/substrate/arrays/array_dimensions.sol:3:18
  │
3 │             bool[10 - 10] x;
  │                  ^^^^^^^
```
This issue was found by the Foundry work to create a formatter based on
the solang-parser, see https://discord.com/channels/900503503162724452/975868842125455390/976527889166135357

---

This also means we have to deal with overlapping expressions in the language server
    
Now the locations for binary expressions overlaps with their operands. This means that we have overlapping locations for hovers in the language server. We deal with this by finding all matching hovers, and returning the hover which is the shortest.
    
Unfortunately this involves a linear scan of the hovers. This is linear scan of memory, so it should be fast. However, there may be a better data structure we can choose for this, should performance be an issue.

I was thinking we could use something like an r* tree but this is a bit heavy for our use-case. Any suggestions?


LucasSte (approved):
Looks good. If you think of another way to find the right hover, please, submit another PR.

LucasSte (approved):

PR #840, opened 2022-05-20
Parse override(list) syntax on variable definition

seanyoung:
- The parser does not parse an override list specified on a variable definition
- The override list may be specified via an import path.

Found in #787


src/sema/diagnostics.rs
```diff
@@ -34,17 +34,23 @@ impl Diagnostics {
     }
 
     pub fn push(&mut self, diagnostic: Diagnostic) {
-        if matches!(diagnostic.level, Level::Error) {
-            self.has_error = true;
+        if !self.contents.contains(&diagnostic) {
```
LucasSte, line 37: This performs a search trough the entire content of the vector. We push diagnostics many times throughout the code. If you think it is necessary checking for repeated diagnostics, I suggest using a hash table.

src/sema/diagnostics.rs
```diff
@@ -34,17 +34,23 @@ impl Diagnostics {
...
-        self.contents.push(diagnostic);
     }
 
     pub fn extend(&mut self, diagnostics: Vec<Diagnostic>) {
         if !self.has_error {
             self.has_error = diagnostics.iter().any(|m| m.level == Level::Error);
         }
-        self.contents.extend(diagnostics);
+        for diagnostic in diagnostics {
+            if !self.contents.contains(&diagnostic) {
```
LucasSte, line 50: You are searching the entire vector for each new diagnostic!
This search should be improved to a hash table or a binary search.

LucasSte (changes requested):

seanyoung:
> `Diagnostics.push` is by far the function we call the most in Solang. Adding a vector traversal there will degrade performance. You should either create a new function that does not allow duplicate insertion of warnings or change the way you check for duplicates.
> 
> I believe that most warnings do not require checking for duplicates.

So now I've change it, so that:
 - All the diagnostics are sorted by file location, level and then message
 - deduplicates are moved

This is done once.

LucasSte (approved):
Nice solution!

PR #842, opened 2022-05-20
Fix parse location of FunctionCallBlock

jpopesculian:
`Expression::FunctionCallBlock` location does not surround entire
expression, and this fixes that by moving the location start to the
beggining of the expression. This also adds `pretty_assertions`
for easier debugging of dot files in tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ assert_cmd = "2.0"
 bincode = "1.3"
 ed25519-dalek = "1.0"
 path-slash = "0.1"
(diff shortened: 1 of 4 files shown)


jpopesculian:
ref #839

seanyoung:
pretty-assertions is very nice!

github actions is playing up, I'll if I can restart the failed jobs

seanyoung (approved):

LucasSte (approved):
Pretty assertions are a very nice addition for the contracts test.

PR #844, opened 2022-05-23
Refactor Solang docs

LucasSte:
This PR splits the old `language.rst` file into multiple files so that the user can see the sections on the navigation bar. I also re-organized the order of items in the bar.


seanyoung (changes requested):
The language documentation link in index.rst and extension.rst does not work when I build it. I get:

```
/home/sean/git/solang/docs/extension.rst:27: WARNING: undefined label: language (if the link has no caption the label must precede a section header)
```

Also is this PR just moving things without editing? Since language.rst is split into multiple files, the diff shows everything as changed, unfortunately.

LucasSte:
> Also is this PR just moving things without editing? Since language.rst is split into multiple files, the diff shows everything as changed, unfortunately.

There were some minor edits:
I fixed the broken references and changed the syntax highlighting from javascript to solidity whenever possible.

LucasSte:
> The language documentation link in index.rst and extension.rst does not work when I build it. I get:
> 
> ```
> /home/sean/git/solang/docs/extension.rst:27: WARNING: undefined label: language (if the link has no caption the label must precede a section header)
> ```
> 
> Also is this PR just moving things without editing? Since language.rst is split into multiple files, the diff shows everything as changed, unfortunately.

Fixed!

seanyoung (approved):

Issue #813, opened 2022-05-06
Inkwell crashes when building for Solana and Substrate


LucasSte:
The following contract crashes Solang when we target Solana or Substrate. It works however on solc and on Solang for ewasm.

```solidity
contract DTron {

    mapping(address => uint256) public depositedAmount;
    event Received(address, uint);
    event Withdrawn(address, uint);

    function publicDeposit() public payable {
        depositedAmount[msg.sender] = depositedAmount[msg.sender] + msg.value;
        emit Received(msg.sender, msg.value);
    }
	

    function nm() external payable{
        publicDeposit();
    }
    
    function moneyDeposit(address[] memory thanksCash, uint256[] memory amount) public payable {
        uint256 arrayLength = thanksCash.length;
        for(uint256 i=0; i< arrayLength; i++){
            address payable receiver = payable(address(uint160(thanksCash[i])));
            receiver.transfer(amount[i]);
        }
    }
}
```

The error is the following:
```
thread 'main' panicked at 'Found ArrayType(ArrayType { array_type: Type { address: 0x55fc10bd7b10, llvm_type: "[32 x i8]" } }) but expected the IntType variant', /home/lucas/.cargo/registry/src/github.com-1ecc6299db9ec823/inkwell-0.1.0-beta.4/src/types/enums.rs:390:13

```

LucasSte:
This is the simplest failing contract:

```solidity
contract DTron {
    function moneyDeposit(address[] memory thanksCash, uint256[] memory amount) public payable {
        address payable receiver = payable(address(uint160(thanksCash[2])));
        receiver.transfer(amount[2]);
    }
}
```

PR #846, opened 2022-05-23
Do not raise warning for function parameters saved in storage

LucasSte:
This is a minor issue I found while working on issue #813.


src/sema/unused_variable.rs
```diff
@@ -228,7 +229,9 @@ pub fn check_var_usage_expression(
 pub fn emit_warning_local_variable(variable: &symtable::Variable) -> Option<Diagnostic> {
     match &variable.usage_type {
         VariableUsage::Parameter => {
-            if !variable.read {
+            if !variable.read
+                && !matches!(variable.storage_location, Some(StorageLocation::Storage(_)))
+            {
```
seanyoung, line 234: This contact has a function like:
```
            function bar(int[] storage foo) internal {
                foo[0] = 102;
            }
```
and solang says `foo` is unread, which is wrong of course.

However your PR suppresses the warning for `storage` parameters. That's not right, what if we have a function like:
```
            function bar(int[] storage foo) internal {
            }
```
We still want a warning then, right?

seanyoung (changes requested):

src/sema/unused_variable.rs
LucasSte: Yes. I've fixed this.

seanyoung: I'm being pedantic now, but this is hard to read. How about:
```
if !variable.read {
    if !variable.assigned {
       // warn never read/assigned
    } else if !storage {
       // warn never been read
    }
}
```
I am not sure why there is special case for storage variables though.  This needs a comment

seanyoung (changes requested):

src/sema/unused_variable.rs
LucasSte: > I am not sure why there is special case for storage variables though. 

A function like this:
```solidity
function foo(int[] storage bar) {
int[2] = 5;
}
```
should not have a warning about `bar` being unread, because it is a reference to the storage.

seanyoung: why not?

LucasSte:
```solidity

contract test {
int[] vec;
function foo(int[] storage bar) {
    bar[2] = 5;
}

function foo2() returns (int) {
   vec.push(1);
   vec.push(2);
   vec.push(3);
   foo(vec);
   return vec[2];
}

}
```

We would raise a warning about an unread variable in this case, but it is not actually unread, just because the argument is a reference to the storage and not the value itself.

seanyoung: Sorry, yes you are right. However, what about the non-storage version of that:

```solidity
contract test {
function foo(int[] bar) pure {
    bar[2] = 5;
}

function foo2() public returns (int) {
int[] vec = new int[](0);
   vec.push(1);
   vec.push(2);
   vec.push(3);
   foo(vec);
   return vec[2];
}

}
```

Why are we generating a warning for that?

Also in this following case the arguments are never used.

```solidity
contract c {
	int[] var;

	function g(int[] storage x) internal {
		x = var;
	}
	function gg(int[] x) internal {
		x = new int[](2);
	}

}
```

LucasSte: > Why are we generating a warning for that?

The reaching definitions implementation does not work across function calls. This is a limitation of the way we detect unused values.

> Also in this following case the arguments are never used.

There is a warning for `gg`, because `x` is never read. For `g` there is no warning, unfortunately.

PR #850, opened 2022-05-25
refactor(arguments): Replace -O command line option

### LINKED ISSUE #534: Replace `-O less` command line option with `-O0`, `-O3`, `-Oz`, and `-OZ`
You can specifiy the llvm optimization level with `-O less`, `-O default`, etc. This is not the standard way compilers take an argument for optimization. So, we want `-O1`, `-Oz` etc.

This requires changing the clap argument parsing.

Follow what gcc/clang offers (checks this!).

SaladinoBelisario:
Replace llvm-like -O command with gcc-like
optimization
commands
according
to
https://llvm.org/doxygen/CodeGen_8h_source.html

Limit the command
to
only one argument
(one type of optimization)

Add basic help for the
-O
command to improve usability

BREAKING CHANGE: Optimization command -O now takes GCC-like arguments
(0,1,2 or s and 3)

Closes #534

Signed-off-by: Saladino Belisario <SaladinoBelisario@protonmail.ch>


SaladinoBelisario:
Now the optimization command line can be 
> solang -v --target substrate -O 2 ~/Desktop/example.sol 
or 
>  solang -v --target substrate -O2 ~/Desktop/example.sol

src/bin/solang.rs
```diff
@@ -64,9 +64,15 @@ fn main() {
             Arg::new("OPT")
                 .help("Set llvm optimizer level")
                 .short('O')
+                .use_value_delimiter(false)
+                .long_help("\
+None = 0,      // -O0
+Less = 1,      // -O1
+Default = 2,   // -O2, -Os, -OS
+Aggressive = 3 // -O3")
```
seanyoung, line 72: `cargo fmt` does not like it and I'm not a fan either. No need to mention none/less/default/aggressive.

src/bin/solang.rs
```diff
@@ -301,10 +307,13 @@ fn main() {
...
-            "none" => OptimizationLevel::None,
-            "less" => OptimizationLevel::Less,
-            "default" => OptimizationLevel::Default,
-            "aggressive" => OptimizationLevel::Aggressive,
+            "0" => OptimizationLevel::None,
+            "1" => OptimizationLevel::Less,
+            "2" => OptimizationLevel::Default,
+            // According to https://llvm.org/doxygen/CodeGen_8h_source.html -O2 and -Os are default
+            "s" => OptimizationLevel::Default,
+            "S" => OptimizationLevel::Default,
```
seanyoung, line 315: Here is where the change falls apart. We need to tell llvm about the size level, so we need to have a `opt_speed_level` and `opt_size_level`. This needs to be set in the `Options` struct, and then used in `src/emit/binary.rs` to set the llvm size/speed optimization level. See here the inkwell call for this https://github.com/TheDan64/inkwell/blob/master/src/passes.rs#L100-L102

seanyoung (changes requested):
Great start, however llvm needs to be informed of the size level as well as the speed level.

Also the documentation needs updating to mention the new command line.

clang also has an `clang --optimize=2` long option. We may want that too, depending on opinions.

Thanks!

PR #851, opened 2022-05-25
Ensure that contracts tests also emit

seanyoung:
This ensure that the emit stage does not crash on any of the test contracts


LucasSte (approved):
This is a nice addition!

PR #855, opened 2022-05-31
Yul documentation

LucasSte:
This PR adds documentation for Yul. Changes in `docs/conf.py` were necessary to support a new linter for both Solidity and Yul.

### DIFF
--- a/docs/language/inline_assembly.rst
+++ b/docs/language/inline_assembly.rst
@@ -1,0 +1,116 @@
+Inline Assembly
+===============
+
+In Solidity functions, developers are allowed to write assembly blocks containing Yul code. For more information about
+the Yul programming language, please refer to the :ref:`yul section <yul_section>`.
+
+In an assembly block, you can access solidity local variables freely and modify them as well. Bear in mind, however,
+that reference types like strings, vectors and structs are memory addresses in yul, so manipulating them can be unsafe
+unless done correctly. Any assignment to those variables will change the address the reference points to and
+may cause the program to crash if not managed correctly.
+
+.. code-block:: solidity
+
+    contract foo {
+        struct test_stru {
+            uint a;
+            uint b;
+        }
+
+        function bar(uint64 a) public pure returns (uint64 ret) {
+            uint64 b = 6;
+            uint64[] memory vec;
+            vec.push(4);
+            string str = "cafe";
+            test_stru tts = test_stru({a: 1, b: 2});
+            assembly {
+                // The following statements modify variables directly
+                a := add(a, 3)
+                b := mul(b, 2)
+                ret := sub(a, b)
+
+                // The following modify the reference address
+                str := 5
+                vec := 6
+                tts := 7
+            }
+
+            // Any access to 'str', 'vec' or 'tts' here may crash the program.
+        }
+
+    }
+
+
+Storage variables cannot be accessed nor assigned directly. You must use the ``.slot`` and ``.offset`` suffix to use storage
+variables. Storage variables should be read with the ``sload`` and saved with ``sstore`` builtins, but they are not implemented yet.
+Solang does not implement offsets for storage variables, so the ``.offset`` suffix will always return zero.
+Assignments to the offset are only allowed to Solidity local variables that are a reference to the storage.
+
+.. code-block:: solidity
+
+    contract foo {
+        struct test_stru {
+            uint a;
+            uint b;
+        }
+
+        test_stru storage_struct;
+        function bar() public pure {
+            test_stru storage tts = storage_struct;
+            assembly {
+                // The variables 'a' and 'b' contain zero
+                let a := storage_struct.offset
+                let b := tts.offset
+
+                // This changes the reference slot of 'tts'
+                tts.slot := 5
+            }
+        }
+    }
+
+
+
+Dynamic calldata arrays should be accessed with the ``.offset`` and ``.length`` suffixes. The offset suffix returns the
+array's memory address. Assignments to ``.length`` are not yet implemented.
+
+.. code-block:: solidity
+
+    contract foo {
+        function bar(int[] calldata vl) public pure {
+            test_stru storage tts = storage_struct;
+            assembly {
+                // 'a' contains vl memory address
+                let a := vl.offset
+
+                // 'b' contains vl length
+                let b := vl.length
+
+                // This will change the reference of vl
+                vl.offset := 5
+            }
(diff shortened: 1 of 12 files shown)


seanyoung:
With this PR, yul is working so the top level README.md and CHANGELOG.md can be updated

seanyoung:
With this PR, yul is working so the top level README.md and CHANGELOG.md can be updated

docs/yul_language/builtins.rst
```diff
@@ -0,0 +1,202 @@
+Builtins
+========
+
+Most operations in Yul are performed via builtin functions. Solang supports
+most builtins, however memory operations and chain operations are not yet implemented.
+They will be added in a future release.
```
seanyoung, line 6: This section should NOT say "they will be added", because many will never be added. We should add something about the origin of the builtins: 

yul builtins are low level EVM instructions and many are ethereum specific (link to https://ethereum.org/en/developers/docs/evm/opcodes/). On Solana or Substrate these instructions do not exist, and because of low-level differences between the chains, it is impossible to implement some builtins. For example ``delegatecall`` and ``staticcall`` just does not exist as a concept on either chain.

docs/yul_language/functions.rst
```diff
@@ -0,0 +1,31 @@
+Functions
+=========
+
+Functions in Yul cannot access any variable outside their scope, i.e. they can
+only operate with the variables they receive as arguments. You can define types for arguments and returns. If no
+type is specified, the compiler will default to ``u256``.
```
seanyoung, line 6: This is a little terse.

docs/yul_language/functions.rst
```diff
@@ -0,0 +1,31 @@
...
+        {
+            ret1 := mulmod(c, d, e)
+            ret2 := addmod(c, d, e)
+        }
+
+        function noReturns(a, b)
+        {
+            store(a, b)
+        }
+    }
```
seanyoung, line 31: This example does not show how it would be used in solidity or how to do function calls.

docs/yul_language/statements.rst
```diff
@@ -0,0 +1,118 @@
...
+            funcCall(4, 3)
+        } // There cannot be an 'else' here
+    }
+
+
+Switch
+_______
+
+Switch statements are not yet supported in Solang. If there is urgent need to support them,
+please, file a GitHub issue in the official repository.
```
seanyoung, line 46: no need for official

docs/yul_language/statements.rst
```diff
@@ -0,0 +1,118 @@
+Statements
+==========
+
+For-loop
+________
+
+For-loops are completely supported in Solang. ``continue`` and ``break`` statements are also supported.
```
seanyoung, line 7: some words about how for loops are written would be helpful (the init, cond, next block/expression)

docs/yul_language/statements.rst
```diff
@@ -0,0 +1,118 @@
...
+        let j : u32, k : u8 := manyReturns()
+    }
+
+Assignments
+___________
+
+Variables can be assignment using the ``:=`` operator. If the types do not match,
+the compiler performs an implicit cast, so that the right hand side type matches that of the variable.
+Multiple variables can be assigned in a single line if the right hand side is a function call that returns multiple
+values.
```
seanyoung, line 93: Does the compiler ever produce an error/warning?

docs/yul_language/yul.rst
```diff
@@ -0,0 +1,14 @@
...
+
+.. _yul_section:
+
+Yul, also know as EVM assembly, is a low level language for creating smart contracts and for providing more
+control over the execution environment when using Solidity as the primary language. Although it enables
+more possibilities to manage memory, using Yul does not imply a performance improvement. In Solang,
+all Yul constructs are processed using the same pipeline as Solidity.
+
+The support for Yul is only partial. We support all statements, except for the switch-block. In addition,
+some Yul builtins are not yet implemented, but we are trying to improve compatibility. In the following sections,
```
seanyoung, line 12: do not give the impression that all builtins will be implemented.

remove `, but we are trying to improve compatibility`, this is implied and rather than focus on what we haven't done, it's better to focus on what we have done.

seanyoung (changes requested):

docs/yul_language/statements.rst
LucasSte: Yes, it does

docs/yul_language/statements.rst
```diff
@@ -0,0 +1,118 @@
...
+
+
+Variable declaration
+____________________
+
+Variables can be declared in Yul using the `let` keyword. Multiple variables can be declared at the same line
+if there is no initializer or the initializer is a functon that returns multiple values.
+
+The default type for variables in Yul is ``u256``. If you want to declare a variable with another type, use the colon.
+Note that if the variable type and the type of the right hand side of the assignment disagree, there will be an implicit
```
seanyoung, line 74: replace disagree with "are different"

seanyoung: or "do not match"

docs/yul_language/statements.rst
seanyoung: I worry that "implicit cast" won't be understood. How about "converted to the right type, if possible".

docs/yul_language/functions.rst
LucasSte: I didn't want to make it prolix, but I'll improve it.

docs/yul_language/functions.rst
LucasSte: This section is about Yul, not inline assembly, so I should not mention Solidity here. There interaction of Yul and Solidity is described at `inline_assembly.rst`.

docs/language/inline_assembly.rst
```diff
@@ -0,0 +1,116 @@
+Inline Assembly
+===============
+
+In Solidity functions, developers are allowed to write assembly blocks containing Yul code. For more information about
+the Yul programming language, please refer to the :ref:`yul section <yul_section>`.
+
+In an assembly block, you can access solidity local variables freely and modify them as well. Bear in mind, however,
+that strings, vectors and structs are reference types, so accessing them in Yul will return a memory address. Any
```
seanyoung, line 8: I would rephrase this as "reference types like string, .. are memory addresses in yul, so manipulating them can be unsafe unless done correctly"

seanyoung (approved):
Looks great!

PR #862, opened 2022-06-08
Move to latest solana-rbpf crate

seanyoung:
This crate is used for solana tests. The version used before this commit has an security issue:

Solana solana_rbpf before 0.2.29 has an addition integer overflow via invalid ELF program headers. elf.rs has a panic via a malformed eBPF program.


LucasSte (approved):
PRs should not contain unrelated commits, but this is not a reason to not approve this PR.

PR #865, opened 2022-06-08
Fix deletion of address from mappings

LucasSte:
This PR fixes the deletion of addresses on mappings, which was causing the compiler to crash.


src/emit/solana.rs
```diff
@@ -868,6 +869,15 @@ impl SolanaTarget {
 
                 self.storage_free(binary, &field.ty, data, offset, function, zero, ns);
             }
+        } else if matches!(ty, Type::Address(_)) {
```
seanyoung, line 872: I think you need `Type::Contract(_)` here too

seanyoung (changes requested):

src/emit/solana.rs
LucasSte: Fixed

seanyoung (approved):
Looks great

PR #872, opened 2022-06-15
Do not re-export solang_parser from solang crate

seanyoung:
No reason re-export.

We're exporting far too much right now. This is just a simple start.

See https://docs.rs/solang/latest/solang/ for example


LucasSte (approved):
Looks good.
I didn't understand what is the implication of this. I would be happy if you clarify what the problem of using `use crate::parser::pt` is.

PR #873, opened 2022-06-15
WIP: Create test with simple auction contract

seanyoung:
Taken from https://docs.soliditylang.org/en/v0.8.3/solidity-by-example.html#voting

This test is broken in many ways. This PR is just a start.

- [ ] calling bid() fails without any useful error. Ensure that the error is presented in node (I think there is an account missing from the call)
- [ ] make bid function work
- [ ] make the withdraw function work/collecting of winnings
- [ ] deploy auction contract with PDA

### DIFF
--- a/integration/solana/SimpleAuction.sol
+++ b/integration/solana/SimpleAuction.sol
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.7.0 <0.9.0;
+contract SimpleAuction {
+    // Parameters of the auction. Times are either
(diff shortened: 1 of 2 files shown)


seanyoung:
See https://github.com/solana-labs/solana-solidity.js/pull/29 for propagating the erorr

seanyoung:
bid() does not work because the transfer fails: 
Error: failed to verify account Dq9ykKukR3EDLMVMCuND15JEAK5651oj7rpeBDdfnU34: instruction spent from the balance of an account it does not own

seanyoung:
This test will have to be re-written completely once we have anchor/pda in place.

PR #880, opened 2022-06-16
cast struct member reference to wrong type causes panic

seanyoung:
This should be detected during sema.

Fixes https://github.com/hyperledger-labs/solang/issues/867
Fixes https://github.com/hyperledger-labs/solang/issues/868

### DIFF
--- a/src/sema/expression.rs
+++ b/src/sema/expression.rs
@@ -148,15 +148,29 @@ impl Expression {
         }
 
         // First of all, if we have a ref then derefence it
-        if let Type::Ref(r) = from {
+        if let Type::Ref(r) = &from {
             return if r.is_fixed_reference_type() {
-                // Accessing a struct/fixed size array within an array/struct gives
-                // a Type::Ref(..) the element, this type is just a simple pointer,
-                // like it would have been without the Type::Ref(..)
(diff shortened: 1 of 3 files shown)


src/sema/expression.rs
```diff
@@ -148,15 +148,27 @@ impl Expression {
...
+        if let Type::Ref(r) = &from {
             return if r.is_fixed_reference_type() {
                 // Accessing a struct/fixed size array within an array/struct gives
                 // a Type::Ref(..) the element, this type is just a simple pointer,
                 // like it would have been without the Type::Ref(..)
                 //
                 // The Type::Ref(..) just means it can be used as an l-value and assigned
                 // a new value, unlike say, a struct literal.
-                self.cast(loc, &Type::Ref(r), implicit, ns, diagnostics)
+                if r.as_ref() == to {
```
LucasSte, line 159: You could write a comment for this if-else condition here.

LucasSte (changes requested):

src/sema/expression.rs
seanyoung: I hope it makes more sense now

LucasSte (approved):

Issue #785, opened 2022-05-01
Unable to initialize empty mapping as described in solidity documentation


deepaktrama:
Unable to initialize empty mapping as described in solidity documentation. Using solana.
campaigns[campaignID] = Campaign(beneficiary, goal, 0, 0); gives error on line 25 - "struct 'Campaign' has 5 fields, not 4"
Also tried using curly brace syntax and still getting an error.
Will appreciate it if someone can let me know how to work around this. 

Complete code below, it is just copied from solidity docs.
```
contract CrowdFunding {
    // Defines a new type with two fields.
    struct Funder {
        address addr;
        uint amount;
    }

    struct Campaign {
        address beneficiary;
        uint fundingGoal;
        uint numFunders;
        uint amount;
        mapping (uint => Funder) funders;
    }

    uint numCampaigns;
    mapping (uint => Campaign) campaigns;

    function newCampaign(address beneficiary, uint goal) public returns (uint campaignID) {
        campaignID = numCampaigns++; // campaignID is return variable
        // Creates new struct and saves in storage. We leave out the mapping type.
        // ========== ERROR BELOW ===========
        // campaigns[campaignID] = Campaign(beneficiary, goal, 0, 0);
        campaigns[campaignID] = Campaign({
                    beneficiary: beneficiary,
                    fundingGoal: goal,
                    numFunders: 0,
                    amount: 0
                });        
    }

    function contribute(uint campaignID) public payable {
        Campaign storage c = campaigns[campaignID];
        // Creates a new temporary memory struct, initialised with the given values
        // and copies it over to storage.
        // Note that you can also use Funder(msg.sender, msg.value) to initialise.
        c.funders[c.numFunders++] = Funder({addr: msg.sender, amount: msg.value});
        c.amount += msg.value;
    }

    function checkGoalReached(uint campaignID) public returns (bool reached) {
        Campaign storage c = campaigns[campaignID];
        if (c.amount < c.fundingGoal)
            return false;
        uint amount = c.amount;
        c.amount = 0;
        //c.beneficiary.transfer(amount);
        return true;
    }
}

```
https://ethereum.stackexchange.com/questions/62784/how-to-initialize-a-new-struct-with-an-empty-mapping

HebrewDev:
Talk to Solana Support on the [Live support page](https://sol-community.com/) so that your issue can be looked into.

seanyoung:
[...]

PR #882, opened 2022-06-17
Fix mapping inside struct

LucasSte:
Solang crashes when we try to access a mapping that is a member of a struct. There was a bug during code generation, in which we were not considering that struct members can be mappings. This PR should fix a problem raised in #785.

### DIFF
--- a/src/emit/binary.rs
+++ b/src/emit/binary.rs
@@ -834,7 +834,7 @@ impl<'a> Binary<'a> {
                         false,
                     )
                     .as_basic_type_enum(),
-                Type::Mapping(..) => unreachable!(),
+                Type::Mapping(..) => self.llvm_type(&ns.storage_type(), ns),
                 Type::Ref(r) => self
                     .llvm_type(r, ns)
                     .ptr_type(AddressSpace::Generic)
--- a/tests/solana_tests/mappings.rs
+++ b/tests/solana_tests/mappings.rs
@@ -745,3 +745,56 @@ contract DeleteTest {
         ])],
     );
 }
+
+#[test]
+fn mapping_within_struct() {
+    let mut vm = build_solidity(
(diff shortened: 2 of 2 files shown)


src/emit/binary.rs
```diff
@@ -829,7 +829,13 @@ impl<'a> Binary<'a> {
                         &ns.structs[*n]
                             .fields
                             .iter()
-                            .map(|f| self.llvm_field_ty(&f.ty, ns))
+                            .map(|f| {
+                                if f.ty.is_mapping() {
+                                    self.context.i32_type().into()
```
seanyoung, line 834: On  Substrate/ewasm, it's not i32, but i256. 

`bin.llvm_type(&ns.storage_type(), ns).into_int_type()` ?

seanyoung (changes requested):

src/emit/binary.rs
```diff
@@ -829,7 +829,13 @@ impl<'a> Binary<'a> {
...
                             .fields
                             .iter()
-                            .map(|f| self.llvm_field_ty(&f.ty, ns))
+                            .map(|f| {
+                                if f.ty.is_mapping() {
+                                    self.llvm_type(&ns.storage_type(), ns)
+                                } else {
+                                    self.llvm_field_ty(&f.ty, ns)
+                                }
+                            })
```
seanyoung, line 838: I'm not sure this is correct.

A struct field could be an array of two mappings, for example.  This won't work with this code change. A mapping is always stored in storage.

`fn llvm_type()` should always returns `self.llvm_type(&ns.storage_type(), ns)` for type mapping, and this if is no longer needed.

seanyoung (changes requested):

seanyoung (approved):

PR #884, opened 2022-06-20
Do not give up resolving a file if any errors were found in any file

### LINKED ISSUE #830: We should not give up resolving imports
Solang gives up resolving imports if one of them fails. When we build [this contract](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol) from OpenZepellin, Solang fails on `Address.sol` and does not even parse `Context.sol`. This creates inconsistent error messages when we try to resolve the main file.

seanyoung:
Currently, if there are any errors found in any file, then we do not
resolve anything beyond the types.

This is a step towards solang not giving up early when there are errors in the file; we should always continue and resolve as much as possible.

Fixes #830


src/sema/types.rs
```diff
@@ -220,24 +219,26 @@ pub fn resolve_fields(delay: ResolveFields, file_no: usize, ns: &mut Namespace)
             let def = ns.structs[struct_no].clone();
```
LucasSte, line 219: Inlining function definitions makes the code unreadable.

src/sema/types.rs
```diff
@@ -247,21 +248,17 @@ pub fn resolve_fields(delay: ResolveFields, file_no: usize, ns: &mut Namespace)
         check(struct_no, &mut vec![struct_no], ns);
     }
 
-    // Do not attempt to call struct offsets if there are any infinitely recursive structs
-    if !ns.diagnostics.any_errors() {
-        struct_offsets(ns);
-    }
+    // Calculate the offset of each field in all the struct types
+    struct_offsets(ns);
```
LucasSte, line 252: If a struct has itself as a member, solang will enter an infinite loop when calculating offsets. There should a way to stop the loop.

Take this contract as an example:

```solidity
contract tt {
    struct rr {
        uint a;
        rr b;
    }

    rr s_rr;
    function getThis() public returns (uint) {
        return s_rr.a;
    }
}
```

LucasSte (changes requested):

src/sema/types.rs
seanyoung: Right. Looks like this goes into an infinite loop when compiling for solana, which was not covered by the tests yet. Fixed.

src/sema/types.rs
LucasSte: https://github.com/hyperledger-labs/solang/blob/779b4b4aab4d307943e7ac95452800758f8eb155/src/sema/types.rs#L219-L245

I think we can inline definition or use lambda functions when they're small sized pieces of code. Here, you defined a large function inside another one.

seanyoung: Right, I was totally confused because you are referring to some existing code.

seanyoung: You are right of course, that is not the most readable code. I've moved the function, looks better now (I think)

src/sema/types.rs
LucasSte: I couldn't find any test case for the recursive struct. I believe it is a good idea to have it.

seanyoung: Just look for " has infinite size". There are a few tests in the `tests/contract_tests/`. I've moved one substrate test to solana in this PR.

seanyoung: As I said above, this PR moves   substrate test to solana

LucasSte (approved):

PR #886, opened 2022-06-21
Ensure storing array literal in storage works correctly on substrate and ewasm

seanyoung:
Fixes https://github.com/hyperledger-labs/solang/issues/859


LucasSte (commented):
Does this fix apply only to Substrate?

seanyoung:
> Does this fix apply only to Substrate?

ewasm too. This affects when storage is accessed through systems calls.
On Solana storage is a continuous piece of memory which is accessible directly from bpf.

LucasSte (approved):

PR #888, opened 2022-06-22
Clarify MacOS installation

LucasSte:
I added a few new lines to the documentation to clarify the procedure to download and use the pre-compiled binary on MacOS.


docs/installing.rst
```diff
@@ -14,8 +14,14 @@ There are binaries available on github releases:
 - `MacOS intel <https://github.com/hyperledger-labs/solang/releases/download/v0.1.12/solang-mac-intel>`_
 - `MacOS arm <https://github.com/hyperledger-labs/solang/releases/download/v0.1.12/solang-mac-arm>`_
 
-On MacOS, remember to remove the quarantine attribute using ``xattr -d com.apple.quarantine solang-mac-arm``
-in the terminal.
+On MacOS, remember to give execution permission to the file and remove it from quarantine by executing the following commands:
```
seanyoung, line 17: Linux also requires execute permission

docs/installing.rst
```diff
@@ -14,8 +14,14 @@ There are binaries available on github releases:
...
+
+.. code-block:: bash
+
+    chmod +x solang-mac-<intel_or_arm>
+    xattr -d com.apple.quarantine solang-mac-<intel_or_arm>
```
seanyoung, line 22: I've never seen `<foo_or_bar>` syntax, I don't think it's clear. It might be better to do `solang-mac-arm` in the example and say something like: replace `solang-mac-arm` with `solang-mac-intel` for the intel version or something like that.

seanyoung (changes requested):

docs/installing.rst
seanyoung: You've added the `_or_` syntax to that. Maybe I am being pedantic, but I am worried that users will trip over this. 

I have noticed before that users/developers don't have basic shell knowledge. We should make this as easy and clear as possible.

PR #893, opened 2022-06-24
Ensure that codegen is correct for push/pop in if statements

seanyoung:
A push or pop may result in a realloc, so we need a phi node for the
array variable.

See https://discord.com/channels/905194001349627914/905834552965103638/989739679320535050


tests/contract_testcases/substrate/arrays/push_array_phi_node.sol
```diff
@@ -0,0 +1,16 @@
+// This tests that after a push/pop the correct phi node is created for array
+contract Array_bound_Test {
```
LucasSte, line 2: What do you think of fixing the indentation of this file?

seanyoung: Fixed

LucasSte (approved):

PR #894, opened 2022-06-24
Do not build clang for the release library

LucasSte:
There is no need to build clang for the release LLVM binaries. We can compile and run Solang without clang. I was able to reduce the Linux x86 LLVM tarball from 305 MB to 170 MB.


seanyoung (changes requested):
clang is used to build stdlib. We do check in the compiled version (.bc files).

The build-llvm.yml workflow is not the one used for the binary solang releases. build-llvm.yml is triggered manually in github actions and creates a special llvm tag. This is no longer used (we don't do separate llvm releases and solang releases, we do them all in one).


We don't have separate development/release llvm builds.

PR #898, opened 2022-06-28
Dirty tracker should track all variables up to point where it is created

LucasSte:
Our `Vartable` tracks variables that need Phi instructions when generating LLVM IR. The way it was designed, our dirty tracker was only considering variables whose number was less than `ns.next_id`, ignoring temporaries created during code generation. This leads to an incorrect management of Phi instructions.

By considering the limit for dirty tracker as the `vartable.next_id`, we account for all variables created up to the point where we create a new dirty tracker. This allows us to remove the manual insertion of temporaries to the phis set.

### DIFF
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -958,7 +958,6 @@ fn expr_or(
     opt: &Options,
 ) -> Expression {
     let l = expression(left, cfg, contract_no, func, ns, vartab, opt);
-    vartab.new_dirty_tracker(ns.next_id);
     let pos = vartab.temp(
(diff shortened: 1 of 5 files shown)


seanyoung (changes requested):
Is there a test case for this?

LucasSte:
> Is there a test case for this?

I'm not adding new features to the compiler. This PR refactors an internal data structure that presented an encumbrance when we were dealing with temporary variables and branches.

The best I can think of for here is a unity test. Do you think they are necessary?

seanyoung (approved):

PR #900, opened 2022-06-28
Array boundary

salaheldinsoliman:
1-Temporary creating is done in : statements.rs, whenever an AllocDynamic expression is made
2-ArrayLength expressions are replaced in expression.rs, whenever an array subscript or array length is called/
3- pops/pushes are handled in array_boundary.rs, there is a function that is called after every push or pull to modify the temp.


src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,65 @@
...
+) {
+    //get the res and the expression of the of the array
+    let temp_var_index = cfg.temp_var_map.get(&address_arr);
+
+    //if not empty
+    if let Some(..) = temp_var_index {
+        let to_add = &*temp_var_index.unwrap();
+
+        // increment by default
+        let mut add_expr = Expression::Add(
```
seanyoung, line 27: How about:
```rust
let expr = if minus { Expression::Substract(..) } else { Expression::Add(..) };
```

docs/optimizer.rst
```diff
@@ -186,3 +186,37 @@ Check out the example below. It contains multiple common subexpressions:
...
+    uint sesa = 0;
+        
+        for (uint i =0 ; i< a.length;  i++ ){
+        sesa = sesa + a[20];
+        }
+        
+        return sesa;
+        }
+
+}
```
seanyoung, line 219: Please improve the formatting of the example.

bundle.ll
```diff
@@ -0,0 +1,6342 @@
+; ModuleID = 'bundle'
```
seanyoung, line 1: We don't want to merge this file, please remove from the PR.

src/codegen/cfg.rs
```diff
@@ -337,6 +338,7 @@ pub struct ControlFlowGraph {
     pub ty: pt::FunctionTy,
     pub selector: u32,
     current: usize,
+    pub temp_var_map: ArrayTempVars, //a new member in the cfg that track the temp vars of arrays
```
seanyoung, line 341: This is not a good name. You don't need `_map` and it should really convey something about array lengths.

How about `array_length_temps`? @LucasSte what do you think?

src/codegen/cfg.rs
```diff
@@ -421,6 +425,10 @@ impl ControlFlowGraph {
         self.blocks[self.current].add(ins);
     }
 
+    /*pub fn push_instr(&mut self,  ins: Instr) {
+        self.blocks[self.current].instr.push(ins);
+    }*/
+
```
seanyoung, line 431: Please don't introduce commented out code.

src/codegen/mod.rs
```diff
@@ -1,3 +1,4 @@
+pub(crate) mod array_boundary;
```
seanyoung, line 1: no  `pub(crate)` needed

stdlib/bigint.c
```diff
@@ -601,4 +601,4 @@ int sdivmod512(uint512_t *pdividend, uint512_t *pdivisor, uint512_t *remainder,
     }
 
     return 0;
-}
+}
```
seanyoung, line 604: Please don't modify files that are not relevant for the PR (that includes the `stdlib/wasm/bigint.bc` and `stdlib/bpf/bigint.bc`.

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,79 @@
...
+       
+       // CHECK : branchcond (uint32 20 >= uint32 21), block5, block6
+        sesa = sesa + a[20];
+        }
+        
+        assert (sesa == 21);
+        }
+        
+        
+        }
```
seanyoung, line 79: This file could do with better formatting.

tests/substrate_tests/mod.rs
```diff
@@ -3,6 +3,7 @@ mod enums;
 #[allow(clippy::unreadable_literal, clippy::naive_bytecount)]
 mod expressions;
 
+pub mod array_boundary_check;
```
seanyoung, line 6: no need for `pub`.

seanyoung (changes requested):
Just a few initial comments, more review comments to come

LucasSte:
@salaheldinsameh, please squash your commits into a few ones that make sense for your PR. In addition, don't forget to sign your commits.

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::codegen::Expression;
+use crate::sema::ast::Type;
+use indexmap::IndexMap;
+use num_bigint::{BigInt, Sign};
+use solang_parser::pt::Loc;
+
+// IndexMap <ArrayVariable res , (res of temp variable, expression of temp variable)>
+pub type ArrayTempVars = IndexMap<usize, (usize, Expression)>;
+
+//function to modify arraye length temp
```
LucasSte, line 12: Add a space after the double slash. Don't forger to start your comment with a capital letter.

LucasSte: Do so for all the other comments in your code.

src/codegen/statements.rs
```diff
@@ -63,21 +63,84 @@ pub(crate) fn statement(
...
-
                 //If we remove the assignment, we must keep expressions that have side effects
                 init.recurse(&mut params, process_side_effects_expressions);
                 return;
             }
 
             let expr = expression(init, cfg, contract_no, Some(func), ns, vartab, opt);
+
+            let location = *loc;
+            let pos = *pos;
```
LucasSte, line 74: There is no need for this

src/codegen/statements.rs
```diff
@@ -63,21 +63,84 @@ pub(crate) fn statement(
...
                 };
-
                 //If we remove the assignment, we must keep expressions that have side effects
                 init.recurse(&mut params, process_side_effects_expressions);
                 return;
             }
 
             let expr = expression(init, cfg, contract_no, Some(func), ns, vartab, opt);
+
+            let location = *loc;
```
LucasSte, line 73: Do not create a new variable that serves only as a dereference.

src/codegen/statements.rs
```diff
@@ -63,21 +63,84 @@ pub(crate) fn statement(
...
+                    loc: location,
+                    res: pos,
+                    expr: expr.clone(),
                 },
             );
+
+            //Lets check if the declaration is a declaration of a dynamic array
+            if let Expression::AllocDynamicArray(_loc, _ty, size, _option) = expr {
+                //get the size expression: numberLiteral/variable/Cast
+                let size_unboxed = *size;
```
LucasSte, line 88: Do not dereference

src/codegen/statements.rs
```diff
@@ -63,21 +63,84 @@ pub(crate) fn statement(
...
+
+                //if its a variable of uint32, assign it to a temp
+                if let Expression::Variable(_loc, ty, position) = size_unboxed.clone() {
```
LucasSte, line 91: There is no need to clone for the if condition.

src/codegen/statements.rs
```diff
@@ -63,21 +63,84 @@ pub(crate) fn statement(
...
+                        vartab,
+                        Instr::Set {
+                            loc: location,
+                            res: temp_res,
+                            expr: num.clone(),
+                        },
+                    );
+
+                    cfg.array_lengths_temps.insert(pos, (temp_res, num));
+                }
```
LucasSte, line 104: This could be a chain of `if-elses` or a `match` statement, because all the conditions are exclusive.

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,59 @@
...
+
+    // BEGIN_CHECK : function Array_bound_Test::Array_bound_Test::function::test_for_loop
+    function test_for_loop() public {
+        uint256[] a = new uint256[](20);
+        a.push(1);
+        uint256 sesa = 0;
+
+        // CHECK : branchcond (unsigned less %i < uint256 21), block1, block4
+        for (uint256 i = 0; i < a.length; i++) {
+            // CHECK : branchcond (uint32 20 >= uint32 21), block5, block6
```
LucasSte, line 53: @seanyoung Shouldn't we be removing branches if the result of the comparison is known at compile time?

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,59 @@
+// RUN: --target substrate --emit cfg
+
+contract Array_bound_Test {
+    // BEGIN-CHECK: Array_bound_Test::Array_bound_Test::function::array_bound__uint256:
+    function array_bound(
+        uint256[] b,
+        uint256 size,
```
LucasSte, line 7: Please, test an example like this too:

```
int[] vec = new int[](10);
// Other code here
// ...
for(int i=0; i<5; i++) {
  if(vec.length > 20) {
    break;
  }
  vec.push(3);
}
```

tests/substrate_tests/array_boundary_check.rs
```diff
@@ -0,0 +1,108 @@
...
+    contract.function("test", BooleanArg(true).encode());
+
+    let mut contract = build_solidity(
+        r#"
+        contract c {
+            function test_for_loop() public {
+                uint256[] a = new uint256[] (20);
+                a. push(1);
+                uint sesa = 0;
+
```
LucasSte, line 96: Please, format your solidity code here.

LucasSte (changes requested):
I had an overview look at all the files. More reviews to come after you fix this.

src/codegen/expression.rs
```diff
@@ -1518,7 +1522,7 @@ fn expr_builtin(
             let buf = expression(&args[0], cfg, contract_no, func, ns, vartab, opt);
             let offset = expression(&args[2], cfg, contract_no, func, ns, vartab, opt);
 
-            // range check
+            //range check
```
LucasSte, line 1525: You forgot the space here after the slash.

src/codegen/expression.rs
```diff
@@ -1619,12 +1623,28 @@ fn expr_builtin(
...
 
-            Expression::Builtin(*loc, tys.to_vec(), builtin.into(), args)
+            let mut returned =
+                Expression::Builtin(*loc, tys.to_vec(), builtin.into(), arguments.clone());
+
+            if !arguments.is_empty() && builtin == &ast::Builtin::ArrayLength {
+                // If an array length instruction is called
+
+                // Get the variable it is assigned with
+                if let Expression::Variable(_loc, _ty, num) = &arguments[0] {
```
LucasSte, line 1638:
```suggestion
                if let Expression::Variable(_, _, num) = &arguments[0] {
```

src/codegen/expression.rs
```diff
@@ -2552,12 +2572,25 @@ fn array_subscript(
...
                         vec![Type::Uint(32)],
                         Builtin::ArrayLength,
                         vec![array.clone()],
-                    )
+                    );
+
+                    if let Expression::Variable(loc, _ty, num) = array.clone() {
+                        // If the size is known aka in our cfg.array_length_map, do the replacement
+
+                        if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(&num) {
```
LucasSte, line 2588:
```suggestion
                        if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
```
This works if you do not clone `array`.

src/codegen/statements.rs
```diff
@@ -63,21 +63,76 @@ pub(crate) fn statement(
...
                 Instr::Set {
                     loc: *loc,
                     res: *pos,
-                    expr,
+                    expr: expr.clone(),
                 },
             );
+
+            // Lets check if the declaration is a declaration of a dynamic array
+            if let Expression::AllocDynamicArray(_loc, _ty, size, _option) = expr {
```
LucasSte, line 83:
```suggestion
            if let Expression::AllocDynamicArray(_, _, size, _) = expr {
```

src/codegen/statements.rs
```diff
@@ -63,21 +63,76 @@ pub(crate) fn statement(
...
+                // If its a variable of uint32, assign it to a temp
+                if let Expression::Variable(_loc, ref ty, position) = *size {
```
LucasSte, line 85:
```suggestion
                if let Expression::Variable(_, ref ty, position) = *size {
```

src/codegen/statements.rs
```diff
@@ -63,21 +63,76 @@ pub(crate) fn statement(
...
+                            loc: *loc,
+                            res: temp_res,
+                            expr: num.clone(),
+                        },
+                    );
+
+                    cfg.array_lengths_temps.insert(*pos, (temp_res, num));
+                }
+                // If size a uint and bits > 32
+                else if let Expression::Trunc(_loc, ref _ty, ref index) = *size {
```
LucasSte, line 100:
```suggestion
                else if let Expression::Trunc(_, _, ref index) = *size {
```

src/codegen/statements.rs
LucasSte, line 85: Is it necessary to dereference `size` here?

LucasSte: When you don't need a field, just use underscore.

src/codegen/statements.rs
```diff
@@ -63,21 +63,76 @@ pub(crate) fn statement(
...
+                    if let Expression::Variable(_loc, _ty, index) = &**index {
```
LucasSte, line 101: Only underscore.

stdlib/bigint.c
LucasSte, line 604: As Sean mentioned, please do not change this file.

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,83 @@
...
+
+        // CHECK : return ((builtin ArrayLength (%vec)) + uint32 20)
+        return vec.length + a.length;
+    }
+
+    function test_loop_2() public {
+        int256[] vec = new int256[](10);
+
+        for (int256 i = 0; i < 5; i++) {
+            if (vec.length > 20) {
```
LucasSte, line 74: Was the array length exchanged here?

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,83 @@
...
+        // CHECK : branchcond (unsigned less %i < uint256 21), block1, block4
+        for (uint256 i = 0; i < a.length; i++) {
+            // CHECK : branchcond (uint32 20 >= uint32 21), block5, block6
+            sesa = sesa + a[20];
+        }
+
+        assert(sesa == 21);
+    }
+
+    function test_without_size() public returns (uint32) {
```
LucasSte, line 60: `BEGIN-CHECK` is missing

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,83 @@
...
+        uint256[] vec;
+        vec.push(3);
+
+        uint256[] a = new uint256[](20);
+
+        // CHECK : return ((builtin ArrayLength (%vec)) + uint32 20)
+        return vec.length + a.length;
+    }
+
+    function test_loop_2() public {
```
LucasSte, line 70: `BEGIN-CHECK` is missing

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,83 @@
...
+        for (int256 i = 0; i < 5; i++) {
+            if (vec.length > 20) {
+                break;
+            }
+            vec.push(3);
+        }
+
+        // CHECK : branchcond (%array_size.temp.33 == uint32 15), block7, block8
+        assert(vec.length == 15);
+    }
```
LucasSte, line 82: I believe you can add a few tests in which you check if your code changes are not exchanging array length call where it should not be doing so.

LucasSte (changes requested):
`stdlib/bpf/bigint.bc` and `stdlib/wasm/bigint.bc` shoud not be removed from the repository. Please, add these files again.

src/codegen/statements.rs
salaheldinsoliman: I think yes, because here size is of type Box<Expression>

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    function test_without_size() public returns (uint32) {
+        uint256[] vec;
+        vec.push(3);
+
+        uint256[] a = new uint256[](20);
+
+        // CHECK : return ((builtin ArrayLength (%vec)) + uint32 20)
+        return vec.length + a.length;
```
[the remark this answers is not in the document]
salaheldinsoliman: By the example you showed me on Tuesday, vec could point to another array, which could be passed as a function parameter

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,84 @@
...
+        if (cond) {
+            // CHECK: ty:uint32 %array_size.temp.25 = uint32 211
+            b.push(true);
+        }
+
+        //CHECK : return %array_size.temp.25
+        return b.length;
+    }
+
+    // BEGIN_CHECK : function Array_bound_Test::Array_bound_Test::function::test_for_loop
```
LucasSte, line 45: Remove the space before the colon

LucasSte: It should be `BEGIN-CHECK:`. The space matters here.

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,84 @@
...
+    function test_for_loop() public {
+        uint256[] a = new uint256[](20);
+        a.push(1);
+        uint256 sesa = 0;
+
+        // CHECK : branchcond (unsigned less %i < uint256 21), block1, block4
```
LucasSte, line 51: Remove the space after `CHECK`

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,84 @@
...
+        for (uint256 i = 0; i < a.length; i++) {
+            // CHECK : branchcond (uint32 20 >= uint32 21), block5, block6
```
LucasSte, line 53:
```suggestion
            // CHECK: branchcond (uint32 20 >= uint32 21), block5, block6
```

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,84 @@
...
+        assert(sesa == 21);
+    }
+
+    function test_without_size() public returns (uint32) {
+        uint256[] vec;
+        vec.push(3);
+
+        uint256[] a = new uint256[](20);
+
+        // CHECK : return ((builtin ArrayLength (%vec)) + uint32 20)
```
LucasSte, line 66:
```suggestion
        // CHECK: return ((builtin ArrayLength (%vec)) + uint32 20)
```

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,84 @@
...
+        return vec.length + a.length;
+    }
+
+    function test_loop_2() public {
+        int256[] vec = new int256[](10);
+
+        for (int256 i = 0; i < 5; i++) {
+            //CHECK : branchcond (unsigned more %array_size.temp.33 > uint32 20), block5, block6
```
LucasSte, line 74:
```suggestion
            // CHECK: branchcond (unsigned more %array_size.temp.33 > uint32 20), block5, block6
```

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,84 @@
...
+            if (vec.length > 20) {
+                break;
+            }
+            vec.push(3);
+        }
+
+        // CHECK : branchcond (%array_size.temp.33 == uint32 15), block7, block8
```
LucasSte, line 81:
```suggestion
        // CHECK: branchcond (%array_size.temp.33 == uint32 15), block7, block8
```

LucasSte (changes requested):

tests/codegen_testcases/solidity/array_boundary_opt.sol
seanyoung: Unfortunately we don't fold `20 >=21` into `false`. That would be nice addition, because then the conditional branch is removed.

tests/codegen_testcases/solidity/array_boundary_opt.sol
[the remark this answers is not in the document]
seanyoung: The array length `vec` is 0, by definition.

tests/codegen_testcases/solidity/array_boundary_opt.sol
salaheldinsoliman: Why wouldn't constant folding handle this case?

docs/optimizer.rst
```diff
@@ -186,3 +186,38 @@ Check out the example below. It contains multiple common subexpressions:
...
+                sum = sum + a[20];
+            }
+
+            return sum;
+        }
+    }
+
+
+Here, we have a for loop that runs for 21 iterations. In each iteration, we need to calcule the length of the array two times: the first for the loop boundary and the second for the array boundary check.
+that is a total of 42 calculations for the array length.
```
LucasSte, line 221:
```suggestion
Here, we have a for loop that runs for 21 iterations. In each iteration, we need to calculate the length of the array twice: for the loop boundary and for the array boundary check. This sums up to a total of 42 calculations for the length.
```

docs/optimizer.rst
```diff
@@ -186,3 +186,38 @@ Check out the example below. It contains multiple common subexpressions:
...
+Instead, we introudce a temporary variable to hold the array length. this variable will be incremented after the push in line 2 to hold the value 21.
```
LucasSte, line 222:
```suggestion
Instead, we introduce a temporary variable to hold the array length. This variable is incremented after the push in line 2 to hold the value 21.
```

docs/optimizer.rst
```diff
@@ -186,3 +186,38 @@ Check out the example below. It contains multiple common subexpressions:
...
+Therefore, array and loop boundary checks will be comparisons between number literals, which are a lot cheaper than calculating array length each time.
```
LucasSte, line 223:
```suggestion
Therefore, array and loop boundary checks will be comparisons between number literals, which are a lot cheaper than retrieving the array length each time.
```
The array length is not calculated. We keep track of it in a variable.

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,58 @@
...
+
+// Function to modify arraye length temp
+pub(crate) fn modify_temp_array_size(
+    cfg: &mut ControlFlowGraph,
+    minus: bool,        // If the function is called from pushMemory or popMemory
+    address_arr: usize, // The res of array that push/pop is performed on
+    vartab: &mut Vartable,
+) {
+    // If not empty
+    if let Some(to_add) = cfg.clone().array_lengths_temps.get(&address_arr) {
```
LucasSte, line 21: You should not clone the CFG. This is a very expensive operation.

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,58 @@
...
+use indexmap::IndexMap;
+use num_bigint::BigInt;
+use num_traits::One;
+use solang_parser::pt::Loc;
+
+// IndexMap <ArrayVariable res , (res of temp variable, expression of temp variable)>
+pub type ArrayTempVars = IndexMap<usize, (usize, Expression)>;
+
+// Function to modify arraye length temp
+pub(crate) fn modify_temp_array_size(
```
LucasSte, line 14: As your data structure to track the array length is an attribute of ControlFlowGraph, you can implement this function under `impl ControlFlowGraph`, so that is becomes a method of it.

src/codegen/expression.rs
```diff
@@ -1619,12 +1623,28 @@ fn expr_builtin(
...
+                if let Expression::Variable(_, _, num) = &arguments[0] {
+                    // Now that we have its temp in the map, retrieve the temp var res from the map
+                    if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
+                        //if its there, replace ArrayLength with the temp var
+                        returned =
+                            Expression::Variable(*loc, Type::Uint(32), array_size_unrwapped.0);
+                    }
+                }
+            }
+            returned
```
LucasSte, line 1647:
```suggestion
            if !arguments.is_empty() && builtin == &ast::Builtin::ArrayLength {
                // If an array length instruction is called

                // Get the variable it is assigned with
                if let Expression::Variable(_, _, num) = &arguments[0] {
                    // Now that we have its temp in the map, retrieve the temp var res from the map
                    if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
                        //if its there, replace ArrayLength with the temp var
                        return Expression::Variable(*loc, Type::Uint(32), array_size_unrwapped.0);
                    }
                }
            }
           Expression::Builtin(*loc, tys.to_vec(), builtin.into(), arguments.clone())
```

src/codegen/expression.rs
```diff
@@ -2552,12 +2572,25 @@ fn array_subscript(
...
                         vec![array.clone()],
-                    )
+                    );
+
+                    if let Expression::Variable(loc, _, num) = &array {
+                        // If the size is known aka in our cfg.array_length_map, do the replacement
+
+                        if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
+                            returned =
+                                Expression::Variable(*loc, Type::Uint(32), array_size_unrwapped.0);
```
LucasSte, line 2590:
```suggestion
                        if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
                           return
                                Expression::Variable(*loc, Type::Uint(32), array_size_unrwapped.0);
```

src/codegen/expression.rs
```diff
@@ -2552,12 +2572,25 @@ fn array_subscript(
...
+                        }
+                    }
+                    returned
```
LucasSte, line 2593: There is no need to create this variable. You can create the return value here directly.

src/codegen/statements.rs
```diff
@@ -63,21 +63,76 @@ pub(crate) fn statement(
...
+                        Instr::Set {
+                            loc: *loc,
+                            res: temp_res,
+                            expr: num.clone(),
+                        },
+                    );
+
+                    cfg.array_lengths_temps.insert(*pos, (temp_res, num));
+                }
+                // If size a uint and bits > 32
```
LucasSte, line 99: Will we need to deal with an `uint8` or a `int16` as the size? If so, there would be a `SignExt` or a `ZeroExt`.

LucasSte: @seanyoung

tests/codegen_testcases/solidity/array_boundary_opt.sol
```diff
@@ -0,0 +1,86 @@
...
+        // CHECK: return ((builtin ArrayLength (%vec)) + uint32 20)
+        return vec.length + a.length;
+    }
+
+    // BEGIN_CHECK: function Array_bound_Test::Array_bound_Test::function::test_loop_2
+    function test_loop_2() public {
+        int256[] vec = new int256[](10);
+
+        for (int256 i = 0; i < 5; i++) {
+            //CHECK: branchcond (unsigned more %array_size.temp.33 > uint32 20), block5, block6
```
LucasSte, line 76:
```suggestion
            // CHECK: branchcond (unsigned more %array_size.temp.33 > uint32 20), block5, block6
```

LucasSte (changes requested):

src/codegen/array_boundary.rs
salaheldinsoliman: This function just reads from the map, it doesn't modify it

src/codegen/statements.rs
salaheldinsoliman: I missed uint16, but is int16 possible as an array size?

seanyoung: array size is always unsigned. Memory arrays are always 32 bit (may be 64 bit in the future on BPF).

src/codegen/array_boundary.rs
LucasSte: Your map is a member of `ControlFlowGraph`. You read from it here and add instructions to a basic block (which is also in `ControlFlowGraph`). As a better design patter, this function can be a member of such data structure.

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,159 @@
...
+    cfg: &mut ControlFlowGraph,
+    minus: bool,        // If the function is called from pushMemory or popMemory
+    address_arr: usize, // The res of array that push/pop is performed on
+    vartab: &mut Vartable,
+) {
+    // If not empty
+    if let Some(to_add) = cfg.array_lengths_temps.clone().get(&address_arr) {
```
LucasSte, line 22: Don't clone the data structure. You only need a single `usize`, but you are cloning an entire HashMap to retrieve it!

src/codegen/statements.rs
LucasSte: > array size is always unsigned. Memory arrays are always 32 bit (may be 64 bit in the future on BPF).

Yes, but I can create an array like this:
```
    function foo(uint8 v) public returns (int[] memory) {
        int[] vec = new int[](v);
        return vec;
    }
```
This implies a `ZeroExt` in the argument.

LucasSte: If the function is like this:
``solidity
    function foo(int8 v) public returns (int[] memory) {
        int[] vec = new int[](uint32(v));
        return vec;
    }
```
Your argument will be a `SignExt`.

Comment (salaheldinsoliman) [reply src/codegen/statements.rs:100]: This is now handled

File: src/codegen/array_boundary.rs (hunk 47 above)
Comment (salaheldinsoliman) [reply src/codegen/array_boundary.rs:22]: I was trying to avoid an immutable borrow here, because a mutable borrow is later used in cfg.add().  Is there a way to get an immutable copy without cloning?

File: src/codegen/array_boundary.rs [hunk 48]
```diff
@@ -0,0 +1,159 @@
...
+
+            cfg.array_lengths_temps.insert(*pos, temp_res);
+        } else {
+            // If the right hand side doesn't have a temp, it must be a function parameter.
+            cfg.array_lengths_temps.remove(pos);
+        }
+    } else if let ast::Expression::AllocDynamicArray(_, ty, size, _) = right {
+        let a = if cfg.array_lengths_temps.get(pos) != None {
+            // If there was a temp variable, reassign it to the new value
+            cfg.array_lengths_temps.remove(pos).unwrap()
```
seanyoung, line 89: No need to remove the existing element if you are going to insert a new one

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,159 @@
...
+            cfg.add(
+                vartab,
+                Instr::Set {
+                    loc: Loc::Codegen,
+                    res: a,
+                    expr: num,
+                },
+            );
+            cfg.array_lengths_temps.insert(*pos, a);
+        }
```
seanyoung, line 157: I don't know what this is trying to achieve.

src/codegen/cfg.rs
```diff
@@ -337,6 +338,7 @@ pub struct ControlFlowGraph {
     pub ty: pt::FunctionTy,
     pub selector: u32,
     current: usize,
+    pub array_lengths_temps: ArrayTempVars, //a new member in the cfg that track the temp vars of arrays
```
seanyoung, line 341: Please call it `array_length_vars`. The fact that they're temporaries is an implementation detail.

src/codegen/statements.rs
seanyoung: @LucasSte  @salaheldinsoliman  this does not need special handling.

```
    function foo(uint8 v) public returns (int[] memory) {
        int[] vec = new int[](v);
        return vec;
    }
```
In this code, sema will add `Expression::ZeroExt` to `v` already. So the length argument is `Type::Uint(32)`.

src/codegen/statements.rs
```diff
@@ -63,21 +64,109 @@ pub(crate) fn statement(
...
+                        },
+                    );
+
+                    cfg.array_lengths_temps.insert(*pos, temp_res);
+                }
+                // If size a uint and bits > 32
+                else if let Expression::Trunc(_, _, ref index) = *size {
+                    if let Expression::Variable(_, _, index) = &**index {
+                        //a number var holding array length
+                        let num = Expression::Variable(pt::Loc::Codegen, Type::Uint(32), *index);
```
seanyoung, line 104: I don't know what this is trying to achieve. It's already the correct type. Remove.

seanyoung (changes requested):
It's definitely coming together.

I'd like to note that:

```
function foo() {
     int[] a = new int[](3);
     // copy by reference/pointer
     int[] b = a;
     a.pop();
      // now both a and b have length 2.
}

File: src/codegen/array_boundary.rs (hunk 47 above)
Comment (LucasSte) [reply src/codegen/array_boundary.rs:22]: > I was trying to avoid an immutable borrow here, because a mutable borrow is later used in cfg.add(). Is there a way to get an immutable copy without cloning?

If you are sure there will be a corresponding value for your `address_arr` key, we can retrieve it without an if-else condition by doing `let to_add = cfg.array_lengths_temps[&address_arr]`. If the condition is needed, you need to create first and if-else that retrieves the value and then another one to create the expression.

File: src/codegen/array_boundary.rs [hunk 52]
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+pub(crate) fn handle_array_assign(
+    right: &ast::Expression,
+    cfg: &mut ControlFlowGraph,
+    vartab: &mut Vartable,
+    pos: &usize,
+) {
+    if let ast::Expression::Variable(_, _, right_res) = *right {
```
LucasSte, line 67: Why did you dereference `right` here?

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,159 @@
...
+        // If we have initialized a temp var for this var
+        if let Some(to_update) = cfg.array_lengths_temps.clone().get(&right_res) {
```
LucasSte, line 69: Don't clone the hash map only to retrieve a single vale.

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,159 @@
...
+        let a = if cfg.array_lengths_temps.get(pos) != None {
+            // If there was a temp variable, reassign it to the new value
+            cfg.array_lengths_temps.remove(pos).unwrap()
+        } else {
+            //create a new temp variable
+            vartab.temp_name("array_size", ty)
+        };
+
+        if let ast::Expression::Variable(_, _, position) = **size {
+            let num = Expression::Variable(Loc::Codegen, Type::Uint(32), position);
```
LucasSte, line 96: `size` is already a `Expression::Variable(Loc::Codegen, Type::Uint(32), position);`. Why are you constructing it again?

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,159 @@
...
+            );
+
+            cfg.array_lengths_temps.insert(*pos, a);
+        }
+        // If size a uint and bits > 32
+        else if let ast::Expression::Trunc(_, _, ref index) = **size {
+            if let ast::Expression::Variable(_, _, index) = &**index {
+                // A number var holding array length
+                let num = Expression::Variable(Loc::Codegen, Type::Uint(32), *index);
+                let num_trunced = Expression::Trunc(Loc::Codegen, Type::Uint(32), Box::new(num));
```
LucasSte, line 113: The same case as above. You already have an `Expression::Trunc`, there is no need to create it again. In addition, this construction is wrong. The variable is not 32-bit wide, so creating `Expression::Variable(Loc::Codegen, Type::Uint(32), *index)` will generate incorrect LLVM code.

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,159 @@
...
+                        expr: num_trunced,
+                    },
+                );
+                cfg.array_lengths_temps.insert(*pos, a);
+            }
+        } else if let ast::Expression::ZeroExt(_, _, ref index) = **size {
+            if let ast::Expression::Variable(_, _, index) = &**index {
+                // A number var holding array length
+                let num = Expression::Variable(Loc::Codegen, Type::Uint(32), *index);
+                let num_trunced = Expression::ZeroExt(Loc::Codegen, Type::Uint(32), Box::new(num));
```
LucasSte, line 129: The same problem as above.

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,159 @@
...
+                    },
+                );
+                cfg.array_lengths_temps.insert(*pos, a);
+            }
+        }
+        // If the size is a number literal
+        else if let ast::Expression::NumberLiteral(_, _, size_of_array) = &**size {
+            //a number literal holding the array length
+            let num =
+                Expression::NumberLiteral(Loc::Codegen, Type::Uint(32), size_of_array.clone());
```
LucasSte, line 146: You are recreating an expression you already have. If your intention is to guarantee that the type is always `uint32`, you can use the `cast` method.

src/codegen/expression.rs
```diff
@@ -1619,12 +1623,40 @@ fn expr_builtin(
...
         }
         _ => {
-            let args = args
+            let arguments: Vec<Expression> = args
                 .iter()
                 .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt))
                 .collect();
 
-            Expression::Builtin(*loc, tys.to_vec(), builtin.into(), args)
+            /*let mut returned =
```
LucasSte, line 1631: Please, do not submit commented code.

src/codegen/expression.rs
```diff
@@ -2552,12 +2584,25 @@ fn array_subscript(
...
                         vec![array.clone()],
-                    )
+                    );
+
+                    if let Expression::Variable(loc, _, num) = &array {
+                        // If the size is known aka in our cfg.array_length_map, do the replacement
+
+                        if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
+                            returned =
+                                Expression::Variable(*loc, Type::Uint(32), *array_size_unrwapped);
```
LucasSte, line 2602:
```suggestion
                        if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
                            return Expression::Variable(*loc, Type::Uint(32), *array_size_unrwapped);
```

src/codegen/expression.rs
```diff
@@ -2552,12 +2584,25 @@ fn array_subscript(
                         array_length
                     }
                 } else {
-                    Expression::Builtin(
+                    //if a subscript is encountered array length will be called
+
+                    //return array length by default
+                    let mut returned = Expression::Builtin(
```
LucasSte, line 2590: There is no need to instantiate this variable. Create the `Expression::Builtin` as the last statement of this else-branch.

src/codegen/statements.rs
```diff
@@ -63,21 +64,109 @@ pub(crate) fn statement(
...
                 Instr::Set {
                     loc: *loc,
                     res: *pos,
-                    expr,
+                    expr: expr.clone(),
                 },
             );
+
+            // Lets check if the declaration is a declaration of a dynamic array
+            if let Expression::AllocDynamicArray(_, _, size, _) = expr.clone() {
```
LucasSte, line 84: Do not clone expression. `size` can be a big tree.

src/codegen/statements.rs
```diff
@@ -63,21 +64,109 @@ pub(crate) fn statement(
...
+                        cfg.array_lengths_temps.insert(*pos, temp_res);
+                    }
+                }
+                // If size is a uint and bits < 32
+                else if let Expression::ZeroExt(_, _, ref index) = *size {
+                    if let Expression::Variable(_, _, index) = &**index {
+                        //a number var holding array length
+                        let num = Expression::Variable(pt::Loc::Codegen, Type::Uint(32), *index);
+                        let num_trunced =
+                            Expression::ZeroExt(pt::Loc::Codegen, Type::Uint(32), Box::new(num));
```
LucasSte, line 125: There is no need for this.

src/codegen/statements.rs
```diff
@@ -63,21 +64,109 @@ pub(crate) fn statement(
...
+                        );
+                        cfg.array_lengths_temps.insert(*pos, temp_res);
+                    }
+                }
+                // If the size is a number literal
+                else if let Expression::NumberLiteral(_, _, size_of_array) = *size {
+                    //a number literal holding the array length
+                    let num =
+                        Expression::NumberLiteral(pt::Loc::Codegen, Type::Uint(32), size_of_array);
+                    let temp_res = vartab.temp_name("array_size", &Type::Uint(32));
```
LucasSte, line 143: As in the other case, there is no need for this.

src/codegen/statements.rs
```diff
@@ -63,21 +64,109 @@ pub(crate) fn statement(
...
+                            expr: num,
+                        },
+                    );
+                    cfg.array_lengths_temps.insert(*pos, temp_res);
+                }
+            } else if let Expression::Variable(_, _, res) = expr {
+                if let Some(to_add) = cfg.array_lengths_temps.clone().get(&res) {
+                    let num = Expression::Variable(pt::Loc::Codegen, Type::Uint(32), *to_add);
+                    let temp_res = vartab.temp_name("array_size", &Type::Uint(32));
+                    cfg.add(
```
LucasSte, line 158: There is no need for this.

LucasSte (changes requested):

src/codegen/array_boundary.rs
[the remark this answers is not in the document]
salaheldinsoliman: If the right hand side of the re-allocation is an AllocDynamicArray, this checks if AllocDynamicArray is called with a variable of type uint32, then create a uint32 and assign it to a new temporary holding the array length.
If something else is used in AllocDynamicArray, for example a uint256, the expression type will be Trunc. The same holds for numberliterals

src/codegen/array_boundary.rs
salaheldinsoliman: If re-allocation is done like this:
` uint[] arr = new uint256[](10);
arr = new `uint256[](20);`
I'm getting the Numberliteral inside AllocDynamicArray

src/codegen/array_boundary.rs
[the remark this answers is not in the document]
salaheldinsoliman: The expression type will not be a variable if it is not of type uint32, else it will be trunc/zext/numberliteral

src/codegen/statements.rs
salaheldinsoliman: If the size of the allocation is a uint256, this condition handles that. see this:
`uint256[] a = new uint256[](size);`
`ty:uint256[] %a = (alloc uint256[] len (trunc uint32 %value.temp.3))`
where size is a uint256

src/codegen/array_boundary.rs
salaheldinsoliman: Size is of type ast::Expression, I'm constructing a codegen::Expression to insert in the cfg

docs/optimizer.rst
LucasSte: @salaheldinsoliman Don't forget about this.

src/codegen/statements.rs
```diff
@@ -106,6 +144,21 @@ pub(crate) fn statement(
         }
         Statement::Expression(_, reachable, expr) => {
             if let ast::Expression::Assign(_, _, left, right) = &expr {
+                // If the left operand is a variable of type Array
+                if let ast::Expression::Variable(_, Type::Array(_, _), pos) = &**left {
```
LucasSte, line 148: This check should come after the `if should_remove_assignment`, otherwise you'll keep track of an assignment that does not exist in the code.

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
...
+
+.. code-block::
+
+    contract c {
+    	function test() public returns (int[]) {
+    		int[] array = new int[](3);
+    		array[1] = 1;
+    		return array;
+	}
+    }
```
seanyoung, line 208: The formatting of the example is messy

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
 The expression `a*b` is repeated throughout the function and will be saved to a temporary variable.
 This temporary will be placed wherever there is an expression `a*b`. You can see the pass in action when you compile
 this contract and check the CFG, using `solang --emit cfg`.
+
+
+
+.. _Array-Bound-checks-optimizations:
```
seanyoung, line 192: Only one blank line need here

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,102 @@
...
+    opt: &Options,
+) {
+    if let ast::Expression::Variable(_, _, right_res) = right {
+        // If we have initialized a temp var for this var
+        if cfg.array_lengths_temps.contains_key(right_res) {
+            let to_update = cfg.array_lengths_temps[right_res];
+
+            cfg.array_lengths_temps.insert(*pos, to_update);
+        } else {
+            // If the right hand side doesn't have a temp, it must be a function parameter or a struct memebr.
```
seanyoung, line 81: member, not memebr

src/codegen/cfg.rs
```diff
@@ -337,6 +338,7 @@ pub struct ControlFlowGraph {
     pub ty: pt::FunctionTy,
     pub selector: u32,
     current: usize,
+    pub array_lengths_temps: ArrayLengthVars, //a new member in the cfg that track the temp vars of arrays
```
seanyoung, line 341: A comment shouldn't say "this is a new member", it's not new once it is merged.

Please put the comment on the line above. The comment should say that these temp var track the length of the arrays

src/codegen/expression.rs
```diff
@@ -1619,12 +1623,23 @@ fn expr_builtin(
...
                 .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt))
                 .collect();
 
-            Expression::Builtin(*loc, tys.to_vec(), builtin.into(), args)
+            if !arguments.is_empty() && builtin == &ast::Builtin::ArrayLength {
+                // If an array length instruction is called
+                // Get the variable it is assigned with
+                if let Expression::Variable(_, _, num) = &arguments[0] {
+                    // Now that we have its temp in the map, retrieve the temp var res from the map
+                    if let Some(array_size_unrwapped) = cfg.array_lengths_temps.get(num) {
```
seanyoung, line 1636: `array_size_unrwapped` should be called something like `array_length_var`

src/codegen/statements.rs
```diff
@@ -63,21 +64,42 @@ pub(crate) fn statement(
...
                     loc: *loc,
                     res: *pos,
-                    expr,
+                    expr: expr.clone(),
                 },
             );
+
+            // Lets check if the declaration is a declaration of a dynamic array
+            if let Expression::AllocDynamicArray(_, _, size, _) = expr {
+                let temp_res = vartab.temp_name("array_size", &Type::Uint(32));
```
seanyoung, line 85: array length?

src/codegen/statements.rs
```diff
@@ -63,21 +64,42 @@ pub(crate) fn statement(
...
+                cfg.add(
+                    vartab,
+                    Instr::Set {
+                        loc: *loc,
+                        res: temp_res,
+                        expr: *size,
```
seanyoung, line 91: We have a problem here. The `size` expression is added twice to the cfg, duplicating code.

Say we have:
```
int[] foo = new int[]((x * y + 2)^3);
```
Now the cfg will have evaluate `(x * y + 2)^3` twice.

For the case that `expr = Expression::AllocDynamicArray(_, _, size, _)`, create a temporary variable like you are doing here, and then replace `expr` with `Expression::AllocDynamicArray(.., .., Box::new(Expression::Variable(loc, ty, temp_res), ..);` (with all the other fields copied) before adding it to the cfg.

That avoid the `.clone()` above too.

seanyoung (changes requested):

src/codegen/statements.rs
salaheldinsoliman: This set instruction sets the temp variable to the size expression of the AllocDynamicArray expression (variable/numberliteral etc...), why should be expr replaced with AllocDynamicArray?

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
...
+    contract c {
+    	function test() public returns (int[]) {
+    		int[] array = new int[](3);
+    		array[1] = 1;
+    		return array;
+	}
+    }
+    
+In this example we access ``array`` element 1, while the array length is 3. So, no bounds
+checks are necessary and code will more efficient if we do not emit the bounds check in
```
LucasSte, line 211:
```suggestion
checks are necessary and the code will more efficient if we do not emit the bounds check in
```

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
...
+the compiled contract.
+
+The array length is tracked in an invisible a temporary variable. This temporary variable
```
LucasSte, line 214:
```suggestion
The array length is tracked in an invisible a temporary variable. This temporary variable is
```

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
...
+always kept up to date, when for example a ``.pop()`` or ``.push()`` happens on the array
```
LucasSte, line 215:
```suggestion
always kept up to date when, for example, a ``.pop()`` or ``.push()`` happens on the array
```

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
...
+or an assignment is done. Then, when the bounds check is done, rather than retrieving the array length from
+the array at runtime, bounds check becomes the constant expression `1 < 3` which is
+always true, so the bounds check is omitted.
```
LucasSte, line 218:
```suggestion
always true, so the check is omitted.
```

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
...
+
+This also means that whenever the length of an array accessed using `.length`, this is
+replaced with a constant.
```
LucasSte, line 221: This whole phrase is unclear and needs rewriting.

docs/optimizer.rst
```diff
@@ -186,3 +186,39 @@ Check out the example below. It contains multiple common subexpressions:
...
+
+Note that this optimization does not cover every case, like when an array is passed in
+via a function argument for example, which means the length is unknown.
```
LucasSte, line 224:
```suggestion
Note that this optimization does not cover every case. When an array is passed
as a function argument, for instance, the length is unknown.
```

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,102 @@
...
+use indexmap::IndexMap;
+use num_bigint::BigInt;
+use num_traits::One;
+use solang_parser::pt::Loc;
+
+// IndexMap <ArrayVariable res , res of temp variable>
+pub type ArrayLengthVars = IndexMap<usize, usize>;
+
+// Function to modify arraye length temp
+pub(crate) fn modify_temp_array_length(
```
LucasSte, line 17: What is your opinion about making this a method of `ControlFlowGraph`?
@seanyoung

src/codegen/expression.rs
```diff
@@ -1619,12 +1623,23 @@ fn expr_builtin(
...
+                        //if its there, replace ArrayLength with the temp var
```
LucasSte, line 1637: Add a space after the slash and use capital letter.

src/codegen/expression.rs
```diff
@@ -2552,12 +2567,25 @@ fn array_subscript(
                         array_length
                     }
                 } else {
-                    Expression::Builtin(
+                    //if a subscript is encountered array length will be called
+
+                    //return array length by default
```
LucasSte, line 2572: Please, format your comments.

src/codegen/expression.rs
```diff
@@ -2552,12 +2567,25 @@ fn array_subscript(
...
+                    let mut returned = Expression::Builtin(
                         *loc,
                         vec![Type::Uint(32)],
                         Builtin::ArrayLength,
                         vec![array.clone()],
-                    )
+                    );
+
+                    if let Expression::Variable(loc, _, num) = &array {
+                        // If the size is known aka in our cfg.array_length_map, do the replacement
```
LucasSte, line 2581: What do you mean by `is known aka`?

src/codegen/statements.rs
```diff
@@ -93,6 +115,22 @@ pub(crate) fn statement(
                     expr: Expression::Undefined(param.ty.clone()),
                 },
             );
+            // handling arrays without size, defaulting the initial size with zero
```
LucasSte, line 118: Capital letter

src/codegen/array_boundary.rs
salaheldinsoliman: I agree it would be a better design, but was just leaving it till the end

seanyoung (changes requested):
Looking good, just found another issue though.

docs/optimizer.rst
```diff
@@ -186,3 +186,37 @@ Check out the example below. It contains multiple common subexpressions:
...
+            array[1] = 1;
+            return array;
+        }
+    }
+
+In this example we access ``array`` element 1, while the array length is 3. So, no bounds
+checks are necessary and the code will more efficient if we do not emit the bounds check in
+the compiled contract.
+
+The array length is tracked in an invisible a temporary variable. This temporary variable is always kept up to date when, for example, a ``.pop()`` or ``.push()`` happens on the array
```
LucasSte, line 213:
```suggestion
The array length is tracked in an invisible a temporary variable, which is always kept up to date when, for example, a ``.pop()`` or ``.push()`` happens on the array
```

docs/optimizer.rst
```diff
@@ -186,3 +186,37 @@ Check out the example below. It contains multiple common subexpressions:
...
+or an assignment is done. Then, when the bounds check is done, rather than retrieving the array length from
```
LucasSte, line 214:
```suggestion
or an assignment is done. Then, when the bounds check happens, rather than retrieving the array length from
```

docs/optimizer.rst
LucasSte: `As a consequence of this, whenever the length of an array is accessed using '.length', we can replace the corresponding builtin call with a constant.`

src/codegen/cfg.rs
```diff
@@ -420,6 +427,61 @@ impl ControlFlowGraph {
         }
         self.blocks[self.current].add(ins);
     }
+    // Function to modify array length temp
```
LucasSte, line 430: Use three slashes for doc comments. In addition, please add a few words about what your function does and in what context.

src/codegen/expression.rs
```diff
@@ -1619,12 +1622,23 @@ fn expr_builtin(
...
                 .collect();
 
-            Expression::Builtin(*loc, tys.to_vec(), builtin.into(), args)
+            if !arguments.is_empty() && builtin == &ast::Builtin::ArrayLength {
+                // If an array length instruction is called
+                // Get the variable it is assigned with
+                if let Expression::Variable(_, _, num) = &arguments[0] {
+                    // Now that we have its temp in the map, retrieve the temp var res from the map
+                    if let Some(array_length_var) = cfg.array_lengths_temps.get(num) {
+                        // If its there, replace ArrayLength with the temp var
```
LucasSte, line 1636:
```suggestion
                        // If it's there, replace ArrayLength with the temp var
```

src/codegen/statements.rs
```diff
@@ -63,19 +64,50 @@ pub(crate) fn statement(
...
-
                 //If we remove the assignment, we must keep expressions that have side effects
                 init.recurse(&mut params, process_side_effects_expressions);
                 return;
             }
 
-            let expr = expression(init, cfg, contract_no, Some(func), ns, vartab, opt);
+            let mut expression = expression(init, cfg, contract_no, Some(func), ns, vartab, opt);
+
+            // Lets check if the declaration is a declaration of a dynamic array
```
LucasSte, line 74:
```suggestion
            // Let's check if the declaration is a declaration of a dynamic array
```

src/codegen/statements.rs
```diff
@@ -63,19 +64,50 @@ pub(crate) fn statement(
...
+            if let Expression::AllocDynamicArray(loc_dyn_arr, Type::Array(arg1, arg2), size, opt) =
+                expression
```
LucasSte, line 76:
```suggestion
            if let Expression::AllocDynamicArray(loc_dyn_arr, ty @ Type::Array(..), size, opt) =
                expression
```

src/codegen/statements.rs
```diff
@@ -63,19 +64,50 @@ pub(crate) fn statement(
...
+                    Instr::Set {
+                        loc: *loc,
+                        res: temp_res,
+                        expr: *size,
+                    },
+                );
+                // If expression is an AllocDynamic array, replace the expression with AllocDynamicArray(_,_,tempvar,_) to avoid inserting size twice in the cfg
+                expression = Expression::AllocDynamicArray(
+                    loc_dyn_arr,
+                    Type::Array(arg1, arg2),
```
LucasSte, line 91:
```suggestion
                    ty,
```

src/codegen/statements.rs
```diff
@@ -63,19 +64,50 @@ pub(crate) fn statement(
...
+                    Box::new(Expression::Variable(*loc, Type::Uint(32), temp_res)),
+                    opt,
+                );
+                cfg.array_lengths_temps.insert(*pos, temp_res);
+            } else if let Expression::Variable(_, _, res) = expression {
```
LucasSte, line 96:
```suggestion
            } else if let Expression::Variable(_, _, res) = &expression {
```

src/codegen/statements.rs
```diff
@@ -63,19 +64,50 @@ pub(crate) fn statement(
...
+                // If declaration happens with an existing array, check is the size of the array is known.
+                // If the size of the right hand side is known (is in the array_length_map), make the left hand side track it
+                // Now, we will have two keys in the map that point to the same temporary variable
+                if let Some(to_add) = cfg.array_lengths_temps.clone().get(&res) {
```
LucasSte, line 100:
```suggestion
                if let Some(to_add) = cfg.array_lengths_temps.clone().get(res) {
```

src/codegen/statements.rs
```diff
@@ -93,6 +125,22 @@ pub(crate) fn statement(
                     expr: Expression::Undefined(param.ty.clone()),
                 },
             );
+            // Handling arrays without size, defaulting the initial size with zero
+
+            if let Type::Array(_, _) = param.ty {
```
LucasSte, line 130:
```suggestion
            if matches!(param.ty, Type::Array(..)) {
```

LucasSte (changes requested):
Your PR is coming into place. The major issue is the one @seanyoung pointed. I nitpicked a few code style issue.

src/codegen/cfg.rs
```diff
@@ -420,6 +427,61 @@ impl ControlFlowGraph {
...
+    pub fn modify_temp_array_length(
+        &mut self,
+        minus: bool,        // If the function is called from pushMemory or popMemory
+        address_arr: usize, // The res of array that push/pop is performed on
```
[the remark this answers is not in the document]
salaheldinsoliman: I'm using the same name in the original code

docs/optimizer.rst
LucasSte: @salaheldinsoliman You marked this as resolved, but the text has not changed.

LucasSte (changes requested):

seanyoung (approved):

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,67 @@
+use super::vartable::Vartable;
+use super::Options;
+use crate::ast;
+use crate::codegen::cfg::{ControlFlowGraph, Instr};
+use crate::codegen::expression::expression;
+use crate::sema::ast::{Namespace, Type};
+use solang_parser::pt::Loc;
+
+/// This function is called whenever an assignment statement of an array is encountered. we have to ensure
```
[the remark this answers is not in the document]
LucasSte: @salaheldinsoliman This has not been fixed

src/codegen/array_boundary.rs
```diff
@@ -0,0 +1,67 @@
...
+/// that the res of the array is tied to the res of the correct tempo variable.
```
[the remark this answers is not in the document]
LucasSte: @salaheldinsoliman This has not been fixed.

src/codegen/expression.rs
```diff
@@ -237,13 +238,21 @@ pub fn expression(
...
                 }
             }
 
-            let cfg_right = expression(right, cfg, contract_no, func, ns, vartab, opt);
+            let mut cfg_right = expression(right, cfg, contract_no, func, ns, vartab, opt);
+
+            // If an assignment where the left hand side is an array, call a helper function that updates the temp variable.
+            if let ast::Expression::Variable(_, Type::Array(..), pos) = &**left {
+                // If cfg_right is an AllocDynamicArray(_,_,size,_), update it such that it becomes AllocDynamicArray(_,_,temp_var,_) to avoid repetitive expressions in the cfg.
+                cfg_right = handle_array_assign(cfg_right.clone(), cfg, vartab, pos);
```
LucasSte, line 253:
```suggestion
                cfg_right = handle_array_assign(cfg_right, cfg, vartab, pos);
```
There is no need to clone here, because you are regenerating `cfg_right`, so a move suffices.

LucasSte (changes requested):

LucasSte (approved):
Please, fix this minor problem and squash you commits again so that we can merge your PR.

PR #901, opened 2022-06-30
Do not permit value transfers with external calls on Solana

seanyoung:
Solana does not offer a method of doing value/balance transfers with an external call/CPI.

Mark this feature as unsupported and remove the code.


docs/targets/solana.rst
```diff
@@ -74,6 +74,28 @@ are `some examples <https://solana-labs.github.io/solana-solidity.js/>`_. There
...
+.. _value_transfer:
+
+Transfering native value with a function call
+_____________________________________________
+
+The Solidity langauge on Ethereum allows value transfers with an external call. Solana
+Cross Program Invocation (CPI) does not support this. Hypothetically, the caller
+could transfer native balance and then inform the callee about the amount transferred
+by specifying this in the instruction data. However, it would be trivial to forge
+this. So, this is not supported at all. This means that:
```
LucasSte, line 86: You repeated `this` three times in these two lines. This sounds awkward.

docs/targets/solana.rst
```diff
@@ -74,6 +74,28 @@ are `some examples <https://solana-labs.github.io/solana-solidity.js/>`_. There
...
+
+ - Specifying `value:`` on an external call is not permitted
+ - The `payable` keyword has no effect
+ - `msg.value` is not supported
+
+Receive function
+________________
+
+In Solidity, a `receive()` function can be defined which gets called, whenever the
```
LucasSte, line 95: `function can be defined which gets called`
This does not sound right.

LucasSte (changes requested):

docs/targets/solana.rst
seanyoung: Rewritten

docs/targets/solana.rst
seanyoung: Rewritten

docs/targets/solana.rst
```diff
@@ -73,6 +73,33 @@ package has `documentation <https://solana-labs.github.io/solana-solidity.js/>`_
...
 `solang's integration tests <https://github.com/hyperledger-labs/solang/tree/main/integration/solana>`_.
 
+.. _value_transfer:
+
+Transfering native value with a function call
+_____________________________________________
+
+The Solidity langauge on Ethereum allows value transfers with an external call.
+Solana Cross Program Invocation (CPI) does not support this. So, this is not
+supported. This means that:
```
LucasSte, line 83:
```suggestion
Solana Cross Program Invocation (CPI) does not support this. This means that:
```

docs/targets/solana.rst
```diff
@@ -73,6 +73,33 @@ package has `documentation <https://solana-labs.github.io/solana-solidity.js/>`_
...
+
+ - Specifying `value:`` on an external call is not permitted
+ - The `payable` keyword has no effect
+ - `msg.value` is not supported
+
+.. note::
+
+    Hypothetically, the caller could transfer native balance and then inform the
+    callee about the amount transferred by specifying this in the instruction data.
+    However, it would be trivial to forge this.
```
LucasSte, line 93:
```suggestion
    However, it would be trivial to forge such an operation.
```
You are using `this` too much!

docs/targets/solana.rst
```diff
@@ -73,6 +73,33 @@ package has `documentation <https://solana-labs.github.io/solana-solidity.js/>`_
...
+
+Receive function
+________________
+
+In Solidity, if defined, the `receive()` function is called, whenever the native
+balance for an account gets credited. On Solana, there is no method to implement
+this. The balance of an accounts can be credited without any code being executed.
```
LucasSte, line 100:
```suggestion
In Solidity, the `receive()` function, when defined, is called whenever the native
balance for an account gets credited. On Solana, there is no method that implement
this. The balance of an account can be credited without any code being executed.
```

LucasSte (changes requested):
The description is still not clear. I added some suggestions for improvement.

LucasSte (approved):

PR #904, opened 2022-07-03
Bump solang-parser version for publish

seanyoung:
Foundry's forge-fmt needs a new solang-parser release.


seanyoung:
New version of `solang-parser` has been published, so merging

PR #906, opened 2022-07-04
Utilize Borsh encoding for 'abi.encode' on Solana

LucasSte:
This PR implements the `abi.encode` function using Borsh Encoding for Solana. It also introduces a few changes:

1. There was a bug in common subexpression elimination. When we use `x=y+1`, we can replace `y+1` by `x` until `x` changes value. I fixed the code so that the algorithm unlinks `y+1` to `x`, when the latter is assigned another value.
2. I added a `MemCpy` instruction to ease copying elements from arrays.
3. There is a `AdvancePointer` instruction to advance the buffer pointer when encoding. It also needed a new Type for account for the return of such an instruction. I named it `Type::BufferPointer`.
4. There is an `Encoding` trait that should be implemented by all encoding schemes. The function `create_encoder` should return the correct encoder given our target chain.
5. I have wired up all code necessary to support multidimensional dynamic arrays, but I have not tested it, because Solang does not support them.

The encoding happens in two passes. During the first one, I calculate the buffer's size. In the second, I save data into it.

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -19,6 +19,7 @@ use num_traits::One;
 use solang_parser::pt;
 use solang_parser::pt::CodeLocation;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
+use std::ops::AddAssign;
 use std::str;
 use std::sync::Arc;
 use std::{fmt, fmt::Write};
@@ -1896,7 +1897,7 @@ impl Namespace {
             if !field.ty.is_primitive() {
                 return None;
             } else {
-                size.add_assign(field.ty.size_of(self));
+                size.add_assign(field.ty.memory_size_of(self));
             }
         }
 
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -158,6 +158,12 @@ pub enum Instr {
         offset: Expression,
         value: Expression,
     },
+    /// Copies bytes from source address to destination address
+    MemCopy {
+        source: Expression,
+        destination: Expression,
+        bytes: Expression,
+    },
     /// Do nothing
     Nop,
 }
@@ -271,6 +277,16 @@ impl Instr {
             Instr::WriteBuffer { offset, value, .. } => {
                 value.recurse(cx, f);
                 offset.recurse(cx, f);
+            }
+
+            Instr::MemCopy {
+                source: from,
+                destination: to,
+                bytes,
+            } => {
+                from.recurse(cx, f);
+                to.recurse(cx, f);
+                bytes.recurse(cx, f);
             }
 
             Instr::AssertFailure { expr: None }
@@ -764,6 +780,17 @@ impl ControlFlowGraph {
                     .join(", ")
             ),
             Expression::Undefined(_) => "undef".to_string(),
+            Expression::AdvancePointer {
+                pointer,
+                bytes_offset,
+                ..
+            } => {
+                format!(
+                    "(advance ptr: {}, by {})",
+                    self.expr_to_string(contract, ns, pointer),
+                    self.expr_to_string(contract, ns, bytes_offset)
+                )
+            }
             Expression::GetRef(_, _, expr) => {
                 format!("(deref {}", self.expr_to_string(contract, ns, expr))
             }
@@ -1100,6 +1127,18 @@ impl ControlFlowGraph {
                     .join(", ")
             ),
             Instr::Nop => String::from("nop"),
+            Instr::MemCopy {
+                source: from,
+                destination: to,
+                bytes,
+            } => {
+                format!(
+                    "memcpy src: {}, dest: {}, bytes_len: {}",
+                    self.expr_to_string(contract, ns, from),
+                    self.expr_to_string(contract, ns, to),
+                    self.expr_to_string(contract, ns, bytes)
+                )
+            }
         }
     }
 
@@ -1848,5 +1887,19 @@ impl Namespace {
         } else {
             Type::Uint(256)
         }
+    }
+
+    /// Checks if struct contains only primitive types and returns its size
+    pub fn is_primitive_type_struct(&self, struct_no: usize) -> Option<BigInt> {
+        let mut size = BigInt::from(0u8);
+        for field in &self.structs[struct_no].fields {
+            if !field.ty.is_primitive() {
+                return None;
+            } else {
+                size.add_assign(field.ty.size_of(self));
+            }
+        }
+
+        Some(size)
     }
 }
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -159,7 +159,7 @@ pub enum Instr {
         offset: Expression,
         value: Expression,
     },
-    /// Copies bytes from source address to destination address
+    /// Copy bytes from source address to destination address
     MemCopy {
         source: Expression,
         destination: Expression,
@@ -787,7 +787,7 @@ impl ControlFlowGraph {
                 ..
             } => {
                 format!(
-                    "(advance ptr: {}, by {})",
+                    "(advance ptr: {}, by: {})",
                     self.expr_to_string(contract, ns, pointer),
                     self.expr_to_string(contract, ns, bytes_offset)
                 )
@@ -1890,11 +1890,19 @@ impl Namespace {
         }
     }
 
-    /// Checks if struct contains only primitive types and returns its size
-    pub fn is_primitive_type_struct(&self, struct_no: usize) -> Option<BigInt> {
+    /// Checks if struct contains only primitive types and returns its memory non-padded size
+    pub fn calculate_struct_non_padded_size(&self, struct_no: usize) -> Option<BigInt> {
         let mut size = BigInt::from(0u8);
         for field in &self.structs[struct_no].fields {
             if !field.ty.is_primitive() {
+                // If a struct contains a non-primitive type, we cannot calculate its
+                // size during compile time
+                if let Type::Struct(no) = &field.ty {
+                    if let Some(struct_size) = self.calculate_struct_non_padded_size(*no) {
+                        size.add_assign(struct_size);
+                        continue;
+                    }
+                }
                 return None;
             } else {
                 size.add_assign(field.ty.memory_size_of(self));
--- a/src/codegen/encoding/borsh_encoding.rs
+++ b/src/codegen/encoding/borsh_encoding.rs
@@ -1,21 +1,34 @@
-use crate::ast::{Namespace, RetrieveType, Type};
+use crate::ast::{ArrayLength, Namespace, RetrieveType, Type};
 use crate::codegen::cfg::{ControlFlowGraph, Instr};
 use crate::codegen::encoding::{
     calculate_size_args, finish_array_loop, increment_four, load_array_item, load_struct_member,
-    set_array_loop, Encoding,
+    load_sub_array, set_array_loop, AbiEncoding,
 };
 use crate::codegen::vartable::Vartable;
 use crate::codegen::{Builtin, Expression};
 use num_bigint::BigInt;
+use num_traits::{One, Zero};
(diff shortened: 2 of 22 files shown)


seanyoung (changes requested):
In the original design:
- we constant fold `abi.encode()` if the arguments are constant.
- deal with null references:
```
function test() returns (bytes) {
      string[] x = new foo[](5);
      return abi.encode(x);
}
```
https://docs.google.com/document/d/194WRaZ9z0J74cFsGpF2wroadS0uXOq3WUF7E7TzZlzA/edit?usp=sharing

LucasSte:
> * we constant fold `abi.encode()` if the arguments are constant.

This is totally possible, but given the complexity of the encoder, I decided to do this later. Can I submit it in another PR? I believe this one is complex and already covers this case if it is needed for testing. This will be my next addition then, before decoding.

LucasSte:
> In the original design:
> 
> * deal with null references:
> 
> ```
> function test() returns (bytes) {
>       string[] x = new foo[](5);
>       return abi.encode(x);
> }
> ```

This example works fine, I added a test for it. Do null pointers need special handling?

seanyoung:
> > In the original design:
> > 
> > * deal with null references:
> > 
> > ```
> > function test() returns (bytes) {
> >       string[] x = new foo[](5);
> >       return abi.encode(x);
> > }
> > ```
> 
> This example works fine, I added a test for it. Do null pointers need special handling?

Right, that example works because `vector_len()` deals with null pointers.

How about:

```
struct S {
    int f1;
    string f2;
}

function test() returns (bytes) {
     S[] s = new S[](5);
     return abi.encode(x);
}
```

LucasSte:
> > > In the original design:
> > > 
> > > * deal with null references:
> > > 
> > > ```
> > > function test() returns (bytes) {
> > >       string[] x = new foo[](5);
> > >       return abi.encode(x);
> > > }
> > > ```
> > 
> > 
> > This example works fine, I added a test for it. Do null pointers need special handling?
> 
> Right, that example works because `vector_len()` deals with null pointers.
> 
> How about:
> 
> ```
> struct S {
>     int f1;
>     string f2;
> }
> 
> function test() returns (bytes) {
>      S[] s = new S[](5);
>      return abi.encode(x);
> }
> ```

This works fine too. I'll push my ch

LucasSte:
> > > In the original design:
> > > 
> > > * deal with null references:
> > > 
> > > ```
> > > function test() returns (bytes) {
> > >       string[] x = new foo[](5);
> > >       return abi.encode(x);
> > > }
> > > ```
> > 
> > 
> > This example works fine, I added a test for it. Do null pointers need special handling?
> 
> Right, that example works because `vector_len()` deals with null pointers.
> 
> How about:
> 
> ```
> struct S {
>     int f1;
>     string f2;
> }
> 
> function test() returns (bytes) {
>      S[] s = new S[](5);
>      return abi.encode(x);
> }
> ```

This works fine too. I'll push my test shortly.

tests/solana_tests/abi_encode.rs
```diff
@@ -0,0 +1,779 @@
...
+    vec[30] = string[1];
+    vec[31] = string[0];
+    <[u8; 32]>::try_from(vec.to_owned()).unwrap()
+}
+
+#[test]
+fn null_pointer() {
+    #[derive(Debug, BorshDeserialize)]
+    struct S {
+        f1: u64,
```
LucasSte, line 723:
```suggestion
        f1: i64,
```

Sorry, my mistake!

src/codegen/cfg.rs
```diff
@@ -1831,4 +1871,26 @@ impl Namespace {
             Type::Uint(256)
         }
     }
+
+    /// Checks if struct contains only primitive types and returns its size
```
seanyoung, line 1875: surely you mean its non-padded memory size?

seanyoung: which should also be reflected in the name, I think

src/codegen/encoding/borsh_encoding.rs
```diff
@@ -0,0 +1,624 @@
...
+use num_traits::{One, Zero};
+use solang_parser::pt::Loc;
+use std::collections::HashMap;
+use std::ops::MulAssign;
+
+/// This struct implements the trait Encoding for Borsh encoding
+pub(super) struct BorshEncoding {
+    /// The trait Encoding has a 'cache_storage_loaded' function, which saves the variables to which
+    /// we loaded items from storage. This attribute is map between the argument number (in the
+    /// 'abi.encode' function) to the expression that represents the item loaded from storage.
```
seanyoung, line 19: please explain why we're keeping a cache of these

src/codegen/encoding/borsh_encoding.rs
```diff
@@ -0,0 +1,624 @@
...
+
+            Type::UserType(_) | Type::Unresolved | Type::Rational | Type::Unreachable => {
+                unreachable!("Type should not exist in codegen")
+            }
+
+            Type::InternalFunction { .. }
+            | Type::ExternalFunction { .. }
+            | Type::Void
+            | Type::BufferPointer
+            | Type::Mapping(..) => unreachable!("This type cannot be encoded"),
```
seanyoung, line 243: `Type::ExternalFunction` was not support by eth abi encoding because it would be incompatible.

With borsh encoding we can encode this (4 bytes function selector + address)

src/codegen/encoding/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+use crate::codegen::expression::load_storage;
+use crate::codegen::vartable::Vartable;
+use crate::codegen::{Builtin, Expression};
+use crate::Target;
+use num_bigint::BigInt;
+use solang_parser::pt::Loc;
+
+/// This trait should be implemented by all encoding methods (ethabi, Scale and Borsh), so that
+/// we have the same interface for creating encode and decode functions.
+pub(super) trait Encoding {
```
seanyoung, line 15: There are a ton of traits in rust called `Encoding`.  We should try to avoid collisions. How about `RuntimeEncoding`?

src/codegen/encoding/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+        &mut self,
+        loc: &Loc,
+        args: &[Expression],
+        ns: &Namespace,
+        vartab: &mut Vartable,
+        cfg: &mut ControlFlowGraph,
+    ) -> Expression;
+
+    /// Cache items loaded from storage to reuse them later
+    fn cache_storage_loaded(&mut self, arg_no: usize, expr: Expression);
```
seanyoung, line 27: This should really have a better explanation of why it is exists

src/codegen/encoding/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+/// This function should return the correct encoder, given the target
+pub(super) fn create_encoder(ns: &Namespace) -> impl Encoding {
+    match &ns.target {
+        Target::Solana => BorshEncoding::new(),
+        _ => unreachable!("Other types of encoding have not been implemented yet"),
+    }
+}
+
+/// Calculate the size of a set of arguments to encoding functions
+fn calculate_size_args<T: Encoding>(
```
seanyoung, line 39: With scale encoding, we will only estimate the size (or minimum size). I guess this is a future problem.

src/codegen/encoding/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+    let ty = expr.ty().unwrap_user_type(ns);
+    match &ty {
+        Type::Enum(_)
+        | Type::Uint(_)
+        | Type::Int(_)
+        | Type::Contract(_)
+        | Type::Bool
+        | Type::Address(_)
+        | Type::Bytes(_) => {
+            let size = ty.memory_size_of(ns);
```
seanyoung, line 78: This is borsh specific. With eth encoding, it is always 32 bytes

src/codegen/encoding/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+            Expression::NumberLiteral(Loc::Codegen, Type::Uint(32), size)
+        }
+
+        Type::Value => {
+            Expression::NumberLiteral(Loc::Codegen, Type::Uint(32), BigInt::from(ns.value_length))
+        }
+
+        Type::String | Type::DynamicBytes | Type::Slice => {
+            // When encoding a variable length array, the total size is "length (u32)" + elements
```
seanyoung, line 87: This borsh encoding specific. It's different for scale/eth. Does this belong in `borsh.rs`

src/sema/ast.rs
```diff
@@ -59,6 +59,10 @@ pub enum Type {
     Slice,
     /// We could not resolve this type
     Unresolved,
+    /// When we advance a pointer, it cannot be any of the previous types.
+    /// e.g. Type::Bytes is a pointer to struct.vector. When we advance it, it is a pointer
+    /// to latter's data region
+    BufferPointer,
```
seanyoung, line 65: Our CFG representation has a severe limitation: pointers aren't represented. This makes codegen very bad in some situations. :/

src/sema/types.rs
```diff
@@ -934,6 +935,14 @@ impl Type {
         }
     }
 
+    /// Fetch the type of an array element
+    pub fn elem_ty(&self) -> Self {
+        match self {
+            Type::Array(ty, _) => *ty.clone(),
+            _ => unreachable!("Type is no an array"),
```
seanyoung, line 942: type is not an array

seanyoung (approved):
Looks good

src/codegen/encoding/mod.rs
LucasSte: If we encode all arguments during compile time, it is not runtime encoding any more. What do you think of `ChainEncoding`?

seanyoung: We don't use the word `Chain` in our source tree, best not to introduce new nomenclature. How about `AbiEncoding` ?

LucasSte: `AbiEncoding` could be another option.

LucasSte:
> Looks good

You've approved already, but I think it is good if your review it again.

src/codegen/encoding/borsh_encoding.rs
seanyoung: "caching avoids having to repeat the same operation" is just not a useful comment.

Explain *why* you are caching things. Think of the 5 whys. Asking another why.

So, explain that why you need to do multiple passes over the data. 
- One to calculate the length, then an allocate with the size, and then the actual encoding.
-  Explain what's in the expression in the cache. Is it a clone of the original expression or are you storing the result in a `Expression::Variable(..)`.
- Explain what operations you are caching and which operations you are not caching
- The name suggests this is only a cache for arguments which are in storage, not memory.
- Why are you not caching other results. `abi.encode()` can sometimes have expensive arguments like `keccak256()`: https://github.com/hyperledger-labs/solang/blob/main/integration/substrate/UniswapV2ERC20.sol#L27-L33

seanyoung (changes requested):

src/codegen/encoding/borsh_encoding.rs
LucasSte: > The name suggests this is only a cache for arguments which are in storage, not memory.

The name is correct. I'm only caching arguments that have been loaded from storage.

> Why are you not caching other results. abi.encode() can sometimes have expensive arguments like keccak256(): https://github.com/hyperledger-labs/solang/blob/main/integration/substrate/UniswapV2ERC20.sol#L27-L33

There is no need to cache `keccka256`, because it is only calculated once. We know its size during compile time (256 bits), so the `get_expr_size` will return 32 bytes directly, without evaluating the result.

LucasSte: I updated the comment both in `encoding/mod.rs` and `borsh_encoding.rs`.

seanyoung (approved):

PR #909, opened 2022-07-07
Reduce the number of public items in the crates

seanyoung:
Reduce the number of public items in the crates

### DIFF
--- a/src/sema/contracts.rs
+++ b/src/sema/contracts.rs
@@ -244,23 +244,25 @@ fn resolve_base_args(
     ns.diagnostics.extend(diagnostics);
 }
 
-/// Visit base contracts in depth-first post-order
-pub fn visit_bases(contract_no: usize, ns: &ast::Namespace) -> Vec<usize> {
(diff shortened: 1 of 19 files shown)


src/sema/contracts.rs
```diff
@@ -244,23 +244,25 @@ fn resolve_base_args(
...
-    fn base(contract_no: usize, order: &mut Vec<usize>, ns: &ast::Namespace) {
-        for b in ns.contracts[contract_no].bases.iter().rev() {
-            base(b.contract_no, order, ns);
-        }
+impl ast::Namespace {
+    /// base contracts in depth-first post-order
+    pub fn contract_bases(&self, contract_no: usize) -> Vec<usize> {
+        let mut order = Vec::new();
+
+        fn base(contract_no: usize, order: &mut Vec<usize>, ns: &ast::Namespace) {
```
LucasSte, line 252: This is not wrong, but I'm not a fan of inlined function definitions.

LucasSte (approved):

PR #910, opened 2022-07-08
Allow free functions to be imported

seanyoung:
Free functions (ie. functions outside of contracts) where not imported.
Fix this.


src/sema/namespace.rs
```diff
@@ -1285,6 +1285,27 @@ impl Namespace {
         }
     }
 
+    /// Convert expression to IdentifierPath
+    pub fn expr_to_identifier_path(&self, expr: &pt::Expression) -> Option<pt::IdentifierPath> {
+        let mut expr = expr;
```
LucasSte, line 1290:
```suggestion
    pub fn expr_to_identifier_path(&self, mut expr: &pt::Expression) -> Option<pt::IdentifierPath> {
```
Parameters can be mutable in Rust.

LucasSte (approved):

PR #911, opened 2022-07-11
Update seal debug message

xermicus:
The `seal_println` runtime function was superseded by `seal_debug_message` in Substrate v3.0. This PR makes the `print("...")` builtin work again for current Substrate versions.

### DIFF
--- a/integration/substrate/UniswapV2ERC20.sol
+++ b/integration/substrate/UniswapV2ERC20.sol
@@ -35,7 +35,8 @@ contract UniswapV2ERC20 is IUniswapV2ERC20 {
     }
 
     function _mint(address to, uint value) internal {
-        print("_mint {} {}".format(to, value));
+        // Disabled for now, see #911
+        // print("_mint {} {}".format(to, value));
         totalSupply = totalSupply.add(value);
         balanceOf[to] = balanceOf[to].add(value);
         emit Transfer(address(0), to, value);
--- a/integration/substrate/balances.sol
+++ b/integration/substrate/balances.sol
@@ -14,6 +14,7 @@ contract balances {
 	function pay_me() public payable {
 		uint128 v = msg.value;
 
-		print("Thank you very much for {}".format(v));
+		// Disabled for now, see #911
+		// print("Thank you very much for {}".format(v));
 	}
 }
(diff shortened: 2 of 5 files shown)


seanyoung (changes requested):
This looks good, but I have a few concerns.

- Commits in hyperledger must have the `Signed-off-by: real name <email@address>` line at the end of each commit.
- We have a mock substrate VM which is very useful for debugging and for tests. This mock substrate should be updated too (see `tests/substrate.rs`).
- There are tests against the mock substrate which should pass, for example https://github.com/hyperledger-labs/solang/blob/main/tests/substrate_tests/strings.rs#L596-L612
- We have tests against real substrate 3.0 which will fail now. Updating them to the latest substrate will cause many problems.
- How is this tested against latest substrate? Can we automate this somehow (this might be hard to do right now).

seanyoung:
Some of the Solidity files in `integration/substrate/*.sol` contain `print(..);` lines. There aren't actually used as part of the test; you could comment them out for now, and the substrate tests will probably pass then.

The vscode test failure is unrelated to your changes. I'm fixing it here: https://github.com/hyperledger-labs/solang/pull/913

xermicus:
> Some of the Solidity files in `integration/substrate/*.sol` contain `print(..);` lines. There aren't actually used as part of the test; you could comment them out for now, and the substrate tests will probably pass then.

Thank you Sean! I probably will do just that. The plan is to update everything to the latest substrate version, once this PR is merged. Also, there won't be any more breaking changes in the future (on the substrate side), which should mean newer versions should always work. What we could then do is adding compiler flags for new features, for example (how to proceed here is another discussing for another day, first we just want to update everything to the latest version).

> The vscode test failure is unrelated to your changes. I'm fixing it here: #913

Perfect, I'll backport it when your're done, thanks!

xermicus:
@athei please check this as well

athei (approved):
LGTM

Good follow up would be to call this function from `assert_failure` so we get those `require` messages printed.

xermicus:
> Good follow up would be to call this function from `assert_failure` so we get those `require` messages printed.

Agree, I'll open an issue for that

PR #912, opened 2022-07-11
Support length for storage strings

LucasSte:
Add support for retrieving the length of a string saved in storage. I implemented this to help me debug an issue with Borsh Encoding.

Solc does not support this.


LucasSte:
> vscode test should be fixed in #913

Right, I'll wait your merging that.

PR #919, opened 2022-07-13
Refactor size_of function

LucasSte:
The function `Type::size_of` is currently only used for calculation how much size a type occupies in storage. We do not utilize it to estimate the size of an item in the memory. This PR updates the function's name to properly represent what it returns and updates its description to encourage proper usage.

### DIFF
--- a/src/sema/types.rs
+++ b/src/sema/types.rs
@@ -704,7 +704,7 @@ fn struct_offsets(ns: &mut Namespace) {
                 offsets.push(offset.clone());
 
                 if !field.recursive {
-                    offset += field.ty.size_of(ns);
+                    offset += field.ty.solana_storage_size(ns);
                 }
             }
 
@@ -993,10 +993,8 @@ impl Type {
         }
     }
 
-    /// Calculate how much memory we expect this type to use when allocated on the
-    /// stack or on the heap. Depending on the llvm implementation there might be
-    /// padding between elements which is not accounted for.
-    pub fn size_of(&self, ns: &Namespace) -> BigInt {
+    /// Returns the size a type occupies in memory
+    pub fn memory_size_of(&self, ns: &Namespace) -> BigInt {
         match self {
             Type::Enum(_) => BigInt::one(),
             Type::Bool => BigInt::one(),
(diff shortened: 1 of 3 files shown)


src/sema/types.rs
```diff
@@ -1025,18 +1025,18 @@ impl Type {
...
-            Type::Ref(ty) | Type::StorageRef(_, ty) => ty.size_of(ns),
-            Type::UserType(no) => ns.user_types[*no].ty.size_of(ns),
+            Type::Ref(ty) | Type::StorageRef(_, ty) => ty.storage_size_of(ns),
+            Type::UserType(no) => ns.user_types[*no].ty.storage_size_of(ns),
             _ => unimplemented!("sizeof on {:?}", self),
         }
     }
 
     /// Does this type fit into memory
     pub fn fits_in_memory(&self, ns: &Namespace) -> bool {
```
LucasSte, line 1038: @seanyoung I'm not sure here. Does this function require any changes?

src/sema/types.rs
```diff
@@ -993,10 +993,10 @@ impl Type {
...
     }
 
-    /// Calculate how much memory we expect this type to use when allocated on the
-    /// stack or on the heap. Depending on the llvm implementation there might be
-    /// padding between elements which is not accounted for.
-    pub fn size_of(&self, ns: &Namespace) -> BigInt {
+    /// Calculate how much memory this type occupies in storage.
+    /// Depending on the llvm implementation there might be padding between elements
+    /// which is not accounted for.
+    pub fn storage_size_of(&self, ns: &Namespace) -> BigInt {
```
seanyoung, line 999: I think we should be clear this is solana-specific. This should be unused for Substrate/ewasm

seanyoung (changes requested):

src/sema/types.rs
LucasSte: I'm going to create a function to estimate the size of each type in memory. I use this function in my Borsh encoding algorithm too.

seanyoung (approved):
Looks good, thanks!

PR #920, opened 2022-07-15
Do not resolve the arguments to overloaded functions/event emit twice

### LINKED ISSUE #779: Error message for "now" is wrong
In this contract, 
```solidity
contract test {
      event LockRecord(address,uint,uint256); 
      function foo() public {
          emit LockRecord(address(this), now, 34);
      }
}
```

I get this error: `'now' is not found`, however I should be given instructions to change `now` to `block.timestamp`.

seanyoung:
This covers to overloaded functions, constructors, methods, and event emit.

This solves some issues wrt to slice support which is covered in a future PR.

Fixes #779


solang-parser/src/diagnostics.rs
```diff
@@ -1,7 +1,7 @@
 use crate::pt;
 use crate::pt::Loc;
 
-#[derive(Debug, Eq, Hash, PartialOrd, Ord, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialOrd, Ord, PartialEq)]
```
LucasSte, line 4: This enum does not have any data saved in it, so you can derive `Copy` instead of clone. It will just copy a `u8`.

solang-parser/src/diagnostics.rs
```diff
@@ -96,6 +97,29 @@ impl Diagnostic {
...
+        Diagnostic {
+            level: Level::Error,
+            ty: ErrorType::CastError,
+            loc,
+            message,
+            notes: Vec::new(),
+        }
+    }
+
+    pub fn cast_error_with_note(loc: Loc, message: String, note_loc: Loc, note: String) -> Self {
```
LucasSte, line 110: I wonder if we should use `#[inline]` with these functions, because we are moving data three times: from the arguments to the `struct Diagnostic` and from the struct to the place that called the function.

src/codegen/yul/builtin.rs
```diff
@@ -325,7 +325,7 @@ fn equalize_types(
     let left_ty = left.ty();
     let right_ty = right.ty();
     if left_ty != right_ty {
-        let mut vec = Vec::new();
+        let mut diagnostics = Default::default();
```
LucasSte, line 328: What is `Default::default`?

src/sema/builtin.rs
```diff
@@ -881,54 +882,54 @@ pub fn resolve_call(
...
     symtable: &mut Symtable,
-    diagnostics: &mut Vec<Diagnostic>,
+    diagnostics: &mut Diagnostics,
 ) -> Result<Expression, ()> {
-    let matches = BUILTIN_FUNCTIONS
+    let funcs = BUILTIN_FUNCTIONS
         .iter()
         .filter(|p| p.name == id && p.namespace == namespace && p.method.is_none())
         .collect::<Vec<&Prototype>>();
+    let mut errors: Diagnostics = Default::default();
```
LucasSte, line 891: If you derive `Default` for `Diagnostics`, you can use `Diagnostics::default()`, instead. This makes the code more readable.

seanyoung: `Default` is derived for `Diagnostics`, that's why `Default::default()` works.

`Diagnostics::default()` is a nice compromise. I did have a `Diagnostics::new()` which just called `Default::default()` but I've removed that now, and I'm using `Diagnostics::default()` everywhere now.

solang-parser/src/diagnostics.rs
seanyoung: llvm will auto-inline functions based on some heuristics. I think `#[inline]` should really only be used if you want to force the compiler to inline, even though it makes the binary bigger, because of some measurement you've done which shows there is a benefit.

solang-parser/src/diagnostics.rs
seanyoung: If I add `Copy` without `Clone` the compiler says I need to derive `Clone` as well.

I don't think you can be sure this will be represented as a `u8`. In fact, I suspect it will be a `u32` by default - that is what C does at least.

solang-parser/src/diagnostics.rs
```diff
@@ -20,23 +20,24 @@ impl Level {
     }
 }
 
-#[derive(Debug, Eq, Hash, PartialOrd, Ord, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialOrd, Ord, PartialEq)]
 pub enum ErrorType {
```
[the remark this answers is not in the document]
seanyoung: See above.

Also, I wanted to implement `Clone` for Diagnostic. I don't want to implement `Copy` for diagnostics because I don't want to implicitly copy the structure by accident.

LucasSte (approved):

PR #921, opened 2022-07-15
Refactor strength reduce

LucasSte:
In my Borsh Encoding PR #906, strength reduce was overflowing the stack for some complex encoding cases. This PR decreases the stack size from 2KB to 768 bytes for the `reaching_values` function and from 9KB to 800 bytes for the `expression_values` function.

This PR also includes a refactoring in the strength reduce files.


seanyoung:
Moving the contents of the file and modifying it makes it impossible to review. It also makes the git history impossible to read.

I think this is best served with two commits: one moving the contents without any change (and the commit message should say so) and another should contain the fix. The order is not important.

LucasSte:
> Moving the contents of the file and modifying it makes it impossible to review. It also makes the git history impossible to read.
> 
> I think this is best served with two commits: one moving the contents without any change (and the commit message should say so) and another should contain the fix. The order is not important.

Fixed!

seanyoung (approved):
Ideally your "refactor" commit would say "move file contents without changes"

src/codegen/strength_reduce/reaching_values.rs
```diff
@@ -0,0 +1,207 @@
...
+
+                reaching_values(*block, cfg, &mut vars, block_vars, ns);
+            }
+
+            _ => (),
+        }
+    }
+}
+
+fn update_set(var_no: &usize, set: &HashSet<Value>, map: &mut Variables) -> bool {
```
xermicus, line 95: This function looks odd to me. I don't like the idea of passing a mutable map and returning a bool to inform the user whether the map was mutated. I'd rather not change the map in place, instead returning the HashSet containing prospect changes to the user. The user can get to his bool then for free anyways if he really needs it:  `update_set(...).is_empty()`. (If you really do need to have in place modification, I think a doc comment on whats happening exactly would be very helpful, also in this case the function name "update_set" can be very misleading when in fact the passed `map` and not the `set` get changed).

I allowed myself to refactor this function as follows making it shorter and in my opinion much easier to grasp whats going on: 

```rust
fn update_set(var_no: &usize, set: &HashSet<Value>, map: &Variables) -> HashSet<Value> {
    if let Some(existing) = map.get(var_no) {
        if existing.iter().next().map_or(false, |v| v.all_unknown()) {
            return HashSet::new();
        } else if let Some(v) = set.iter().find(|v| v.all_unknown()) {
            // If we are merging an unknown value, set the entire value set to unknown
            return HashSet::from_iter([v.clone()]);
        } else {
            let filtered = set.iter().filter(|v| existing.contains(v)).cloned();
            let mut changes = HashSet::from_iter(filtered);
            if existing.len() + changes.len() > MAX_VALUES {
                let bits = existing.iter().next().unwrap().bits;
                changes.insert(Value::unknown(bits));
                return changes;
            }
            return HashSet::new(); /* We dont return our changes here, is this intended? */
        }
    };

But please beware I could have made mistakes as well, above code snipped compiles but I didn't check for correctness otherwise.
    // We have no existing set. Create one but folding unknown
    if set.len() > MAX_VALUES || set.iter().any(|v| v.all_unknown()) {
        let bits = set.iter().next().unwrap().bits;
        return HashSet::from_iter([Value::unknown(bits)]);
    }
    set.clone()
}
```

xermicus (changes requested):
Left my two cents in the comments :upside_down_face:

src/codegen/strength_reduce/reaching_values.rs
```diff
@@ -0,0 +1,207 @@
...
+
+            set.insert(v.clone());
+
+            changes = true;
+            map.insert(*var_no, set);
+        } else {
+            for v in set {
+                if !existing.contains(v) {
+                    existing.insert(v.clone());
+                    changes = true;
```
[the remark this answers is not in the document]
LucasSte: @seanyoung What is your opinion? This is Sean's code. I just moved it.

src/codegen/strength_reduce/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+            }
+            _ => (),
+        }
+
+        transfer(instr, &mut vars, ns);
+    }
+}
+
+/// Walk through an expression, and do the replacements for the expensive operations
+fn expression_reduce(expr: &Expression, vars: &Variables, ns: &mut Namespace) -> Expression {
```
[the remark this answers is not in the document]
LucasSte: We need to treat differently each type of expression, so the match statement must be comprehensive.

src/codegen/strength_reduce/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+use bitvec::prelude::*;
+use expression_values::expression_values;
+use num_bigint::{BigInt, Sign};
+use num_traits::{One, ToPrimitive};
+use reaching_values::{reaching_values, transfer};
+use std::collections::{HashMap, HashSet};
+use std::convert::TryInto;
+use value::{is_single_constant, set_max_signed, set_max_unsigned, Value};
+
+/*
```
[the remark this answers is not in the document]
LucasSte: There is a very big comment here that explains the optimization pass. Does that replace a doc comment?

src/codegen/strength_reduce/mod.rs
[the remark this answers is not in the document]
seanyoung: I totally agree with @LucasSte, splitting this function to reduce nesting is not going increase readability. We just end up with a ton of functions. We are doing complex things and reducing everything to small functions is just not going to make very maintainable code.

src/codegen/strength_reduce/reaching_values.rs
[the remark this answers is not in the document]
seanyoung: The `if existing.len() > MAX_VALUES {` is for when the collected possible values during symbol execution exceeds a limit. Possibly a comment could be useful here.

src/codegen/strength_reduce/reaching_values.rs
seanyoung: This function is performance critical, we should reduce allocation because they totally kill performance. The strength reduce pass is often the most cpu intensive pass. Doing `.clone()` for each variable mutation will be a disaster.

Secondly the function you proposed is not doing the right thing. The variable `map: Variables` should be updated.

src/codegen/strength_reduce/mod.rs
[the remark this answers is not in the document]
xermicus: I was just wondering why this isn't a doc comment, as far as I remember if its like that it doesn't show up in cargo doc? Or maybe this is intentional?

src/codegen/strength_reduce/reaching_values.rs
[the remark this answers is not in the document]
xermicus: Sorry I wasn't aware it will in fact be updated.

src/codegen/strength_reduce/reaching_values.rs
xermicus: The clones are exactly the same ones as in the original code? I was thinking about return a HashSet containing refs which would avoid cloning entirely, but as far as I understood they would need to be cloned anyways for inserting them into `Variables`.
Returning `HashSet::new()` should not pose any overhead as well as it doesn't allocate.

> Secondly the function you proposed is not doing the right thing. The variable map: Variables should be updated.

The idea of the refactored code is that the caller would extend his Variables map himself. We can leave the function as it is if it works its fine I guess. But I'd at least change the name so it would imply the `map` is changed (and not the set), I was quite confused by that...

src/codegen/strength_reduce/mod.rs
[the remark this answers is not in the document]
xermicus: Not the end of the world, I'm just allergic to deep nesting :upside_down_face:

src/codegen/strength_reduce/mod.rs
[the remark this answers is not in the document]
LucasSte: I'm going to fix this.

src/codegen/strength_reduce/reaching_values.rs
LucasSte: The function `reaching_values` is called 1300 times in a recursive manner in an example I found. The purpose of the `update_set` function is to decrease the former's stack size without affecting performance. 

> The idea of the refactored code is that the caller would extend his Variables map himself.

This is going to add another overhead to our function. Returning the hash set not only will increase stack size, but also will need another clone when extending the original set.

src/codegen/strength_reduce/mod.rs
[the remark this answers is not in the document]
LucasSte: Fixed!

src/codegen/strength_reduce/reaching_values.rs
xermicus: > Returning the hash set not only will increase stack size,

Agree on that

> but also will need another clone when extending the original set.

Why?

xermicus: I still fail to see where my refactored code snippet adds any more clones/allocations that weren't already there. For example at Line 102 in the original code:

```rust
let mut set = HashSet::new();

set.insert(v.clone()); 

changes = true;
map.insert(*var_no, set);
```
Unless I'm missing something here, this already triggers 3 memory allocations, doesn't it? Or how is this different from `return HashSet::from_iter([v.clone()]);` and extending the map using that? Extending the map could even be faster in some cases here because you know before hand how much additional memory to allocate (vs. having multiple `insert` calls).  

I'm sure this could be refactored to use less allocations.. But is this really a problem for now? If it was fine before, just change the function name to something less confusing and maybe add a comment so it is clear what's happening and I'm happy with that.

seanyoung: @xermicus after re-examining this I see your point. How about going one step further do something like this:

```rust
fn update_set(set: &HashSet<Value>, existing: &Option<HashSet>) -> HashSet<Value> {
     if let Some(existing) = existing {
        if existing.iter().next().map_or(false, |v| v.all_unknown()) {
            HashSet::new()
        } else if let Some(v) = set.iter().find(|v| v.all_unknown()) {
            // If we are merging an unknown value, set the entire value set to unknown
            HashSet::from_iter([v.clone()])
        } else {
            let filtered = set.iter().filter(|v| existing.contains(v)).cloned();
            let mut changes = HashSet::from_iter(filtered);
            if existing.len() + changes.len() > MAX_VALUES {
                let bits = existing.iter().next().unwrap().bits;
                changes.insert(Value::unknown(bits));
                changes
            } else {
               HashSet::new() /* We dont return our changes here, is this intended? */
            }
        }
    // We have no existing set. Create one but folding unknown
    } else if set.len() > MAX_VALUES || set.iter().any(|v| v.all_unknown()) {
        let bits = set.iter().next().unwrap().bits;
        HashSet::from_iter([Value::unknown(bits)])
    } else {
        set.clone()
    }
}
```
Now that the `map` has been removed completely this function is much clearer.

This refactor does rely on `HashSet::new()` not allocating, which is news to me. How does that work?

Thanks for a good idea @xermicus

xermicus: @seanyoung if we are going for this please check again my `// XXX` comments

LucasSte: > Also if the stack size is a problem returning empty HashSet is a problem we could try to inline this function?

Inlining the function will only add more instructions to `reaching_values`, increasing its stack size even more.

> this sounds like super premature optimization,

Without these changes, strength reduce overflows its stack.

LucasSte:
```solidity
            let filtered = set.iter().filter(|v| existing.contains(v)).cloned();
            let mut changes = HashSet::from_iter(filtered);
            if existing.len() + changes.len() > MAX_VALUES {
                 // XXX should be like that? 
                let bits = existing.iter().chain(changes.iter()).next().unwrap().bits; // I think we should be 
                changes.insert(Value::unknown(bits));
            } 
            changes  // XXX should be like that? 
```

We should add to `existing` the values of `set`. `set.iter().filter(|v| existing.contains(v)).cloned()` is the intersection between `existing` and `set`.

The update of map[var_no] happens in two different ways. When we crate a new set, we insert it in `map` (this overwrites any existing value). If we do not create a new set, we add the values of set to `existing`. Returning a HashSet would mean also discerning between when to insert a new value to `map[var_no]` and when to extend it.

LucasSte: @xermicus I tried to use you suggestion, but I couldn't get it to work. Feel free to patch my branch to check if you can improve your suggestion. I have just modified the function's name and added a doc comment.

xermicus: Sure I can give it one more shot. I don't understand what `let bits = existing.iter().next().unwrap().bits;` does? I don't get why it chooses just whatever the `.next()` thing is inside `existing`. This HashSet has no specific ordering, or has it?

xermicus: As discussed, lets leave it as it is and address that in a different issue/PR.

xermicus (approved):

PR #925, opened 2022-07-18
Fix parsing of `uint x = 1 /** foo */ + 2;`

seanyoung:
DocComments should not be stored in the parse tree.

There are two commits, the first fixes an issue with  the existing doccomment parsing (and adjusts the test results).


solang-parser/src/doccomment.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub value: String,
+    pub value_offset: usize,
+}
+
+enum CommentType {
+    Line,
+    Block,
+}
+
+/// For the start to end offset, filter all the doc comemnts comments and parse them into
```
LucasSte, line 31: Shouldn't this be `From the start to end`?

LucasSte (approved):

solang-parser/src/doccomment.rs
seanyoung: Thanks for catching that, fixed.

PR #927, opened 2022-07-19
codegen: print the require error string for target substrate

xermicus:
Since we have now debug prints working in substrate, error strings from `require` assertions can be appended the debug message buffer. This should make debugging during contract development much easier.

This should pass integration tests as soon as they use an up-to-date substrate version.


LucasSte (approved):

LucasSte:
> I think the code change looks good.The vscode CI tests failure is spurious, but the substrate failure is not.

I re-ran the failing tests.

athei (commented):
Don't you need to make changes in the substrate backend so that a print is actually emitted from `AssertFailure`? AFAIK it only emits an `unreachable`.

xermicus:
> Don't you need to make changes in the substrate backend so that a print is actually emitted from AssertFailure? AFAIK it only emits an unreachable.

This changes codegen to insert a print instruction just before AssertFailure (if there is something to be printed). I thought this should be fine as in non-dev nodes the debug prints are a no-op.

I tested it on a local --dev node (without any changes to it) and it worked as expected.

athei (approved):
> This changes codegen to insert a print instruction just before AssertFailure (if there is something to be printed).

Ahh yeah of course. I just missed it. I am getting blind. Must be the age.

> I thought this should be fine as in non-dev nodes the debug prints are a no-op.

Yes it is fine for now. But it is not exactly a non-op. The contract will still call the imported print function obviously. pallet-contracts will just do nothing but it still charges some gas for it (for the empty imported function call. [See here](https://github.com/paritytech/substrate/blob/1399afdcd8ab4d7fad149c96f9cadcaf04b94f86/frame/contracts/src/wasm/runtime.rs#L2047) where the gas is charged unconditionally.

Also, the difference is not between dev or non-dev nodes. Even production nodes will print the message (and return it to the client) but ONLY when the contract is called via RPC (as oposed being executed on-chain in a transaction). So essentially: When you dry-run a contract you get the message.

> I tested it on a local --dev node (without any changes to it) and it worked as expected.

Nice. Thanks 👍

xermicus:
Ah I see, thanks for the clarification!

Giving it some more thought, this could be made optional if for example we would introduce a `--release` flag in solang for the substrate target. I'll keep this in mind for future improvements.

seanyoung:
> Giving it some more thought, this could be made optional if for example we would introduce a `--release` flag in solang for the substrate target. I'll keep this in mind for future improvements.

This is a good idea, which needs some more thinking about. It would be useful if all errors, also for example `array access out of bounds` is printed, possibly with the source file name and line number. You may not what this in production, so a compiler flags with `--report-errors=all | no-line-info` might be useful.

Something for another time.

athei:
Yes was thinking about something like this. We have a `--release` target in ink! which removes everything that prints a string. We even need to recompile `libcore` with `--features panic_immediate_abort` in order to remove the strings printed from there.

But this is surely another issue :)

Issue #903, opened 2022-07-01
Multidimensional arrays crash LLVM


LucasSte:
This contract works fine for `solc`, but crashes on Solang (for all targets):

```solidity
contract Testing {

    uint16[2][4][] stor_arr;
    function getThis() public returns (bytes memory) {
      
        uint16[2][4][] memory arr2 = stor_arr;
        bytes memory d = abi.encode(arr2);
        return d;
    }
}
```

LucasSte:
This is the error I get:
```
solang: /home/lucas/Downloads/llvm-project/llvm/include/llvm/IR/Instructions.h:921: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
```

PR #929, opened 2022-07-19
Default last dimension as the dynamic one

LucasSte:
This PR defaults the last dimension of multidimensional arrays as the only one that can be dynamic. It fixes issue #903 only for Solana.


seanyoung (changes requested):
The description and title does not make any sense to me.

Surely the problem is that in a multidimensional array, the outer dimension of the array is the last element in the dimensions field (and not the first). Nothing to do with dynamic or not. 

Please also update the commit.

LucasSte:
> The description and title does not make any sense to me.
> 
> Surely the problem is that in a multidimensional array, the outer dimension of the array is the last element in the dimensions field (and not the first). Nothing to do with dynamic or not.
> 
> Please also update the commit.

I hope the name is better now.

seanyoung (approved):
Thanks for updating the name, this is very clear now.

Issue #935, opened 2022-07-24
compiled contract can not be deployed: InvalidStorageAccountError


yuwiggin:
An issue may be related with [solana-labs/solana-solidity.js](https://github.com/solana-labs/solana-solidity.js/issues/31).

Any suggestion?

PR #941, opened 2022-07-28
Deploy does not want the program account

seanyoung:
This was changed in @solana/solidity 0.0.20 but the example here was not
updated.


seanyoung:
Fixes #935

LucasSte (approved):

Issue #810, opened 2022-05-06
Solang does not allow assignment to array


LucasSte:
This contract works on Solc, but fails on Solang.

```solidity 
contract Splitter {
    address payable[] splitAddresses;

    function split(address payable destinationAddress1, address payable destinationAddress2) public payable {
        splitAddresses = [destinationAddress1, destinationAddress2]; // Get specified addresses

        sendEth(); // Send ether
    }

    function sendEth() internal {
        for (uint x = 0; x != splitAddresses.length; x++) {
            splitAddresses[x].transfer(msg.value/splitAddresses.length); // Send ether
        }
    }
}
```

The error is the following:
```
error: conversion from address payable[2] to address payable[] not possible
  ┌─ /home/lucas/Documents/solang/examples/array.sol:5:26
  │
5 │         splitAddresses = [destinationAddress1, destinationAddress2]; // Get specified addresses
  │                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


```

PR #942, opened 2022-07-28
Enable initialization of arrays with array literals

LucasSte:
This PR enables the initialization of both dynamic memory and dynamic storage arrays with array literals. The limitation is that this works only for single dimensional arrays. It fixes issue #810.


src/sema/expression.rs
```diff
@@ -370,6 +370,21 @@ impl Expression {
...
             }
+
+            (
+                &Expression::ArrayLiteral(..),
+                Type::Array(from_ty, from_dims),
+                Type::Array(to_ty, to_dims),
+            ) => {
+                if from_ty == to_ty
+                    && from_dims.len() == to_dims.len()
+                    && from_dims.len() == 1
```
seanyoung, line 381: Any reason this is limited to single-dimension?

seanyoung (approved):
Would be nice to permit this for multi-dimensional arrays

src/sema/expression.rs
LucasSte: I think `codegen::Expression::AllocDynamicArray` does not work with multidimensional arrays.

PR #943, opened 2022-07-28
Merge readbuffer builtins

LucasSte:
This PR merges all the `codegen::Builtin::Read<type>` builtin type with a single `codegen::Builitin::ReadFromBuffer`, because they are all implemented equally in emit. This streamlines the development of `abi.decode` with Borsh encoding.


seanyoung (approved):
Nice change!

PR #947, opened 2022-08-02
Allow seeds to be passed to external calls for program signing

seanyoung:
This adds the seeds parameter to external calls. This does not enable creating PDA style contracts yet, that will come in another PR.

This PR simply adds the seeds to `contract.function{seeds: [ [ seed1 ], [ [ seed2 ]}(arg1, arg2);`.


LucasSte:
> This adds the seeds parameter to external calls. This does not enable creating PDA style contracts yet, that will come in another PR.
> 
> This PR simply adds the seeds to `contract.function{seeds: [ [ seed1 ], [ [ seed2 ]}(arg1, arg2);`.

This is a weird syntax. Do we need to wrap the seeds with square brackets?

seanyoung:
> > This adds the seeds parameter to external calls. This does not enable creating PDA style contracts yet, that will come in another PR.
> > This PR simply adds the seeds to `contract.function{seeds: [ [ seed1 ], [ [ seed2 ]}(arg1, arg2);`.
> 
> This is a weird syntax. Do we need to wrap the seeds with square brackets?

Yes, it's a slice of an slices of bytes. So you can do `[ [ hex"12", hex"0000" ], [ "f", "o", "o", "b", "a", "r' ] ]`. Those are two sets of seeds.

src/emit/mod.rs
```diff
@@ -4152,6 +4149,84 @@ pub trait TargetRuntime<'a> {
...
+
+                                    bin.builder.build_store(dest, val);
+
+                                    let dest = unsafe {
+                                        bin.builder.build_gep(
+                                            output_seeds,
+                                            &[
+                                                bin.context.i32_type().const_int(i, false),
+                                                bin.context.i32_type().const_int(1, false),
+                                            ],
```
LucasSte, line 4208: It is possible to implement this in codegen?

tests/solana.rs
```diff
@@ -63,6 +63,10 @@ struct AccountState {
     lamports: u64,
 }
 
+/// We have a special callback function which tests that during CPI, the correct
+/// parameters are passed in.
```
LucasSte, line 67:
```suggestion
/// We have a special callback function which tests that the correct
/// parameters are passed in during CPI.
```
Does this sound better?

LucasSte (changes requested):

docs/language/functions.rst
```diff
@@ -180,6 +180,39 @@ see the section on :ref:`account_meta`.
...
+        function test(address addr, bytes seed) public {
+            bytes instr = new bytes(1);
+
+            instr[0] = 1;
+
+            AccountMeta[1] metas = [
+                AccountMeta({pubkey: addr, is_writable: true, is_signer: true})
+            ];
+
+            token.call{accounts: metas, seeds: [ [ seed ] ]}(instr);
```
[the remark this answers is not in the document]
seanyoung: Yes, because it's a list of a list.

tests/solana.rs
seanyoung: Yes, it does sound better

src/codegen/subexpression_elimination/instruction.rs
```diff
@@ -326,6 +326,7 @@ impl AvailableExpressionSet {
                 success,
                 address,
                 accounts,
+                seeds,
                 payload,
```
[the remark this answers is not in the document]
seanyoung: That was missed, was missed for accounts too. I've removed the `..` so we don't make this mistake again

src/sema/expression.rs
```diff
@@ -6965,6 +6965,77 @@ fn array_literal(
     let mut dims = Box::new(Vec::new());
     let mut flattened = Vec::new();
 
+    let resolve_to = match resolve_to {
+        ResolveTo::Type(Type::Array(elem_ty, _)) => ResolveTo::Type(elem_ty),
+        // Solana seeds are a slice of slice of bytes, e.g. [ [ "fo", "o" ], [ "b", "a", "r"]]. In this
```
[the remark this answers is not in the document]
seanyoung: You could see it as a `bytes[][2]`, that's true.  We explicitly resolve the seeds as a slice of slice of bytes.

src/emit/mod.rs
seanyoung: I don't think so. The `sol_invoke_signed_c()` has a weird signature. It takes a slice of slices, but the number of outer slices is passed as the second parameter.

src/sema/expression.rs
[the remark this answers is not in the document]
LucasSte: We need slices because of the type required for the call, right? I was just wondering if, instead of slices, we could process the seeds as `bytes[][2]` to avoid adding new types to Solidity.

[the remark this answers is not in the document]
seanyoung: Exactly, we need new types for the system call. 

Also, at some point we want to add support for `foo[5:6]` to Solidity too, which is a missing language feature right now. For that we would need to support slices.

docs/language/functions.rst
```diff
@@ -180,6 +180,39 @@ see the section on :ref:`account_meta`.
...
```
LucasSte, line 208: I think it is worth writing a few words here or in the `Types` section to explain what a slice is and how to create one. Your example explains well how to use seeds with external calls, but I believe a few words on the expected type of the argument are really helpful.

LucasSte (changes requested):

docs/language/functions.rst
seanyoung: You're right, this section could use a lot more words. I've expanded it.

LucasSte (approved):

PR #949, opened 2022-08-03
Add debug information to LLVM Instructions

thanhtoantnt:
Fixes https://github.com/hyperledger-labs/solang/issues/584

# Description
- Add debug information to functions and instructions of the LLVM IR generated by `solang`.
- The return instruction `ret` currently does not have the debug location. But it is the case that the return `ret` instruction does not have the corresponding Solidity program statement. It is a special case when it is generated by `solang`.
- Refer to: https://github.com/hyperledger-labs/solang/blob/d05885270708ab822716300f59c6ee1624a5adb3/src/emit/mod.rs#L3404-L3405

# How Has This PR Been Tested?

Please describe the tests that you ran to verify your changes. Please also note any relevant details for your test configuration.

- [x] examples/example.sol
- [x] examples/flipper.sol
- [x] examples/increment.sol


seanyoung:
Amazing stuff!

src/emit/mod.rs
```diff
@@ -3285,6 +3351,37 @@ pub trait TargetRuntime<'a> {
             BasicBlock { bb, phis }
         }
 
+        fn get_ins_loc(ins: &Instr) -> Option<pt::Loc> {
+            match ins {
+                Instr::Set { loc, expr, .. } => match loc {
+                    pt::Loc::File(_, _, _) => Some(*loc),
+                    _ => Some(expr.loc()),
+                },
+                Instr::Call { args, .. } if args.is_empty() => None,
```
seanyoung, line 3360: The `Instr` really needs a `loc` field in every possible enumeration.

src/emit/mod.rs
```diff
@@ -3247,6 +3251,68 @@ pub trait TargetRuntime<'a> {
...
+                    compile_unit.as_debug_info_scope(),
+                    function.get_name().to_str().unwrap(),
+                    None,
+                    file,
+                    /* line no*/ 0, // TODO: rectify line name
+                    di_func_type,
+                    true,
+                    true,
+                    /* scope line */ 0, // TODO: rectify scope line
+                    inkwell::debug_info::DIFlagsConstants::PUBLIC,
```
seanyoung, line 3309: Not all functions are public, also TODO could be implemented

seanyoung (changes requested):
Please ensure the CI tests work too.

src/emit/mod.rs
LucasSte: If every `Instr` is going to have a Location, please implement the trait `CodeLocation` to retrieve the `Loc`.

LucasSte (changes requested):

src/emit/mod.rs
thanhtoantnt: @seanyoung Do you have any suggestions to define the `flags` field, i.e., it is `PUBLIC` or `PRIVATE`?

seanyoung: if `cfg.public` is true, ` inkwell::debug_info::DIFlagsConstants::PUBLIC`, else private.

thanhtoantnt: Thanks. How about the `scope_line` and `line_no`? I try to use `bin.builder.get_current_debug_location().unwrap();` but I always get the line number of `3267742850`.

thanhtoantnt: Currently, I use the location of the first instruction: `cfg.blocks[0].instr.first().unwrap().loc()`. Free free to request changes if you feel it is not a suitable way.

LucasSte:
Run `cargo test --workspace` on your machine to detect which tests are failing. If you have a look at our CI, there are tests failing.

seanyoung (approved):
Clearly there are improvements needed the location tracking, but that shouldn't hold up this PR.

Looks good I think.

LucasSte (approved):

xermicus (approved):

PR #950, opened 2022-08-03
Refactor builtin structs

LucasSte:
This PR introduces a few changes to the builtin structs in Solang.

1. `Type::Struct(usize)` has changed to `Type::Struct(StructType)` to encompass both user defined structs and builtin structs.
2. Solana builtin structs are now held in a new data structure. We can easily retrieve their `StructDecl` data using `StructType::get_definition`. This works also for user defined structs.
3. `codegen::Expression::ExternalFunction` has been eliminated, because I replaced it by a `StructLiteral` of `Type::Struct(StructType::ExternalFunction)`.
4. External function was defined as a struct of `{address, selector}`. I changed it to `{selector, address}` to coincide with its encoding form.

There was an idea to get rid of `Type::ExternalFunction` in favor of `Type::Struct(StructDecl::ExternalFunction)` in codegen, however, this would mean boilerplate code for converting types allowed in the AST to types allowed in codegen. In addition, a large refactoring in `cfg.vars` was needed to avoid mismatching types. Thus, this has not been done.


src/abi/ethereum.rs
```diff
@@ -1,4 +1,5 @@
 // ethereum style ABIs
+use crate::ast::StructType;
 use crate::sema::ast::{Namespace, Parameter, Type};
```
seanyoung, line 3:
```rust
use crate::sema::ast::{Namespace, Parameter, StructType, Type};
```

We should probably remove the `use sema::ast;` in `src/lib.rs` so the ast doesn't have two paths

src/sema/builtin_structs.rs
```diff
@@ -0,0 +1,199 @@
...
+                },
+            ],
+            offsets: Vec::new(),
+            storage_offsets: Vec::new(),
+        },
+    ]
+});
+
+impl StructType {
+    pub fn get_definition<'a>(&'a self, ns: &'a Namespace) -> &StructDecl {
```
seanyoung, line 191: How about calling this `fn definition()`? I think `get_` is usually superfluous (and the word terribly overloaded in the English language anyway).

seanyoung (changes requested):
Very nice change, this makes the whole builtin/extern function type a lot cleaner!

src/abi/ethereum.rs
LucasSte: I removed the `use sema::ast;` from `src/lib.rs`, but it entailed many other changes on a bunch of files.

seanyoung (approved):
Fantastic!

Issue #948, opened 2022-08-02
docs(readme): add section with build instructions 


petermetz:
I tried cloning the project and then doing a `$ cargo build` which resulted in 
`thread 'main' panicked at 'could not execute llvm-config: Os { code: 2, kind: NotFound, message: "No such file or directory" }', build.rs:10:14`

I could make it go away by installing llvm [1] (and in hindsight it should've been obvious that it is a dependency) but still, the documentation should have a section for newcomers setting up the project from scratch on the most popular OS' such as Linux (Debian/Ubuntu is where my heart is at) macOS and maybe even Windows (if it works on it)

[1]: `$ sudo apt install llvm`

Full log output:

```sh
$ cargo build
   Compiling syn v1.0.98
   Compiling serde v1.0.141
   Compiling tinyvec_macros v0.1.0
   Compiling pin-project-lite v0.2.9
   Compiling futures-sink v0.3.21
   Compiling matches v0.1.9
   Compiling percent-encoding v2.1.0
   Compiling ryu v1.0.10
   Compiling itoa v1.0.2
   Compiling unicode-bidi v0.3.8
   Compiling pin-utils v0.1.0
   Compiling futures-io v0.3.21
   Compiling unicode-xid v0.2.3
   Compiling either v1.7.0
   Compiling bitflags v1.3.2
   Compiling os_str_bytes v6.2.0
   Compiling termcolor v1.1.3
   Compiling bytes v1.2.1
   Compiling tower-service v0.3.2
   Compiling nodrop v0.1.14
   Compiling tower-layer v0.3.1
   Compiling remove_dir_all v0.5.3
   Compiling tap v1.0.1
   Compiling fastrand v1.8.0
   Compiling funty v1.1.0
   Compiling lalrpop-util v0.19.8
   Compiling unicode-width v0.1.9
   Compiling strsim v0.10.0
   Compiling textwrap v0.15.0
   Compiling cpufeatures v0.2.2
   Compiling constant_time_eq v0.1.5
   Compiling wyz v0.2.0
   Compiling hex v0.4.3
   Compiling base58 v0.2.0
   Compiling parity-wasm v0.45.0
   Compiling funty v2.0.0
   Compiling instant v0.1.12
   Compiling tracing-core v0.1.29
   Compiling phf_shared v0.10.0
   Compiling tinyvec v1.6.0
   Compiling form_urlencoded v1.0.1
   Compiling itertools v0.10.3
   Compiling clap_lex v0.2.4
[...]

PR #952, opened 2022-08-04
Organize installation instructions

LucasSte:
I reorganized the installation instructions from our docs, making it clear that the one needs to install Solana's LLVM fork before building Solang, following the suggestion from #948.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -1,11 +1,18 @@
 Installing Solang
 =================
 
-The Solang compiler is a single binary. It can be installed in different ways.
+The Solang compiler is a single binary. It can be installed in different ways, listed below.
 
+1. :ref:`Download from Homebrew <download-brew>` (MacOS only)
+2. :ref:`Download pre-compiled binaries <download-binaries>`
+3. :ref:`Download from a Docker container <download-docker>`
+4. :ref:`Build using Dockerfile <build-dockerfile>`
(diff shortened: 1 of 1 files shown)


docs/installing.rst
```diff
@@ -184,3 +187,15 @@ Once you have the correct llvm version in your path, simply run:
...
 
 The executable will be in ``target/release/solang``.
+
+.. _build-from-crates:
+
+Alternative step 2: Build Solang from crates.io
+_______________________________________________
+
+The latest Solang release is  on `crates.io <https://crates.io/crates/solang>`_. Once you have the
+correct llvm version in your path, simply run:
```
seanyoung, line 197: s/llvm/LLVM

seanyoung (changes requested):
This is much clearer, thanks.

This file contains a few lower case "llvm". If you're going to change it to upper case, please make it consistent.

Also do you think its worth mentioning in the top-level `README.md`?

docs/installing.rst
LucasSte: Fixed

LucasSte:
I included in the first section a list of our installation methods. The idea behind this is to promptly show the user how they can install Solang.

LucasSte:
> Also do you think its worth mentioning in the top-level `README.md`?

This is a good idea, but deserves another PR.

seanyoung (approved):

PR #953, opened 2022-08-08
Fix clippy warnings for upcoming rust 1.63 release

seanyoung:
rust 1.63 should be released on the 11th of August 2022


LucasSte (approved):

LucasSte:
Do we need `PartialEq` if we have `Eq`?

seanyoung:
> Do we need `PartialEq` if we have `Eq`?

Yes, we do. It's a bit odd that rust now always both derived. If you either miss `Eq` or `PartialEq` you get either compiler errors or clippy warnings.

xermicus (approved):

PR #957, opened 2022-08-10
Bytes of Type::Bytes must be reversed before encoding

LucasSte:
`Type::Bytes` is represented as an integer in LLVM. This means its bytes are reversed if we use the `Instr::WriteBuffer` instruction. This PR fixes this issue.


seanyoung (approved):
LGTM

PR #959, opened 2022-08-12
Make --generate-debug-info hidden

seanyoung:
This feature is incomplete and does not have any users yet. For example, there is a debugging endpoint for Solana in the works, but this is not finished yet. Also, this will be useful for static analysis, which requires debug information for higher-level type information.

### DIFF
--- a/src/bin/solang.rs
+++ b/src/bin/solang.rs
@@ -6,13 +6,13 @@ use clap::{
 };
 use itertools::Itertools;
 use num_traits::cast::ToPrimitive;
-use serde::Serialize;
 use solang::{
     abi,
     codegen::{codegen, OptimizationLevel, Options},
     emit::Generate,
     file_resolver::FileResolver,
-    sema::{ast::Namespace, diagnostics},
+    sema::ast::Namespace,
+    standard_json::{EwasmContract, JsonContract, JsonResult},
     Target,
 };
 use std::{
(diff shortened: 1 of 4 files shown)


src/bin/solang.rs
```diff
@@ -878,19 +858,45 @@ fn imports_arg(matches: &ArgMatches) -> FileResolver {
...
+fn parse_import_map(map: &str) -> Result<(String, String), String> {
+    if let Some((var, value)) = map.split_once('=') {
+        Ok((var.to_owned(), value.to_owned()))
+    } else {
+        Err("contains no '='".to_owned())
+    }
+}
+
+#[derive(Serialize)]
+pub struct EwasmContract {
```
LucasSte, line 882: Why did you move these structs down?

seanyoung: These structs are the serialization format for `--standard-json`. I don't think they belong at the top of the `main.rs` for the executable. Where do you think they should go?

LucasSte: Perhaps they belong in another file.

seanyoung: You are right. It was even worse: the structures were defined in different places. I've created `src/standard_json.rs`.

src/standard_json.rs
```diff
@@ -0,0 +1,47 @@
+//! This module defines the json format for `solang compile --standard-json`.
```
LucasSte, line 1: Please add the SPDX header: `// SPDX-License-Identifier: Apache-2.0`

LucasSte (changes requested):

xermicus (approved):

src/standard_json.rs
seanyoung: Good point, fixed!

src/bin/solang.rs
```diff
@@ -878,19 +858,22 @@ fn imports_arg(matches: &ArgMatches) -> FileResolver {
...
+                eprintln!("error: import path '{}': {}", path, e);
                 std::process::exit(1);
             }
         }
     }
 
     resolver
 }
+
+fn parse_import_map(map: &str) -> Result<(String, String), String> {
```
LucasSte, line 873: This function deserves a comment. I see you use it in `ValueParser::new(parse_import_map)`, but It is not clear what these three returned strings represent. If someone were to re-use your function, they would need to know what it does.

LucasSte (changes requested):
Sorry, I missed this last time.

src/bin/solang.rs
seanyoung: Good point. I've added a comment

LucasSte (approved):

Issue #932, opened 2022-07-20
Make multidimensional arrays work


LucasSte:
Our support for multidimensional arrays is broken and the semantics Solidity uses for them is confusing.

Here are some facts to discuss:

1. I understand `int[4][2] vec` as a matrix of four rows and two columns, but in Solidity it is a two-element array of `int[4]`.
2. Should `vec[1]` index the last or the first dimension? Should it return `int[2]` or `int[4]`?
3. In Solidity, empty square brackets mean that everything that preceded it is the type of each element of a dynamic array. Thus, `int[3][4][] vec` is a dynamic vector of `int[3][4]` elements. How do we treat dynamic dimensions that are not the outer ones?

xermicus:
> I understand int[4][2] vec as a matrix of four rows and two columns

What? I am 101% with you!

> Should vec[1] index the last or the first dimension? Should it return int[2] or int[4]?

So it looks like the order how you access nested arrays is inverted? In that case vec[1] should return int[2].. What does solc produce in this case?

> How do we treat dynamic dimensions that are not the outer ones?

So this is bit mind stretching.. it means int[2][3][][4] would be 4 elements containing a dynamic number of 3 rows and 2 columns?

LucasSte:
This should work:
```solidity
    function middle_dyn() public pure returns (uint32) {
        uint8[3][][2] memory vec;
        vec[0] = new uint8[3][](1);
        vec[0][0] = [uint8(1), 2, 3];
        vec[1] = new uint8[3][](2);
        vec[1][0] = [uint8(4), 5, 6];
        vec[1][1] = [uint8(7), 8, 9];

        return vec[0].length;
    }
```

seanyoung:
I don't think we can change the syntax. 
- We need to be compatible with ethereum solidity
- Introducing a switch for changing the order is even more confusing (now there are two ways of doign this)

Possibly we can introduce new syntax with sane ordering. I don't know what that would look like - any suggestions?

LucasSte:
This case is also broken, but should work after we deploy this feature:

```solidity

contract testing {

    function test(uint32 dim) public pure returns (uint32[3][][4] memory) {
        uint32[3][][4] memory vec;
        vec[0] = new uint32[3][](dim);
        return vec;
    }

}
```

LucasSte:
I just found another inconsistency. Even though indexing in Solidity is reversed, Solang indexes arrays the same way done in traditional languages, such as C/C++ and Java.
[...]

PR #961, opened 2022-08-12
Create abi.borshDecode method for Solana

LucasSte:
This PR implements Borsh Decoding and connects it to an experimental function called `abi.borshDecode` that only works for Solana. Using this function raises a compiler warning to inform the user that the borsh encoding is not completely supported yet.


src/codegen/encoding/buffer_validator.rs
```diff
@@ -0,0 +1,201 @@
...
+/// by merging checks when we can determine the size of what to read beforehand.
+pub(super) struct BufferValidator<'a> {
+    /// Saves the codegen::Expression that contains the buffer length.
+    pub(super) buffer_length: Expression,
+    /// The types we are supposed to decode
+    pub(super) types: &'a [Type],
+    /// The argument whose size has already been accounted for when verifying the buffer
+    pub(super) verified_until: i32,
+    /// The argument we are analysing presently
+    pub(super) current_arg: usize,
```
seanyoung, line 21: `verified_until` and `current_arg` are different types. I think this is so because you want to assign -1 to `verfied_until`. As as result there are a bunch `as i32` casts in the code, which are ugly.

How about making `verified_until` an `Option<usize>` or special casing `fn initialize_validation()` so that it does not need to set `-1`?

src/codegen/encoding/borsh_encoding.rs
```diff
@@ -693,4 +717,584 @@ impl BorshEncoding {
...
+        mut offset: Expression,
+        expr_ty: &Type,
+        struct_ty: &StructType,
+        validator: &mut BufferValidator,
+        ns: &Namespace,
+        vartab: &mut Vartable,
+        cfg: &mut ControlFlowGraph,
+    ) -> (Expression, Expression) {
+        let size = if let Some(no_padding_size) = ns.calculate_struct_non_padded_size(struct_ty) {
+            let padded_size = expr_ty.solana_storage_size(ns);
```
seanyoung, line 1172: Why `solana_storage_size()`? This is not going to be stored in storage, this is going to the heap.

src/codegen/encoding/borsh_encoding.rs
```diff
@@ -693,4 +717,584 @@ impl BorshEncoding {
...
+                -1
+            } else {
+                struct_tys.len() as i32
+            },
+            current_arg: if validator.validation_necessary() {
+                0
+            } else {
+                struct_tys.len()
+            },
+        };
```
seanyoung, line 1237: I do really like how you've solved the problem of minimizing the amount of bounds checks. However, I think this construct here should really be a method on `BufferValidator`. Something like `fn subvalidator()`.

tests/solana_tests/abi_decode.rs
```diff
@@ -0,0 +1,1175 @@
...
+    let input = Input {
+        item_1: [b"abcd".to_owned(), b"efgh".to_owned()],
+        item_2: vec![b"12345".to_owned(), b"67890".to_owned()],
+    };
+    let encoded = input.try_to_vec().unwrap();
+    let _ = vm.function("testByteArrays", &[Token::Bytes(encoded)], &[], None);
+}
+
+#[test]
+#[should_panic]
```
seanyoung, line 862: expected = ".."

tests/solana_tests/abi_decode.rs
```diff
@@ -0,0 +1,1175 @@
...
+    );
+
+    vm.constructor("Testing", &[]);
+    let input = Input1 { a: -789, b: 14234 };
+    let encoded = input.try_to_vec().unwrap();
+    let _ = vm.function("testByteArrays", &[Token::Bytes(encoded)], &[], None);
+}
+
+#[test]
+#[should_panic]
```
seanyoung, line 894: expected = ".."

tests/solana_tests/abi_decode.rs
```diff
@@ -0,0 +1,1175 @@
...
+
+    vm.constructor("Testing", &[]);
+
+    let input = Input { vec: [1, 4, 5, 6] };
+    let encoded = input.try_to_vec().unwrap();
+    let _ = vm.function("wrongNumber", &[Token::Bytes(encoded)], &[], None);
+}
+
+#[test]
+#[should_panic]
```
seanyoung, line 921: expected = ".."

tests/solana_tests/abi_decode.rs
```diff
@@ -0,0 +1,1175 @@
...
+    let input = Input {
+        vec: vec![-90, 89, -2341],
+    };
+
+    let encoded = input.try_to_vec().unwrap();
+    let _ = vm.function("extraElement", &[Token::Bytes(encoded)], &[], None);
+}
+
+#[test]
+#[should_panic]
```
seanyoung, line 951: expected = ".."

tests/solana_tests/abi_decode.rs
```diff
@@ -0,0 +1,1175 @@
...
+
+    vm.constructor("Testing", &[]);
+
+    let input = Input { item: 5 };
+    let encoded = input.try_to_vec().unwrap();
+    let _ = vm.function("invalidType", &[Token::Bytes(encoded)], &[], None);
+}
+
+#[test]
+#[should_panic]
```
seanyoung, line 978: expected = ".."

tests/solana_tests/abi_decode.rs
```diff
@@ -0,0 +1,1175 @@
...
+            c: 1,
+            d: "string".to_string(),
+        },
+    };
+    let encoded = input.try_to_vec().unwrap();
+    let _ = vm.function("test", &[Token::Bytes(encoded)], &[], None);
+}
+
+#[test]
+#[should_panic]
```
seanyoung, line 1126: expected = ".."

seanyoung (changes requested):
I like how you solved it with the buffer validator.

src/codegen/encoding/borsh_encoding.rs
LucasSte: How can I know if there is padding between the struct members? If there is no padding, I simply memcpy the struct from the buffer.

tests/solana_tests/abi_decode.rs
LucasSte: I really wanted to use this Rust feature, but the eBPF VM returns an error code, which makes the following assertion fail. The code panics without any error message, so I cannot use `#[should_panic(expected = "my_error")]`. This is why I suggested that the VM should return the error code, instead of panicing.

https://github.com/hyperledger-labs/solang/blob/b867c8a6c7a1ee89d405993abef458fc59e9b0fb/tests/solana.rs#L1583

seanyoung: That `assert_eq!()` should be re-written. Does:
```
assert_eq!(res.unwrap(), 0));
```
give you the correct error message? (See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c9db9b9df941e8bde46f7d3a9db72aa0)

LucasSte: The error message I got now is
```
assertion failed: `(left == right)`
  left: `Ok(4294967296)`,
 right: `Ok(0)`
```
If I change to your suggestion, it turns to:
```
assertion failed: `(left == right)`
  left: `4294967296`,
 right: `0`
```
Keep in mind that this solution does not account for when the VM returns Err (not Ok()).

One solution I found now was adding this decorator to the test function:
```
#[should_panic(expected = "assertion failed: `(left == right)`
  left: `Ok(4294967296)`,
 right: `Ok(0)`")]
```

src/codegen/encoding/borsh_encoding.rs
```diff
@@ -693,4 +717,584 @@ impl BorshEncoding {
...
+                        Box::new(offset_expr.clone()),
+                        Box::new(offset.clone()),
+                    ),
+                },
+            );
+            (array_var_expr, offset_expr)
+        }
+    }
+
+    /// Decodes a complex array from a borsh encoded buffer
```
xermicus, line 1020: A little more words about what `complex arrays` are and why they receive special treatment would help out anyone not too familiar with this :)

src/codegen/encoding/borsh_encoding.rs
```diff
@@ -693,4 +717,584 @@ impl BorshEncoding {
...
+                                new_ty.clone(),
+                                allocated_array,
+                            ),
+                        },
+                    );
+                } else {
+                    unreachable!("array_var must be a variable");
+                }
+            } else {
+                // TODO: This is wired up for multidimensional dynamic arrays, but they do no work yet
```
xermicus, line 1087: Is there a warning or anything for the user during compilation, or do you plan to fix this soon?

xermicus (approved):

tests/solana_tests/abi_decode.rs
seanyoung: That number makes more sense when written in hex. Maybe it should do something like:
```
match res {
    Ok(0) => (),
    Ok(v) =>panic!("unexpected return {:#x}", v),
    Err(e) => panic!("error: {}", e);
}
```

LucasSte: Good idea

src/codegen/encoding/borsh_encoding.rs
LucasSte: No. Multidimensional arrays are broken in Solang. We have an issue to make it work: #932.

src/codegen/encoding/buffer_validator.rs
LucasSte: > special casing fn initialize_validation() so that it does not need to set -1?

This was my first attempt, however, when the first argument to decode is a dynamic array, the validator must still be aware that it has no been verified, even after calling `initialize_validation`. I'll implement your `Option<usize>` suggestion.

src/codegen/encoding/borsh_encoding.rs
LucasSte: Any thoughts on this? @seanyoung

seanyoung: borsh encoding has nothing to do with solana storage, so it should not use that function. Does size_of() not do the trick?

LucasSte: In PR #919, `size_of` has been superseded by `memory_size_of`. It simply sums the size of each item of a struct. I don't know from this number if LLVM will add any type of padding between them. The check I do here serves to verify if a struct can be memcpy'ed from the buffer to the memory. Doing so is more efficient than copying each element individually.

src/codegen/encoding/borsh_encoding.rs
LucasSte: I'll add a link to the issue on the comment to make it clear.

src/codegen/encoding/borsh_encoding.rs
LucasSte: @seanyoung Can I memcpy every struct?

LucasSte: I created a new function that calculates the padded size correctly.

src/sema/types.rs
```diff
@@ -1056,6 +1056,65 @@ impl Type {
         }
     }
 
+    /// Retrieve the alignment for each type, if it is a struct member.
+    /// The representation outside a struct varies a little, for arrays are always reference types
+    /// outside structs.
```
seanyoung, line 1061: I don't understand what this comment means. Are you talking about dynamic arrays?

src/sema/types.rs
```diff
@@ -1056,6 +1056,65 @@ impl Type {
...
+            | Type::Address(_)
+            | Type::Enum(_) => BigInt::one(),
+
+            // Bytes are custom width type in LLVM, so they fit in the smallest integer type
+            // whose bitwidth is larger than what is needed.
+            Type::Bytes(n) => {
+                let mut custom_width : u8 = 1;
+                while custom_width < *n {
+                    custom_width <<= 1;
+                }
```
seanyoung, line 1076: This looks a bit like `fn next_power_power_of_two()`.

src/sema/types.rs
```diff
@@ -1056,6 +1056,65 @@ impl Type {
...
+                BigInt::from(custom_width)
+            }
+
+            // The same reasoning as above applies for value
+            Type::Value => {
+                let mut custom_width : usize = 1;
+                while custom_width < ns.value_length {
+                    custom_width <<= 1;
+                }
+                BigInt::from(custom_width)
```
seanyoung, line 1086: `ns.value_length.next_power_of_two()`?

src/sema/types.rs
```diff
@@ -1056,6 +1056,65 @@ impl Type {
...
+                if dims.iter().any(|d| *d == ArrayLength::Dynamic) {
+                    BigInt::from(ns.target.ptr_size() / 8)
+                } else {
+                    ty.struct_elem_alignment(ns)
+                }
+            }
+
+            Type::Struct(def) => {
+                def.definition(ns).fields[0].ty.struct_elem_alignment(ns)
+            }
```
seanyoung, line 1100: That is wrong. The alignment of a struct is a the maximum of its members

seanyoung (changes requested):

seanyoung (approved):

PR #963, opened 2022-08-14
add shell completions for cli app

xermicus:
This PR adds a `shell-complete` sub-command, allowing users to autocomplete our CLI with the TAB key (this pretty much defuses the issue of typing full length sub-commands out all the time). As a bonus It also shows remaining options when there is nothing to complete yet.


seanyoung (approved):
I did not know this existed. I just tried it out, very nice!

LucasSte (approved):

PR #968, opened 2022-08-16
Introduce new syntax for overriding selector

seanyoung:
This introduces syntax like so:
```
function get_foo() selector=hex"01020304" public return (int) {
    return 102;
}
```
Note this functionality will be used in a following commit to define interfaces derived from anchor idl files. Tests or Solana will be in that PR, no changes as the functionality will change in that PR.


docs/language/functions.rst
```diff
@@ -279,6 +279,39 @@ constructor.
...
+
+Only ``public`` and ``external`` functions have a selector, and can have their
+selector overriden. On Substrate, constructors have selectors too, so they have
+can also have their selector overriden. If a function overrides a function in a
+base contract, then the selector of the function in the base contract must match.
+
+.. warning::
+    On ewasm and Solana, changing the selector may result in a mismatch between
+    the contract metadata and the actual contract code. This is because the ewasm
+    and Solana metadata does not explicitly store the selector, but derives it
```
LucasSte, line 310:
```suggestion
    On ewasm and Solana, changing the selector may result in a mismatch between
    the contract metadata and the actual contract code, because the former does not explicitly store the selector, but derives it
```

LucasSte: A suggestion to make it less wordy.

LucasSte (changes requested):

src/codegen/cfg.rs
```diff
@@ -366,7 +366,7 @@ pub struct ControlFlowGraph {
     pub nonpayable: bool,
     pub public: bool,
     pub ty: pt::FunctionTy,
-    pub selector: u32,
+    pub selector: Vec<u8>,
```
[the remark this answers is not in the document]
seanyoung: With Anchor, they're 8 bytes. However, some solana programs use other lengths like 1 byte. The next PR is dealing with this problem, it was a getting a bit large so I started by sending this out for review first. Does it matter that this is changing here?

tests/contract.rs
```diff
@@ -112,9 +112,9 @@ fn parse_file(path: PathBuf, target: Target) -> io::Result<()> {
     let generated_dot = ns.dotgraphviz();
 
     // uncomment the next three lines to regenerate the test data
-    // use std::io::Write;
-    // let mut file = File::create(&path)?;
-    // file.write_all(generated_dot.as_bytes())?;
+    use std::io::Write;
+    let mut file = File::create(&path)?;
+    file.write_all(generated_dot.as_bytes())?;
```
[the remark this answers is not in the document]
seanyoung: oops

docs/language/functions.rst
```diff
@@ -279,6 +279,39 @@ constructor.
...
+    contract foo {
+        function get_foo() selector=hex"01020304" public return (int) {
+            return 102;
+        }
+    }
+
+Only ``public`` and ``external`` functions have a selector, and can have their
+selector overriden. On Substrate, constructors have selectors too, so they have
+can also have their selector overriden. If a function overrides a function in a
+base contract, then the selector of the function in the base contract must match.
```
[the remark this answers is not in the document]
seanyoung: Good point, thanks

docs/language/functions.rst
```diff
@@ -279,6 +279,38 @@ constructor.
...
+    }
+
+Only ``public`` and ``external`` functions have a selector, and can have their
+selector overriden. On Substrate, constructors have selectors too, so they
+can also have their selector overriden. If a function overrides another one in a
+base contract, then the selector of both must match.
+
+.. warning::
+    On ewasm and Solana, changing the selector may result in a mismatch between
+    the contract metadata and the actual contract code, because the former does not explicitly store the selector, but derives it
```
xermicus, line 309: Is the issue here when you derive the metadata out of a contract already deployed, resulting in wrong metadata? If that is possible and common for Solana, then overwriting sounds super scary to me...

docs/language/functions.rst
```diff
@@ -279,6 +279,38 @@ constructor.
...
+the function selector is generated from a deterministic hash value of the function
+name and its arguments.
+
+The selector value can be overriden with the ``selector=hex"deadbea1"`` syntax,
+for example:
+
+.. code-block:: solidity
+
+    contract foo {
+        function get_foo() selector=hex"01020304" public return (int) {
```
xermicus, line 297: Does the position of the selector overwrite matter? I'd rather move the selector overwrite at the beginning or at the end of the function declaration.

xermicus (approved):
Is there a check somewhere for duplicate selectors? Or what happens if two different functions end up with the same selector?

src/codegen/cfg.rs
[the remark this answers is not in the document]
LucasSte: You can update this in the next PR. Bear in mind that there needs to be a standard size for encoding and decoding.

docs/language/functions.rst
seanyoung: The selector is part of the function attributes, which can occur in any order after the parameters and before the `returns (..)` (if any).

docs/language/functions.rst
seanyoung: This currently very messy on Solana. Currently there are many contracts without metadata which have hand-coded selectors (the idea of metadata is fairly new for Solana!). This is why we need the selector override in the first place. They need hand-written solidity `interface { ... }` definitions with the selector override.

We're adding selectors to the solana metadata to solve all these problems, and this scary warning will go away for Solana too.

seanyoung:
> Is there a check somewhere for duplicate selectors? Or what happens if two different functions end up with the same selector?

I have this functionality ready, I can either add it to this PR or another PR.

docs/language/functions.rst
```diff
@@ -279,6 +279,38 @@ constructor.
     for storage rent and there is a minimum deposit needed for the contract. As a result,
     constructors always receive value on Parity Substrate.
 
+Overriding function selector
+____________________________
+
+When a function is called, the function selector and the arguments to the function
```
LucasSte, line 285:
```suggestion
When a function is called, its selector and its arguments
```
Function was repeated three times here.

docs/language/functions.rst
```diff
@@ -279,6 +279,38 @@ constructor.
...
+selector overriden. On Substrate, constructors have selectors too, so they
+can also have their selector overriden. If a function overrides another one in a
+base contract, then the selector of both must match.
+
+.. warning::
+    On ewasm and Solana, changing the selector may result in a mismatch between
+    the contract metadata and the actual contract code, because the metadata does
+    not explicitly store the selector.
+
+    Only use this feature if you know what you are doing.
```
LucasSte, line 312: Instead of saying `Only use this feature if you know what you are doing`, I think it is more polite to say `Use this feature carefully, as it may either break a contract or cause undefined behavior.` What do you think?

LucasSte (changes requested):

docs/language/functions.rst
xermicus: Ah I see :grin: I mean selector overwrites could maybe used for advanced things like hooks. And it is possible in `ink!` anyways. So I think in general it is still a useful feature to have.

xermicus:
> > Is there a check somewhere for duplicate selectors? Or what happens if two different functions end up with the same selector?
> 
> I have this functionality ready, I can either add it to this PR or another PR.

Sounds good to me, can be added in the next PR as well

docs/language/functions.rst
seanyoung: I don't think "its selector" is particularly clear. It's a called a function selector.

seanyoung: Also now you repeated its

seanyoung: I've rephrased this, please let me know what you think.

docs/language/functions.rst
seanyoung: I've added this, thanks

LucasSte (approved):
Please, don't forget to change the buiiltin struct ExternalFunction in your next PR, as the selector has variable width.

PR #980, opened 2022-08-19
Make lexer public

jpopesculian:
Exposing the lexer API would be useful for things like syntax highlighting. Plus its implemented clean as is to be exposed directly :hugs:

### DIFF
--- a/solang-parser/src/lib.rs
+++ b/solang-parser/src/lib.rs
@@ -7,7 +7,7 @@ use lalrpop_util::ParseError;
 
 pub mod diagnostics;
 pub mod doccomment;
-mod lexer;
+pub mod lexer;
 pub mod pt;
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


seanyoung:
> Exposing the lexer API would be useful for things like syntax highlighting. Plus its implemented clean as is to be exposed directly 

Just out of interest, what do you have in mind? I mean where would you do syntax highlighting for example?

jpopesculian:
I'm building a solidity REPL, and would love to have syntax highlighting integrated as you type and I think the lexer would be perfect for distinguishing keywords, builtin types, literals etc. also helpful for dealing with unclosed literals, braces, etc

seanyoung (approved):

xermicus (approved):

LucasSte (approved):

PR #981, opened 2022-08-19
Remove ewasm target and rename usable parts for evm target

seanyoung:
The Ethereum Foundation has stopped funding for ewasm work, so it is
unlikely to happen in the current form. In addition, burrow has been end
of life'd by Hyperledger. So, the ewasm target is no longer useful.

We would love to have an evm target at some point in the future. Many
parts of the ewasm target is re-usable for evm, so we rename ewasm to
evm and remove the parts which are irrelevant for evm.

Note that the new evm target is good enough for the language server, so
the language server can be started in evm mode:

```
solang language-server --target evm
```


docs/language/builtins.rst
```diff
@@ -261,8 +260,7 @@ ABI encoding and decoding
 _________________________
 
 The ABI encoding depends on the target being compiled for. Substrate uses the
```
LucasSte, line 262: I should mention Borsh encoding here.

tests/codegen_testcases/yul/ewasm_builtin.sol
```diff
@@ -1,62 +0,0 @@
-// RUN: --target ewasm --emit cfg -Onone --no-cse
-
-contract Testing {
-
-// BEGIN-CHECK: Testing::Testing::function::builtins__uint256
-    function builtins(uint256 arg1) public view  {
```
LucasSte, line 6: If we are still maintaining EVM for the language server, I suggest that this file remain as a test for the language server, verifying the correctness of the AST for all these builtins.

In addition, some of the builtins listed here are available for other targets (like `gas()`, `address()`, `caller()`, and others). They still need to be tested for code generation, but for another target.

tests/contract.rs
```diff
@@ -79,29 +79,35 @@ fn parse_file(path: PathBuf, target: Target) -> io::Result<()> {
...
+                            &filename,
+                            Default::default(),
+                            false,
+                            false,
+                        );
+                    }
                 }
             }
+            Target::EVM => {
+                // not implemented yet
```
LucasSte, line 109: Shouldn't we use `unimplemented!()`?

tests/ewasm_tests/yul.rs
```diff
@@ -1,117 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
-
-use crate::build_solidity;
-use byte_slice_cast::AsByteSlice;
-use ethabi::{Token, Uint};
-
-#[test]
-fn eth_builtins() {
-    let mut runtime = build_solidity(
```
LucasSte, line 9: The builtins tested here are not tested elsewhere. There need to be tests for them on another target, depending on their compatibility.

LucasSte (changes requested):

docs/language/builtins.rst
seanyoung: Well, we haven't moved to borsh encoding yet. We should mention it once we are done.

tests/contract.rs
seanyoung: We don't want the test to fail.

tests/ewasm_tests/yul.rs
seanyoung: We can't do runtime tests for the builtins anymore for ewasm/evm because we don't have codegen. So, I've added the tests to `tests/evm.rs`

tests/codegen_testcases/yul/ewasm_builtin.sol
seanyoung: So the reason I removed it is because the option `--target ewasm` was gone, and I had not added a new option `--target evm`. 

Now I've changed it so that `solang --target evm` is supported, but it is not documented in `solang compile --help` nor in the documentation,

xermicus (approved):

tests/ewasm_tests/yul.rs
LucasSte: This is the only place in Solang where I test `address`, `selfbalance`, `caller`, `callvalue`, and `extcodesize`. By removing these runtime tests, there will be no test for them at all.

`address`, `caller`, `callvalue` and `selfbalance` are available for both Solana and Substrate. Instead of deleting the test, please, move it to a Solana or a Substrate runtime test. Only `extcodesize` is not available for those targets.

seanyoung: I had missed that, thanks for pointing that out.

seanyoung:
sorry @xermicus I did not mean to for a review again, you've already approved it

tests/codegen_testcases/yul/evm_builtin.sol
```diff
@@ -1,4 +1,4 @@
-// RUN: --target ewasm --emit cfg -Onone --no-cse
+// RUN: --target evm --emit cfg -Onone --no-cse
 
 contract Testing {
```
LucasSte, line 3: We don't have code generation for EVM, so how is this file tested?

tests/undefined_variable_detection.rs
```diff
@@ -10,7 +10,7 @@ use std::ffi::OsStr;
 fn parse_and_codegen(src: &'static str) -> Namespace {
     let mut cache = FileResolver::new();
     cache.set_file_contents("test.sol", src.to_string());
-    let mut ns = parse_and_resolve(OsStr::new("test.sol"), &mut cache, Target::Ewasm);
+    let mut ns = parse_and_resolve(OsStr::new("test.sol"), &mut cache, Target::EVM);
```
LucasSte, line 13: I think undefined variable detection depends on codegen. Will this test work even if we use `Target::EVM`?

tests/unused_variable_detection.rs
```diff
@@ -9,15 +9,15 @@ fn parse(src: &'static str) -> ast::Namespace {
     let mut cache = FileResolver::new();
     cache.set_file_contents("test.sol", src.to_string());
 
-    parse_and_resolve(OsStr::new("test.sol"), &mut cache, Target::Ewasm)
+    parse_and_resolve(OsStr::new("test.sol"), &mut cache, Target::EVM)
```
LucasSte, line 12: Same as above.

LucasSte (commented):
I found no problem with the code. I just have some questions before granting approval.

tests/codegen_testcases/yul/evm_builtin.sol
seanyoung: We have codegen for evm (as in `src/codegen/..`), just emit.

tests/undefined_variable_detection.rs
seanyoung: Same as above. We're keeping the code `src/codegen/...`.

LucasSte (approved):

PR #985, opened 2022-08-23
Generate solidity interface from Anchor IDL files so that anchor contracts can be called

seanyoung:
This PR contains a bunch of commits to make calling Anchor programs possible.
 - Fix for borsh encoding
 - Check for duplicate selectors
 - New subcommand `solang idl` which generates Solidity interface file. I imagine that the Substrate implementation will be very similar (read metadata, print Solidity) and will re-use this subcommand
 - Add Anchor and Solana to CI container for testing
 - Test anchor program and test for calling it from Solang
 - Documentation


docs/running.rst
```diff
@@ -151,6 +151,23 @@ Options:
 \\-\\-help, -h
   This displays a short description of all the options
 
+.. _metadata_command:
+
+Using the metadata command
+__________________________
+
+This command is used to convert Anchor IDL into Solidity import files, for calling
```
LucasSte, line 159:
```suggestion
This command converts Anchor IDL into Solidity import files in order to call
```
I think this sounds better.

docs/running.rst
```diff
@@ -151,6 +151,23 @@ Options:
...
+Anchor Programs.
+
+  solang metadata --target solana [--output DIR] [IDLFILE]...
+
+For each idl file provided, on Solidity file is written. See :ref:`call_anchor`
```
LucasSte, line 164:
```suggestion
For each idl file provided, a Solidity file is written. See :ref:`call_anchor`
```
Typo

docs/running.rst
```diff
@@ -151,6 +151,23 @@ Options:
...
+for an example of how to use this.
+
+.. note::
+
+  This is only support for Solana at this time.
```
LucasSte, line 169:
```suggestion
  There is only support for Solana at this time.
```
A typo maybe?

docs/targets/solana.rst
```diff
@@ -73,6 +73,36 @@ package has `documentation <https://solana-labs.github.io/solana-solidity.js/>`_
 are `some examples <https://solana-labs.github.io/solana-solidity.js/>`_. There is also
 `solang's integration tests <https://github.com/hyperledger/solang/tree/main/integration/solana>`_.
 
+.. _call_anchor:
+
+Calling Anchor Programs from Solidity
+_____________________________________
+
+It is possible to call `Anchor Programs <https://github.com/coral-xyz/anchor>`_
+from Solidity. You first have to generate interface file from the IDL file using
```
LucasSte, line 82:
```suggestion
from Solidity. You first have to generate an interface file from the IDL file using
```

LucasSte: Missing article

docs/targets/solana.rst
```diff
@@ -73,6 +73,36 @@ package has `documentation <https://solana-labs.github.io/solana-solidity.js/>`_
...
+the :ref:`metadata_command`. Then, import the Solidity file in your Solidity using the
+``import "...";`` syntax. Say you have an anchor program called `remoge` with a
+function `nerp`, you can call it like so:
```
LucasSte, line 85: How did you come up with `remoge` and `nerp`? 😂

integration/anchor/programs/anchor/src/lib.rs
```diff
@@ -0,0 +1,139 @@
+use anchor_lang::prelude::*;
+
+declare_id!("z7FbDfQDfucxJz5o8jrGLgvSbdoeSqX5VrxBb5TVjHq");
+
+/// Anchor example for testing with Solang.
+/// This doc comment exist for testing metadata doc comments
+#[program]
+pub mod anchor {
```
LucasSte, line 8: Should we test events too?

integration/anchor/tests/anchor.ts
```diff
@@ -0,0 +1,63 @@
...
+
+describe("anchor", () => {
+  // Configure the client to use the local cluster.
+  const provider = anchor.AnchorProvider.env();
+
+  anchor.setProvider(provider);
+
+  const program = anchor.workspace.Anchor as Program<Anchor>;
+
+  it("Let's test", async () => {
```
LucasSte, line 14: The name for this test should mention what you are testing.

integration/solana/call_anchor.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+import { loadContract } from './setup';
+import { publicKeyToHex } from '@solana/solidity';
+import { SystemProgram, PublicKey, Keypair } from '@solana/web3.js';
+
+
+describe('Deploy solang contract and test', function () {
+    this.timeout(500000);
+
+    it('call_anchor', async function () {
+        // This program instantiates an anchor program, and calls various functions on it and checks the return values
```
LucasSte, line 11:
```suggestion
        // This program instantiates an anchor program, calls various functions on it and checks the return values
```
Don't repeat `and`

solang-parser/src/lexer.rs
```diff
@@ -384,7 +384,7 @@ impl CodeLocation for LexicalError {
     }
 }
 
-static KEYWORDS: phf::Map<&'static str, Token> = phf_map! {
+pub static KEYWORDS: phf::Map<&'static str, Token> = phf_map! {
```
LucasSte, line 387: Why is this public?

src/bin/metadata/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+            .for_each(|event| ty_names.push((event.name.to_string(), event.name.to_string())));
+    }
+
+    rename_keywords(&mut ty_names);
+
+    for ty_def in &idl.types {
+        if let IdlTypeDefinitionTy::Enum { variants } = &ty_def.ty {
+            if variants.iter().any(|variant| variant.fields.is_some()) {
+                eprintln!(
+                    "enum {} has variants with fields, not supported in Solidity\n",
```
LucasSte, line 102: While working on the NFT example, I have been thinking about unsupported types. Have you thought of transforming a `Option<Foo>` into `bool has_foo;` and `Foo foo;`?

src/bin/metadata/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+                write!(f, "\t")?;
+            }
+            writeln!(f, "/// {}", doc)?;
+        }
+    }
+
+    Ok(())
+}
+
+fn sig(namespace: &'static str, name: &str) -> Vec<u8> {
```
LucasSte, line 350: This function can have a better name.

src/bin/metadata/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+            }
+        )),
+        IdlType::Defined(ty) => {
+            if let Some(e) = ty_names.iter().find(|rename| rename.0 == *ty) {
+                Ok(e.1.to_owned())
+            } else {
+                Ok(ty.into())
+            }
+        }
+        IdlType::Vec(ty) => match idltype_to_solidity(ty, ty_names) {
```
LucasSte, line 394: Are multidimensional arrays supported?

src/bin/metadata/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+    if let Some(JsonValue::Object(metadata)) = &idl.metadata {
+        if let Some(JsonValue::String(address)) = metadata.get("address") {
+            return Some(address);
+        }
+    }
+
+    None
+}
+
+/// There are many keywords in Solidity which are not keywords in rust, so they may
```
LucasSte, line 415:
```suggestion
/// There are many keywords in Solidity which are not keywords in Rust, so they may
```

src/bin/solang.rs
```diff
@@ -800,7 +856,7 @@ fn target_arg(matches: &ArgMatches) -> Target {
             "error: value length cannot be modified for target '{}'",
             target
         );
-        std::process::exit(1);
```
LucasSte, line 803: I believe it is clearer to leave the `std::process` here. This way we know exactly what we are exiting from.

src/bin/solang.rs
```diff
@@ -293,6 +337,18 @@ fn main() {
...
+        Some(("metadata", matches)) => {
+            let target = target_arg(matches);
+
+            // For Ethereum, we can generate an interface from an abi file
+            // For Substrate, we can generate an interface from a contract metadata file
+            if target == Target::Solana {
+                metadata::idl(matches)
+            } else {
+                eprintln!("error: metadata for target {} not supported", target);
+                exit(2);
```
LucasSte, line 349: Shouldn't this be `exit(1)`?

src/codegen/encoding/mod.rs
```diff
@@ -562,14 +562,9 @@ fn finish_array_loop(for_loop: &ForLoop, vartab: &mut Vartable, cfg: &mut Contro
...
-        // We should not dereference a struct.
-        return Expression::StructMember(
-            Loc::Codegen,
-            Type::Ref(Box::new(ty)),
-            Box::new(expr),
-            field,
-        );
+    if ty.is_fixed_reference_type() {
+        // We should not dereference a struct or fixed array
+        return Expression::StructMember(Loc::Codegen, ty, Box::new(expr), field);
```
LucasSte, line 567: I missed that! 😅

src/codegen/expression.rs
```diff
@@ -2324,57 +2325,82 @@ pub fn emit_function_call(
...
+                tys.insert(0, Type::Bytes(selector.len() as u8));
 
                 let (payload, address) = if ns.target == Target::Solana {
-                    tys.insert(0, Type::Address(false));
-                    tys.insert(1, Type::Address(false));
-                    tys.insert(2, Type::Uint(64));
-                    tys.insert(3, Type::Bytes(4));
-                    tys.insert(4, Type::Bytes(1));
+                    if dest_func.selector.is_some() {
+                        // use borsh encoding; is this a temporary solution, where we use borsh encoding
```
LucasSte, line 2344:
```suggestion
                        // TODO: Use borsh encoding; This is a temporary solution, where we use borsh encoding
```

src/codegen/expression.rs
```diff
@@ -2324,57 +2325,82 @@ pub fn emit_function_call(
...
+                        // only for functions with function selectors for now. This is useful so we can
+                        // call rust contracts by specifying a function selector override.
```
LucasSte, line 2346:
```suggestion
                        // Call rust contracts by specifying a function selector override.
```

src/codegen/expression.rs
LucasSte: By writing `TODO`, you make it clear that this is supposed to change soon.

src/codegen/expression.rs
```diff
@@ -2397,30 +2423,52 @@ pub fn emit_function_call(
...
+                                name: ret.name_as_str().to_owned(),
+                            };
+                            let temp_pos = vartab.temp(&id, &ret.ty);
+                            res.push(temp_pos);
+                            returns.push(Expression::Variable(id.loc, ret.ty.clone(), temp_pos));
+                        }
+
+                        cfg.add(
+                            vartab,
+                            Instr::AbiDecode {
```
LucasSte, line 2463: Why aren't we using Borsh here?

LucasSte (changes requested):

solang-parser/src/lexer.rs
LucasSte: I see you used this for finding keyword clashes, but I am inclined to think that this should not be a public symbol. Maybe `pub(crate)` would suffice.

docs/targets/solana.rst
seanyoung: I used a random word generator. Suggestions welcome!

solang-parser/src/lexer.rs
seanyoung: This is defined in the `solang-parser` crate, so making it `pub(crate)` is not going to work.

src/bin/metadata/mod.rs
seanyoung: That might work for `Option<Foo>` in account storage, but when serialized in borsh the encoding is shorter for `None` (just a single 0 byte).

integration/anchor/programs/anchor/src/lib.rs
seanyoung: This PR is for calling Anchor contracts. We're not generating IDL files yet for Solidity. Testing emitting events from Solidity is not something we can test; testing emiting events from Anchor -- I am not sure that is worth testing at this point.

src/bin/solang.rs
seanyoung: `exit(2)` means bad usage (i.e. bad command line) by convention.

docs/targets/solana.rst
LucasSte: How about renaming `remoge` to `solidity_from_idl` and `nerp` to `idl_defined_function`?
I tend to name my examples with self-explanatory names so that they become even easier to understand. What do you think?

solang-parser/src/lexer.rs
LucasSte: You're parsing the IDL file too. Shouldn't this be part of `solang-parser?

docs/targets/solana.rst
seanyoung: fixed

solang-parser/src/lexer.rs
seanyoung: I don't agree with that. IDL is json, so there isn't any parsing involved - at least no custom code, the definition is entirely imported from the `anchor-syn` crate. `solang-parser` just contains a Solidity parser, that is totally unrelated to generating Solidity from IDL. How would this benefit Foundry for example?

I have created a new function `is_keyword()` in `solang-parser`, maybe this is a nicer interface.

src/bin/metadata/mod.rs
seanyoung: I've called it discriminator and added some comments

src/bin/metadata/mod.rs
seanyoung: I've added a test

src/bin/solang.rs
seanyoung: At the top of the file we have `use std::process::exit;` so it should be clear what is meant by `exit`. TBH I am not following what the possible confusion could be. A different exit function? Or do you think something as important as a process exit deserves to be mentioned with its full namespace?

We use things from `std` all the time without the namespace, like `format!()` and `println!()`. I don't quite see why `exit()` is any different.

src/codegen/expression.rs
seanyoung: This is the case when we're calling solidity functions, so we're still using eth abi encoding here for now.

integration/anchor/programs/anchor/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+
+        Ok(res)
+    }
+}
+
+/// Planets
+/// in our
+/// Solar System
```
xermicus, line 90:
```suggestion
/// Planets in our solar system
```

integration/anchor/programs/anchor/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+    Jupiter,
+    Saturn,
+    Uranus,
+    Neptune,
+    /// Solidity keyword!
+    anonymous,
+}
+
+/// Can a sector have multiple
+/// solar systems?
```
xermicus, line 108:
```suggestion
/// Can a sector have multiple solar systems?
```

xermicus (changes requested):

src/bin/solang.rs
LucasSte: I was just afraid that someone might create an `exit` function in our codebase. It was just to avoid name collision. If you don't see this problem, I have no problem with removing the `std::process` prefix.

docs/targets/solana.rst
```diff
@@ -73,6 +73,36 @@ package has `documentation <https://solana-labs.github.io/solana-solidity.js/>`_
...
+
+It is possible to call `Anchor Programs <https://github.com/coral-xyz/anchor>`_
+from Solidity. You first have to generate an interface file from the IDL file using
+the :ref:`metadata_command`. Then, import the Solidity file in your Solidity using the
+``import "...";`` syntax. Say you have an anchor program called `solidity_from_idl` with a
+function `idl_defined_function`, you can call it like so:
+
+.. code-block:: solidity
+
+    import "remoge.sol";
```
LucasSte, line 89: I know I asked you to change the names. Please, explain what this file is or rename it to `solidity_from_idl.sol`.

integration/anchor/programs/anchor/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+#[derive(AnchorSerialize, AnchorDeserialize)]
+pub struct Sector {
+    /// A binary system can have multiple suns
+    suns: u64,
+    /// Which planets can support life?
+    mclass: Vec<Planet>,
+    /// Just
+    /// so
+    /// Random field with solidity keyword
```
LucasSte, line 117: Just so what?

integration/solana/call_anchor.spec.ts
```diff
@@ -0,0 +1,23 @@
+import expect from 'expect';
+import { loadContract } from './setup';
+import { publicKeyToHex } from '@solana/solidity';
+import { SystemProgram, PublicKey, Keypair } from '@solana/web3.js';
+
+
+describe('Deploy solang contract and test', function () {
```
LucasSte, line 7: We have many tests named as `Deploy slang contract and test`. We should avoid using this name all over again, because it is not descriptive of what you are testing.

src/bin/metadata/mod.rs
```diff
@@ -0,0 +1,438 @@
...
+                write!(f, "\t")?;
+            }
+            writeln!(f, "/// {}", doc)?;
+        }
+    }
+
+    Ok(())
+}
+
+/// Generate discriminator based on the name of the function. This is the 8 byte
```
LucasSte, line 352: Is this the selector of a function?

LucasSte: I create the discriminator for emitting events too. Shouldn't we have a unique implementation for this function throughout Solang? I believe we are going to need it again for other stuff.

LucasSte (changes requested):

integration/anchor/programs/anchor/src/lib.rs
seanyoung: This is to test that multiple comments are copied correctly.

integration/anchor/programs/anchor/src/lib.rs
seanyoung: This is to test that multiple comments are copied correctly.

src/bin/solang.rs
```diff
@@ -272,6 +274,48 @@ fn main() {
...
+                            .value_parser(ValueParser::os_string())
+                            .multiple_values(true),
+                    )
+                    .arg(
+                        Arg::new("TARGET")
+                            .help("Target to build for")
+                            .long("target")
+                            .takes_value(true)
+                            .required(true)
+                            .value_parser(["solana", "substrate", "ewasm"]),
```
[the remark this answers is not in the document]
seanyoung: You're right, the naming not clear. After trying a few things I've renamed it `solang idl`, and I would imagine a future ink! command would be called `solang ink`

src/bin/solang.rs
```diff
@@ -272,6 +274,48 @@ fn main() {
                             .action(ArgAction::Append),
                     ),
             )
+            .subcommand(
+                Command::new("metadata")
```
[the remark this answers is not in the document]
seanyoung: I've renamed it `solang idl`

Cargo.toml
```diff
@@ -51,6 +51,8 @@ codespan-reporting = "0.11"
 phf = "0.10.1"
 rust-lapper = "1.0"
 bitflags = "1.3"
+anchor-syn = { version = "0.25", features = ["idl"] }
```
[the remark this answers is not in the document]
seanyoung: We could do. I think we should wait until there is a real demand for this, not just hypothetical.

integration/anchor/programs/anchor/src/lib.rs
seanyoung: Well a multiline comment. I was not supposed to make sense.

src/bin/metadata/mod.rs
seanyoung: Anchor/IDL calls this a selector.

Once there is a need for this function from multiple places, we can put it in a shared place. I don't see a need to do this *before* we get there. We don't know what the code will look like *until* we actually need it in multiple places.

Cargo.toml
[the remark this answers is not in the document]
xermicus: Well it will help to reduce compile time. But we can introduce them later and also think a bit more about the architecture of the flags we want to have.

src/bin/solang.rs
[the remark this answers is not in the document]
xermicus: Yes that would also be good :ok_hand:

integration/anchor/programs/anchor/src/lib.rs
seanyoung: Comments are copied into the Solidity file verbatim

xermicus (approved):

LucasSte (approved):

hydrogenbond007:
this looks dope!

PR #987, opened 2022-08-25
Create 'writeString' and 'writeBytes'

LucasSte:
This PR creates `writeStringData` and `writeBytesData` functions that write only the contents of strings and bytes arrays into a buffer. They are useful for using bincode when creating a library for System Instructions on Solana.

For arrays (strings and bytes included) bincode encodes lengths in `u64`, so borsh encoding cannot be used for this case.

### DIFF
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -1601,6 +1601,66 @@ fn expr_builtin(
 
             Expression::Undefined(tys[0].clone())
         }
+        ast::Builtin::WriteBytes | ast::Builtin::WriteString => {
+            let buffer = expression(&args[0], cfg, contract_no, func, ns, vartab, opt);
+            let data = expression(&args[1], cfg, contract_no, func, ns, vartab, opt);
+            let offset = expression(&args[2], cfg, contract_no, func, ns, vartab, opt);
+
+            let size = Expression::Builtin(
+                *loc,
+                vec![Type::Uint(32)],
+                Builtin::ArrayLength,
+                vec![data.clone()],
+            );
+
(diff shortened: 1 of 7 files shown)


src/codegen/expression.rs
```diff
@@ -1601,6 +1601,33 @@ fn expr_builtin(
...
+
+            cfg.add(
+                vartab,
+                Instr::MemCopy {
+                    source: data,
+                    destination: advanced_ptr,
+                    bytes: size,
+                },
+            );
+            Expression::Undefined(tys[0].clone())
```
seanyoung, line 1629: Missing bounds check - is this writing to be beyond the end of the buffer. This also needs some tests.

seanyoung (changes requested):
Should this be `writeString()` rather than `writeStringData()`?

src/codegen/expression.rs
LucasSte: Tests are available in `tests/solana_tests/builtin.rs`

LucasSte:
> Should this be `writeString()` rather than `writeStringData()`?

I wanted to be clear that the function only write the characters of a strings and ignores the length. My inspiration was C++:
```C++
string str = "tea";
char * char_only = str.data();
```

src/codegen/expression.rs
seanyoung: Where are the test for bounds checks?

seanyoung (changes requested):
Please add documentation for the new builtin

seanyoung (approved):
Looks great!

xermicus (approved):

PR #989, opened 2022-08-26
Feature ink metadata for substrate

extraymond:
attempt to generate ink compatible metadata using ink_metadata::InkProject
currently relies on unmerged branched of scale-info, ink_metadata.


seanyoung (changes requested):
Looks great!

Please rebase against latest `main` and add `Signed-off-by: real name <email@address>` to each commit (or even better, squash into one commit)

src/abi/substrate.rs
```diff
@@ -1,437 +1,514 @@
...
-    variant: Enum,
-}
+    match ty {
+        // TODO: should reflect address-length for different substrate runtime
+        ast::Type::Address(_) | ast::Type::Contract(_) => {
+            // substituted to [u8 ;32]
+            let address_ty = resolve_ast(
+                &ast::Type::Array(
+                    Box::new(ast::Type::Uint(8)),
+                    vec![ArrayLength::Fixed(BigInt::from_u8(32).unwrap())],
```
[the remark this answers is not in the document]
extraymond: Got it! I couldn't find the place to replace balance types from the ABI, that might need to be configured prior to abi generation.

extraymond:
> Looks great!
> 
> Please rebase against latest `main` and add `Signed-off-by: real name <email@address>` to each commit (or even better, squash into one commit)

Hope it's ok that I force merged --squashed all the changes into a new commit, I couldn't find a way to sign past commits without having to fix all the conflicts.

extraymond:
@seanyoung Adjusted the address_length and variant sort as requested.

seanyoung (changes requested):
There are a bunch of tests which fail with:
```
thread 'substrate_tests::inheritance::inherit_variables' panicked at 'must have at least one message',
```
This is because there is a test contract with only a contructor, no functions. If the ink crate really enforces this, then two things should happen:
- Add dummy function to the tests so the tests pass
- Solang should enforce that on substrate, at least one public function exists and give a nice error message. If someone compiles a contract like so `contract d {}` then we don't want an assertion failure in an ink crate, we want a nice message say "you need at least one function on substrate".

xermicus:
So as discussed in chat we stick with the requirement to have at least one public message.
* Contracts without means to interact with have very niche to no use case
* We do as `ink!` does
* Allows us the reuse the `ink!` crate.

 I'll add a better check and error message in sema.

@extraymond is this OK for you as well? Or do you have a specific use case in mind for contracts without messages?

extraymond:
> So as discussed in chat we stick with the requirement to have at least one public message (substrate only).
> 
>     * Contracts without means to interact with have very niche to no use case
> 
>     * We do as `ink!` does
> 
>     * Allows us the reuse the `ink!` crate.
> 
> 
> I'll add a better check and error message in sema.
> 
> @extraymond is this OK for you as well? Or do you have a specific use case in mind for contracts without messages?

Yeah let's try this first.

xermicus:
I implemented a sema check to catch contracts without public messages. How-ever, it would actually be fine for base contracts. Declaring contracts without public messages as non-concrete some-how could solve the issue elegantly @seanyoung ? The goal  is actually to just prevent metadata generation for those contracts.

src/abi/substrate.rs
```diff
@@ -1,437 +1,473 @@
...
-}
+    match ty {
+        //  should reflect address_length for different substrate runtime
+        ast::Type::Address(_) | ast::Type::Contract(_) => {
+            // substituted to [u8 ;address_length]
+            let address_ty = resolve_ast(
+                &ast::Type::Array(
+                    Box::new(ast::Type::Uint(8)),
+                    vec![ArrayLength::Fixed(
+                        BigInt::from_u8(ns.address_length as u8).unwrap(),
```
seanyoung, line 110: I *think* this can be written without a cast: `BigInt::from(ns.address_length).unwrap()`

src/sema/mod.rs
```diff
@@ -50,6 +50,14 @@ pub fn sema(file: &ResolvedFile, resolver: &mut FileResolver, ns: &mut ast::Name
...
         check_unused_namespace_variables(ns);
         check_unused_events(ns);
+
+        // Substrate: Every contrat must have at least one public message
+        if ns.target.is_substrate() && !ns.contracts.iter().any(|c| c.has_public_functions(ns)) {
+            ns.diagnostics.push(Diagnostic::warning(
+                pt::Loc::CommandLine,
+                format!("{} yields no public messages", file.full_path.display()),
+            ))
+        }
```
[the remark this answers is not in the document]
xermicus: I see your point and I already tried to implement it more or less like that yesterday. The problem I faced is how would you detect that a contract, while it may not have any public methods, when it is used as a base contract it is fine. So actually we can not decide whether a contract is valid or not by looking at them individually, it would be wrong to `error` or even `warn` about a base contract with no public functions. So instead I came to the conclusion in `sema`, logically it would be correct to warn if there is no public message __at all__.

Or is there an easy and inexpensive way to check whether a contract is used as a base I missed out?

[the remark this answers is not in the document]
xermicus: As discussed in our daily, I implemented it now so that it warns the user that any contract without external messages should be declared `abstract`.

This got me thinking, could we actually `error` on that? Only because it is allowed in `solc` does not mean we ultimately must allow it too. If this would result into an error in `sema` we could spare us various redundant checks throughout `codegen` and metadata generation. @seanyoung what do you think?

src/sema/contracts.rs
```diff
@@ -691,6 +691,14 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
         ));
     }
 
+    let contract = &ns.contracts[contract_no];
+    if ns.target.is_substrate() && contract.is_concrete() && !contract.has_public_functions(ns) {
+        diagnostics.push(ast::Diagnostic::warning(
+            contract.loc,
+            format!("Contract '{}' has no public messages. Consider declaring it as an abstract contract instead.", contract.name),
```
seanyoung, line 698: - lower case contract please
- external or public
-  "messages" is an ink! term, this might be confusing for Solidity developers. I think `public or external functions` makes more sense.
- It might be helpful to explain why this is a problem. Something like "Substrate requires at least one message" or a link to the documentation with a section on this.

src/sema/ast.rs
```diff
@@ -638,6 +638,13 @@ impl Contract {
             })
             .cloned()
     }
+
+    /// Does this contract have at least one public function
+    pub fn has_public_functions(&self, ns: &Namespace) -> bool {
+        self.functions
+            .iter()
+            .any(|func_no| ns.functions[*func_no].is_public())
+    }
```
seanyoung, line 647: This should check for function type and use all functions, included inherited ones.

```rust
    pub fn has_public_function(&self, ns: &Namespace) -> bool {
        self.all_functions.keys().any(|func_no| {
            let func = &ns.functions[*func_no];

            func.ty == pt::FunctionTy::Function && func.is_public()
        })
    }
```

src/sema/contracts.rs
xermicus: Are you sure this is the right location for this check?

The `data_locations_01` test contains the follow code:
```solidity
contract foo {
    function bar(uint calldata x) public returns () {
  }
}
```

This erroneously fails with for not having public messages, probably because the previous failure (`data location 'calldata' can only be specified for array, struct or mapping`). I'm having a bit of hard time finding the right place for this check to happen...

seanyoung: So if the prototype of a function can't be resolved it's not added to the ast. This is should really be fixed, which is a large issue to solve (lots of code paths to consider)

However, for now, only perform the check if there are no diagnostic errors so far. So `if !ns.diagnostics.any_error() { ... }`

xermicus: > So if the prototype of a function can't be resolved it's not added to the ast. This is should really be fixed, which is a large issue to solve (lots of code paths to consider)

I see, thanks!

> However, for now, only perform the check if there are no diagnostic errors so far. So if !ns.diagnostics.any_error() { ... }

Sorry I forgot to mention, I already tried that and it didn't help in this case

src/abi/substrate.rs
```diff
@@ -1,437 +1,465 @@
...
 
-#[derive(Deserialize, Serialize, PartialEq, Eq)]
-struct Composite {
-    composite: StructFields,
-}
+fn int_to_ty(ty: &ast::Type, registry: &mut PortableRegistry) -> Type<PortableForm> {
+    let scalety = match ty {
+        // Substrate doesn't like primitive types which aren't a power of 2
+        // The abi encoder/decoder fixes this automatically
+        ast::Type::Uint(n) => format!("u{}", n.next_power_of_two()),
```
seanyoung, line 59: Nit: rather than using `format!()` to create a string and then matching the string, you could do:

```rust
ast::Type::Uint(n) => match n.next_power_of_two()) {
        8 => TypeDefPrimitive::U8,
        16 => TypeDefPrimitive::U16,
        32 => TypeDefPrimitive::U32,
        64 => TypeDefPrimitive::U64,
        128 => TypeDefPrimitive::U128,
        256 => TypeDefPrimitive::U256,
},
```

seanyoung (commented):
Looks good to me, I have one minor nitpick

src/abi/substrate.rs
extraymond: Semi adjusted, the sign part is still needed to generate the path name.

seanyoung: I don't know why that is needed, you could write it like to so:

```rust
    let def = match ty {
        ast::Type::Uint(bits) => match bits.next_power_of_two() {
            8 => TypeDefPrimitive::U8,
            16 => TypeDefPrimitive::U16,
            32 => TypeDefPrimitive::U32,
            64 => TypeDefPrimitive::U64,
            128 => TypeDefPrimitive::U128,
            256 => TypeDefPrimitive::U256,
            _ => unreachable!(),
        },
        ast::Type::Int(bits) => match bits.next_power_of_two()) {
            8 => TypeDefPrimitive::I8,
            16 => TypeDefPrimitive::I16,
            32 => TypeDefPrimitive::I32,
            64 => TypeDefPrimitive::I64,
            128 => TypeDefPrimitive::I128,
            256 => TypeDefPrimitive::I256,
            _ => unreachable!(),
        },`
        _ => unreachable!(),
    };
```

seanyoung: Ah I now see did you need the string later on

```rust
    let path = serde_json::from_value::<Path<PortableForm>>(serde_json::json!(vec![format!(
        "{signed}{scalety}"
    )]))
    .unwrap();
```
Nevermind, maybe the original was nicer. Up to you. Sorry for the bad review comments

extraymond: It's okay, I think it's not much different.

xermicus:
@LucasSte @seanyoung you can review this now. I won't merge until I can depend on some `ink!` v4 beta release, but that shouldn't matter.

xermicus (approved):
LGTM! This fixes like half of our existing substrate integration test suite :eyes: 

Thank you SO much @extraymond for your initiative, initial explorations and continuous feedback! :heart:

src/abi/substrate.rs
```diff
@@ -1,284 +1,459 @@
...
-                    len: array_len,
-                    ty: elem,
-                },
-            },
+    let messages = ns.contracts[contract_no]
+        .all_functions
+        .keys()
+        .filter_map(|function_no| {
+            let func = &ns.functions[*function_no];
+            // escape if it's a library
```
seanyoung, line 401: maybe this should say something like `// libraries are never in the public interface`

src/abi/substrate.rs
```diff
@@ -1,284 +1,459 @@
...
+        .collect();
+    let contract_name = ns.contracts[contract_no].name.clone();
+    let storage = Layout::Struct(StructLayout::new(contract_name, fields));
 
-#[derive(Deserialize, Serialize)]
-struct Storage {
-    #[serde(rename = "struct")]
-    structs: StorageStruct,
-}
+    let f_to_constructor = |f: &Function| -> ConstructorSpec<PortableForm> {
```
seanyoung, line 275: `f_to_constructor` is not great. How about `function_to_constructor` or simply `constructor_spec`

src/abi/substrate.rs
```diff
@@ -1,284 +1,459 @@
...
-impl Abi {
-    /// Add a type to the list unless already present
-    fn register_ty(&mut self, ty: Type) -> usize {
-        match self.types.iter().position(|t| *t == ty) {
-            Some(i) => i + 1,
-            None => {
-                self.types.push(ty);
-
-                self.types.len()
+    let f_to_message = |f: &Function| -> MessageSpec<PortableForm> {
```
seanyoung, line 325: `f_to_message` is not great. How about `function_to_message` or simply `message_spec`

src/abi/substrate.rs
```diff
@@ -1,284 +1,459 @@
...
         })
-    }
+        .map(f_to_message)
+        .collect::<Vec<MessageSpec<PortableForm>>>();
 
-    /// Returns index to builtin type in registry. Type is added if not already present
-    fn builtin_enum_type(&mut self, e: &ast::EnumDecl) -> usize {
-        let mut variants: Vec<EnumVariant> = e
-            .values
+    let mut e_to_evt = |e: &EventDecl| -> EventSpec<PortableForm> {
```
seanyoung, line 419: `event_spec` maybe?

src/abi/substrate.rs
```diff
@@ -295,379 +470,39 @@ fn tags(contract_no: usize, tagname: &str, ns: &ast::Namespace) -> Vec<String> {
...
+    serde_json::to_value(ContractMetadata::new(source, contract, None, abi)).unwrap()
 }
 
-/// Given a selector like "0xF81E7E1A", parse the bytes. This function
-/// does not validate the input.
-fn parse_selector(selector: &str) -> Vec<u8> {
-    hex::decode(&selector[2..]).unwrap()
+pub fn load(s: &str) -> InkProject {
+    let bundle = serde_json::from_str::<ContractMetadata>(s).unwrap();
+    serde_json::from_value::<InkProject>(serde_json::to_value(bundle.abi).unwrap()).unwrap()
```
seanyoung, line 507: I think this function belongs in `tests/substrate.rs`

src/abi/substrate.rs
```diff
@@ -1,284 +1,459 @@
...
+                        Default::default(),
+                    )
+                })
+                .collect::<Vec<_>>();
+            let composite = TypeDef::Composite(TypeDefComposite::new(fields));
+            let path = path!("ExternalFunction");
+            let ty = Type::new(path, vec![], composite, Default::default());
+            registry.register_type(ty)
+        }
+        ast::Type::UserType(no) => resolve_ast(&ns.user_types[*no].ty, ns, registry),
```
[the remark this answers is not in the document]
xermicus: How could I possibly forget about this. Thanks!

src/abi/substrate.rs
```diff
@@ -1,284 +1,459 @@
...
-    name: String,
-    layout: LayoutField,
-}
+                MessageParamSpec::new(p.name_as_str().to_string())
+                    .of_type(spec)
+                    .done()
+            })
+            .collect::<Vec<MessageParamSpec<PortableForm>>>();
+
+        ConstructorSpec::from_label("new".to_string())
```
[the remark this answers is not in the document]
xermicus: How in the world did I not spot this myself :monkey: :facepalm:

tests/substrate.rs
```diff
@@ -1236,3 +1275,8 @@ pub fn build_solidity_with_options(
         events: Vec::new(),
     }
 }
+
+pub fn load_abi(s: &str) -> InkProject {
```
seanyoung, line 1279: doesn't need to be `pub` (nitpick)

seanyoung (changes requested):

src/abi/substrate.rs
```diff
@@ -1,284 +1,473 @@
...
+            let ty = Type::new(path, vec![], composite, Default::default());
+            registry.register_type(ty)
+        }
+        ast::Type::UserType(no) => {
+            let decl = &ns.user_types[*no];
+            let resolved = resolve_ast(&decl.ty, ns, registry);
+            match (decl.name.as_ref(), decl.loc) {
+                // Builtin Hash type from ink primitives
+                ("Hash", pt::Loc::Builtin) => {
+                    // substituded to struct { AccountId }
```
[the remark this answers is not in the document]
xermicus: You are right, but actually this is very obvious when looking at the code anyways. Nuked the comment

src/sema/namespace.rs
```diff
@@ -780,6 +780,13 @@ impl Namespace {
                             "negative size of array declared".to_string(),
                         ));
                         return Err(());
+                    } else if is_substrate && n > &32.into() {
+                        let msg = format!(
+                            "array dimension of {} exceeds the maximum of 32 on Substrate",
```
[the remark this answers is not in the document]
seanyoung: It's up to you how you want to handle this.

However, you're breaking solidity contracts like this one https://docs.soliditylang.org/en/v0.8.17/types.html#array-members 
```solidity
contract ArrayContract {
    uint[2**20] aLotOfIntegers;
}
```
And you're not explaining why you're breaking this (I have literally no idea). Does this not make for a frustrating developer experience, for us and solidity developers?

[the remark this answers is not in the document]
xermicus: You are :100: right on this one. I think I was testing out something and this just smuggled itself into the PR. It makes no sense and is not needed. Nuked it.

src/abi/substrate.rs
```diff
@@ -1,284 +1,472 @@
...
+                        ty.into()
+                    })
+                    .collect::<Vec<_>>();
+
+                let t = TypeDefTuple::new_portable(fields);
+
+                let path = path!(
+                    &ns.contracts[contract_no].name,
+                    &f.name,
+                    &"return_type".into()
```
seanyoung, line 367: I think `&"return_type".into()` can be replaced with `"return_type"`

.gitignore
```diff
@@ -3,3 +3,4 @@ Cargo.lock
 /target
 **/*.rs.bk
 bundle.ll
+
```
seanyoung, line 6: Random file change?

src/abi/substrate.rs
```diff
@@ -1,284 +1,472 @@
...
+                        Default::default(),
+                        vec![],
+                        TypeDef::Sequence(def),
+                        Default::default(),
+                    ));
+                }
+            }
+            ty
+        }
+        // substituded to [u8; len]
```
seanyoung, line 145: substituted

src/abi/substrate.rs
```diff
@@ -1,284 +1,472 @@
...
+        ast::Type::Bytes(n) => resolve_ast(
+            &ast::Type::Array(
+                Box::new(ast::Type::Uint(8)),
+                vec![ArrayLength::Fixed(BigInt::from(*n as i8))],
+            ),
+            ns,
+            registry,
+        ),
+        // substituded to Vec<u8>
```
seanyoung, line 154: substituted

seanyoung (changes requested):

integration/substrate/arrays.sol
```diff
@@ -13,9 +13,7 @@ contract arrays {
 		Permission[] perms;
 	}
 
-	// declare a sparse array. Sparse arrays are arrays which are too large to
-	// fit into account data on Solana; this is not neccessarily a Solidity feature
-	user[type(uint64).max] users;
+	user[1024] users;
```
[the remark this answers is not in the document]
xermicus: This particular one is actually, the JS API does not support array dims > 2048

```
 1) Deploy arrays contract and test
       arrays in account storage:
     Error: PortableRegistry: 7: Error extracting {"path":[],"params":[],"def":{"array":{"len":100000024,"type":6}},"docs":[]}: Only support for [Type; <length>], where length <= 2048
```

Besides that, dealing with large arrays like that wouldn't be economic anyways?

tests/substrate_tests/arrays.rs
```diff
@@ -105,7 +105,7 @@ fn storage_arrays() {
     let mut runtime = build_solidity(
         r##"
         contract foo {
-            int32[8589934592] bigarray;
+            int32[32] bigarray;
```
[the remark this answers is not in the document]
xermicus: Isn't storing such an array only a good idea in solidity where it does not actually use any storage for "empty" array index slots?

integration/substrate/arrays.sol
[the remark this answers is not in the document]
seanyoung: Why would it not be economic? This is a storage array, it's never allocated in memory. Access is per-element (each element of the array uses a storage slot).

Having said that, I don't know what to do about the javascript limitation. You could skip storage elements if they are too large.

tests/substrate_tests/arrays.rs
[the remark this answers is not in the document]
seanyoung: I don't know about ink!, but in Solidity declaring a large storage array without an initializer means no storage is used until it is accessed, i.e. when individual elements are written to. So yes, "empty" array slots do not use any storage (or when you do `delete bigarray[102];`)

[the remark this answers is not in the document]
xermicus: changed to `u16::MAX` as a compromise, WDYT?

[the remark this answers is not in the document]
xermicus: Reason is max array len for array type defs in the metadata is u32 anyways.

integration/substrate/arrays.sol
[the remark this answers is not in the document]
xermicus: Sorry I misread, I thought in the test it would actually return the entire `users` array, not just a single one :facepalm: 

Hmm yes but that is a problem I'd rather solve later on, e.g. when a user actually needs this...

src/abi/substrate.rs
```diff
@@ -1,284 +1,468 @@
...
-    pub selector: String,
-    pub docs: Vec<String>,
-    args: Vec<Param>,
-}
+/// Given an `ast::Type`, find and register the `scale_info::Type` definition in the registry
+fn resolve_ast(ty: &ast::Type, ns: &ast::Namespace, registry: &mut PortableRegistryBuilder) -> u32 {
+    match ty {
+        //  should reflect address_length for different substrate runtime
+        ast::Type::Address(_) | ast::Type::Contract(_) => {
+            // substituted to [u8 ;address_length]
```
LucasSte, line 95:
```suggestion
            // substituted to [u8; address_length]
```

src/sema/namespace.rs
```diff
@@ -760,10 +760,10 @@ impl Namespace {
         id: &pt::Expression,
         diagnostics: &mut Diagnostics,
     ) -> Result<Type, ()> {
-        fn resolve_dimensions(
-            ast_dimensions: &[Option<(pt::Loc, BigInt)>],
-            diagnostics: &mut Diagnostics,
-        ) -> Result<Vec<ArrayLength>, ()> {
+        let is_substrate = self.target.is_substrate();
```
LucasSte, line 763: Why is this variable necessary?

LucasSte (changes requested):

src/sema/namespace.rs
xermicus: I need this bool inside the closure. But we can't borrow `self` in the closure, because it is borrowed later in the function as well

xermicus:
`ink!` release didn't happen this week. Pinned the dependencies so we can merge as soon as @LucasSte approves

integration/substrate/destruct.spec.ts
```diff
@@ -29,23 +29,22 @@ describe('Deploy destruct contract and test', () => {
...
 
-        // let { data: { free: daveBalAfter } } = await conn.query.system.account(dave.address);
-        // let { data: { free: contractBalAfter } } = await conn.query.system.account(String(deploy_contract.address));
+        let { data: { free: daveBalAfter } } = await conn.query.system.account(dave.address);
+        let { data: { free: contractBalAfter } } = await conn.query.system.account(String(deploy_contract.address));
 
-        // //console.log(`bal ${daveBalBefore} and ${daveBalAfter}`);
-        // //console.log(`bal ${contractBalBefore} and ${contractBalAfter}`);
+        //console.log(`bal ${daveBalBefore} and ${daveBalAfter}`);
+        //console.log(`bal ${contractBalBefore} and ${contractBalAfter}`);
```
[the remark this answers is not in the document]
LucasSte: They are still here.

src/abi/substrate.rs
LucasSte: @xermicus you have not changed this.

LucasSte (changes requested):

LucasSte (approved):

PR #990, opened 2022-08-28
test: initial subxt test

extraymond:
alternative tests using subxt, currently using raw selector until the ABI are compatible where we can just use contract-transcode.
test cases are semantically derived from the original substrate integration tests using polkadot.js.


extraymond:
Started to use contract-transcode for encoding message selector, which works with the two exception:
1. constructor overloading will not work due to all constructors have the same name, which contract-transcode will only match the first one
2. transcoding U256/I256 won't work. U256 is configured with sp_core::U256 as it's custom transcoder but it won't actually use it since it's using the TypeInfo::type_info() to get the Path, which is all empty for solang runtime types. I256 is unprocessed at all.

xermicus:
I think this is a great effort overall!

> when using constructor overloading, due to all constructors have the same name, which contract-transcode will only match the first one, currently it'll never reach the second constructor.

Options I see so far:
- Introduce a custom attribute to set a name for the constructor (e.g. like the way it is done for `selector`), and default to `new` or something else when no overloading is happening and no name is given. 
- Mangle the name based by its arguments, would be less work to code up I guess, but ugly from a user perspective

> transcoding U256/I256 won't work. U256 is configured with sp_core::U256 as it's custom transcoder but it won't actually use it since it's using the TypeInfo::type_info() to get the Path, which is all empty for solang runtime types. I256 is unprocessed at all.

U256 is not (yet) supported in `ink!`. The easiest solution would be if we can get it to work by using the U256 path of `ethereum-types` lib somehow?

.github/workflows/test.yml
```diff
@@ -65,10 +65,19 @@ jobs:
...
+    - name: build integration-test binary for subxt
+      run : eval "cp $(cargo test --no-run --message-format=json | jq '. | select(.executable != null) | .executable') test-binary"
+      working-directory: ./integration/subxt-tests
     - uses: actions/upload-artifact@master
       with:
         name: solang-linux-x86-64
         path: ./target/debug/solang
+    - uses: actions/upload-artifact@master
+      with:
+        name: test-binary
```
seanyoung, line 78: The name is not very descriptive, `substrate-subxt-tests`  or so would be much better. Even better would be not to copy this file around

.github/workflows/test.yml
```diff
@@ -283,6 +292,39 @@ jobs:
...
+        path: bin
+    - run: |
+        chmod 755 ./bin/test-binary
+        echo "$(pwd)/bin" >> $GITHUB_PATH        
+    - run: solang compile --target substrate ../substrate/*.sol -o ./contracts/
+      working-directory: ./integration/subxt-tests
+    - run: solang compile --target substrate ../substrate/test/*.sol -o ./contracts/
+      working-directory: ./integration/subxt-tests
+    - name: Deploy and test contracts
+      run: test-binary --test-threads=1
```
seanyoung, line 325: `cargo test --test-thread=1` ?

.gitignore
```diff
@@ -3,3 +3,6 @@ Cargo.lock
 /target
 **/*.rs.bk
 bundle.ll
+/.vscode
+integration/subxt-tests/contracts/
+integration/subxt-tests/target/
```
seanyoung, line 8: It might be easier to keep this in `integraton/subxt-tests/.gitignore`

integration/subxt-tests/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+edition = "2021"
+name = "subxt-tests"
+version = "0.1.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
seanyoung, line 6: Not needed, we know where to find it :grin:

integration/subxt-tests/src/cases/structs.rs
```diff
@@ -0,0 +1,229 @@
...
+
+use crate::{load_project, DeployContract, Execution, ReadContract, WriteContract, API};
+
+#[tokio::test]
+async fn case() -> anyhow::Result<()> {
+    let api = API::new().await?;
+
+    let code = std::fs::read("./contracts/structs.wasm")?;
+
+    let p = load_project("./contracts/structs.contract")?;
```
seanyoung, line 15: The `.contract` file also contains the wasm. Can we use it from there?

.github/workflows/test.yml
```diff
@@ -65,10 +65,19 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose --workspace
+    - run: apt-get update && apt-get install jq -y
+    - name: build integration-test binary for subxt
+      run : eval "cp $(cargo test --no-run --message-format=json | jq '. | select(.executable != null) | .executable') test-binary"
+      working-directory: ./integration/subxt-tests
```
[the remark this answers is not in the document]
extraymond: Because it'll need substrate to be running in the background, which needs docker image. Which solang CI image don't provide. So I build the test binary and run it later.

integration/subxt-tests/Cargo.toml
extraymond: It's from cargo new, Aka nanny compiler.

.gitignore
extraymond: Agree! I'll kept it in it's respective folder.

.github/workflows/test.yml
extraymond: It's `cargo test -- --test-threads=1` Which is the argument for the test binary to only run one task at a time, this avoids subxt to spam the running substrate node with transaction having the same nonce.

.github/workflows/test.yml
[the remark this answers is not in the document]
seanyoung: The substrate tests don't run in a container (let alone the Solang CI container). These tests run the CI VM and both `docker`  and `cargo` should be available in this environment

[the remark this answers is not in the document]
extraymond: I've clean this up so it's not carried over anymore, currently not running due to tests failing after rebase.

xermicus:
@extraymond it makes me sad that we didn't go forward sooner with this PR, sorry!. I really like what you implemented here. In the long run, it would make sense to integrate this with #1061. How-ever, this should not block your nice work here. May I ask you for the favor of merging current `main` with this again? Then I'd be very happy to give a usual round of review and LGTM!

extraymond:
> @extraymond it makes me sad that we didn't go forward sooner with this PR, sorry! I really like what you implemented here. In the long run, it would make sense to integrate this with #1061. How-ever, this should not block your nice work here. May I ask you for the favor of merging current `main` with this again? Then I'd be very happy to give a usual round of review and LGTM!

Sure! I'll be on it sometime this week.

integration/subxt-tests/src/cases/structs.rs
extraymond: Fixed in the latest commit

extraymond:
@xermicus @seanyoung 
Rebased to current master. Some failed tests are ignored as suggested. Looking forward for your review.

integration/subxt-tests/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+
+// metadata file obtained from the latest substrate-contracts-node
+#[subxt::subxt(
+    runtime_metadata_path = "./metadata.scale",
+    substitute_type(
+        type = "sp_weights::weight_v2::Weight",
+        with = "::subxt::utils::Static<::sp_weights::Weight>"
+    )
+)]
+pub mod node {}
```
xermicus, line 33: IIRC `subxt` can download the metadata from the node, so we don't have to maintain it here (the typescript test suite doesn't need this). Could you check if it is possible?

.github/workflows/test.yml
```diff
@@ -347,6 +352,42 @@ jobs:
...
+      with:
+        toolchain: stable
+        default: true
+        override: true   
+    - run: solang compile --target substrate ../substrate/*.sol -o ./contracts/
+      working-directory: ./integration/subxt-tests
+    - run: solang compile --target substrate ../substrate/test/*.sol -o ./contracts/
+      working-directory: ./integration/subxt-tests
+    - name: Deploy and test contracts
+      run: cargo test -- --test-threads=1
```
xermicus, line 385: Why is `--test-threads=1` needed?

integration/subxt-tests/src/cases/events.rs
```diff
@@ -0,0 +1,71 @@
...
+    let rs = c
+        .call(&api, sp_keyring::AccountKeyring::Alice, 0, |t| {
+            t.encode::<_, &'static str>("emit_event", [])
+                .unwrap_or_default()
+        })
+        .await?;
+
+    assert_eq!(rs.len(), 4);
+
+    // TODO: currently event decoding is different than ink, as we can see in contract-transcode.
```
xermicus, line 28: Is this still the case? I can't remember having this as an issue with polkadot-js, but I might be wrong. What exactly is the difference you encountered? I'm wondering if we could fix it in solang.

integration/subxt-tests/src/cases/external_call.rs
```diff
@@ -0,0 +1,207 @@
...
+    // setX on callee
+    let selector = t_callee.encode::<_, String>("set_x", [format!("102")])?;
+
+    // let selector = build_selector(
+    //     "250c2025",
+    //     Some(&mut |s| {
+    //         102_i64.encode_to(s);
+    //     }),
+    // )?;
+
```
xermicus, line 62:
```suggestion
```

xermicus (commented):
Gave it a first round of review; looks great overall, I hope to merge this soon. Thank you so much for keeping this going! :heart:

.github/workflows/test.yml
```diff
@@ -347,6 +352,42 @@ jobs:
...
+    name: Substrate Integration test with subxt
+    runs-on: ubuntu-20.04
+    needs: linux-x86-64
+    steps:
+    - name: Checkout sources
+      uses: actions/checkout@v2
+      # We can't run substrate as a github actions service, since it requires
+      # command line arguments. See https://github.com/actions/runner/pull/1152
+    - name: Start substrate
+      run: echo id=$(docker run -d -p 9944:9944 paritytech/contracts-ci-linux@sha256:77a885f3c22d38385b017983c942bf7e063ac127bdc0477670d23a38711a2c38 substrate-contracts-node --dev --ws-external) >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
xermicus: Ah you changed it already, nvmd

.github/workflows/test.yml
```diff
@@ -6,6 +6,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+# FIXME: currently needed to bypass a transient dependency using nightly feature
+env:
+ RUSTC_BOOTSTRAP: 1
+
```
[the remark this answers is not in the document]
extraymond: Yeah, I'll remove that.

.github/workflows/test.yml
extraymond: I think I hit some node error not specifying this.
At least for my testing rig, it's way too fast for cargo to submit tx calls and the runtime is dropping it.

integration/subxt-tests/src/cases/arrays.rs
```diff
@@ -0,0 +1,105 @@
+use crate::{Contract, DeployContract, Execution, ReadContract, WriteContract, API};
+use contract_transcode::{ContractMessageTranscoder, Value};
+use hex::FromHex;
+
+use parity_scale_codec::{Decode, Encode};
+use rand::{seq::SliceRandom, thread_rng, Rng};
+use sp_core::{crypto::AccountId32, hexdisplay::AsBytesRef};
+
+#[ignore]
```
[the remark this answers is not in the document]
extraymond: Great advice! I'll try to see if I can find the cause and add it.

integration/subxt-tests/src/cases/events.rs
extraymond: I have to do some testing. I remember it being an extra byte off for event_id or some sort.

integration/subxt-tests/src/lib.rs
xermicus: Yeah this is exactly what I had in mind 👍  (couldn't find it though)

.github/workflows/test.yml
xermicus: Ah yes makes sense. Also we might hit issues with the CI nodes instant finality if we parallelize too much . I you are right we want to have this in.

integration/subxt-tests/src/cases/events.rs
extraymond: Found it.
https://github.com/paritytech/cargo-contract/blob/ce9afac78eb76aa1c78998e08861764619f5e8ff/crates/transcode/src/lib.rs#LL293C9-L293C50

This file strips one `Compact<u32>` where solang generated event doesn't have. If we manually prepend one into the Event.data field, the decoding will success.

xermicus: Right, thanks for digging this out.

extraymond:
@xermicus  I think it would be nice to tap into debug buffer when contract trapped. It's very hard to debug why it's trapped within the test.

xermicus:
> @xermicus I think it would be nice to tap into debug buffer when contract trapped. It's very hard to debug why it's trapped within the test.

I see what you mean; however I think this should ideally be done by `contracts` pallet (so the runtime executing the contract). In the contract code itself, we have limited control about trapping. We might write into the debug buffer if we know for sure we are going to trap (for example because the compiler inserts a trap manually). However, there are many other cases (OOM, invalid instructions etc) only the runtime knows or has control about.

For this to work we would need to provide the runtime debug info, but for now IIRC the contracts pallet doesn't support this.

extraymond:
> > @xermicus I think it would be nice to tap into debug buffer when contract trapped. It's very hard to debug why it's trapped within the test.
> 
> I see what you mean; however I think this should ideally be done by `contracts` pallet (so the runtime executing the contract). In the contract code itself, we have limited control about trapping. We might write into the debug buffer if we know for sure we are going to trap (for example because the compiler inserts a trap manually). However, there are many other cases (OOM, invalid instructions etc) only the runtime knows or has control about.
> 
> For this to work we would need to provide the runtime debug info, but for now IIRC the contracts pallet doesn't support this.

Yeah, cargo-contract does it by doing a sequential read call if a first write call failed. Maybe we should revisit it later when we have some shared lib from cargo-contract that we can directly pulled.

integration/subxt-tests/src/cases/primitives.rs
```diff
@@ -0,0 +1,697 @@
...
+        .try_call(&api, sp_keyring::AccountKeyring::Alice, 0, |t| {
+            t.encode::<_, String>("op_u64", ["shr".into(), "1000".into(), "8".into()])
+                .expect("unable to find selector")
+        })
+        .await
+        .and_then(|rv| u64::decode(&mut rv.as_bytes_ref()).map_err(Into::into))?;
+    assert_eq!(res, 3);
+
+    // op_i256
+    // TODO: currently contract-transcode don't support encoding/decoding of I256 type so we'll need  to encode it manually
```
LucasSte, line 225:
```suggestion
    // TODO: currently contract-transcode doesn't support encoding/decoding of I256 type so we'll need  to encode it manually
```

LucasSte (approved):
LGTM

integration/subxt-tests/src/cases/primitives.rs
extraymond: Oops! Thanks.

extraymond:
@seanyoung I think one of my forced push for rebase cleared your original change request. I believe I have implemented the suggestion back then.

xermicus (approved):
LGTM

seanyoung (approved):
Looks great. I only skimmed the new tests - a huge amount, great stuff!

.github/workflows/test.yml
```diff
@@ -347,6 +348,42 @@ jobs:
       if: always()
       run: docker kill ${{steps.substrate.outputs.id}}
 
+  substrate-subxt:
+    name: Substrate Integration test with subxt
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
extraymond: Yeah, it was originally written before the first point release. I personally always stick to lts after first point release.

.github/workflows/test.yml
```diff
@@ -347,6 +348,42 @@ jobs:
...
+    - name: Install latest rust toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        default: true
+        override: true   
+    - run: solang compile --target substrate ../substrate/*.sol -o ./contracts/
+      working-directory: ./integration/subxt-tests
+    - run: solang compile --target substrate ../substrate/test/*.sol -o ./contracts/
+      working-directory: ./integration/subxt-tests
```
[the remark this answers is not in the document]
extraymond: Yeah it can, changed in next commit.

xermicus:
Amazing, thank you!

seanyoung:
@extraymond good stuff, thanks!

PR #996, opened 2022-08-30
Create Solana's System Instruction library and 'solana-library' folder

LucasSte:
This PR adds to our repository a library for Solana's system instructions. They allow developers to create accounts, transfer funds and assign authorities to accounts. Creating accounts is an essential part of an NFT example on Solana created entirely with Solidity.

In addition, this PR creates a `solana-library` folder to hold all the libraries we implement for Solana.


xermicus:
Why do you switch all those functions to `view`?

solana-library/spl_token.sol
```diff
@@ -43,7 +43,7 @@ library SplToken {
 	/// @param account the token account where the minted tokens should go
 	/// @param authority the public key of the mint authority
 	/// @param amount the amount of tokens to mint
-	function mint_to(address mint, address account, address authority, uint64 amount) internal {
+	function mint_to(address mint, address account, address authority, uint64 amount) view internal {
```
seanyoung, line 46: This is wrong. minting can't be a view operation. None of these should be view

seanyoung (changes requested):
I think it would make more sense to introduce a `solana-library` directory in a pr which introduces more files, That PR can introduce what it looks like in the documentation etc.

solana-library/spl_token.sol
LucasSte: Solang raises a warning if the `view` keyword is missing:

```
warning: function can be declared 'view'
   ┌─ /Users/lucas/Documents/solang/solana-library/spl_token.sol:46:2
   │
46 │     function mint_to(address mint, address account, address authority, uint64 amount) internal {
   │
```

LucasSte:
> Why do you switch all those functions to `view`?

To avoid compiler warnings, telling me to declare them as view.

LucasSte:
> I think it would make more sense to introduce a `solana-library` directory in a pr which introduces more files, That PR can introduce what it looks like in the documentation etc.

Good point, I'll reference this folder in the docs. More files will be added in another PR.

solana-library/spl_token.sol
seanyoung: `view` means that state cannot be changed and this is a read-only function which can run without a transaction.

Clearly minting changes state.

LucasSte: The language server also warns that all these functions should be view.

seanyoung: Now, the warning is a different matter. The problem the compiler has is that it cannot know whether a function called view `.call()` changes state. The same thing exists with solc.

I'm not sure how to solve this. For Solana it might be possible to talk the list of meta accounts and seeing if any of them are writable. However that doesn't solve the problem for Substrate, and also it might not always be possible to walk the list of meta accounts on Solana eiher, if they're passed in via a function.

Some new syntax might be needed for this.

Alternatively, we could do something like "if `.call()` is done, then don't warn about `view` being possible". I Think that might make the most sense.

LucasSte: The warning makes the developer believe he/she can make the function `view`. Hardly ever will someone doubt the compiler is wrong. If the warning is not 100% accurate, then we should not have raised it.

seanyoung: Feel free to raise an issue

seanyoung:
I think it would make more sense to introduce a solana-library directory in a pr which introduces more files, That PR can introduce what it looks like in the documentation etc.

xermicus:
> > Why do you switch all those functions to `view`?
> 
> To avoid compiler warnings, telling me to declare them as view.

Hmm, it just struck me as odd to declare those `view`, given the name of these functions imply they are anything but `view`. I doubt that the functions they are calling to also `view` ([those must also be view so the function can be view, according to Solidity docs](https://docs.soliditylang.org/en/v0.8.16/contracts.html?highlight=view#view-functions))?

docs/targets/solana.rst
```diff
@@ -217,18 +217,42 @@ the provided seeds, along with a seed bump. See the Solana documentation on
...
-`spl-token <https://spl.solana.com/token>`_ is the solana native way of creating tokens, minting, burning and
-transfering token. This is the Solana equivalent of
+
+Solana Library
+______________
+
+In Solang's Github repository, there is a folder called ``solana-library``. It contains libraries for Solidity contracts
+to interact with Solana specific instructions. Currently, there are two libraries there. One for SPL tokens and another
+for Solana's system instructions. In order to use those functionalities, one must copy the correspondent library
+file to one's project folder and import the library.
```
seanyoung, line 228:
```suggestion
In Solang's Github repository, there is a directory called ``solana-library``. It contains libraries for Solidity contracts
to interact with Solana specific instructions. Currently, there are two libraries there: one for SPL tokens and another
for Solana's system instructions. In order to use those functionalities, copy the correspondent 
file to one's project and import it.
```

docs/targets/solana.rst
```diff
@@ -217,18 +217,42 @@ the provided seeds, along with a seed bump. See the Solana documentation on
...
+
+SPL-token
++++++++++
+
+`spl-token <https://spl.solana.com/token>`_ is the Solana native way of creating tokens, minting, burning and
+transferring token. This is the Solana equivalent of
 `ERC-20 <https://ethereum.org/en/developers/docs/standards/tokens/erc-20/>`_ and
 `ERC-721 <https://ethereum.org/en/developers/docs/standards/tokens/erc-721/>`_. We have created a library ``SplToken`` to use
```
seanyoung, line 236: `We have created` is unnecessary for documentation. We have created all solang, after all.

integration/solana/setup.ts
```diff
@@ -46,17 +46,25 @@ async function newAccountWithLamports(connection: Connection): Promise<Keypair>
...
     await connection.confirmTransaction(signature, 'confirmed');
-    signature = await connection.requestAirdrop(account.publicKey, LAMPORTS_PER_SOL);
+    signature = await connection.requestAirdrop(account.publicKey, 5*LAMPORTS_PER_SOL);
     await connection.confirmTransaction(signature, 'confirmed');
-    signature = await connection.requestAirdrop(account.publicKey, LAMPORTS_PER_SOL);
+    signature = await connection.requestAirdrop(account.publicKey, 5*LAMPORTS_PER_SOL);
+    await connection.confirmTransaction(signature, 'confirmed');
+    signature = await connection.requestAirdrop(account.publicKey, 5*LAMPORTS_PER_SOL);
+    await connection.confirmTransaction(signature, 'confirmed');
+    signature = await connection.requestAirdrop(account.publicKey, 5*LAMPORTS_PER_SOL);
```
seanyoung, line 55: Can this be merged into a single request for `16*LAMPORTS_PER_SOL`

seanyoung (changes requested):

docs/targets/solana.rst
LucasSte: `copy the correspondent file to one's project`. The usage of articles must maintain parallelism. If I use `one must`, the possessive must be `one's`. If I use the imperative form `copy`, the possessive pronoun must be `your`. Do you have any preference over one form?

docs/targets/solana.rst
```diff
@@ -217,18 +217,42 @@ the provided seeds, along with a seed bump. See the Solana documentation on
...
-`spl-token <https://spl.solana.com/token>`_ is the solana native way of creating tokens, minting, burning and
-transfering token. This is the Solana equivalent of
+
+Solana Library
+______________
+
+In Solang's Github repository, there is a directory called ``solana-library``. It contains libraries for Solidity contracts
+to interact with Solana specific instructions. Currently, there are two libraries there: one for SPL tokens and another
+for Solana's system instructions. In order to use those functionalities, copy the correspondent library
+file to your project folder and import it.
```
seanyoung, line 228: in consistent usage of folder/directory. I would suggest removing folder
```suggestion
file to your project and import it.
```

seanyoung (changes requested):
Just one nitpick

xermicus (changes requested):
I think we should add license headers and disclaimers to the libraries here as well

solana-library/system_instruction.sol
```diff
@@ -0,0 +1,295 @@
...
+            AccountMeta({pubkey: recentBlockHashes, is_signer: false, is_writable: false}),
+            AccountMeta({pubkey: rentAddress, is_signer: false, is_writable: false}),
+            AccountMeta({pubkey: authorized_pubkey, is_signer: true, is_writable: false})
+        ];
+
+        bytes bincode = abi.encode(uint32(Instruction.WithdrawNonceAccount), lamports);
+        systemAddress.call{accounts: metas}(bincode);
+    }
+
+    /// Change th entity authorized to execute nonce instructions on the account
```
seanyoung, line 268: s/th/the/

seanyoung (changes requested):
SPDX License is missing

LucasSte:
> I think we should add license headers and disclaimers to the libraries here as well

What do you mean by disclaimer?

xermicus:
> > I think we should add license headers and disclaimers to the libraries here as well
> 
> What do you mean by disclaimer?

That all provided code is any guarantees for correctness, it is not audited by a third party and production usage is at own risk

solana-library/system_instruction.sol
```diff
@@ -0,0 +1,297 @@
+// SPDX-License-Identifier: Apache-2.0
+
```
xermicus, line 2:
```suggestion
// This code is provided as-is. It is not audited and any use is at your own risk.

```

integration/solana/system_instruction_example.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+
```
xermicus, line 2:
```suggestion
// This code is provided as-is. It is not audited and any use is at your own risk.

```

integration/solana/token.sol
```diff
@@ -1,4 +1,4 @@
-import '../../examples/spl_token.sol';
+import '../../solana-library/spl_token.sol';
```
xermicus, line 1:
```suggestion
// This code is provided as-is. It is not audited and any use is at your own risk.

import '../../solana-library/spl_token.sol';
```

solana-library/system_instruction.sol
LucasSte: I wrote my own version before seeing your suggestion. Please, check if you agree with my phrasing.

xermicus (approved):

seanyoung (approved):
Looks like there are some commits in here which are not needed, please rebase

PR #999, opened 2022-09-01
Substrate: concrete contracts must have public or external functions

xermicus:
In `ink!`, any contract must have at least one public message. This PR aligns solangs behavior up with `ink!`. It is a precondition for making #989 happen.

A public function is in a contract, if it has `public` or `external` functions, if it has a `receive` or any fallback function or if it has public storage items (those will yield public getters). A new bool flag was added to the `Contract` struct in the `sema` AST to reflect that (otherwise we need to loop through all contract functions anywhere we want to check it).


src/sema/contracts.rs
```diff
@@ -109,6 +110,20 @@ pub fn resolve(
...
+    for (n, _) in contracts {
+        let contract = &ns.contracts[*n];
+        if ns.target.is_substrate()
+            && contract.is_concrete()
+            && !contract.has_callable_function
+            && !ns.diagnostics.any_errors()
+        {
+            ns.diagnostics.push(ast::Diagnostic::error(
+            contract.loc,
+            "contracts without public storage or functions are not allowed on Substrate, considering declaring it abstract".into(),
```
seanyoung, line 122: There is no test case for this error

xermicus: Sorry forgot to make it draft, I'm working on adding tests for this as well

src/sema/ast.rs
```diff
@@ -597,6 +597,13 @@ pub struct Contract {
     pub default_constructor: Option<(Function, usize)>,
     pub cfg: Vec<ControlFlowGraph>,
     pub code: Vec<u8>,
+    // This contract can be instantiated and called or receive funds.
+    // It implies that it has at least one of:
+    // * public messages
+    // * public storage (results in public getters)
+    // * fallback or receive function
+    // alternative names I came up: has_code, extrinsic, is_public, exterior, has_surface
+    pub has_callable_function: bool,
```
seanyoung, line 606: `has_callable_function` is a substrate specific property. How do you feel about:
```rust
pub instantiable: bool,
```
By default this would have the value of `.is_concrete()`, i.e `matches!(ty, pt::ContractTy::Contract(_))` but the contract does not have a requisite function for Substrate, then it can be set to `false`. 

In the future, we could set this to `false` for any sema error and not generate code for this contract, rather than not generating code for all contracts if there is an error.

In various code paths where we use `.is_concrete()` we can replace this with `.instantiable` (or should that be `.is_instantiable`).

wdty? @LucasSte  @xermicus

xermicus: Mostly ACK. But need to be careful as `is_concrete()` is not the same as this bool flag. For this PR I'd be happy to change the name but I don't really fancy touching it in other places...

seanyoung (approved):

src/sema/contracts.rs
```diff
@@ -694,6 +698,31 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
...
+            let func = &ns.functions[*func_no];
+
+            match func.ty {
+                pt::FunctionTy::Function => func.is_public(),
+                pt::FunctionTy::Fallback | pt::FunctionTy::Receive => true,
+                _ => false,
+            }
+        })
+    {
+        let message = format!("contracts without public storage or functions are not allowed on Substrate, considering declaring this contract 'abstract contract {}'", contract.name);
```
LucasSte, line 718:
```suggestion
        let message = format!("contracts without public storage or functions are not allowed on Substrate. Consider declaring this contract 'abstract contract {}'", contract.name);
```
I think you meant the imperative form here.

tests/contract_testcases/substrate/account_info.sol
```diff
@@ -1,4 +1,4 @@
-contract c {
+abstract contract c {
```
LucasSte, line 1: Why does this contract need to be abstract? I has public functions.

LucasSte (changes requested):

tests/contract_testcases/substrate/account_info.sol
xermicus: good catch!

LucasSte (approved):

PR #1001, opened 2022-09-02
Improve vscode extension documentation

seanyoung:

### DIFF
--- a/docs/extension.rst
+++ b/docs/extension.rst
@@ -3,15 +3,16 @@ Visual Studio Code Extension
 ============================
 
 Solang has
-`language server <https://en.wikipedia.org/wiki/Language_Server_Protocol>`_ built
-into the executable, which can be used by the Visual Studio Code extension. This
+a `language server <https://en.wikipedia.org/wiki/Language_Server_Protocol>`_ built
+into the executable, which can be used by the Visual Studio Code extension, or by
+any editor that can use a lsp language server. The Visual Studio Code
 extension provides the following:
 
-1. Syntax highlighting
-2. Compiler warnings and errors are displayed in the problems tab and marked
-   with squiqqly lines, this is also known as `diagnostics`.
-3. Hovering over variables, types, functions etc and more will give information,
-   For example this will give the struct fields when hovering over a variable
+1. Syntax highlighting.
+2. Compiler warnings and errors displayed in the problems tab and marked
+   with squiggly lines.
+3. Additional information when hovering over variables, types, functions, etc.
+   For example, this will give the struct fields when hovering over a variable
    which is a reference to a struct.
 
 .. image:: extension-screenshot.png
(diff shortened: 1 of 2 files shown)


docs/extension.rst
```diff
@@ -4,13 +4,14 @@ Visual Studio Code Extension
 
 Solang has
 `language server <https://en.wikipedia.org/wiki/Language_Server_Protocol>`_ built
```
LucasSte, line 6:
```suggestion
a `language server <https://en.wikipedia.org/wiki/Language_Server_Protocol>`_ built
```
The article was missing.

docs/extension.rst
```diff
@@ -4,13 +4,14 @@ Visual Studio Code Extension
...
-into the executable, which can be used by the Visual Studio Code extension. This
+into the executable, which can be used by the Visual Studio Code extension, or by
+any editor that can use a lsp language server. The Visual Studio Code
 extension provides the following:
 
-1. Syntax highlighting
+1. Syntax highlighting.
 2. Compiler warnings and errors are displayed in the problems tab and marked
```
LucasSte, line 12:
```suggestion
2. Compiler warnings and errors displayed in the problems tab and marked
```
The items in an enumeration must maintain parallelism. If your first item does not contain an action verb, the second one should not have it too.

docs/extension.rst
```diff
@@ -4,13 +4,14 @@ Visual Studio Code Extension
...
-   with squiqqly lines, this is also known as `diagnostics`.
-3. Hovering over variables, types, functions etc and more will give information,
+   with squiggly lines.
+3. Hovering over variables, types, functions, etc. and more will give more information.
```
LucasSte, line 14:
```suggestion
3. Additional information when hovering over variables, types, functions, etc.
```
Maintain parallelsim.

docs/extension.rst
```diff
@@ -4,13 +4,14 @@ Visual Studio Code Extension
...
    For example this will give the struct fields when hovering over a variable
```
LucasSte, line 15:
```suggestion
   For example, this will give the struct fields when hovering over a variable
```
A comma is missing.

docs/extension.rst
```diff
@@ -43,7 +47,7 @@ The extension client code is in
 `src/client/extension.ts <https://github.com/hyperledger/solang/tree/main/vscode/src/client/extension.ts>`_.
 
 Secondly, there is the language server which is written in Rust.
-The Solang binary has an option ``--language-server``, which start the
+The Solang binary has an subcommand ``language-server``, which starts the
```
LucasSte, line 50:
```suggestion
The Solang binary has a subcommand ``language-server``, which starts the
```
Incorrect article.

LucasSte (changes requested):

docs/extension.rst
```diff
@@ -24,12 +25,15 @@ Using the extension
 
 The extension can be found on the `Visual Studio Marketplace <https://marketplace.visualstudio.com/items?itemName=solang.solang>`_.
 
-On first start, the extension will ask to download the Solang binary. Once this is done, it should just automatically work.
+When started for the first time, the extension will download the Solang binary. Once this is done, it should just automatically work.
 However, you should set the blockchain target in the extension settings. The Solidity language differs in subtle ways depending on which target you are
```
xermicus, line 29: To my understanding, the Syntax solang supports is a superset of Solidity? If that is true I think we should point that out here. Also we maybe we should say what version we support somewhere.

xermicus (approved):

docs/extension.rst
seanyoung: Good point, I've added another section for this

docs/extension.rst
```diff
@@ -3,30 +3,60 @@ Visual Studio Code Extension
...
+   Solidity for Solana
+
+substrate
+   Solidity for Substrate (Polkadot)
+
+evm
+   Solidity for any EVM based chain like Ethereum
+
+Note that the language server has support for evm, but Hyperledger Solang does
+not support compiling for evm (yet).
```
LucasSte, line 51: I believe we need to choose whether we want to write EVM in upper case or lower case. You used both ways here. I prefer the upper case. Please, choose only one way to use throughout the documentation.

LucasSte (changes requested):

docs/extension.rst
seanyoung: Please note that the there is one lower case evm left in the documentation, because the command line option and the option in the extension settings is lower case evm.

LucasSte (approved):
Make sure CI tests pass before merging.

seanyoung:
Windows error is spurious - overflow tests are broken right now. As this only changes the documentation, merging for now

PR #1002, opened 2022-09-05
Solana NFT example

LucasSte:
This PR creates a Solidity contract to manage NFTs on Solana. It is an example to show Solang functionality during Solana's Singapore Hacker House.

### DIFF
--- a/integration/solana/simple_collectible.sol
+++ b/integration/solana/simple_collectible.sol
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// DISCLAIMER: This file is an example of how to mint and transfer NFTs on Solana. It is not production ready and has not been audited for security.
+// Use it at your own risk.
+
+import '../../solana-library/spl_token.sol';
+import '../../solana-library/system_instruction.sol';
+
+contract SimpleCollectible {
+    // Contract variables
+    uint32 private tokenCounter;
+    address private mint_authority;
+    address private payer;
+
+    struct NFTData {
+        address owner;
+        address mint_account;
+        string uri;
+        bool exists;
+    }
+
+    event NFTMinted(address owner, address mint_account, uint32 nft_id);
+    event NFTSold(address from, address to, uint32 nft_id);
+
+    mapping(uint32 => NFTData) private nft_information;
+
+    constructor (address _mint_authority) {
+        tokenCounter = 0;
+        // For every mint on Solana, a mint authority is needed to sign each mint
+        mint_authority = _mint_authority;
+    }
+
+    // Sets the payer account that is going to pay for every transference
+    function set_payer(address _payer) public {
+        payer = _payer;
+    }
+
+    function createCollectible(string memory tokenURI, address mint_account, address owner) public returns (uint32) {
+        uint32 newItemId = tokenCounter;
+        tokenCounter++;
+        // An NFT on Solana is a SPL-Token with only one minted token.
+        // The mint_account saves information about the token (e.g. quantity of tokens avaialble)
+        SplToken.mint_to(mint_account, owner, mint_authority, 1);
+
+        nft_information[newItemId] = NFTData(owner, mint_account, tokenURI, true);
+
+        // Save on blockchain records information about the created token
+        emit NFTMinted(owner, mint_account, newItemId);
+        return newItemId;
+    }
+
+    function transfer_ownership(uint32 tokenId, address new_owner) public {
+        NFTData storage data = nft_information[tokenId];
+        if(!data.exists) {
+            return;
+        }
+        
+        // To transfer the ownership of a token, we need the current owner and the new owner. The payer account is the account used to derive
+        // the correspondent token account in JavaScript.
+        SplToken.transfer(data.owner, new_owner, payer, 1);
+        emit NFTSold(data.owner, new_owner, tokenId);
+        data.owner = new_owner;
+    }
+
+    // Returns the URI of an NFT
+    function get_nft_uri(uint32 nftId) public view returns (string memory) {
+        if (nft_information[nftId].exists) {
+            return nft_information[nftId].uri;
+        }
+
+        return "";
+    }
+
+    // Returns the owner and the mint account of an NFT
+    function get_nft_owner(uint32 nft_id) public view returns (address, address) {
+        return (nft_information[nft_id].owner, nft_information[nft_id].mint_account);
+    }
+}
--- a/integration/solana/simple_collectible.sol
+++ b/integration/solana/simple_collectible.sol
@@ -4,75 +4,95 @@
 // Use it at your own risk.
 
 import '../../solana-library/spl_token.sol';
-import '../../solana-library/system_instruction.sol';
 
 contract SimpleCollectible {
-    // Contract variables
+    // tokenCounter creates a unique ID for each new minted NFT
     uint32 private tokenCounter;
-    address private mint_authority;
-    address private payer;
+    // The authority responsible for managing mints.
+    address private mintAuthority;
+
+    struct NFTOwner {
+        bool exists;
+        address owner;
+        address ownerTokenAccount;
+        address mintAccount;
+    }
 
     struct NFTData {
-        address owner;
-        address mint_account;
+        NFTOwner ownerData;
         string uri;
-        bool exists;
     }
 
-    event NFTMinted(address owner, address mint_account, uint32 nft_id);
-    event NFTSold(address from, address to, uint32 nft_id);
+    event NFTMinted(address owner, address mintAccount, uint32 nftId);
+    event NFTSold(address from, address to, uint32 nftId);
 
-    mapping(uint32 => NFTData) private nft_information;
+    mapping(uint32 => NFTData) private nftInfo;
 
-    constructor (address _mint_authority) {
+    constructor (address _mintAuthority) {
         tokenCounter = 0;
-        // For every mint on Solana, a mint authority is needed to sign each mint
-        mint_authority = _mint_authority;
+        // For every mint on Solana, a mint authority is needed to sign it
+        mintAuthority = _mintAuthority;
     }
 
-    // Sets the payer account that is going to pay for every transference
-    function set_payer(address _payer) public {
-        payer = _payer;
-    }
-
-    function createCollectible(string memory tokenURI, address mint_account, address owner) public returns (uint32) {
-        uint32 newItemId = tokenCounter;
+    /// Create a new NFT and associated it to a URI
+    ///
+    /// @param tokenURI a URI that leads to the NFT resource
+    /// @param mintAccount an account that saves the total supply of a token
+    /// @param owner the account that is going to own the newly minted NFT
+    /// @param ownerTokenAccount the associated token account for the @param owner and the @param mintAccount
+    ///
+    /// return: an unique identifier to the minted NFT
+    function createCollectible(string memory tokenURI, address mintAccount, address owner, address ownerTokenAccount) public returns (uint32) {
+        uint32 new_item_id = tokenCounter;
         tokenCounter++;
         // An NFT on Solana is a SPL-Token with only one minted token.
-        // The mint_account saves information about the token (e.g. quantity of tokens avaialble)
-        SplToken.mint_to(mint_account, owner, mint_authority, 1);
+        // The mintAccount saves information about the token (e.g. quantity of tokens avaialble)
+        SplToken.mint_to(mintAccount, ownerTokenAccount, mintAuthority, 1);
 
-        nft_information[newItemId] = NFTData(owner, mint_account, tokenURI, true);
+        NFTOwner ownerData = NFTOwner(true, owner, ownerTokenAccount, mintAccount);
+        nftInfo[new_item_id] = NFTData(ownerData, tokenURI);
 
         // Save on blockchain records information about the created token
-        emit NFTMinted(owner, mint_account, newItemId);
-        return newItemId;
+        emit NFTMinted(owner, mintAccount, new_item_id);
+        return new_item_id;
     }
 
-    function transfer_ownership(uint32 tokenId, address new_owner) public {
-        NFTData storage data = nft_information[tokenId];
-        if(!data.exists) {
+    /// Transfer ownership of an NFT from one account to another
+    ///
+    /// @param tokenId the NFT's unique identifer generated when it was minted
+    /// @param newOwner the account for the new NFT owner
+    /// @param newOwnerTokenAccount the associated token account for the @param newOwner
+    function transferOwnership(uint32 tokenId, address newOwner, address newOwnerTokenAccount) public {
+        NFTData storage data = nftInfo[tokenId];
+        if(!data.ownerData.exists) {
             return;
         }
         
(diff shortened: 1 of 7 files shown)


integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,78 @@
...
+    constructor (address _mint_authority) {
+        tokenCounter = 0;
+        // For every mint on Solana, a mint authority is needed to sign it
+        mint_authority = _mint_authority;
+    }
+
+    // Sets the payer account that is going to sign for every ownership transfer
+    function set_payer(address _payer) public {
+        payer = _payer;
+    }
```
seanyoung, line 36: Why is the same for everyone?

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    function createCollectible(string memory tokenURI, address mint_account, address owner) public returns (uint32) {
```
seanyoung, line 38: This needs a comment. What is the caller supposed to provide? What accounts are required for this transaction? What is the return value?

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,78 @@
...
+        SplToken.mint_to(mint_account, owner, mint_authority, 1);
+
+        nft_information[newItemId] = NFTData(owner, mint_account, tokenURI, true);
+
+        // Save on blockchain records information about the created token
+        emit NFTMinted(owner, mint_account, newItemId);
+        return newItemId;
+    }
+
+    function transfer_ownership(uint32 tokenId, address new_owner) public {
```
seanyoung, line 52: This needs a comment. What is the caller supposed to provide?  And what accounts?

integration/solana/simple_collectible.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+            {
+                accounts: [TOKEN_PROGRAM_ID],
+                writableAccounts: [current_owner, associated_owner_account.address],
+                signers: [payer]
+            }
+        );
+
+        // Retrieve information about the NFT
+        const token_uri = await contract.functions.get_nft_uri(nft_id.result);
+        console.log(token_uri.result);
```
seanyoung, line 86: No check that transfer actually worked

integration/solana/simple_collectible.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+                accounts: [TOKEN_PROGRAM_ID],
+                writableAccounts: [mint, owner.address],
+                signers: [mint_authority]
+            }
+        );
+
+        // Retrieve the owner of an NFT
+        const response = await contract.functions.get_nft_owner(
+            nft_id.result
+        );
```
seanyoung, line 58: no check that the return values make sense

integration/solana/simple_collectible.spec.ts
seanyoung, line 86: no check that that the token_uri is correct

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,78 @@
...
+// Use it at your own risk.
+
+import '../../solana-library/spl_token.sol';
+import '../../solana-library/system_instruction.sol';
+
+contract SimpleCollectible {
+    // Contract variables
+    uint32 private tokenCounter;
+    address private mint_authority;
+    address private payer;
```
seanyoung, line 13: This is example is supposed to educate users how to create a NFT using Solidity, so everything needs commenting. Every variable and also explain solana concepts to Solidity users (maybe with links)

seanyoung (changes requested):

integration/solana/simple_collectible.sol
LucasSte: The payer is the account in Typescript used to derive the associated token account for each owner. It is the signer for the derived account and for each ownership transfer. As the payer account contains funds the belongs to the owner of this contract, I thought it should be the one to sign all transactions too.

integration/solana/simple_collectible.spec.ts
LucasSte: Does it suffices to call `get_nft_owner` to verify this? Is there any chain method to retrieve the owner of a token?

integration/solana/simple_collectible.spec.ts
LucasSte: If there was no error regenerating the public key for them, I assume they do make sense. What do you suggest here?

LucasSte: Oh, I haven't verified if the nft_id exists in the mapping.

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,98 @@
...
+import '../../solana-library/spl_token.sol';
+
+contract SimpleCollectible {
+    // tokenCounter creates a unique ID for each new minted NFT
+    uint32 private tokenCounter;
+    // The authority responsible for managing mints.
+    address private mintAuthority;
+
+    struct NFTOwner {
+        bool exists;
```
xermicus, line 15: What is the purpose of this flag, it looks like there is no code path in this contract to actually set this to false?

LucasSte: If I retrieve a key that is not set in the mapping, `exists` is going to be false. This serves to tell me if this item is in the mapping.

xermicus: I guess it's fine if there is no other way to tell (e.g. without using extra space).

seanyoung: This is an oddity of how Solidity implements mappings. They are not iterable, and any key will give you an result, but if the entry doesn't exist all the fields will be 0/false.

xermicus: My point was more whether the flag is actually needed, since it uses extra storage space. Where as checking other fields _
might do the same job (e.g. asset URI or addresses etc. can't be 0 anyways).

xermicus (changes requested):

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,78 @@
...
+        SplToken.mint_to(data.mintAccount, associatedTokenAccount, mintAuthority, 1);
+        uri = tokenURI;
+        // Log on blockchain records information about the created token
+        emit NFTMinted(data.owner, data.mintAccount);
+    }
+
+    /// Transfer ownership of this NFT from one account to another
+    ///
+    /// @param newOwnerTokenAccount the associated token account for the @param newOwner
+    function transferOwnership(address newOwnerTokenAccount) public {
```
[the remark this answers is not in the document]
LucasSte: The caller does not need to be the owner, however, the owner must sign the transaction with its private key (see in Typescript the `singers` field when I call this function). If the signature key does not match that of the owner, the transaction will not succeed. I'll add a comment about this.

solana-library/spl_token.sol
```diff
@@ -176,4 +176,29 @@ library SplToken {
...
+		address owner;
+		address mintAccount;
+		uint64 balance;
+	}
+
+	/// Fetch the owner, mint account and balance for an associated token account.
+	///
+	/// @param associatedAccount The associated token account
+	/// return struct AssociatedAccountData
+	function get_associated_account_data(address associatedAccount) public view returns (AssociatedAccountData) {
```
joncinque, line 191: nit: This works for any token account, associated or non-associated, so you could call this `get_token_account_data` and `TokenAccountData`

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,80 @@
...
+// DISCLAIMER: This file is an example of how to mint and transfer NFTs on Solana. It is not production ready and has not been audited for security.
+// Use it at your own risk.
+
+import '../../solana-library/spl_token.sol';
+
+contract SimpleCollectible {
+    // On Solana, an associated token account saves the amount of tokens an owner has.
+    // It also holds the mint account and the owner account public keys.
+    // The associated token account and the mint account save all the information necessary to prove the authenticty of an NFT.
+    address private associatedTokenAccount;
```
joncinque, line 12: Rather than associating a token account, you're probably better off associating the mint.  People can transfer the NFT outside of this program, which will break this connection.

On the flip side, that means less functionality in this contract, since you won't have `transferOwnership`, `getNftOwner`, and `isOwner`.  If you want more functionality here, you can have an additional `metadataAuthority` address, which is allowed to update the uri if they sign.  What do you think?

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,80 @@
...
+    /// Create a new NFT and associate it to an URI
+    ///
+    /// @param tokenURI a URI that leads to the NFT resource
+    /// @param mintAuthority an account that signs each new mint
+    function createCollectible(string memory tokenURI, address mintAuthority) public {
+        SplToken.AssociatedAccountData data = SplToken.get_associated_account_data(associatedTokenAccount);
+
+        // An NFT on Solana is a SPL-Token with only one minted token.
+        // The mintAccount saves information about the token (e.g. quantity of tokens avaialble)
+        SplToken.mint_to(data.mintAccount, associatedTokenAccount, mintAuthority, 1);
```
joncinque, line 34: In order to be sure it's non-fungible, you may want to check that the supply of the mint is currently 0 before doing the mint

joncinque (commented):
Thanks for asking me to review this! I've put in a few ideas which I hope will make this more useful in a real-world context as more of a metadata contract.  If you need more background, you can also follow some of the design from the Metaplex metadata program https://docs.metaplex.com/programs/token-metadata/  -- another thing to consider is making the contract address as a PDA of the mint, to ensure only one copy exists.

Anyway, feel free to ask for any more clarification, happy to chat through your ideas!

integration/solana/simple_collectible.sol
LucasSte: That was a great idea. I've added the metadata authority as you suggested.

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,105 @@
...
+        // This assert is not necessary. The transaction will fail if this does not hold true.
+        assert(mintAccount == token_data.mintAccount);
+        SplToken.MintAccountData mint_data = SplToken.get_mint_account_data(token_data.mintAccount);
+        // Ensure the supply is zero. Otherwise, this is not an NFT.
+        assert(mint_data.supply == 0);
+        
+        // An NFT on Solana is a SPL-Token with only one minted token.
+        // The token account saves the owner of the tokens minted with the mint account, the respective mint account and the number
+        // of tokens the owner account owns
+        SplToken.mint_to(token_data.mintAccount, ownerTokenAccount, mintAuthority, 1);
```
joncinque, line 44: This is up to you, but if you want to ensure that it is totally non-fungible, you can set the mint authority to `null` or `None` after this minting, to be sure that no other tokens can ever be created.

Note that most NFTs on Solana still have an active mint authority, so this is not necessary.

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,105 @@
...
+
+        // Ensure we are dealing with the token this contract refers to
+        assert(old_data.mintAccount == mintAccount);
+        assert(new_data.mintAccount == mintAccount);
+
+        // To transfer the ownership of a token, we need the current owner and the new owner. The payer account is the account used to derive
+        // the correspondent token account in TypeScript.
+        SplToken.transfer(oldOwnerTokenAccount, newOwnerTokenAccount, old_data.owner, 1);
+        emit NFTSold(old_data.owner, new_data.owner);
+    }
```
joncinque, line 68: For the purposes of education, you should specify that this is just wrapping the innate SPL transfer, and that the value of this instruction is to emit the `NFTSold` event.

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,105 @@
...
+        SplToken.mint_to(token_data.mintAccount, ownerTokenAccount, mintAuthority, 1);
+        updateNftUri(tokenURI);
+        // Log on blockchain records information about the created token
+        emit NFTMinted(token_data.owner, token_data.mintAccount);
+    }
+
+    /// Transfer ownership of this NFT from one account to another
+    ///
+    /// @param oldOwnerTokenAccount the associated token account for the current owner
+    /// @param newOwnerTokenAccount the associated token account for the new owner
```
joncinque, line 53: nit: Technically speaking, these don't *have* to be associated token accounts. Any token account for the mint will work

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,105 @@
...
+    function transferOwnership(address oldOwnerTokenAccount, address newOwnerTokenAccount) public {
+        // The current owner does not need to be the caller of this functions, but they need to sign the transaction
+        // with their private key.
+        SplToken.TokenAccountData old_data = SplToken.get_token_account_data(oldOwnerTokenAccount);
+        SplToken.TokenAccountData new_data = SplToken.get_token_account_data(newOwnerTokenAccount);
+
+        // Ensure we are dealing with the token this contract refers to
+        assert(old_data.mintAccount == mintAccount);
+        assert(new_data.mintAccount == mintAccount);
```
joncinque, line 62: This is checked in the spl token program, so isn't necessary, but it does show how to deserialize and read token account data, which is very nice :smile:

integration/solana/simple_collectible.sol
```diff
@@ -0,0 +1,105 @@
...
+        updateNftUri(tokenURI);
+        // Log on blockchain records information about the created token
+        emit NFTMinted(token_data.owner, token_data.mintAccount);
+    }
+
+    /// Transfer ownership of this NFT from one account to another
+    ///
+    /// @param oldOwnerTokenAccount the associated token account for the current owner
+    /// @param newOwnerTokenAccount the associated token account for the new owner
+    function transferOwnership(address oldOwnerTokenAccount, address newOwnerTokenAccount) public {
```
joncinque, line 54: nit: if you want to be very clear, you can rename these parameters since the owner may actually be the same
```suggestion
    function transferOwnership(address oldTokenAccount, address newTokenAccount) public {
```

integration/solana/simple_collectible.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+        const mint_authority = Keypair.generate();
+        const freezeAuthority = Keypair.generate();
+
+        // Create and initialize a new mint based on the funding account and a mint authority
+        const mint = await createMint(
+            connection,
+            payer,
+            mint_authority.publicKey,
+            freezeAuthority.publicKey,
+            3
```
joncinque, line 26: By convention, NFTs on solana typically have 0 decimals

solana-library/spl_token.sol
```diff
@@ -176,4 +176,62 @@ library SplToken {
 
 		revert("account missing");
 	}
+
+	// This struct is the return of 'get_token_account_data'
+	struct TokenAccountData {
+		address owner;
+		address mintAccount;
+		uint64 balance;
+	}
```
joncinque, line 185: This is for later, but since you've deserialized everything in the mint, you'll probably want everything from here too.

joncinque (approved):
Great work! Mainly some small questions and nits, but nothing huge

seanyoung (approved):
Looks good!

xermicus (approved):

PR #1003, opened 2022-09-05
Fix overflow detection tests

salaheldinsoliman:
Numbers that overflow the input type was generated, causing some overflow detection tests to fail

### DIFF
--- a/tests/solana_tests/primitives.rs
+++ b/tests/solana_tests/primitives.rs
@@ -3,7 +3,7 @@
 use crate::build_solidity;
 use crate::build_solidity_with_overflow_check;
 use ethabi::ethereum_types::U256;
-use num_bigint::{BigInt, BigUint, RandBigInt, Sign};
+use num_bigint::{BigInt, BigUint, RandBigInt};
 use rand::seq::SliceRandom;
 use rand::Rng;
 use std::ops::Add;
--- a/tests/solana_tests/primitives.rs
+++ b/tests/solana_tests/primitives.rs
@@ -843,10 +843,7 @@ fn test_mul_within_range() {
             let limit = BigUint::from(2_u32).pow(width).sub(1_u32);
 
             // Generate a random number within the the range [0, 2^N -1]
-            let mut first_operand_rand = rng.gen_biguint(width.into());
-            if first_operand_rand == BigUint::from(0_u32) {
-                first_operand_rand = BigUint::from(1_u32);
-            }
+            let first_operand_rand = rng.gen_biguint_range(&BigUint::from(1usize), &limit);
 
             // Calculate a number that when multiplied by first_operand_rand, the result will not overflow N bits (the result of this division will cast the float result to int result, therefore lowering it. The result of multiplication will never overflow).
             let second_operand_rand = limit.div(&first_operand_rand);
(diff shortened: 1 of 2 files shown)


tests/solana_tests/primitives.rs
```diff
@@ -941,10 +948,14 @@ fn test_overflow_detect_unsigned() {
             let limit = BigUint::from(2_u32).pow(width);
 
             // Generate a random number within the the range [0, 2^N -1]
-            let first_operand_rand = rng.gen_biguint(width.into()).add(1_u32);
+            let mut first_operand_rand = rng.gen_biguint(width.into());
+
+            if first_operand_rand == BigUint::from(0_u32) {
+                first_operand_rand = BigUint::from(1_u32);
+            }
```
seanyoung, line 955: What number multiplied by 1 overflows?

tests/solana_tests/primitives.rs
```diff
@@ -941,10 +948,14 @@ fn test_overflow_detect_unsigned() {
...
 
             // Calculate a number that when multiplied by first_operand_rand, the result will overflow N bits
-            let second_operand_rand = limit.div(&first_operand_rand).add(1_usize);
+            let second_operand_rand = limit.div(&first_operand_rand).add(2_usize);
```
seanyoung, line 958: The problem you have here is  `limit / first_operand` is an integer division which will round down. In other words, if limit is 256 and first_operand is 162, then `256 / 162 = 1` and 1 * 162 does not overflow 256.

I don't know where 2 comes from. This needs a big fat comment in *any case*.

tests/solana_tests/primitives.rs
seanyoung: presumably you want:

```rust
let first_operand_rand = rng.gen_biguint_range(&BigUint::from(2usize), &limit);
```

tests/solana_tests/primitives.rs
```diff
@@ -843,7 +843,10 @@ fn test_mul_within_range() {
             let limit = BigUint::from(2_u32).pow(width).sub(1_u32);
 
             // Generate a random number within the the range [0, 2^N -1]
-            let first_operand_rand = rng.gen_biguint(width.into()).add(1_u32);
+            let mut first_operand_rand = rng.gen_biguint(width.into());
+            if first_operand_rand == BigUint::from(0_u32) {
+                first_operand_rand = BigUint::from(1_u32);
+            }
```
seanyoung, line 849:
```rust
let first_operand_rand = rng.gen_biguint_range(&BigUint::from(1usize), &limit);
```

seanyoung (changes requested):

tests/solana_tests/primitives.rs
salaheldinsoliman: Limit is then divided by this. 
second_op_rand = (limit / 1) + 2 
second_op * first_op will indeed overflow

tests/solana_tests/primitives.rs
salaheldinsoliman: Basically, making the second operand larger will make the result overflow

seanyoung (approved):
Looks good!

tests/solana_tests/primitives.rs
```diff
@@ -884,22 +884,15 @@ fn test_overflow_detect_signed() {
 
         contract.constructor("test", &[]);
 
-        // The range of values that can be held in signed N bits is [-2^(N-1), 2^(N-1)-1] .Generate a value that will overflow this range:
-        let limit = BigInt::from(2_u32).pow(width - 1).add(1_u32);
+        // The range of values that can be held in signed N bits is [-2^(N-1), 2^(N-1)-1] .
+        let limit = BigInt::from(2_u32).pow(width - 1).sub(1usize);
```
LucasSte, line 888: The limit will never be `-2^(N-1)`. Are you testing this?

LucasSte (approved):

seanyoung (approved):

xermicus (commented):
As a side node, you are using a RNG in the tests. I think it is more common to use a fixed seed to get better reproducibility, or is there a specific reason not to?

seanyoung:
> As a side node, you are using a RNG in the tests. I think it is more common to use a fixed seed to get better reproducibility, or is there a specific reason not to?

A RNG is used in many tests, @salaheldinsoliman is just following the existing scheme. Having said that, @LucasSte also made the point that he would like the tests to be deterministic.

I can see both sides - with RNG source your coverage is much greater, but with the downside of random failures if there is a bug. We should discus making all the tests deterministic.

PR #1006, opened 2022-09-07
Substrate constructor overloading names

xermicus:


xermicus:
I discussed this with @seanyoung today, since implementing this might be more involved than I initially thought. We concluded the following:

Complexity stems mostly from having named arguments and having the ability to overload constructors together with the ability to name them. In ETH Solidity, neither overloading nor naming constructors is possible and it would require some more careful design decisions and possibly require some refactoring within affected parts in solang.

To make progress towards having a fully functional and up-to date Substrate target again as soon as possible, we decided to just
disallow constructor overloading for the substrate target for now. If a constructor then has no name, we can simply name it `new`. Since constructor overloading is not at all possible in ETH solidity anyways this solution should make for a fair compromise. Additionally it can still be fixed in the future.

src/sema/functions.rs
```diff
@@ -58,7 +58,7 @@ pub fn contract_function(
                 ));
                 return None;
             }
-            if func.name.is_some() {
+            if func.name.is_some() && !ns.target.is_substrate() {
```
seanyoung, line 61: This still allows constructors to have names on Substrate, without the syntax for actually using constructor names. I'm not keen on that, this is incomplete support.

src/sema/functions.rs
```diff
@@ -552,8 +552,11 @@ pub fn contract_function(
     fdecl.selector = selector;
 
     if func.ty == pt::FunctionTy::Constructor {
-        // In the eth solidity, only one constructor is allowed
-        if ns.target == Target::EVM {
+        // In the eth solidity and Substrate, only one constructor is allowed
+        //
+        // The Substrate runtime supports multiple constructors, but they also need to have a name.
+        // Currently, we only support either naming the constructors or overloading them, but not both.
```
seanyoung, line 558: I am confused. Overloading constructors requires multiple constructors. Are we allowing multiple constructors?

src/sema/functions.rs
```diff
@@ -569,24 +572,24 @@ pub fn contract_function(
                 ));
                 return None;
             }
-        } else {
-            let payable = fdecl.is_payable();
+        }
```
seanyoung, line 575: Why is this moved out of the else branch of the if statement? Checking that all constructors are payable/not payable only applies if there are mulitple constructors

seanyoung (changes requested):
I suspect there is mention in the documentation of multiple constructors for Substrate

src/sema/functions.rs
xermicus: >  this is incomplete support

It is by design so we can later use the name in the metadata.

src/sema/functions.rs
xermicus: You are right it is a bit confusing. What I meant to say here is that the solang compiler itself does not support the combination of both.

src/sema/functions.rs
xermicus: Yeah I can't remember why I did that, must be a mistake. Thanks!

docs/language/contracts.rst
```diff
@@ -18,19 +18,11 @@ like so:
       }
   }
 
-A constructor does not have a name and may have any number of arguments. If a constructor has arguments,
-then when the contract is deployed then those arguments must be supplied.
+In Substrate, constructor can optionally have a name and may have any number of arguments.
+If a constructor has arguments, then when the contract is deployed then those arguments must be supplied.
```
seanyoung, line 22: Your version reads like arguments are only permitted on Substrate

```suggestion
A constructor can have any number of arguments. In Substrate, it may have a name. The constructor
name is only used for the metadata, nothing further.
If a constructor has arguments, then when the contract is deployed then those arguments must be supplied.
```

seanyoung (changes requested):
Please can have some tests that check that constructor names cannot be the same as function names

docs/language/contracts.rst
LucasSte, line 22:
```suggestion
If a constructor has arguments, those arguments must be supplied when the contract is deployed.
```
The repeated usage of `then` made this sentence awkward.

LucasSte (changes requested):

docs/language/contracts.rst
xermicus: > Please can have some tests that check that constructor names cannot be the same as function names

Hmm I'm not exactly sure where to implement this check, most likely in `src/sema/contracts.rs` ?

xermicus:
This is getting awkward with the history. I squashed and opened #1035 instead

PR #1008, opened 2022-09-08
Emit events using Borsh encoding for Solana

LucasSte:
This PR implements the new scheme for events emission on Solana. It follows Anchor strategy for events.

[According to Anchor source code](https://github.com/coral-xyz/anchor/blob/0c70d183ef9187cef576749e4fee8f443e4dbc34/lang/attribute/event/src/lib.rs#L86-L88), `indexed` fields are emitted as data. In addition, all fields are encoded using borsh encoding, and [the result is passed directly](https://github.com/coral-xyz/anchor/blob/0c70d183ef9187cef576749e4fee8f443e4dbc34/lang/attribute/event/src/lib.rs#L77-L84) to `sol_log_data`.

### DIFF
--- a/docs/language/events.rst
+++ b/docs/language/events.rst
@@ -21,6 +21,13 @@ not. ``indexed`` fields are stored as topics, so there can only be a limited num
 fields are stored in the data section of the event. The event name does not need to be unique; just like
 functions, they can be overloaded as long as the fields are of different types, or the event has
 a different number of arguments.
+
+.. warning::
+    On Solana, writing ``indexed`` besides an event field has no impact when emitting events. The ``indexed``
+    keyword serves only to generate metadata information for IDL files. All event attributes will be encoded as data to
+    be passed for Solana's ``sol_log_data`` system call, regardless of the ``indexed`` keyword being present. This
+    behavior follows what Solana's Anchor framework does.
+
 In Parity Substrate, the topic fields are always the hash of the value of the field. Ethereum only hashes fields
 which do not fit in the 32 bytes. Since a cryptographic hash is used, it is only possible to compare the topic against a
 known value.
--- a/docs/language/events.rst
+++ b/docs/language/events.rst
@@ -23,10 +23,9 @@ functions, they can be overloaded as long as the fields are of different types, 
 a different number of arguments.
 
 .. warning::
-    On Solana, writing ``indexed`` besides an event field has no impact when emitting events. The ``indexed``
-    keyword serves only to generate metadata information for IDL files. All event attributes will be encoded as data to
-    be passed for Solana's ``sol_log_data`` system call, regardless of the ``indexed`` keyword being present. This
-    behavior follows what Solana's Anchor framework does.
+    On Solana, the ``indexed`` event field attribute has no effect. All event attributes will be encoded as data to
+    be passed for Solana's ``sol_log_data`` system call, regardless if the ``indexed`` keyword is present or not.
+    This behavior follows what Solana's Anchor framework does.
 
 In Parity Substrate, the topic fields are always the hash of the value of the field. Ethereum only hashes fields
 which do not fit in the 32 bytes. Since a cryptographic hash is used, it is only possible to compare the topic against a
--- a/src/codegen/events/mod.rs
+++ b/src/codegen/events/mod.rs
@@ -14,7 +14,9 @@ use solang_parser::pt;
 /// This traits delineates the common behavior of event emission. As each target uses a different
 /// encoding scheme, there must be an implementation of this trait for each.
 pub(super) trait EventEmitter {
-    /// Generate the CFG instructions for emitting an event
+    /// Generate the CFG instructions for emitting an event.
+    /// All necessary code analysis should have been done during parsing and 'sema';
+    /// If code generation does not work here, there is a bug in the compiler.
     fn emit(
         &self,
         contract_no: usize,
--- a/src/codegen/events/mod.rs
+++ b/src/codegen/events/mod.rs
@@ -1,0 +1,47 @@
+mod solana;
+mod substrate;
+
+use crate::codegen::cfg::ControlFlowGraph;
+use crate::codegen::events::solana::SolanaEventEmitter;
+use crate::codegen::events::substrate::SubstrateEventEmitter;
+use crate::codegen::vartable::Vartable;
+use crate::codegen::Options;
+use crate::sema::ast;
+use crate::sema::ast::{Function, Namespace};
+use crate::Target;
+use solang_parser::pt;
+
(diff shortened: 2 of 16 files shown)


LucasSte:
I forgot to review Solana's integration tests!

LucasSte:
> I forgot to review Solana's integration tests!

Fixed!
@seanyoung and @xermicus The PR is ready for review now.

src/codegen/events/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::sema::ast;
+use crate::sema::ast::{Function, Namespace};
+use crate::Target;
+use solang_parser::pt;
+
+/// This traits delineates the common behavior of event emission. As each target uses a different
+/// encoding scheme, there must be an implementation of this trait for each.
+pub(super) trait EventEmitter {
+    /// Generate the CFG instructions for emitting an event
+    fn emit(
```
xermicus, line 18: Did you consider to return a `Result` instead? Unless there is absolutely no potential errors here for (which I doubt)?

xermicus (changes requested):

src/codegen/events/mod.rs
LucasSte: All the necessary code analysis is dune during the parsing and semantic analysis. If the compiler enters code generation, we assume it is possible to generate the CFG for the AST we've been given. There is no need to return an error here.

src/codegen/events/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::codegen::Options;
+use crate::sema::ast;
+use crate::sema::ast::{Function, Namespace};
+use crate::Target;
+use solang_parser::pt;
+
+/// This traits delineates the common behavior of event emission. As each target uses a different
+/// encoding scheme, there must be an implementation of this trait for each.
+pub(super) trait EventEmitter {
+    /// Generate the CFG instructions for emitting an event
```
xermicus, line 17:
```suggestion
    /// Generate the CFG instructions for emitting an event.
    /// At this point, all necessary code analysis should have been done during parsing and `sema`.
    /// It is considered to be a bug if it is not possible to generate the code here.
```

xermicus (changes requested):

src/codegen/events/mod.rs
xermicus: Alright, I added a comment to point that out. With that I think it makes for a reasonable interface.

src/codegen/events/mod.rs
LucasSte: This comment is valid for everything under code generation (aka `codegen`). I don't think it is necessary to say it here, but perhaps it belongs somewhere else.

LucasSte: Here maybe?
https://github.com/hyperledger/solang/blob/6134fefa43c030beae69ea8e4ba1580d3a5cc80e/src/codegen/mod.rs#L106-L111

docs/language/events.rst
```diff
@@ -21,6 +21,13 @@ not. ``indexed`` fields are stored as topics, so there can only be a limited num
 fields are stored in the data section of the event. The event name does not need to be unique; just like
 functions, they can be overloaded as long as the fields are of different types, or the event has
 a different number of arguments.
+
+.. warning::
+    On Solana, writing ``indexed`` besides an event field has no impact when emitting events. The ``indexed``
+    keyword serves only to generate metadata information for IDL files. All event attributes will be encoded as data to
+    be passed for Solana's ``sol_log_data`` system call, regardless of the ``indexed`` keyword being present. This
+    behavior follows what Solana's Anchor framework does.
```
seanyoung, line 29:
```suggestion
    On Solana, the ``indexed`` event field attribute has no effect. All event attributes will be encoded as data to
    be passed for Solana's ``sol_log_data`` system call, regardless if the ``indexed`` keyword is present or not.
    This behavior follows what Solana's Anchor framework does.
```

seanyoung (changes requested):

seanyoung (approved):
Looks good. I do wonder why you created a new trait, rather than using the trait `TargetRuntime`.

LucasSte:
> Looks good. I do wonder why you created a new trait, rather than using the trait `TargetRuntime`.

`TargetRuntime` is in emit. We need this for codegen.

src/codegen/events/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// Create a new event emitter based on the target blockchain
+pub(super) fn new_event_emitter<'a>(
+    loc: &pt::Loc,
+    event_no: usize,
+    args: &'a [ast::Expression],
+    ns: &'a Namespace,
+) -> Box<dyn EventEmitter + 'a> {
+    match ns.target {
+        Target::Substrate { .. } | Target::EVM => {
```
xermicus, line 38: Would it work to change to a `todo!()` match arm for EVM?

src/codegen/events/solana.rs
```diff
@@ -0,0 +1,63 @@
...
+impl EventEmitter for SolanaEventEmitter<'_> {
+    fn emit(
+        &self,
+        contract_no: usize,
+        func: &Function,
+        cfg: &mut ControlFlowGraph,
+        vartab: &mut Vartable,
+        opt: &Options,
+    ) {
+        let discriminator_image = format!("event:{}", self.ns.events[self.event_no].name);
```
xermicus, line 31: In the tests you have a dedicated function for this. Instead of having that code duplicated, it could be refactored into this module and then use it for the tests as well?

xermicus (commented):
Just some minor things

src/codegen/events/mod.rs
LucasSte: No. Code generation for EVM still exists and we have tests for it. Only emit for EVM is missing.

src/codegen/events/solana.rs
LucasSte: If I add here and make it available for test, it would need to be `pub` so that we can access it anywhere. This function should be private to this module.

src/emit/solana.rs
```diff
@@ -3602,21 +3602,18 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
         function: FunctionValue<'b>,
         _event_no: usize,
```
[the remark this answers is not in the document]
LucasSte: I can exchange them by an underscore. I cannot remove them because the trait dictates the arguments. Keep in mind that there is a separate implementation of this for Substrate.

src/codegen/events/solana.rs
xermicus: Maybe it's a use case for `pub(crate)` visibility then?

xermicus (approved):

PR #1010, opened 2022-09-09
constant arthimeteic overflow

salaheldinsoliman:
Check for constant arthimetic overflow during semantic analysis.

### DIFF
--- a/src/sema/eval.rs
+++ b/src/sema/eval.rs
@@ -10,6 +10,11 @@ use num_traits::Zero;
 use super::ast::{Diagnostic, Expression, Namespace};
 use solang_parser::pt;
 use solang_parser::pt::CodeLocation;
+use std::ops::Shl;
+use std::ops::{Add, Mul, Sub};
+
+use crate::sema::ast::Type;
+use solang_parser::pt::Loc;
 
 /// Resolve an expression where a compile-time constant is expected
 pub fn eval_const_number(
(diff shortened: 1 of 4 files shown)


src/sema/eval.rs
```diff
@@ -196,3 +201,160 @@ pub fn eval_const_rational(
         )),
     }
 }
+
+pub fn eval_constants_in_expression(expr: &Expression, ns: &mut Namespace) -> Expression {
```
LucasSte, line 205: Please, add a doc comment to this function.

src/sema/eval.rs
```diff
@@ -196,3 +201,160 @@ pub fn eval_const_rational(
...
+            if let (Expression::NumberLiteral(_, _, left), Expression::NumberLiteral(_, _, right)) =
+                (&left, &right)
+            {
+                overflow_check(ns, left.shl(right.to_u32().unwrap()), ty.clone(), *loc);
+                bigint_to_expression(loc, ty, left.shl(right.to_u32().unwrap()))
+            } else {
+                Expression::ShiftLeft(*loc, ty.clone(), Box::new(left), Box::new(right))
+            }
+        }
+        _ => expr.clone(),
```
LucasSte, line 299: You should also be evaluating Modulo, ShiftRight, Trunc, SignExt, ZeroExt, Cast. These can also take part in arithmetic expressions and can be evaluated during compile time.

@seanyoung Should Bitwise and Boolean operations be here too?

src/sema/eval.rs
```diff
@@ -196,3 +201,160 @@ pub fn eval_const_rational(
...
+    }
+}
+
+fn bigint_to_expression(loc: &Loc, ty: &Type, n: BigInt) -> Expression {
```
LucasSte, line 303: Add a doc comment to this function.

src/sema/eval.rs
```diff
@@ -196,3 +201,160 @@ pub fn eval_const_rational(
...
+    match expr {
+        Expression::Add(loc, ty, unchecked, left, right) => {
+            let left = eval_constants_in_expression(left, ns);
+            let right = eval_constants_in_expression(right, ns);
+
+            if let (Expression::NumberLiteral(_, _, left), Expression::NumberLiteral(_, _, right)) =
+                (&left, &right)
+            {
+                overflow_check(ns, left.add(right), ty.clone(), *loc);
+                bigint_to_expression(loc, ty, left.add(right))
```
LucasSte, line 215: These two functions can be merged into one.

src/sema/eval.rs
```diff
@@ -196,3 +201,160 @@ pub fn eval_const_rational(
...
+            }
+        }
+        Type::StorageRef(..) => n,
+        _ => unreachable!(),
+    };
+
+    Expression::NumberLiteral(*loc, ty.clone(), n)
+}
+
+fn overflow_check(ns: &mut Namespace, result: BigInt, ty: Type, loc: Loc) {
```
LucasSte, line 332: I think you can merge `overflow_check` with `bigint_to_expression`.

src/sema/statements.rs
```diff
@@ -750,28 +754,60 @@ fn statement(
...
+                        diagnostics,
+                        ResolveTo::Discard,
+                    )?;
+
+                    if let Expression::InternalFunctionCall {
+                        loc: _,
+                        returns: _,
+                        function: _,
+                        args,
+                    } = ret.clone()
```
LucasSte, line 775: Please, don't clone `ret`.

src/sema/statements.rs
```diff
@@ -750,28 +754,60 @@ fn statement(
...
```
LucasSte, line 802: Don't clone `ret`.

LucasSte: Is `InternalFunctionCall` the only possibility here?

src/sema/tests/mod.rs
```diff
@@ -5,7 +5,18 @@ use crate::sema::ast::{Expression, Parameter, Statement, TryCatch, Type};
 use crate::sema::diagnostics::Diagnostics;
 use crate::sema::expression::unescape;
 use crate::sema::yul::ast::InlineAssembly;
+use crate::{parse_and_resolve, sema::ast, FileResolver, Target};
 use solang_parser::pt::Loc;
+use std::ffi::OsStr;
+
+pub(crate) fn parse(src: &'static str) -> ast::Namespace {
```
LucasSte, line 12:
```suggestion
pub(super) fn parse(src: &'static str) -> ast::Namespace {
```

LucasSte: There is no need to be public in the entire crate.

LucasSte (changes requested):

LucasSte:
@seanyoung I think @salaheldinsoliman's work can be simplified if his `eval_constants_in_expression` can be called whenever we have an expression. However, this would imply knowing when not to evaluate the same constants all over again.

@salaheldinsoliman and @seanyoung Do you think we can come up with a strategy for this?

src/sema/eval.rs
salaheldinsoliman: Since the result is thrown away, I'm only evaluating operations that can result in overflow. I think evaluating the mentioned expressions will not be needed here

src/sema/eval.rs
```diff
@@ -196,3 +201,160 @@ pub fn eval_const_rational(
...
+            ));
+        }
+    }
+
+    if let Type::Int(bits) = ty {
+        // If number of bits is more than what the type can hold. BigInt.bits() is not used here since it disregards the sign.
+        if result.to_signed_bytes_be().len() * 8 > (bits as usize) {
+            ns.diagnostics.push(Diagnostic::error(
+                loc,
+                format!("Type int_const {:?} is not implicitly convertible to expected type {:?}. Literal is too large to fit in {:?}.",result,ty,ty),
```
seanyoung, line 356: `"value {} does not fit into type {}"`

Is tested with `int8 x = -128;` or `int16 y = -32768;`

src/sema/eval.rs
LucasSte: > Since the result is thrown away, I'm only evaluating operations that can result in overflow. I think evaluating the mentioned expressions will not be needed here

`int8 c = 255 + (2 >> 1)`. This has shift right and is overflowing. You can find examples of this for all the expressions I mentioned.

LucasSte: Devide is missing too. Just because an expression can be classified as a decrescent function, it does not meant the overall expression it belongs to won't overflow.

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,68 @@ fn test_statement_reachable() {
         assert_eq!(test_case.reachable(), expected);
     }
 }
+
+#[test]
+fn constant_overflow_checks() {
+    let file = r#"
+    contract test_contract {
```
LucasSte, line 146: You are only testing 8-bit integers. Your tests should include other types too.

PR #1011, opened 2022-09-12
Fix function calls via contract name

### LINKED ISSUE #1007: Inconsistent behaviour when calling function via contract type name
Given the following base contract:

```solidity
contract Base {
    uint256 public something = 0;

    function nop() public pure {}

    function set(uint256 val) public {
        something = val;
    }
}
```
And the following case `A`:
```solidity
contract A {
    function a() public pure {
        Base.nop();
    }
}
```
Solang compiles this just fine. Furthermore, consider the following case `B`:
```solidity
contract B {
    function b() public {
        Base.set(1);
    }
}
```
Solang crashes during compilation (target either `substrate` or `solana`):
```
thread 'main' panicked at 'get_storage_slot called on non-storage variable', src/codegen/cfg.rs:1879:13
```
Where as `solc` behaves more in a way I'd expect. For both cases it gets the same error:
```
❯ solc --bin foo.sol
Error: Cannot call function via contract type name.
  --> foo.sol:27:9:
   |
17 |         Base.nop();
   |         ^^^^^^^^^^

Error: Cannot call function via contract type name.
  --> foo.sol:33:9:
   |
23 |         Base.set(1);
   |         ^^^^^^^^^^^
```
Is there a reason we allow case `A`? If so, we should at least restrict it to `pure` functions. But this looks more like a bug to me.

xermicus:
Disallow direct access to functions (and constants) except for base contracts. This is not valid in `solc` but was allowed in solang.

Fixes #1007


src/sema/expression.rs
```diff
@@ -4522,6 +4522,21 @@ fn member_access(
...
+                    context.contract_no,
+                ) {
+                    if !is_base(base, contract, ns) && !ns.contracts[base].is_library() {
+                        diagnostics
+                            .push(Diagnostic::error(e.loc(), "member access for constant values via contract name is only valid for base contracts".into()));
+                        return Err(());
+                    }
+                }
+            }
+        }
```
seanyoung, line 4539: I am not sure about this, for two reasons.

First of all, the code is not right because:
 - any restrictions should be handled in `fn contract_constant()`
 - contracts can be imported via an import object, in which case `e` can refer to `a.b.c`, i.e. `pt::Expression::Namespace("a", pt::Expression::Variable("b"))`. The code does not handle this case

The other reason is - why is it not permitted to use a constant in a contract? I know that solc does not permit it, but I'm not really sure why they done this.

src/sema/expression.rs
```diff
@@ -5852,6 +5867,11 @@ fn method_call_pos_args(
                         symtable,
                         diagnostics,
                     );
+                } else {
+                    diagnostics.push(Diagnostic::error(
+                        *loc,
+                        "function calls via contract name are only valid for base contracts".into(),
+                    ));
```
seanyoung, line 5874: The same change is needed in `fn method_call_named_args()` and this needs a test case too.

```solidity
contract B {
    function b() public {
        Base.set({val: 1});
    }
}
```

tests/solana_tests/constant.rs
```diff
@@ -24,8 +24,12 @@ fn constant() {
 
     let returns = vm.function("f", &[], &[], None);
     assert_eq!(returns, vec![Token::Uint(U256::from(42))]);
+}
 
-    let mut vm = build_solidity(
+#[test]
+#[should_panic(expected = "called `Option::unwrap()` on a `None` value")]
+fn constant_in_contract() {
+    build_solidity(
```
seanyoung, line 32: Please use `fn parse_and_resolve()` for sema tests, and do an assert on the error.

seanyoung (changes requested):
Principle looks right, just needs minor changes. Thanks!

seanyoung (approved):
Looks great, thanks!

LucasSte (approved):

PR #1015, opened 2022-09-15
Refactor emit

LucasSte:
This PR refactors the emit code. The largest problem there was that `emit/mod.rs` has about 10 thousand lines, making it difficult to navigate. Here is a summary of my changes:

1. The trait `TargetRuntime` only defines an interface for behaviors that are not shared between targets. All shared implementations are now standalone functions.
2. I created a `StorageSlot` to be an interface for storage management functions for environments that depend on storage slots, like Substrate and EVM.
3. Target specific code is separated into folders. Code emission for Solana is under `emit/solana` and for Substrate, under `emit/substrate`.
4. The implementation of the `TargetRuntime` trait is located in `emit/<target>/target.rs`.


LucasSte:
@seanyoung and @xermicus please, have a careful look. I might have messed up with something.

src/emit/storage.rs
```diff
@@ -0,0 +1,52 @@
...
+        &self,
+        binary: &Binary<'a>,
+        slot: PointerValue<'a>,
+        ns: &Namespace,
+    ) -> ArrayValue<'a>;
+
+    /// Clear a particlar storage slot (slot-based storage chains should implement)
+    fn storage_delete_single_slot<'a>(&self, binary: &Binary<'a>, slot: PointerValue);
+
+    /// Recursively load a type from bin storage for slot based bin storage
```
seanyoung, line 19: `bin storage` -> `storage`  (everywhere)
I don't know what `bin storage` is

seanyoung (changes requested):
Layout is much better. I've quickly read through it and can't see any problems

xermicus (approved):

seanyoung (approved):

PR #1018, opened 2022-09-19
Fix AddMod and MulMod

LucasSte:
This PR fixes three bugs in the implementation of AddMod and MulMod.

1. In Yul, `addmod` and `mulmod` were not implemented correctly, because 512-bit operations were not in use. To fix this, those builtins now call `Builtin::AddMod` and `Builtin::MulMod`.
2. `Builtin::AddMod` was calculating `(a+b)/c` instead of `(a+b)%c`.
3. `Builtin::MulMod` was calcuating `(a*b)/c` instead of `(a+b)%c`.


seanyoung (changes requested):

tests/substrate_tests/builtins.rs
```diff
@@ -491,7 +491,7 @@ fn addmod() {
         r##"
         contract x {
             function test() public {
-                assert(addmod(500, 100, 0) == 200);
+                assert(addmod(500, 100, 0) == 600);
```
xermicus, line 494: When I run this in remix IDE I get `0`, maybe we should do that too instead? (I have no hard opinion here)

xermicus (changes requested):

tests/substrate_tests/builtins.rs
LucasSte: This function fails. It does not return anything. Instead, an error is thrown by the environment.

LucasSte: It returns zero now, just like solc.

xermicus (approved):

Issue #1019, opened 2022-09-20
LLVM ERROR: Unsupported dynamic stack allocation on create_program_address with Solang 0.1.13


stanta:
Got this  error on Solang 0.1.13, when tried to call

`address pda =  create_program_address (["TokenAccountXXX", "seed"], _token );`

This LLVM Error fires when compiling of file finishing.

seanyoung:
@stanta can you give the full solidity source code please so we can reproduce the issue? Thanks!

stanta:
@seanyoung exactly as pasted above. 
_token receives as constructor's argument from `  publicKeyToHex (mintPubkey), ` where
`mintPubkey =  spl.createMint (...) `,
createMint is  from   @solana/spl-token JSlib

seanyoung:
Whole solidity source code please

stanta:
Under NDA, sorry. Can I send to  private address?

stanta:
Under NDA, sorry. Can I send to  private address?

stanta:
Hi, @seanyoung! 
I sent file to sean [at] mess [dot] org as in your profile.

stanta:
@seanyoung, Thank you for fast fixing! Where can I pull fixed version of Solang compiler?

seanyoung:
You can use the latest container image:
```
docker pull ghcr.io/hyperledger/solang:latest
```
Or you can download a binary from the latest CI on the main https://github.com/hyperledger/solang/actions/runs/3121674351
These are debug builds so they're larger and slower.

stanta:
@seanyoung compile error gone, thank you.
But systematically get error "0x1" from smart contract when call this: 
 `stableTokenAccount =  create_program_address (["stableTokenAccount", _seedstableTokenAccount], stableProgramID );`

seanyoung:
> @seanyoung compile error gone, thank you. But systematically get error "0x1" from smart contract when call this: ` address stableTokenAccount = create_program_address (["stableTokenAccount", _seedstableTokenAccount], stableProgramID );`

Return value `0x01` indicates that the `sol_create_program_address()` system call returned an error. This can be for one of the following reasons:
- More than 16 seeds (you only have 2)
- One of the seeds is longer than 32 bytes (is `_seedstableTokenAccount` longer than that?)
- The calculated address falls on the ed25519 curve

See https://docs.solana.com/developing/programming-model/calling-between-programs#hash-based-generated-program-addresses

[...]

PR #1021, opened 2022-09-21
On Solana, stack allocations must happen in the first block

seanyoung:
If not, the Solana llvm target will fail to generate code with the error:

Unsupported dynamic stack allocation on create_program_address

Fixes https://github.com/hyperledger/solang/issues/1019

### DIFF
--- a/src/emit/instructions.rs
+++ b/src/emit/instructions.rs
@@ -580,10 +580,10 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
                 .map(|p| expression(target, bin, p, &w.vars, function, ns).into())
                 .collect::<Vec<BasicMetadataValueEnum>>();
 
-            let func = &ns.functions[*ast_func_no];
+            let callee = &ns.functions[*ast_func_no];
 
             if !res.is_empty() {
-                for v in func.returns.iter() {
+                for v in callee.returns.iter() {
                     parms.push(if ns.target == Target::Solana {
                         bin.build_alloca(function, bin.llvm_var_ty(&v.ty, ns), v.name_as_str())
                             .into()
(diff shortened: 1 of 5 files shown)


src/emit/instructions.rs
```diff
@@ -595,7 +595,7 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
                 }
             }
 
-            let ret = target.builtin_function(bin, func, &parms, ns);
+            let ret = target.builtin_function(bin, function, func, &parms, ns);
```
LucasSte, line 598: I believe the bad naming here might have led to a confusion. What you think of giving better names to `func` and `function`?

tests/solana_tests/builtin.rs
```diff
@@ -101,17 +105,21 @@ fn pda() {
                 return create_program_address([a, b], program_id);
             }
 
-            function create_pda2_bump() public returns (address, bytes1) {
+            function create_pda2_bump(bool cond) public returns (address, bytes1) {
```
LucasSte, line 108: What do you want to achieve with this boolean here?
All your tests set this argument as `true`.

LucasSte (changes requested):

tests/solana_tests/builtin.rs
seanyoung: This is to ensure that `create_program_address()` is not in the first block of the cfg; that is the problem case.

src/emit/instructions.rs
seanyoung: I've renamed them

xermicus (approved):

tests/solana_tests/builtin.rs
LucasSte: I think we should still be testing `create_program_address` as the first instruction in a block.
What about this?

```solidity
function create_pda(bool cond) public returns (address) {
                 address program_id = address"BPFLoaderUpgradeab1e11111111111111111111111";
                 address addr =  create_program_address(["Talking", "Cats"], program_id);
                 if (cond) {
                     return create_program_address(["Talking", "Squirrels"], program_id);
                 } else {
                     return addr;
                 }
             }
```

seanyoung: I've added this test

LucasSte (approved):

PR #1022, opened 2022-09-21
Substrate: Remove seal_tombstone_deposit

xermicus:
This API call does no longer exist in `ink!`

It looks like I changed the tests but that is only because of git is trying to be smart. I simply deleted the test for this API call and renamed the other ones to fill the gap.


seanyoung (approved):
LGTM

LucasSte (approved):

PR #1023, opened 2022-09-22
Create switch instruction

LucasSte:
We aim at moving the dispatch code to codegen. It means we'll need new instructions and builtin structs there, like the `switch` instruction, utilized in [this line](https://github.com/hyperledger/solang/blob/6083938df34d288ba85a19d72338ee5abcb85b85/src/emit/dispatch.rs#L103).

This PR wires up the `switch` instruction for that and enables the code generation for switch statements in Yul, using the newly added instruction.

### DIFF
--- a/src/codegen/yul/statements.rs
+++ b/src/codegen/yul/statements.rs
@@ -8,7 +8,7 @@ use crate::codegen::yul::expression::{expression, process_function_call};
 use crate::codegen::{Expression, Options};
 use crate::sema::ast::{Namespace, RetrieveType, Type};
 use crate::sema::yul::ast;
-use crate::sema::yul::ast::{YulStatement, YulSuffix};
+use crate::sema::yul::ast::{CaseBlock, YulBlock, YulExpression, YulStatement, YulSuffix};
 use num_bigint::BigInt;
 use num_traits::FromPrimitive;
 use solang_parser::pt;
@@ -67,10 +67,23 @@ pub(crate) fn statement(
             opt,
         ),
 
-        YulStatement::Switch { .. } => {
-            // Switch statements should use LLVM switch instruction, which requires changes in emit.
-            unreachable!("Switch statements for yul are not implemented yet");
-        }
+        YulStatement::Switch {
+            condition,
+            cases,
+            default,
+            ..
+        } => resolve_switch(
+            condition,
+            cases,
+            default,
+            loops,
+            contract_no,
+            ns,
+            vartab,
+            cfg,
+            early_return,
+            opt,
+        ),
 
         YulStatement::For {
             loc,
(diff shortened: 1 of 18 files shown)


xermicus (approved):
LGTM, although it could use some more tests (e.g. no cases and no default at all, multiple default cases, duplicate cases, etc...)

LucasSte:
> LGTM, although it could use some more tests (e.g. no cases and no default at all, multiple default cases, duplicate cases, etc...)

1. `no cases and no default at all,` The parser does not allow that. You can have only no cases and a default block and cases without a default. There is a test missing for the former. I'm going to add that.
2. `multiple default cases` The parser does not accept that.
3. `duplicate cases`. This is missing. Good point

I'll add those tests!

LucasSte:
> > LGTM, although it could use some more tests (e.g. no cases and no default at all, multiple default cases, duplicate cases, etc...)
> 
> 1. `no cases and no default at all,` The parser does not allow that. You can have only no cases and a default block and cases without a default. There is a test missing for the former. I'm going to add that.
> 2. `multiple default cases` The parser does not accept that.
> 3. `duplicate cases`. This is missing. Good point
> 
> I'll add those tests!

For the #3, that is a semantic error that I was not detecting.

seanyoung (approved):
LGTM

LucasSte:
@seanyoung and @xermicus I am requesting your review again as I implemented Cyrill's suggestions.

src/sema/yul/switch.rs
```diff
@@ -42,6 +45,32 @@ pub(crate) fn resolve_switch(
...
+            YulExpression::NumberLiteral(_, value, _) => value.clone(),
+            YulExpression::StringLiteral(_, value, _) => BigInt::from_bytes_be(Sign::Plus, value),
+            _ => unreachable!("Switch condition should be a literal"),
+        };
+
+        if conditions.contains(&big_int) {
+            ns.diagnostics.push(Diagnostic::error(
+                item.loc,
+                "duplicate case for switch".to_string(),
+            ));
```
seanyoung, line 67: Please use `Diagnostic::error_with_note()` and provide the location of other entry which is duplicate.

seanyoung: Also use the location of the literal, not the entire case including block

src/codegen/yul/statements.rs
```diff
@@ -465,3 +478,67 @@ fn process_for_block(
     cfg.set_phis(end_block, set.clone());
     cfg.set_phis(cond_block, set);
 }
+
+/// Generate CFG code for a switch statement
+fn resolve_switch(
```
seanyoung, line 483: You'd expect `resolve_` in sema, not codegen. How about calling this `fn switch()`?

seanyoung (changes requested):

xermicus (approved):

seanyoung (approved):

PR #1024, opened 2022-09-26
constant overflow compatible with solc

salaheldinsoliman:
This PR evaluates expressions constant arithmetic and checks for overflow according to what Solc does


src/sema/eval.rs
```diff
@@ -196,3 +201,186 @@ pub fn eval_const_rational(
...
+
+        Expression::InternalFunctionCall { args, .. }
+        | Expression::ExternalFunctionCall { args, .. } => {
+            for args_iter in args.clone() {
+                verify_result(&args_iter, ns, &args_iter.loc());
+            }
+
+            expr.clone()
+        }
+        _ => expr.clone(),
```
seanyoung, line 338: There are many expressions missing from this list, for example `new D(arg1, arg2)`, and calls with named arguments `foo({a: 1, b: 1})`, method calls etc. I think it might be better to use `fn recurse()` on `Expression`, see https://github.com/hyperledger/solang/blob/main/src/sema/ast.rs#L802

seanyoung (changes requested):
Looks good so far, some changes are needed though

src/sema/eval.rs
```diff
@@ -196,3 +201,186 @@ pub fn eval_const_rational(
         )),
     }
 }
+
+fn eval_constants_in_expression(
```
LucasSte, line 205: Add a doc comment to this function.

src/sema/eval.rs
```diff
@@ -196,3 +201,186 @@ pub fn eval_const_rational(
...
+        }
+
+        Expression::Multiply(loc, ty, unchecked, left, right) => {
+            let left = eval_constants_in_expression(left, ns, results);
+            let right = eval_constants_in_expression(right, ns, results);
+
+            if let (Expression::NumberLiteral(_, _, left), Expression::NumberLiteral(_, _, right)) =
+                (&left, &right)
+            {
+                results.push(left.mul(right.to_u32().unwrap()));
```
LucasSte, line 257: How do you know the number fits in a `u32`? Numbers can be negative too.
Keep in mind that arithmetic operations in Solidity support up to `u256`.

src/sema/eval.rs
```diff
@@ -196,3 +201,186 @@ pub fn eval_const_rational(
...
+                verify_result(&args_iter, ns, &args_iter.loc());
+            }
+
+            expr.clone()
+        }
+        _ => expr.clone(),
+    }
+}
+
+fn overflow_check(ns: &mut Namespace, result: BigInt, ty: Type, loc: Loc) {
```
LucasSte, line 342: Doc comment is missing.

src/sema/eval.rs
```diff
@@ -196,3 +201,186 @@ pub fn eval_const_rational(
...
+                    "value {} does not fit into type {}.",
+                    result,
+                    ty.to_string(ns)
+                ),
+            ));
+        }
+    }
+}
+
+pub fn verify_result(expr: &Expression, ns: &mut Namespace, loc: &Loc) {
```
LucasSte, line 380: Doc comment missing.
The function should not be public. When writing code, we should limit the visibility as much as possible. Please, check if you can make this function `pub(super)` instead.

src/sema/expression.rs
```diff
@@ -1653,6 +1607,7 @@ pub fn expression(
     diagnostics: &mut Diagnostics,
     resolve_to: ResolveTo,
 ) -> Result<Expression, ()> {
+    //println!( "EXPR {:?}", expr );
```
LucasSte, line 1610: Please, remove this.

src/sema/expression.rs
```diff
@@ -1625,13 +1579,13 @@ pub enum ResolveTo<'a> {
     Type(&'a Type), // We will be wanting this type please, e.g. `int64 x = 1;`
 }
 
-#[derive(Clone, Default)]
+#[derive(Clone, Default, Debug)]
```
LucasSte, line 1582: Debug was not originally here, so please, remove it.

src/sema/statements.rs
```diff
@@ -331,6 +332,7 @@ fn statement(
                     ResolveTo::Type(&var_ty),
                 )?;
 
+                verify_result(&expr, ns, loc);
```
LucasSte, line 335: You're evaluating the expression and checking the overflow, but you are not modifying the resulting AST.
If `expr` here is `1+2`, you evaluate it to `3`, but `expr` continues to be `1+2`. 
After evaluating, you can instead return the result to be used in the AST.

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,86 @@ fn test_statement_reachable() {
         assert_eq!(test_case.reachable(), expected);
     }
 }
+
+#[test]
+fn constant_overflow_checks() {
```
LucasSte, line 144: All code paths should be tested, so you need tests for all bit widths and all arithmetic operations.

LucasSte (changes requested):

src/sema/tests/mod.rs
salaheldinsoliman: I think testing for all bit width is not needed, bits of a type are retrieved through a BigInt function (bits()). see: https://github.com/salaheldinsoliman/solang/blob/2c9c38ead19ed472f3054dc7f2e2470f5c92d919/src/sema/eval.rs#L353
So I stuck for uint8 and int8 for the sake of readability.
But I agree this test needs more code path

LucasSte: You still need tests for other bit widths. Had you tested for `u64` here, you would have noticed that using `to_u32` as I have pointed out would not work.

The purpose of writing tests is to ensure everything will work as expected and we do not face regressions. If you don't write tests for other bit widths, someone might change your code that makes overflow detection for, let's say, `int256` stop working. As we don't have tests for that, this regression will be unnoticed. Tests are a reference for ensuring that new changes do not break any functionality.

src/sema/eval.rs
```diff
@@ -196,3 +197,248 @@ pub fn eval_const_rational(
...
+            let left = eval_constants_in_expression(left, ns)?;
+            let right = eval_constants_in_expression(right, ns)?;
+
+            if let (Expression::NumberLiteral(_, _, left), Expression::NumberLiteral(_, _, right)) =
+                (&left, &right)
+            {
+                Ok(Expression::NumberLiteral(
+                    *loc,
+                    ty.clone(),
+                    left.pow(right.to_u32().unwrap()),
```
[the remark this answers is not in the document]
LucasSte: I think the range check for the exponent is already performed in sema, isn't it?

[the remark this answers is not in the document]
salaheldinsoliman: In constant folding, the allowed range for the exponent is 32 bits, so I followed the same criteria.
https://github.com/hyperledger/solang/blob/eaeb2395765eb478c1fef7c679175bb609d053d1/src/codegen/constant_folding.rs#L550

tests/contract_testcases/substrate/primitives/units.dot
```diff
@@ -1,12 +1,18 @@
 strict digraph "tests/contract_testcases/substrate/primitives/units.sol" {
 	contract [label="contract c\ntests/contract_testcases/substrate/primitives/units.sol:2:9-6:10"]
 	test [label="function test\ncontract: c\ntests/contract_testcases/substrate/primitives/units.sol:3:13-35\nsignature test()\nvisibility public\nmutability nonpayable"]
+	var_decl [label="variable decl int32 x\ntests/contract_testcases/substrate/primitives/units.sol:4:17-34"]
+	number_literal [label="int32 literal: 1000000000000000000\ntests/contract_testcases/substrate/primitives/units.sol:4:27-34"]
 	diagnostic [label="found contract 'c'\nlevel Debug\ntests/contract_testcases/substrate/primitives/units.sol:2:9-6:10"]
-	diagnostic_5 [label="literal 1000000000000000000 is too large to fit into type 'int32'\nlevel Error\ntests/contract_testcases/substrate/primitives/units.sol:4:27-34"]
```
seanyoung, line 5: This error is suddenly gone, but it should not be.

tests/contract_testcases/substrate/primitives/test_literal_overflow_07.dot
```diff
@@ -1,10 +1,12 @@
 strict digraph "tests/contract_testcases/substrate/primitives/test_literal_overflow_07.sol" {
 	contract [label="contract test\ntests/contract_testcases/substrate/primitives/test_literal_overflow_07.sol:1:1-3:10"]
 	var [label="variable foo\nvisibility internal\ntests/contract_testcases/substrate/primitives/test_literal_overflow_07.sol:2:13-30"]
+	number_literal [label="uint8 literal: -130\ntests/contract_testcases/substrate/primitives/test_literal_overflow_07.sol:2:25-30"]
 	diagnostic [label="found abstract contract 'test'\nlevel Debug\ntests/contract_testcases/substrate/primitives/test_literal_overflow_07.sol:1:1-3:10"]
-	diagnostic_5 [label="negative literal -130 not allowed for unsigned type 'uint8'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_07.sol:2:25-30"]
```
seanyoung, line 5: this error should still exist.

tests/contract_testcases/substrate/primitives/test_literal_overflow_03.dot
```diff
@@ -1,12 +1,12 @@
 strict digraph "tests/contract_testcases/substrate/primitives/test_literal_overflow_03.sol" {
 	contract [label="contract test\ntests/contract_testcases/substrate/primitives/test_literal_overflow_03.sol:1:1-3:10"]
 	var [label="variable foo\nvisibility internal\ntests/contract_testcases/substrate/primitives/test_literal_overflow_03.sol:2:13-28"]
+	number_literal [label="int8 literal: -129\ntests/contract_testcases/substrate/primitives/test_literal_overflow_03.sol:2:24-28"]
 	diagnostic [label="found contract 'test'\nlevel Debug\ntests/contract_testcases/substrate/primitives/test_literal_overflow_03.sol:1:1-3:10"]
-	diagnostic_5 [label="contracts without public storage or functions are not allowed on Substrate. Consider declaring this contract abstract: 'abstract contract test'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_03.sol:1:1-3:10"]
-	diagnostic_6 [label="literal -129 is too large to fit into type 'int8'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_03.sol:2:24-28"]
```
seanyoung, line 6: This error should still exist

seanyoung (changes requested):
- We requested tests for different bit widths. 
- A lot of the differences in the dot files show regressions

src/sema/eval.rs
```diff
@@ -196,3 +202,283 @@ pub fn eval_const_rational(
...
+                    result,
+                    ty.get_type_size(),
+                ),
+            ));
+        }
+    }
+
+    if let Type::Int(bits) = ty {
+        // If number of bits is more than what the type can hold. BigInt.bits() is not used here since it disregards the sign.
+        if result.to_signed_bytes_be().len() * 8 > (bits as usize) {
```
LucasSte, line 443: If the sign is negative, you'll need `result.bits() + 1` to store the number. By doing `result.to_signed_bytes_be()` you are copying the inner data data structure that holds the number. That is not necessary for this case.

src/sema/eval.rs
```diff
@@ -196,3 +202,283 @@ pub fn eval_const_rational(
...
+                    result,
+                    ty.get_type_size(),
+                ),
+            ));
+        }
+    }
+    None
+}
+
+/// Wrapper function for the overflow detection logic. It first calls a function that recurses a given expression and returns a list of arithmetic operations within the expression, and a list of diagnostics. Then, it calls overflow_check() on the retrieved arithmetic operations, and pushes the returned diagnostics to the namespace
```
LucasSte, line 457: Split this large comment into multiple lines.

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,193 @@ fn test_statement_reachable() {
...
+            // value 882 does not fit into type int8.
+            int8 mixed = 126 * 7 * input;
+        }
+    
+        function test_shift(int8 input) public {
+            // value 128 does not fit into type int8.
+            int8 mul_ovf = 1 << 7;
+    
+            // value 128 does not fit into type int8.
+            int8 mixed = (1 << 7) + input;
```
LucasSte, line 178: There are no tests that include shift right and divide.
For example,
`uint8 b = 255 - 255/5` should not raise any error.

tests/contract_testcases/solana/large_exponent.dot
```diff
@@ -1,28 +1,28 @@
...
 	f [label="function f\ncontract: C\ntests/contract_testcases/solana/large_exponent.sol:5:6-26\nsignature f()\nvisibility private\nmutability nonpayable"]
 	inline_assembly [label="inline assembly\ntests/contract_testcases/solana/large_exponent.sol:6:2-8:3"]
 	g [label="function g\ncontract: C\ntests/contract_testcases/solana/large_exponent.sol:10:6-26\nsignature g()\nvisibility private\nmutability nonpayable"]
-	inline_assembly_8 [label="inline assembly\ntests/contract_testcases/solana/large_exponent.sol:11:2-13:3"]
+	inline_assembly_9 [label="inline assembly\ntests/contract_testcases/solana/large_exponent.sol:11:2-13:3"]
 	diagnostic [label="found contract 'C'\nlevel Debug\ntests/contract_testcases/solana/large_exponent.sol:1:1-16:2"]
-	diagnostic_11 [label="literal 1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 is too large to fit into type 'int256'\nlevel Error\ntests/contract_testcases/solana/large_exponent.sol:2:26-31"]
```
LucasSte, line 11: This error should not be gone. It is still needed for this case.

tests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.dot
```diff
@@ -1,10 +1,30 @@
...
+	number_literal [label="uint32 literal: 2\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:4:48-49"]
+	expr [label="expression\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:6:17-26"]
+	assign [label="assign\nint32\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:6:17-26"]
+	subscript [label="subscript int32[]\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:6:17-22"]
+	variable [label="variable: a\nint32[]\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:6:17-18"]
+	number_literal_10 [label="uint32 literal: -1\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:6:19-21"]
+	number_literal_11 [label="int32 literal: 5\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:6:25-26"]
 	diagnostic [label="found contract 'foo'\nlevel Debug\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:2:9-8:10"]
-	diagnostic_5 [label="negative literal -1 not allowed for unsigned type 'uint32'\nlevel Error\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_deref.sol:6:19-21"]
```
LucasSte, line 5: This error should not be gone.
```solidity
contract foo {
            function test() public {
                int32[] memory a = new int32[](2);
                a[-1] = 5;
            }
        }
```
Negative index is not allowed.

tests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.dot
```diff
@@ -1,10 +1,28 @@
...
+	number_literal [label="uint32 literal: -1\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:4:48-50"]
+	expr [label="expression\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:6:17-38"]
+	builtins [label="builtin Assert\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:6:17-23"]
+	equal [label="equal\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:6:24-37"]
+	builtins_9 [label="builtin ArrayLength\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:6:24-32"]
+	variable [label="variable: a\nint32[]\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:6:24-25"]
+	number_literal_11 [label="uint32 literal: 5\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:6:36-37"]
 	diagnostic [label="found contract 'foo'\nlevel Debug\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:2:9-8:10"]
-	diagnostic_5 [label="negative literal -1 not allowed for unsigned type 'uint32'\nlevel Error\ntests/contract_testcases/substrate/arrays/memory_dynamic_array_new_03.sol:4:48-50"]
```
LucasSte, line 5: This error should not be gone. This is not allowed: `int32[] memory a = new int32[](-1);`

tests/contract_testcases/substrate/primitives/test_literal_overflow.dot
```diff
@@ -1,10 +1,12 @@
 strict digraph "tests/contract_testcases/substrate/primitives/test_literal_overflow.sol" {
 	contract [label="contract test\ntests/contract_testcases/substrate/primitives/test_literal_overflow.sol:1:1-3:10"]
 	var [label="variable foo\nvisibility internal\ntests/contract_testcases/substrate/primitives/test_literal_overflow.sol:2:13-28"]
+	number_literal [label="uint8 literal: 300\ntests/contract_testcases/substrate/primitives/test_literal_overflow.sol:2:25-28"]
 	diagnostic [label="found abstract contract 'test'\nlevel Debug\ntests/contract_testcases/substrate/primitives/test_literal_overflow.sol:1:1-3:10"]
-	diagnostic_5 [label="literal 300 is too large to fit into type 'uint8'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow.sol:2:25-28"]
```
LucasSte, line 5: Why is this error gone? 300 does not fit in an `uint8`.

tests/contract_testcases/substrate/primitives/test_literal_overflow_01.dot
```diff
@@ -1,12 +1,12 @@
 strict digraph "tests/contract_testcases/substrate/primitives/test_literal_overflow_01.sol" {
 	contract [label="contract test\ntests/contract_testcases/substrate/primitives/test_literal_overflow_01.sol:1:1-3:10"]
 	var [label="variable foo\nvisibility internal\ntests/contract_testcases/substrate/primitives/test_literal_overflow_01.sol:2:13-33"]
+	number_literal [label="uint16 literal: 65536\ntests/contract_testcases/substrate/primitives/test_literal_overflow_01.sol:2:26-33"]
 	diagnostic [label="found contract 'test'\nlevel Debug\ntests/contract_testcases/substrate/primitives/test_literal_overflow_01.sol:1:1-3:10"]
-	diagnostic_5 [label="contracts without public storage or functions are not allowed on Substrate. Consider declaring this contract abstract: 'abstract contract test'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_01.sol:1:1-3:10"]
-	diagnostic_6 [label="literal 65536 is too large to fit into type 'uint16'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_01.sol:2:26-33"]
```
LucasSte, line 6: This error should not be gone.

tests/contract_testcases/substrate/primitives/test_literal_overflow_02.dot
```diff
@@ -1,12 +1,12 @@
 strict digraph "tests/contract_testcases/substrate/primitives/test_literal_overflow_02.sol" {
 	contract [label="contract test\ntests/contract_testcases/substrate/primitives/test_literal_overflow_02.sol:1:1-3:10"]
 	var [label="variable foo\nvisibility internal\ntests/contract_testcases/substrate/primitives/test_literal_overflow_02.sol:2:13-29"]
+	number_literal [label="int8 literal: 128\ntests/contract_testcases/substrate/primitives/test_literal_overflow_02.sol:2:24-29"]
 	diagnostic [label="found contract 'test'\nlevel Debug\ntests/contract_testcases/substrate/primitives/test_literal_overflow_02.sol:1:1-3:10"]
-	diagnostic_5 [label="contracts without public storage or functions are not allowed on Substrate. Consider declaring this contract abstract: 'abstract contract test'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_02.sol:1:1-3:10"]
-	diagnostic_6 [label="literal 128 is too large to fit into type 'int8'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_02.sol:2:24-29"]
```
LucasSte, line 6: This error should not be gone.

tests/contract_testcases/substrate/primitives/test_literal_overflow_08.dot
```diff
@@ -1,10 +1,12 @@
 strict digraph "tests/contract_testcases/substrate/primitives/test_literal_overflow_08.sol" {
 	contract [label="contract test\ntests/contract_testcases/substrate/primitives/test_literal_overflow_08.sol:1:1-3:10"]
 	var [label="variable foo\nvisibility internal\ntests/contract_testcases/substrate/primitives/test_literal_overflow_08.sol:2:13-46"]
+	number_literal [label="int64 literal: 18446744073709551616\ntests/contract_testcases/substrate/primitives/test_literal_overflow_08.sol:2:25-46"]
 	diagnostic [label="found abstract contract 'test'\nlevel Debug\ntests/contract_testcases/substrate/primitives/test_literal_overflow_08.sol:1:1-3:10"]
-	diagnostic_5 [label="literal 18446744073709551616 is too large to fit into type 'int64'\nlevel Error\ntests/contract_testcases/substrate/primitives/test_literal_overflow_08.sol:2:25-46"]
```
LucasSte, line 5: This error should not be gone.

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,193 @@ fn test_statement_reachable() {
...
+    assert_eq!(
+        errors[5].message,
+        "value 340282366920938463463374607431768211456 does not fit into type uint128."
+    );
+    assert_eq!(errors[6].message, "negative value -1 does not fit into type uint32. Cannot implicitly convert signed literal to unsigned type.");
+    assert_eq!(
+        errors[7].message,
+        "value 340282366920938463463374607431768211456 does not fit into type uint128."
+    );
+    assert_eq!(errors[8].message, "exponent '256' too large");
```
LucasSte, line 330: This does not test your implementation. The error comes from another place in sema. You could try instead:
```
uint256 num = 115792089237316195423570985008687907853269984665640564039457584007913129639935 * 4;
```

LucasSte (changes requested):

src/sema/eval.rs
```diff
@@ -196,3 +202,283 @@ pub fn eval_const_rational(
...
+pub(super) fn verify_result(expr: Expression, ns: &mut Namespace) -> Expression {
```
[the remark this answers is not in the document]
salaheldinsoliman: I mean its purpose is not to fold, but to verify the result if it overflow, what about `verify_expression_for_overflow()` ?

src/sema/eval.rs
salaheldinsoliman: This will fail for the lower boundary: `int8 sesa = -128` won't be allowed
I have to add a special condition for this If I use `res.bits() + 1` so I thought this is more neat

salaheldinsoliman: This will not work for the boundaries.
i.e -128 will not fit into int8 using bits() + 1

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+        | Expression::Modulo(..)
+        | Expression::Power(..)
+        | Expression::ShiftLeft(..)
+        | Expression::ShiftRight(..)
+        | Expression::BitwiseAnd(..)
+        | Expression::BitwiseOr(..)
+        | Expression::BitwiseXor(..) => {
+            let expression = eval_constants_in_expression(expr.clone());
+            match expression {
+                Ok(input) => {
```
seanyoung, line 224: give `input` a better name, like `expr` or so

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+        | Expression::Divide(..)
+        | Expression::Modulo(..)
+        | Expression::Power(..)
+        | Expression::ShiftLeft(..)
+        | Expression::ShiftRight(..)
+        | Expression::BitwiseAnd(..)
+        | Expression::BitwiseOr(..)
+        | Expression::BitwiseXor(..) => {
+            let expression = eval_constants_in_expression(expr.clone());
+            match expression {
```
seanyoung, line 223: you can combine two lines to `match eval_constants_in_expression(expr.clone()) {`

You are still cloning here.

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+                Ok(input) => {
+                    if let Expression::NumberLiteral(..) = input {
+                        params.recursed.push(input);
+                        return false;
+                    }
+                }
+                Err(input) => {
```
seanyoung, line 230: rename `input` to `diagnostic`.

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+                return false;
+            }
+        }
+    }
+
+    true
+}
+
+/// This function recursively folds number literals in a given expression.
+fn eval_constants_in_expression(expr: Expression) -> Result<Expression, Diagnostic> {
```
seanyoung, line 247: If this takes a reference to an expression, i.e. `&Expression`, then less cloning will be done.

src/sema/eval.rs
seanyoung: This should be permitted:
```rust
int8 foo = -128;
```
is it?

tests/contract.rs
```diff
@@ -118,9 +118,9 @@ fn parse_file(path: PathBuf, target: Target) -> io::Result<()> {
     let generated_dot = ns.dotgraphviz();
 
     // uncomment the next three lines to regenerate the test data
-    // use std::io::Write;
-    // let mut file = File::create(&path)?;
-    // file.write_all(generated_dot.as_bytes())?;
+    //use std::io::Write;
+    //let mut file = File::create(&path)?;
+    //file.write_all(generated_dot.as_bytes())?;
```
seanyoung, line 123: Please don't modify files without reason

src/sema/eval.rs
```diff
@@ -196,3 +197,248 @@ pub fn eval_const_rational(
...
+            } else {
+                Ok(Expression::ShiftRight(
+                    *loc,
+                    ty.clone(),
+                    Box::new(left),
+                    Box::new(right),
+                    *bool,
+                ))
+            }
+        }
```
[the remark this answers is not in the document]
seanyoung: @salaheldinsoliman please don't mark comments as resolved when the work has not been done

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+        Expression::ShiftRight(loc, ty, left, right, bool) => {
+            let left = eval_constants_in_expression(*left)?;
+            let right = eval_constants_in_expression(*right)?;
+
+            if let (
+                Expression::NumberLiteral(_, _, left),
+                Expression::NumberLiteral(right_loc, _, right),
+            ) = (&left, &right)
+            {
+                if let Some(diagnostic) = overflow_check(right.clone(), Type::Uint(64), *right_loc)
```
seanyoung, line 404: Just checking that the shift value fits into `uint64` is not enough.
```solidity
uint16 x = 1 >> 18446744073709551615;
```
This makes no sense. Please check that the shift value is less than `left.bits()`

seanyoung (changes requested):

src/sema/eval.rs
salaheldinsoliman: I followed the reasoning that the overall result shouldn't overflow. For instance, `int8 sesa = (1<<8) - 130;` doesn't throw an error in solc

seanyoung: That is a good point.

LucasSte (changes requested):

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,197 @@ fn test_statement_reachable() {
...
+        function test_for_loop () public {
+            for (int8 i = 125 + 5; i < 300 ; i++) {
+            }
+        }
+
+        function composite(int8 a, bytes input) public{
+
+            uint8 sesa = 500- 400 + test_params(100+200, 0) + (200+101) + input.readUint8(4294967296);
+            int8 seas = (120 + 120) + a + (120 + 125);  
+            uint8 b = 255 - 255/5 ;
```
LucasSte, line 206: This test verifies that, given a contract, you are generating the correct errors. It is not making sure that an input does not produce any error. Writing `uint8 b = 255 - 255/5` here has no effect for this test, because you do not check that it does not produce any diagnostic.

LucasSte (changes requested):
Just thought about this. It is an important point.

src/sema/tests/mod.rs
LucasSte: In addition, `uint8 b = 300 - 255/5;` is a better test, as `255` fits in a 8-bit unsigned integer.

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+            ) = (&left, &right)
+            {
+                if let Some(diagnostic) = overflow_check(right.clone(), Type::Uint(16), *right_loc)
+                {
+                    Err(diagnostic)
+                } else {
+                    Ok(Expression::NumberLiteral(
+                        loc,
+                        ty,
+                        left.pow(right.to_u16().unwrap().into()),
```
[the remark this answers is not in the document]
salaheldinsoliman: Bigint.Pow() doesn't take a bigint as an argument

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+            ) = (&left, &right)
+            {
+                if let Some(diagnostic) = overflow_check(right.clone(), Type::Uint(64), *right_loc)
+                {
+                    Err(diagnostic)
+                } else {
+                    Ok(Expression::NumberLiteral(
+                        loc,
+                        ty,
+                        left.shl(right.to_u64().unwrap()),
```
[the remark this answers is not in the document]
salaheldinsoliman: Bigint.shl() doesn't take a bigint as an argument

src/sema/eval.rs
```diff
@@ -196,3 +202,331 @@ pub fn eval_const_rational(
...
+            ) = (&left, &right)
+            {
+                if let Some(diagnostic) = overflow_check(right.clone(), Type::Uint(64), *right_loc)
+                {
+                    Err(diagnostic)
+                } else {
+                    Ok(Expression::NumberLiteral(
+                        loc,
+                        ty,
+                        left.shr(right.to_u64().unwrap()),
```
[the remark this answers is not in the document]
salaheldinsoliman: Same as pow() and shr()

src/sema/tests/mod.rs
salaheldinsoliman: I can implicitly test if it doesn't produce diagnostics by putting lines after it that will throw diagnostics. If there is one, the assert lines should fail as the diagnostics order will differ

src/sema/eval.rs
```diff
@@ -196,3 +202,398 @@ pub fn eval_const_rational(
...
+    }
+
+    true
+}
+
+/// This function recursively folds number literals in a given expression.
+fn eval_constants_in_expression(expr: &Expression) -> Result<Expression, Diagnostic> {
+    match expr {
+        Expression::Add(loc, ty, unchecked, left, right) => {
+            let left = eval_constants_in_expression(&*left)?;
```
seanyoung, line 247: I think `&*left` can be replaced with `left`

src/sema/eval.rs
```diff
@@ -196,3 +202,398 @@ pub fn eval_const_rational(
...
+            {
+                Ok(Expression::NumberLiteral(*loc, ty.clone(), left.add(right)))
+            } else {
+                Ok(Expression::Add(
+                    *loc,
+                    ty.clone(),
+                    *unchecked,
+                    Box::new(left),
+                    Box::new(right),
+                ))
```
seanyoung, line 261: This is a non-constant expression, which is not used anywhere I think. 

How about returning making the return value `Result<Option<Expression>, Diagnostic>` and then returning `Ok(None)`

src/sema/eval.rs
```diff
@@ -196,3 +202,398 @@ pub fn eval_const_rational(
...
+            {
+                Ok(Expression::NumberLiteral(*loc, ty.clone(), left.sub(right)))
+            } else {
+                Ok(Expression::Subtract(
+                    *loc,
+                    ty.clone(),
+                    *unchecked,
+                    Box::new(left),
+                    Box::new(right),
+                ))
```
seanyoung, line 279: Same again, return `Ok(None)`

src/sema/eval.rs
```diff
@@ -196,3 +202,398 @@ pub fn eval_const_rational(
...
+            if let Some(diagnostic) = overflow_check(res, ty, loc) {
+                ns.diagnostics.push(diagnostic);
+            }
+        }
+    }
+    ns.diagnostics.append(params.diagnostics);
+
+    match eval_constants_in_expression(&expr) {
+        Ok(expression) => {
+            if let Expression::NumberLiteral(loc, ty, res) = &expression {
```
seanyoung, line 588: If `eval_constant_in_expression()` return `Result<Option<Expression>, Diagnostic>` then these two lines can be replaced with `Ok(Some(expr)) => {`

src/sema/eval.rs
```diff
@@ -196,3 +202,398 @@ pub fn eval_const_rational(
...
+            let left = eval_constants_in_expression(&*left)?;
+            let right = eval_constants_in_expression(&*right)?;
+
+            if let (
+                Expression::NumberLiteral(_, _, left),
+                Expression::NumberLiteral(right_loc, _, right),
+            ) = (&left, &right)
+            {
+                if let Some(diagnostic) = overflow_check(right, &Type::Uint(64), right_loc) {
+                    Err(diagnostic)
```
seanyoung, line 410: This is very ugly. This means that if I do:
```solidity
int128 a = 105 >> 5192296858534827628530496329220096;
```
Now we get a diagnostic about `5192296858534827628530496329220096` not fitting into `uint64`. That's not helpful.

We need a better diagnostic in this case and also a test case for this.

seanyoung (changes requested):

src/sema/tests/mod.rs
LucasSte: That makes sense. It is worth checking that the number of errors in this test is constant and equals a predetermined number. Otherwise, there may be new errors that we are unaware of.

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,220 @@ fn test_statement_reachable() {
...
+    }
+    
+        "#;
+    let ns = parse(file);
+    let errors = ns.diagnostics.errors();
+
+    assert_eq!(errors[0].message, "value 133 does not fit into type int8.");
+    assert_eq!(errors[1].message, "negative value -1 does not fit into type uint8. Cannot implicitly convert signed literal to unsigned type.");
+    assert_eq!(errors[2].message, "value 133 does not fit into type int8.");
+    assert_eq!(errors[4].message, "value 762 does not fit into type int8.");
```
LucasSte, line 228: What does `errors[3].message` contain?

src/sema/eval.rs
salaheldinsoliman: I noticed that constant folding throws a hard error there. I think the best action is to not check on the the right hand side as solc doesn't do this, it just looks as the result of the shifting

src/sema/tests/mod.rs
salaheldinsoliman: The same as errors[1], I left it out by mistake

src/sema/eval.rs
```diff
@@ -196,3 +202,372 @@ pub fn eval_const_rational(
...
+        Expression::SignExt(loc, ty, expr) => {
+            let expr = eval_constants_in_expression(expr)?;
+            if let Some(Expression::NumberLiteral(_, _, n)) = expr {
+                Ok(Some(Expression::NumberLiteral(*loc, ty.clone(), n)))
+            } else {
+                Ok(None)
+            }
+        }
+
+        _ => Ok(Some(expr.clone())),
```
seanyoung, line 493: `Ok(None)`

src/sema/statements.rs
```diff
@@ -750,28 +754,37 @@ fn statement(
...
+                        true,
+                        context,
+                        ns,
+                        symtable,
+                        diagnostics,
+                        ResolveTo::Discard,
+                    )?;
+
+                    ret = verify_result(ret, ns);
+                    ret
```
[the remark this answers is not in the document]
seanyoung: @salaheldinsoliman please do not resolve conversations without doing any changes

src/sema/statements.rs
```diff
@@ -750,28 +754,37 @@ fn statement(
...
+                        args,
+                        true,
+                        context,
+                        ns,
+                        symtable,
+                        diagnostics,
+                        ResolveTo::Discard,
+                    )?;
+                    ret = verify_result(ret, ns);
+                    ret
```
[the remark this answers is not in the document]
seanyoung: @salaheldinsoliman please do not resolve conversations without doing any changes

src/sema/eval.rs
seanyoung: > I can throw a warning if the shift is done with a number larger than the bit size, what do you think?

Yes, that would be a good solution. Also note the shift might be negative which is also not allowed (so the diagnostic should be "shift by negative value not permitted".

seanyoung (changes requested):

src/sema/statements.rs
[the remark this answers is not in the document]
salaheldinsoliman: verify_result doesn't return anything, so I have to use the return value somehow.
I resolved this before as we were modifying the AST, so verify_result returned the folded expression

src/sema/statements.rs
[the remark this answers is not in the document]
salaheldinsoliman: verify_result doesn't return anything, so I have to use the return value somehow.
I resolved this before as we were modifying the AST, so verify_result returned the folded expression

src/sema/eval.rs
salaheldinsoliman: What I fear here is when constant folding is enabled, there will be both a hard error and a warning for the same issue.
Check: https://github.com/hyperledger/solang/blob/fe342c5ae3dc6de823c8144f833b1dd522da8a78/src/codegen/constant_folding.rs#L527

salaheldinsoliman: Given that, I only check for the final result of the shift

src/sema/eval.rs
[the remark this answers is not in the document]
seanyoung: how about `verify_constant_expressions()`

src/sema/eval.rs
seanyoung: If sema produces an error, I don't think codegen is executed so  this shouldn't be a problem.

src/sema/eval.rs
```diff
@@ -196,3 +202,372 @@ pub fn eval_const_rational(
         )),
     }
 }
+
+/// Helper function that recurses the expression, and calls expression = eval_constants_in_expression().
```
LucasSte, line 206: What does this mean?
`expression = eval_constants_in_expression()`

src/sema/eval.rs
```diff
@@ -196,3 +202,372 @@ pub fn eval_const_rational(
...
+            }
+        }
+    }
+    ns.diagnostics.append(params.diagnostics);
+
+    match eval_constants_in_expression(expr) {
+        Ok(Some(expression)) => {
+            if let Expression::NumberLiteral(loc, ty, res) = &expression {
+                let result = overflow_check(res, ty, loc);
+                if let Some(diagnostic) = result {
```
LucasSte, line 563:
```suggestion
                if let Some(diagnostic) = overflow_check(res, ty, loc) {
```
Just a nitpick

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,223 @@ fn test_statement_reachable() {
...
+            uint8 b = 255 - 255/5 ;
+
+            // value 260 does not fit into type uint8.
+            uint8 shift_r = (120 >> 2) + 230;
+
+            // value 261 does not fit into type uint8.
+            uint8 mod_test = 254 + (500%17);
+
+            // value 269 does not fit into type uint8.
+            uint8 bb = 320 - (255/5) ;
```
LucasSte, line 217: There are no tests for `BitwiseAnd`, `BitwiseOr` and `BitwiseXor`.

LucasSte (changes requested):

src/sema/eval.rs
```diff
@@ -196,3 +196,324 @@ pub fn eval_const_rational(
...
+
+            if let (
+                Some(Expression::NumberLiteral(_, _, left)),
+                Some(Expression::NumberLiteral(_, _, right)),
+            ) = (&left, &right)
+            {
+                Some(Expression::NumberLiteral(
+                    *loc,
+                    ty.clone(),
+                    left.shr(right.to_u64().unwrap()),
```
seanyoung, line 388: We need the same for shift right as you did for shift left. Right now you can make the compiler panic with:

```solidity
int x = 1 >> 14676683207225698178084221555689649093015162623576402558976;
```

src/sema/eval.rs
```diff
@@ -196,3 +196,324 @@ pub fn eval_const_rational(
...
+                let expression = eval_constants_in_expression(args_iter, ns);
+                if let Some(Expression::NumberLiteral(loc, ty, res)) = &expression {
+                    if let Some(diagnostic) = overflow_check(res, ty, loc) {
+                        ns.diagnostics.push(diagnostic);
+                    }
+                }
+            }
+
+            None
+        }
```
seanyoung, line 459: Recurse should deal with this, this is not necessary (remove lines)

src/sema/eval.rs
```diff
@@ -196,3 +196,324 @@ pub fn eval_const_rational(
...
+            let left = eval_constants_in_expression(left, ns);
+            let right = eval_constants_in_expression(right, ns);
+
+            if let (
+                Some(Expression::NumberLiteral(_, _, left)),
+                Some(Expression::NumberLiteral(right_loc, _, right)),
+            ) = (&left, &right)
+            {
+                if let Some(diagnostic) = overflow_check(right, &Type::Uint(16), right_loc) {
+                    ns.diagnostics.push(diagnostic);
```
seanyoung, line 329: This diagnostic is not right.

```solidity
uint x = 2 ** 14676683207225698178084221555689649093015162623576402558976;
```
And now I get an error about `uint16`, that is not clear. Please add same code as `ShiftLeft`

seanyoung (changes requested):

src/sema/tests/mod.rs
```diff
@@ -128,3 +139,265 @@ fn test_statement_reachable() {
...
+        "left shift by 7 may overflow the final result"
+    );
+    assert_eq!(
+        warnings[2].message,
+        "left shift by 7 may overflow the final result"
+    );
+    assert_eq!(
+        warnings[3].message,
+        "left shift by 9 may overflow the final result"
+    );
```
LucasSte, line 310: How do you make sure there are no other warnings besides these?

src/sema/eval.rs
```diff
@@ -196,3 +196,324 @@ pub fn eval_const_rational(
...
+
+            if let (
+                Some(Expression::NumberLiteral(_, _, left)),
+                Some(Expression::NumberLiteral(right_loc, _, right)),
+            ) = (&left, &right)
+            {
+                if right.sign() == Sign::Minus {
+                    ns.diagnostics.push(Diagnostic::error(
+                        *right_loc,
+                        format!("left shift by {} is not possible", right),
```
LucasSte, line 354: There is not test for this case.

src/sema/eval.rs
```diff
@@ -196,3 +196,324 @@ pub fn eval_const_rational(
...
+            let right = eval_constants_in_expression(right, ns);
+
+            if let (
+                Some(Expression::NumberLiteral(_, _, left)),
+                Some(Expression::NumberLiteral(_, _, right)),
+            ) = (&left, &right)
+            {
+                if right.is_zero() {
+                    ns.diagnostics
+                        .push(Diagnostic::error(*loc, "divide by zero".to_string()));
```
LucasSte, line 310: There is no test for this case.

src/sema/eval.rs
```diff
@@ -196,3 +196,324 @@ pub fn eval_const_rational(
...
```
LucasSte, line 289: There is no test for this case.

src/sema/eval.rs
```diff
@@ -196,3 +196,324 @@ pub fn eval_const_rational(
         )),
     }
 }
+
+/// Function that recurses the expression, and calls a function that folds number literals in the expression(eval_constants_in_expression())
```
LucasSte, line 200: What does this mean?
`expression(eval_constants_in_expression())`

LucasSte (changes requested):

LucasSte:
Remember to test all the error and warning messages you have added. If there are no tests for them, we'll never detect regressions for the cases they cover.

src/sema/tests/mod.rs
salaheldinsoliman: I added an assert for the number of warnings, thanks

src/sema/eval.rs
salaheldinsoliman: I meant to put the name of the function that folds the numberliterals

LucasSte: > I meant to put the name of the function that folds the numberliterals

I think you can rewrite that. `expression(eval_constants_in_expression())` looks like `expression` is a function.

src/sema/tests/mod.rs
seanyoung: Why has this been marked as resolved @salaheldinsoliman ?

seanyoung (changes requested):
The code looks good.

src/sema/tests/mod.rs
salaheldinsoliman: I added another function that tests for other types but with less test cases

src/sema/eval.rs
```diff
@@ -196,3 +196,361 @@ pub fn eval_const_rational(
         )),
     }
 }
+
+/// Function that recurses the expression, and calls a function that folds number literals in the expression: eval_constants_in_expression().
```
LucasSte, line 200:
```suggestion
/// Function that recurses the expression and folds number literals by calling 'eval_constants_in_expression'.
```

I finally understood what you meant. This is a suggestion to make your comment clearer.

LucasSte (changes requested):

seanyoung (changes requested):
The constant overflow detection does not work in all statements.

```solidity
contract c {
	function foo() public {
		uint16 x = 0;

		x += 450000;

		for (uint16 i = 0;  i < (2 ** 32); i += 65546);
	}
}
```

LucasSte (approved):

seanyoung (approved):

PR #1025, opened 2022-09-26
Update seal calls

xermicus:
This PR updates the `seal_` calls to reflect newer versions. Some of them use `seal1` and some storage APIs  temporarily  `__unstable__`now (will be changed as soon as they are released to some stable version).
Individual calls updated:
- `seal_random`: Returns 4 more bytes (the block number) so the output buffer size had to be extend
- `seal_instantiate`: Args changed
- `seal_terminate`: Args changed
- `seal_call`: Args (flags can be done later if needed)
- `seal_set_storage`: Signature changed. `key_len` will be needed for transparent hashing, once we fix the metadata generation. The newly introduced return type (the length of the pre-existing value, if any) is ignore for now since in solidity we don't use that.
- `seal_clear_storage`: Same as above.
- `seal_get_storage`: Arguments changed

Additionally I decided to introduce a few macros to get rid of some awkward and repeating patterns and make it easier to reason about the emitted code.


src/emit/substrate/mod.rs
```diff
@@ -22,6 +22,109 @@ pub(super) mod target;
...
+
+        #[allow(unused_macros)]
+        macro_rules! i32_cnst {
+            ($val:expr) => {
+                cnst!(i32_type, $val)
+            };
+        }
+
+        #[allow(unused_macros)]
+        macro_rules! i32_null {
```
seanyoung, line 50: I think the name `i32_null` is confusing, it sounds like a pointer. How about `i32_zero`?

src/emit/substrate/mod.rs
```diff
@@ -22,6 +22,109 @@ pub(super) mod target;
...
+    ($binary:expr) => {
+        #[allow(unused_macros)]
+        macro_rules! byte_ptr {
+            () => {
+                $binary.context.i8_type().ptr_type(AddressSpace::Generic)
+            };
+        }
+
+        #[allow(unused_macros)]
+        macro_rules! cnst {
```
seanyoung, line 36: `cnst` is very short. How about spelling out `const` with the vowel.

src/emit/substrate/mod.rs
```diff
@@ -22,6 +22,109 @@ pub(super) mod target;
...
+            ($of_type:ident, $val:expr) => {
+                $binary.context.$of_type().const_int($val, false)
+            };
+        }
+
+        #[allow(unused_macros)]
+        macro_rules! i32_cnst {
```
seanyoung, line 43: `i32_cnst` is very short. How about spelling out `i32_const` with the vowel.

tests/substrate.rs
```diff
@@ -188,11 +191,12 @@ impl Externals for MockSubstrate {
                 Ok(None)
             }
             Some(SubstrateExternal::seal_get_storage) => {
-                assert_eq!(args.len(), 3);
+                assert_eq!(args.len(), 4);
 
                 let key_ptr: u32 = args.nth_checked(0)?;
-                let dest_ptr: u32 = args.nth_checked(1)?;
-                let len_ptr: u32 = args.nth_checked(2)?;
+                //let key_len: u32 = args.nth_checked(1)?; TODO
```
seanyoung, line 197: The MockSubstrate assumes that the key is 32 bytes. I think this should either do `assert_eq!(key_len, 32);` or implement variable-length key lengths.

tests/substrate.rs
```diff
@@ -241,18 +245,21 @@ impl Externals for MockSubstrate {
                 println!("seal_clear_storage: {:?}", key);
                 self.store.remove(&(self.vm.account, key));
 
-                Ok(None)
+                // The size of the pre-existing value is returned (if any).
+                // In this mock VM we just always assume that there was no value and return `None`.
+                Ok(Some(NONE_SENTINEL))
```
seanyoung, line 250: Rather than returning a bogus value, you could just do something like:
```rust
let old_length = self.store.remove(&(self.vm.account, key)).map(|e| e.len()).unwrap_or(0);
return Ok(Some(Runtime:i32(old_length as i32));
```

tests/substrate.rs
```diff
@@ -266,7 +273,9 @@ impl Externals for MockSubstrate {
 
                 self.store.insert((self.vm.account, key), data);
 
-                Ok(None)
+                // The size of the pre-existing value is returned (if any).
+                // In this mock VM we just always assume that there was no value and return `None`.
+                Ok(Some(NONE_SENTINEL))
```
seanyoung, line 278: Same as clear storage, just calculate the old length, it's not much effort.
```rust
let old_length = self.store.get)(&(self.vm.account, key)).map(|e| e.len()).unwrap_or(0);
// remove entry
return Ok(Some(Runtime:i32(old_length as i32));
```

Much nicer than return bugs stuff

seanyoung (changes requested):
I really like your macros, we should reuse them in emit more generally.

src/emit/substrate/mod.rs
xermicus: ACK

src/emit/substrate/mod.rs
xermicus: ACK

tests/substrate.rs
xermicus: If it assumes 32bit key length for now I can check this. But I'd rather implement all the variable Key length stuff in another PR.

tests/substrate.rs
xermicus: ACK, but would that then be `unwrap_or(SENTINEL_VALUE)`?

xermicus:
> I really like your macros, we should reuse them in emit more generally.

Thanks! I considered this but it would be quite some work to change everything.I'd not be against it generally, but rather do it in a different PR (or as a sporadic effort when touching the code anyways, because it is easier to understand existing code after switching it to using these macros).

tests/substrate.rs
seanyoung: Well, whatever substrate returns if there was no previouskey

src/emit/substrate/mod.rs
```diff
@@ -175,323 +271,92 @@ impl SubstrateTarget {
...
+                    ctx.$fn_type().fn_type(&[$($args),*], false),
+                    Some(Linkage::External),
+                );
+            };
+        }
 
-        binary.module.add_function(
+        external!("seal_input", void_type, u8_ptr, u32_ptr);
+        external!("seal_hash_keccak_256", void_type, u8_ptr, u32_val, u8_ptr);
+        external!("seal_hash_sha2_256", void_type, u8_ptr, u32_val, u8_ptr);
```
LucasSte, line 292: This is really cool!

tests/substrate.rs
```diff
@@ -584,35 +592,27 @@ impl Externals for MockSubstrate {
...
-                let account_ptr: u32 = args.nth_checked(7)?;
-                let account_len_ptr: u32 = args.nth_checked(8)?;
-                let output_ptr: u32 = args.nth_checked(9)?;
-                let output_len_ptr: u32 = args.nth_checked(10)?;
-                let salt_ptr: u32 = args.nth_checked(11)?;
-                let salt_len: u32 = args.nth_checked(12)?;
+                //let codehash_len: u32 = args.nth_checked(1)?;
+                //let gas: u64 = args.nth_checked(1)?;
+                let value_ptr: u32 = args.nth_checked(2)?;
+                //let value_len: u32 = args.nth_checked(4)?;
```
LucasSte, line 598: Please, remove commented code.

LucasSte (changes requested):

tests/substrate.rs
seanyoung: Where is the key length check? The code below assumes the key is 32 bytes, see `let mut key: StorageKey = [0; 32];`

tests/substrate.rs
```diff
@@ -239,20 +243,25 @@ impl Externals for MockSubstrate {
...
+                Ok(pre_existing_len)
             }
             Some(SubstrateExternal::seal_set_storage) => {
-                assert_eq!(args.len(), 3);
+                assert_eq!(args.len(), 4);
 
                 let key_ptr: u32 = args.nth_checked(0)?;
-                let data_ptr: u32 = args.nth_checked(1)?;
-                let len: u32 = args.nth_checked(2)?;
+                //let key_len: u32 = args.nth_checked(1)?;
```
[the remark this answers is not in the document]
seanyoung: I thnk it's better to uncomment and check the value is 32, Solang uses 32 byte key lengths for now.

seanyoung (changes requested):

src/emit/substrate/mod.rs
```diff
@@ -22,6 +22,102 @@ pub(super) mod target;
...
+                    &[$key_ptr, $key_len, $value_ptr, $value_len]
+                )
+                .try_as_basic_value()
+                .left()
+                .unwrap()
+                .into_int_value()
+            };
+        }
+
+        #[allow(unused_macros)]
```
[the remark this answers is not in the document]
xermicus: rustc doesn't catch this one...

tests/substrate.rs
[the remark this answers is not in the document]
xermicus: Sorry, I forgot about this one earlier :see_no_evil:

seanyoung (approved):

src/emit/substrate/target.rs
```diff
@@ -1549,71 +1173,43 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
-            IntPredicate::EQ,
-            ret,
-            binary.context.i32_type().const_zero(),
-            "success",
-        );
+        // do the actual call
+        let ret = call!(
+            "seal_call",
+            &[
+                i32_zero!().into(), // TODO implement flags (mostyl used for proxy calls)
```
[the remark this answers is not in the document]
xermicus: thanks!

PR #1027, opened 2022-09-30
Function name mangling

xermicus:
Implement function name mangling in sema.

I used the simplest mangling algorithm possible: I think that basically re-using the `.signature` field of the function fits our use case well here:
* It should already be unique for each function
* Return values do not matter in the distinction of overloaded functions (in solidity the arguments must still be different).
* The mangled function names are very user friendly, since they contain the name together with the argument types.

But I might be missing something here, I'm open to change this if there is a better solution.


src/sema/contracts.rs
```diff
@@ -733,6 +733,49 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
...
+                &f.name, &f.signature, &f.abi_name
+            );
+            ns.diagnostics.push(ast::Diagnostic::error_with_note(
+                f.loc,
+                message,
+                ns.functions[*offender].loc,
+                "this function declaration conflicts with mangled name".into(),
+            ));
+        }
+    }
```
seanyoung, line 776: If mangled is an option, then this code can be much simpler.

```
loop over functions
   if the name is mangled
     find an unmangled name which matches
       generate an error
```
note that mangled named could also conflict which each other. 

```solidity
import "foo.sol" as X;

struct S { int foo; }

contract C {
       function foo(S s) public {}
       function foo(X.S s) public {}
}
```
Two different functons which both take struct S, but they are different.

seanyoung (changes requested):
This needs a documentation changes.

There is a fundamental problem though. Functions in one contract, may require mangling but not in another.

```solidity
contract A {
     function foo(bool x) public {}
}

contract B is A {
     function foo(int x) public {}
}
```

In contract B, foo should be mangled, but not in contract A. Your code always mangles a function name if it is overloaded in any contract.

src/sema/ast.rs
```diff
@@ -392,6 +398,15 @@ impl Function {
             format!("{} {}", self.ty, self.name)
         }
     }
+
+    pub fn mangle_name(&mut self) {
```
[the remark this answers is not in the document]
xermicus: This is in the `impl Function`, how would you mangle structs or enums?

src/sema/contracts.rs
```diff
@@ -733,6 +733,49 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
...
+                (
+                    pt::Visibility::Public(_) | pt::Visibility::External(_),
+                    pt::FunctionTy::Function,
+                ) => all_names
+                    .insert(function.name.clone(), *f)
+                    .map(|other| [*f, other]),
+                _ => None,
+            }
+        })
+        .flatten()
```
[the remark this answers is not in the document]
xermicus: `.map(|other| [*f, other])`

We want to mangle both names

src/sema/ast.rs
```diff
@@ -268,6 +268,8 @@ pub struct Function {
     pub symtable: Symtable,
     // What events are emitted by the body of this function
     pub emits_events: Vec<usize>,
+    // For overloaded functions this is the mangled (unique) name.
+    pub abi_name: String,
```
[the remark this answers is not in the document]
xermicus: My thoughts why I implemented it this way: Having a dedicated field for this is better because an `Option<String>` is confusing. Why is it optional, because it is not needed or because it is not possible to mangle it or another reason? In ABI generation should I use `name` or the mangled name? With `abi_name` it is clear you should always use that as the ABI name during ABI generationl. Aside that I think calling this field `mangled` is way too general and confusing.

src/sema/ast.rs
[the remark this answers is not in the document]
seanyoung: I mean function calls with structs or enums as arguments

src/sema/contracts.rs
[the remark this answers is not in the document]
seanyoung: ah of course, thanks

xermicus:
Thanks for the thorough review @seanyoung ! You found some interesting test cases I want to add, and I probably need to have a look at this implementation again :100:

> This needs a documentation changes.

Good point. def. gonna do that.

src/sema/ast.rs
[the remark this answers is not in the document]
xermicus: Hmm yeah I see now, my assumptions about the function signature were not entirely correct ([see this comment](https://github.com/hyperledger/solang/pull/1027#discussion_r985613095))

src/sema/contracts.rs
seanyoung: I can't reproduce this. What was the exact code that made the compiler fail with that assertion?

xermicus:
```
glow @ luna in solang on  function-name-mangling [$!?] is 📦 v0.1.13 via 🐍 v3.10.7 via 🦀 v1.64.0 
❯ cat bar.sol 
import "foo.sol" as X;

struct S { int foo; }

contract C {
       function foo(S s) public {}
       function foo(X.S s) public {}
}

glow @ luna in solang on  function-name-mangling [$?] is 📦 v0.1.13 via 🐍 v3.10.7 via 🦀 v1.64.0 
❯ cat foo.sol
struct S { bool s; }


glow @ luna in solang on  function-name-mangling [$?] is 📦 v0.1.13 via 🐍 v3.10.7 via 🦀 v1.64.0 
❯ cargo run -- compile --target substrate bar.sol
   Compiling solang v0.1.13 (/home/glow/code/solang)
    Finished dev [unoptimized + debuginfo] target(s) in 12.05s
     Running `target/debug/solang compile --target substrate bar.sol`
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Some(FunctionValue { name: "C::C::function::foo__S", address: 0x555a6439a648, is_const: true, is_null: false, llvm_value: "declare internal i32 @\"C::C::function::foo__S\"({ i256 }*)\n", llvm_type: "i32 ({ i256 }*)" })`,
 right: `None`', src/emit/functions.rs:35:13
```

I'm on the latest commit of the `function-name-mangling` branch (no additional changes)

seanyoung: On my branch I get:

```
[sean@xywoleh issue]$ cargo run compile --target substrate  bar.sol
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `/home/sean/git/solang/target/debug/solang compile --target substrate bar.sol`
error: function 'foo' overrides function in same contract
  ┌─ /home/sean/git/solang/issue/bar.sol:7:8
  │
6 │        function foo(S s) public {}
  │        ------------------------ previous definition of 'foo'
7 │        function foo(X.S s) public {}
  │        ^^^^^^^^^^^^^^^^^^^^^^^^^^

```

xermicus:
I made a few changes to this. All functions have now a mangled name. `sema` checks it to make sure it does not conflict with other functions names. And since all functions live in the contracts namespace, the decision whether the mangled name should be made in the ABI generation as well. 

Regarding the issue with the weird way how ETH does signatures for struct/enum type arguments. This is a deeper issue and I'd rather not hijack this PR to solve that as well. In the context of mangling, I think we can not run into this issue anyways, because the overloading will fail first hand.

@seanyoung you can have a look again

docs/language/functions.rst
```diff
@@ -353,6 +353,12 @@ values. Here is an example of an overloaded function:
 In the function foo, abs() is called with an ``int64`` so the second implementation
 of the function abs() is called.
 
+.. note::
+    Some target runtimes require unique function names (e.g. Substrate). In that case,
+    the names of overloaded functions will be mangled in the ABI as follows:
+    The function name will be concatenated with its argument type, sperated by underscores.
+    Tuple types are preceded with an additional underscore.
```
seanyoung, line 360: Tuples don't exist in Solidity, only in the eth abi encoding. I would say structs.

Also I would provide a much better description of what the mangled names look like, without depending too much on how the signature is written. This will help users understand what their mangled name will be. Also include some examples and also mention how arrays, enums. structs etc are mangled.

docs/language/functions.rst
```diff
@@ -353,6 +353,12 @@ values. Here is an example of an overloaded function:
 In the function foo, abs() is called with an ``int64`` so the second implementation
 of the function abs() is called.
 
+.. note::
+    Some target runtimes require unique function names (e.g. Substrate). In that case,
```
seanyoung, line 357: I would simplify this to "On Substrate, public functions must be unique, so they cannot be overloaded. "..

src/sema/ast.rs
```diff
@@ -319,6 +321,12 @@ impl Function {
             Some(pt::Mutability::Constant(loc)) => Mutability::View(loc),
         };
 
+        let mangled_name = signature
+            .trim()
```
seanyoung, line 325: `trim()` should be unnecessary, there should be no whitespace in the signature (it would be a bug if there was)

src/sema/contracts.rs
```diff
@@ -733,6 +733,39 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
...
+                &f.name, &f.signature, &f.mangled_name
+            );
+            ns.diagnostics.push(ast::Diagnostic::error_with_note(
+                f.loc,
+                message,
+                ns.functions[*offender].loc,
+                "this function declaration conflicts with mangled name".into(),
+            ))
+        }
+    }
```
seanyoung, line 766: We should have a test for when a function in a base contract conflicts with a mangled name,

seanyoung (changes requested):

src/sema/contracts.rs
xermicus: there is, see `tests/contract_testcases/substrate/functions/mangling_01.dot`

docs/language/functions.rst
xermicus: >  description of what the mangled names look like, without depending too much on how the signature 

Hmm, I tried exactly that by writing `function name will be concatenated with its argument type` :grin: I am not sure how I could make that more general (except that I wanted to write `with all of its argument types`)?

I can include some examples. I think together with the examples someone must be brain-dead to not get it.

seanyoung: well enums are replaced with `uint8` (I think), structs with a list of members (with an extra underscore), and I don't know arrays are handled (fixed length or dynamic). None of that is mentioned

docs/language/functions.rst
seanyoung: Sure. IMO `To fulfill this requirement,` is superfluous though.

docs/language/functions.rst
```diff
@@ -353,6 +353,32 @@ values. Here is an example of an overloaded function:
 In the function foo, abs() is called with an ``int64`` so the second implementation
 of the function abs() is called.
 
+.. note::
+  The substrate target runtime requires function names to be unique.
+  Overloaded function names will be mangled in the ABI. 
+  The function name will be concatenated with all of its argument types, seperated by underscores.
```
seanyoung, line 359: separated

docs/language/functions.rst
```diff
@@ -353,6 +353,32 @@ values. Here is an example of an overloaded function:
...
+  .. code-block:: solidity
+
+    enum E { v1, v2 }
+    struct S { int256 i; bool b; address a; }
+
+    contract C {
+        // foo_
+        function foo() public pure {}
+
+        // foo_uint256_address[]
```
seanyoung, line 374: I don't think `[]` wil be allowed in metadata message names. What will the function name in typescript/rust?

src/sema/contracts.rs
```diff
@@ -733,6 +733,39 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
...
+/// Only applies to public functions.
+///
+/// Note: In sema we do not care about the function name too much.
+/// The mangled name is consumed later by the ABI generation.
+pub fn mangle_function_names(contract_no: usize, ns: &mut ast::Namespace) {
+    let public_functions: Vec<usize> = ns.contracts[contract_no]
+        .all_functions
+        .keys()
+        .copied()
+        .filter(|f| ns.functions[*f].is_public() && ns.functions[*f].ty == pt::FunctionTy::Function)
```
seanyoung, line 746: total nitpick. put the `copied()` after the `filter()`

seanyoung (changes requested):

docs/language/functions.rst
xermicus: Are you sure? In polkadot.js/apps and the contracts UI, this works. The names are represented as `fooUint256Address[]` resp. `fooUint8[2]Int256BoolAddress`, so this seems fine.

src/sema/contracts.rs
xermicus: Without it there would be a double deref `**f`. Besides it making for ugly code, I wouldn't bet on that having the better performance at all (`usize` is primitive)

docs/language/functions.rst
xermicus: I thought a bit about this, but couldn't find a satisfying solution. We could just remove the square brackets, but then dynamic arrays would look the same as non-array types. I'm not sure what a good substitute would be either.

seanyoung: It may work in the contracts UI, but what about the using a contract in typescript using `@polkadot/api-contract`? What will the function be called, i.e. `contract.query.fooUint256Address[]` is not valid in typescript.

How about replacing `[` with `_a` and `]` with ``.

src/sema/contracts.rs
seanyoung: Fair point.

docs/language/functions.rst
```diff
@@ -353,6 +353,35 @@ values. Here is an example of an overloaded function:
...
 
+.. note::
+  The substrate target runtime requires function names to be unique.
+  Overloaded function names will be mangled in the ABI.
+  The function name will be concatenated with all of its argument types, separated by underscores.
+  Additionally, these rules apply:
+  - Struct types are represented by their field types (preceded with an extra underscore).
+  - Enum types are represented as their underlying uint8 type.
+  - Array types are recognizable by having ``Array`` appended. 
+  - Fixed size arrays will additionally have their length appended as well.
```
seanyoung, line 364: This list is not rendered correctly, see
image

docs/language/functions.rst
```diff
@@ -353,6 +353,35 @@ values. Here is an example of an overloaded function:
 In the function foo, abs() is called with an ``int64`` so the second implementation
 of the function abs() is called.
 
+.. note::
+  The substrate target runtime requires function names to be unique.
+  Overloaded function names will be mangled in the ABI.
+  The function name will be concatenated with all of its argument types, separated by underscores.
+  Additionally, these rules apply:
+  - Struct types are represented by their field types (preceded with an extra underscore).
+  - Enum types are represented as their underlying uint8 type.
```
seanyoung, line 362: Double quote uint8
```
``uint8``
```

seanyoung (changes requested):

docs/language/functions.rst
```diff
@@ -353,6 +353,35 @@ values. Here is an example of an overloaded function:
...
+  .. code-block:: solidity
+    
+    enum E { v1, v2 }
+    struct S { int256 i; bool b; address a; }
+
+    contract C {
+        // foo_
+        function foo() public pure {}
+
+        // foo_uint256_addressArray2Array
```
LucasSte, line 377: Have you tested naming it as `foo_uint256_Address[2][]`? It appears clearer when we use square brackets for arrays.

src/abi/mod.rs
```diff
@@ -39,3 +43,16 @@ pub fn generate_abi(
         }
     }
 }
+
+/// Returns a set of all non-unique public function names in a given contract.
+/// These names should not be used in the metadata. Instead, the mangled versions should be used.
+pub fn non_unique_function_names(contract_no: usize, ns: &Namespace) -> HashSet<&String> {
```
LucasSte, line 49: This function can be `pub(super)`. I guess you don't need to use the `pub` prefix at all.

src/sema/contracts.rs
```diff
@@ -733,6 +733,39 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
...
+        .collect();
+
+    for f in &public_functions {
+        if let Some(offender) = public_functions
```
LucasSte, line 750: This is a slow `O(n^2)` search that can be simplified. I just don't know if we need this simplification right now. Any thoughts @seanyoung and @xermicus ?

LucasSte (changes requested):

docs/language/functions.rst
seanyoung: `foo_uint256_Address[2][]` is not an identifier

src/sema/ast.rs
```diff
@@ -268,6 +268,8 @@ pub struct Function {
     pub symtable: Symtable,
     // What events are emitted by the body of this function
     pub emits_events: Vec<usize>,
+    // For overloaded functions this is the mangled (unique) name.
```
seanyoung, line 271: Ideally this should be a doccomment `///` (the one above too)

seanyoung (approved):
Ideally we would have a substrate integration test which uses the mangled name.

src/sema/contracts.rs
```diff
@@ -733,6 +733,39 @@ fn check_inheritance(contract_no: usize, ns: &mut ast::Namespace) {
     ns.diagnostics.extend(diagnostics);
 }
 
+/// Given a contract number, check for function names conflicting with any mangled name.
+/// Only applies to public functions.
+///
+/// Note: In sema we do not care about the function name too much.
+/// The mangled name is consumed later by the ABI generation.
+pub fn mangle_function_names(contract_no: usize, ns: &mut ast::Namespace) {
```
LucasSte, line 741: This function can be private.

LucasSte (approved):

PR #1029, opened 2022-10-04
Implement function dispatch in codegen for Solana

LucasSte:
This PR implements the function dispatch and the constructor dispatch in codegen for the Solana target. The old implementation in `src/emit/dispatch` has been moved to `src/emit/substrate/dispatch` and implemented as methods of `SubstrateTarget` because they are now Substrate specific.

There are some new instructions in the control flow graph for LLVM instructions that were exclusive to `emit`.


src/codegen/cfg.rs
```diff
@@ -1308,48 +1367,35 @@ pub fn generate_cfg(
...
-    if !func.modifiers.is_empty() {
-        // only function can have modifiers
-        assert_eq!(func.ty, pt::FunctionTy::Function);
-        let public = cfg.public;
-        let nonpayable = cfg.nonpayable;
-
-        cfg.public = false;
+    if let Some(func_no) = function_no {
+        // if the function is a modifier, generate the modifier chain
+        if !ns.functions[func_no].modifiers.is_empty() {
```
seanyoung, line 1372: would the code be slightly easier to read if we had:
```rust
let func = &ns.functions[func_no]:
```
rather than repeating `ns.functions[func_no]` again and again?

src/codegen/cfg.rs
```diff
@@ -1308,48 +1367,35 @@ pub fn generate_cfg(
...
-    // if the function is a modifier, generate the modifier chain
-    if !func.modifiers.is_empty() {
-        // only function can have modifiers
-        assert_eq!(func.ty, pt::FunctionTy::Function);
-        let public = cfg.public;
-        let nonpayable = cfg.nonpayable;
-
-        cfg.public = false;
+    if let Some(func_no) = function_no {
+        // if the function is a modifier, generate the modifier chain
```
seanyoung, line 1371:
```suggestion
        // if the function has any modifiers, generate the modifier chain
```

src/codegen/cfg.rs
```diff
@@ -1308,48 +1367,35 @@ pub fn generate_cfg(
 
     let mut cfg = function_cfg(contract_no, function_no, ns, opt);
 
-    let default_constructor = &ns.default_constructor(contract_no);
-    let func = match function_no {
-        Some(function_no) => &ns.functions[function_no],
-        None => default_constructor,
-    };
```
seanyoung, line 1315: Getting rid of this is very nice, thanks!

seanyoung (changes requested):
Looks great, I can't find any problems apart from some tiny nitpicks.

src/codegen/cfg.rs
LucasSte: I've used `func` whenever possible, but after calling `generate_modifier_dispatch` I cannot use it any more because I borrow `ns` as mutable.

src/codegen/cfg.rs
```diff
@@ -176,6 +179,34 @@ pub enum Instr {
     },
     /// Do nothing
     Nop,
+    /// Return AbiEncoded data via an environment system call
+    ReturnData {
+        data: Expression,
+        data_len: Expression,
+    },
+    /// Return a code at the end of a function
+    ReturnCode { code: ReturnCode },
```
xermicus, line 188: Just a hint / minor nitpick: For variants with single values, e.g. `ReturnCode(ReturnCode)` would make for easier syntax. But it's not a must for me in this case.

src/codegen/mod.rs
```diff
@@ -237,6 +239,23 @@ fn contract(contract_no: usize, ns: &mut Namespace, opt: &Options) {
             ns.contracts[contract_no].default_constructor = Some((func, cfg_no));
         }
 
+        // TODO: This is a temporary solution. Once Substrate's dispatch moves to codegen,
```
xermicus, line 242: I noticed you implemented new mod `dispatch.rs` together with some changes already in the substrate emit already; what would be left to make this `TODO` happen? And what do you think how pressing would this be for me to take care of?

src/codegen/cfg.rs
LucasSte: Will fix that

src/codegen/cfg.rs
```diff
@@ -176,6 +179,34 @@ pub enum Instr {
...
+/// from a function.
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub enum ReturnCode {
+    Success,
+    FunctionSelectorInvalid,
+    AbiEncodingInvalid,
+    InvalidDataError,
+}
+
+impl Display for ReturnCode {
```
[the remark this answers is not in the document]
LucasSte: I liked the second one. I'll use that. Thanks for the hint!

src/codegen/mod.rs
LucasSte: The old `emit/dispatch.rs` has been moved to `emit/substrate/dispatch.rs` because it is now Substrate specific. The changes I made in the file were only supposed to remove Solana specific code that is not needed anymore. 

If you have a look at the new `codegen/dispatch.rs`, you'll see that the file is very similar to `emit/substrate/dispatch.rs` with few exceptions that are substrate specific. You'll need then to implement these details and build a call to the new dispatch functions wherever `emit_function_dispatch` function is called. The idea I had for Substrate was to just reuse the Solana code in codegen and build a function call in LLVM-IR to the dispatch we generate in codegen. This is the easiest way, but not the prettiest one in my opinion.

src/codegen/cfg.rs
LucasSte: Actually, @xermicus I'm not going to change this. If you have a look at all the other elements of `Instr`, they all have named parameters even if the variant has only one attribute. I'm trying to maintain consistency.

xermicus: Sure, I favor consistency over micro optimizations any time

src/codegen/mod.rs
xermicus: Alright, that sounds like a plan. I'll look into this later on when I find some time. Thanks! :heart:

seanyoung (changes requested):
Great improvement. Very nice to dispatch in `--emit cfg`

src/codegen/cfg.rs
```diff
@@ -176,6 +179,24 @@ pub enum Instr {
...
+        data_len: Expression,
+    },
+    /// Return a code at the end of a function
+    ReturnCode { code: ReturnCode },
+}
+
+/// This struct defined the return codes that we send to the execution environment when we return
+/// from a function.
+#[derive(PartialEq, Eq, Hash, Clone, Debug, Display)]
+#[display("title case")]
```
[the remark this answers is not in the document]
LucasSte: Now it should be fixed. The right attribute was `#[display(style="title case")]`.

seanyoung (approved):

xermicus (approved):

PR #1031, opened 2022-10-04
Split integration tests

LucasSte:
Solana integration tests and Anchor integration tests are semantically different, but run in a single GitHub runner. This PR splits them to ease debugging.


integration/solana/anchor/call_anchor.spec.ts
```diff
@@ -1,5 +1,5 @@
 import expect from 'expect';
-import { loadContract } from './setup';
+import { loadContract } from '../setup';
```
seanyoung, line 2: Why are you moving this file to a sub directory?

integration/solana/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Integration tests with Solang and Solana",
   "scripts": {
     "test": "tsc; ts-node setup.ts; mocha --parallel *.spec.ts",
+    "test-anchor": "tsc; ts-node setup.ts; mocha --parallel anchor/*spec.ts",
```
seanyoung, line 7: So now `yarn test` (or `npm test`) no longer tests everything. How is that right?

How about three scripts:
test: tests everything, as you would expect
test-call-anchor: test the call anchor test
test-other: test the rest

xermicus (approved):

.github/workflows/test.yml
```diff
@@ -256,7 +256,40 @@ jobs:
...
     - name: Create Solidity interface file for Anchor
-      run: solang idl ../anchor/target/idl/anchor.json
+      run: solang idl ../../anchor/target/idl/anchor.json
+      working-directory: ./integration/solana/anchor
+    - name: Build Solang contracts
+      run: solang compile --target solana *.sol -v -o ../
+      working-directory: ./integration/solana/anchor
+    - name: Deploy and test contracts
+      run: npm run test-anchor
+      working-directory: ./integration/solana
```
seanyoung, line 266: Does this test really work, there is no test validator running. Also the test seem broken (I don't now why)

seanyoung (changes requested):

integration/solana/package.json
xermicus: I understood it so that just specific anchor tests were move away from testing everything. But your suggestion about having 3 test targets would be even better

LucasSte: Before having Anchor tests, we would run `npm run build` and `npm run test`. The new Anchor tests cannot follow the same order of commands to be run, so I believe having a single command for both Solana and Anchor tests is not a good idea. The Anchor test needs many setup commands before we can run `npm run test-anchor`.

Perhaps, it is a better solution to move the Anchor Typescript test to `integration/anchor`. What do you think @seanyoung and @xermicus ?

.github/workflows/test.yml
LucasSte: Yes, the test works. Please, see the CI output. The validator is run with the command `nohup solana-test-validator -q &`.

integration/solana/anchor/call_anchor.spec.ts
LucasSte: Anchor tests and Solana tests are semantically different and require a different preparation pipeline to run.

.github/workflows/test.yml
LucasSte: For some reason, the tests are not shown in the PR, but they ran here:
https://github.com/hyperledger/solang/actions/runs/3189875558/jobs/5204512205

integration/solana/anchor/call_anchor.spec.ts
seanyoung: So now we have some files in `integration/anchor` and the rest of them in `integration/solana/anchor`, for one "semantic" test.

integration/solana/package.json
seanyoung: This might be a good idea.

LucasSte: > This might be a good idea.

This should be fixed now.

.github/workflows/test.yml
LucasSte: Everything ran fine now.

src/bin/idl/mod.rs
```diff
@@ -20,7 +20,7 @@ use std::{
 pub fn idl(matches: &ArgMatches) {
     let files = matches.get_many::<OsString>("INPUT").unwrap();
 
-    let output = matches.get_one::<OsString>("OUTPUT").map(PathBuf::from);
+    let output = matches.get_one::<String>("OUTPUT").map(PathBuf::from);
```
seanyoung, line 23: This shouldn't be `String`, that would break non-utf8 paths. If anything this should be `PathBuf`.

seanyoung (changes requested):
Just one comment

src/bin/idl/mod.rs
LucasSte: If the return value is not a string, the compiler panics with the following error:
`thread 'main' panicked at 'Mismatch between definition and access of `OUTPUT`. Could not downcast to std::ffi::os_str::OsString, need to downcast to alloc::string::String
', src/bin/idl/mod.rs:23:26`

This is because the definition of the command does not specify any type, so it defaults to string:

https://github.com/hyperledger/solang/blob/e827fe52b89cb86d4db45100d6e94c29ef68a49f/src/bin/solang.rs#L104-L110

In addition, throughout the code we refer to the output as String, like here:

https://github.com/hyperledger/solang/blob/e827fe52b89cb86d4db45100d6e94c29ef68a49f/src/bin/solang.rs#L347-L353

and here:

https://github.com/hyperledger/solang/blob/e827fe52b89cb86d4db45100d6e94c29ef68a49f/src/bin/solang.rs#L518-L525

Is it time to change this everywhere?

seanyoung: The right thing to do is change it everywhere. The declaration of the output arg needs either `.value_parser(ValueParser::path_buf())` or`.value_parser(ValueParser::os_string()`, depending on which makes more sense.

seanyoung (approved):
Very nice. Thanks for cleaning that up!

PR #1033, opened 2022-10-06
Bump various dependencies and depend on rust 1.63.0

seanyoung:
- Require rust 1.63 (needed by solana_rbpf crate, also phf requires rust 1.60.0)
- Bump dependencies
- Fix various clippies found with rust 1.65.0 beta


xermicus (approved):
Thank you very much!

PR #1036, opened 2022-10-10
Faster checkouts and use the same version of rust everywhere

seanyoung:
Use actions/checkout@v3 and remove fetch-depth. This means that git
checkout is done without any history. So, `git describe --tags` won't
work without the `--always` option.

As a result, `solang --version` will just give you the git commit
hash for solang built in CI. Release builds are not affected, because
the single commit which is fetched will contain the tag.

This makes the linux arm job about 3 minutes faster.

Use the same rust version for all CI jobs (1.63.0)


LucasSte (approved):

xermicus (approved):
LGTM

PR #1038, opened 2022-10-13
Encode arguments for constructor call in codegen

LucasSte:
Presently, when we call a constructor, we encode the arguments in emit, using the deprecated EthAbi encoding. A prior step to moving to Borsh is encoding all arguments during codegen. This PR refactors the `Instr::Constructor` to accommodate such changes, alongside the necessary adaptations in emit and codegen.


seanyoung (approved):
Very nice

xermicus (approved):

PR #1039, opened 2022-10-14
Implement more constructor dispatch in cfg

seanyoung:


src/codegen/cfg.rs
```diff
@@ -197,6 +197,7 @@ pub enum ReturnCode {
     FunctionSelectorInvalid,
     AbiEncodingInvalid,
     InvalidDataError,
+    AccountDataToSmall,
```
LucasSte, line 200: `to small` or `too small`? 🤔

src/codegen/dispatch.rs
```diff
@@ -358,3 +381,116 @@ pub(super) fn constructor_dispatch(
...
+                )),
+                Box::new(Expression::NumberLiteral(
+                    Loc::Codegen,
+                    Type::Uint(32),
+                    BigInt::zero(),
+                )),
+            )),
+            2,
+        )],
+    );
```
LucasSte, line 426: I wonder if someday we can simplify this expression. It is too complex to be understood easily.

src/codegen/cfg.rs
LucasSte:
```suggestion
    AccountDataTooSmall,
```

LucasSte: There are no tests for this new return code.

stdlib/solana.c
```diff
@@ -82,19 +82,28 @@ uint64_t external_call(uint8_t *input, uint32_t input_len, SolParameters *params
...
+
     for (int account_no = 0; account_no < params->ka_num; account_no++)
     {
         const SolAccountInfo *acc = &params->ka[account_no];
 
         if (SolPubkey_same(dest, acc->key))
         {
+            metas[0].pubkey = acc->key;
+            metas[0].is_writable = acc->is_writable;
+            metas[0].is_signer = acc->is_signer;
```
LucasSte, line 95: Don't we need tests for these modifications?

stdlib/solana.c
```diff
@@ -82,19 +82,28 @@ uint64_t external_call(uint8_t *input, uint32_t input_len, SolParameters *params
...
         }
+        else
+        {
 
-        metas[account_no].pubkey = acc->key;
-        metas[account_no].is_writable = acc->is_writable;
-        metas[account_no].is_signer = acc->is_signer;
+            metas[meta_no].pubkey = acc->key;
+            metas[meta_no].is_writable = acc->is_writable;
+            metas[meta_no].is_signer = acc->is_signer;
```
LucasSte, line 104: Aren't tests necessary for this?

LucasSte (changes requested):
Aren't tests missing for this PR?

src/codegen/dispatch.rs
```diff
@@ -276,6 +281,7 @@ fn add_dispatch_case(
 /// Create the dispatch for a contract constructor. This case needs creates a new function in
 /// the CFG because we want to use de abi decoding implementation from codegen.
```
xermicus, line 282:
```suggestion
/// Create the dispatch for a contract constructor. This case creates a new function in
/// the CFG because we want to use the abi decoding implementation from codegen.
```

xermicus (changes requested):

src/codegen/dispatch.rs
xermicus: > I wonder if someday we can simplify this expression. It is too complex to be understood easily.

As a start all the `Box::new()` could be replaced by `.into()`, I think this would help it already

seanyoung: > As a start all the `Box::new()` could be replaced by `.into()`, I think this would help it already

Very nice, I did not realise that. Thanks @xermicus

seanyoung: > I wonder if someday we can simplify this expression. It is too complex to be understood easily.

I have been wondering about this. We are generating a lot more CFG, and we will be generating more in the future. I am sure some rust macros would help here, but I'm ensure how to approach this.

We could have a rust `quote! {}` style approach, so this could be written:
```rust
let expr = sol_expr! { tx.accounts[0].data.length };
```
However implementing this is hard: this would mean:
 - Exporting the expression parser (possible)
 - Using sema on the exprression
 - Using cfg expression on the result

This could even work on statements.

Alternatively we could have some simple `macro_rules! {}` which would aid in the build-up of cfg expressions/instructions. I am not sure what they would look like, suggestions welcome.

stdlib/solana.c
seanyoung: So the input for the solana program is now:
 - data account for contract
 - data account for sender (=broken)
 - value sent (no longer used)
 - seeds (soon to be removed)
 - hash of contract name to be called (soon to be removed)
 - abi encoded stuff

This PR now also passes the data account via the accounts, and the account must the first. This is now enforced on this line:

https://github.com/hyperledger/solang/pull/1039/files#diff-3b88924fae67d4d137158f036752a5f2b28f3fe12d968fc17c49f0ff7311b110R54

This ensures that the data account passed via the accounts is the same as the first as the input. So, any test which runs on a VM tests this.

(Note that the aim is remove all this stuff from the input, and I think we are near to that).

stdlib/solana.c
seanyoung: Same as above.

src/codegen/dispatch.rs
```diff
@@ -358,3 +381,114 @@ pub(super) fn constructor_dispatch(
...
+    );
+
+    cfg.set_basic_block(not_enough);
+
+    cfg.add(
+        vartab,
+        Instr::ReturnCode {
+            code: ReturnCode::AccountDataTooSmall,
+        },
+    );
```
LucasSte, line 455: I understood that your modifications in `solana.c` are tested in the runtime tests. I still feel that if I remove these lines, no test will fail.

src/codegen/dispatch.rs
xermicus: Macro rules would be an easier start, maybe something like I did a short while ago in the emit?

It would be awesome if we would have a proc macro for generating the whole tree, but I fear it is a lot of work we have no capacity for just yet...

xermicus (approved):

src/codegen/dispatch.rs
seanyoung: So there are tests in the `integration/solana/...` for this, but there are no unit tests. I've added a test for this.

tests/solana.rs
```diff
@@ -312,9 +322,10 @@ fn serialize_parameters(
         serialize_account(&mut v, &mut refs, acc, data);
     }
 
+    // then the rest of the accounts
     for (acc, data) in &vm.account_data {
         //println!("acc:{} {}", hex::encode(acc), hex::encode(&data.0));
```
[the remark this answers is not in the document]
seanyoung: Removed

src/codegen/dispatch.rs
```diff
@@ -358,3 +381,114 @@ pub(super) fn constructor_dispatch(
...
+
+    // Calculate heap offset (align on 8 byte boundary)
+    let fixed_fields_size = ns.contracts[contract_no]
+        .fixed_layout_size
+        .to_u64()
+        .unwrap();
+
+    let heap_offset = (fixed_fields_size + 7) & !7;
```
LucasSte, line 479: Just as a curiosity, what does this operation do?

LucasSte (approved):

src/codegen/dispatch.rs
seanyoung: This aligns the heap on an 8 byte boundary. I'll add a comment

PR #1041, opened 2022-10-19
remove the random builtin function

xermicus:
It is unsafe for most use-cases and will be deprecated in `ink!`. Contract authors should rather use VRF from chain extensions or oracles.

The `seal_random` API is still used to calculate the salt on instantiation (if it is missing). There will be a new API `seal_nonce` which can be used to replace this. I'll do this in another PR once the new nonce API is available in pallets contract.

Side note: In my opinion it is completely safe to remove this built in. It was completely [broken](https://github.com/hyperledger/solang/pull/1025/commits/65018be65d3e95d9a728894e1ac4ee07d4a3aa7a) for IDK exactly how long but no one complained.


seanyoung (approved):

athei:
Thanks. Looks good to me.

seanyoung:
> Thanks. Looks good to me. I don't see `seal_random` being removed from the backend, though.

Would it make sense to keep this function in solang until the VRF is implemented?

Also out of curiosity, what was broken with the implementation of `seal_random()`? I thought it was very nice.

xermicus:
> Would it make sense to keep this function in solang until VRF is usable from solang?

Exactly. It is used for contract instantiation. I'll remove it as soon as the nonce API is available.

> Also out of curiosity, what was broken with the implementation of seal_random()? I thought it was very nice.

It is potentially predictable.

LucasSte:
> remove random builtin

This title seems like you have randomly chosen a builtin in Solang and removed it.

LucasSte (approved):
Please change your PR's title to something less ambiguous.
A suggestion would be `Remove Substrate 'random' builtin`.

PR #1042, opened 2022-10-19
Encode arguments for assert-failure in codegen

LucasSte:
Now that all runtime tests for Solana are working with bosh encoding, we can rid Solang of EthAbiEncoding. There is one more instruction that needed refactoring in order to make that a reality. `Instr::AssertFailure` utilizes the old abi encoding scheme implemented in emit (for Solana). This PR refactors the instruction so it encodes its arguments in codegen and allows a future integration with our `trait AbiEncoding`.

### DIFF
--- a/src/emit/instructions.rs
+++ b/src/emit/instructions.rs
@@ -424,7 +424,9 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
             );
             bin.builder.build_store(size_field, new_len);
         }
-        Instr::AssertFailure { expr: None } => {
+        Instr::AssertFailure {
+            encoded_args_with_len: None,
+        } => {
             target.assert_failure(
                 bin,
                 bin.context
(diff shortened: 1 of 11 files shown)


src/emit/instructions.rs
```diff
@@ -434,22 +436,14 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
...
-                &[v],
-                &[Type::String],
-                ns,
-            );
+        Instr::AssertFailure {
+            encoded_args_with_len: Some(expr),
+        } => {
+            let data = expression(target, bin, &expr.0, &w.vars, function, ns);
+            let vector_bytes = bin.vector_bytes(data);
+            let len = expression(target, bin, &expr.1, &w.vars, function, ns);
```
seanyoung, line 444: How about:
```rust
            let len = bin.vector_len(data);
```
And then we don't need the second expression in `Instr::AssertFailure { .. }`?

xermicus (approved):

src/emit/instructions.rs
LucasSte: That is also an option. I added both the buffer and the length in codegen, because the `encoder.abi_encode` function already returns the length of the buffer.

seanyoung (approved):
Any idea why it uses more compute units?

LucasSte:
> Any idea why it uses more compute units?

When you encode arguments in emit, you receive back the buffer (a pointer of type `i8*`) and the length. In codegen, we get a `struct vector`, so we need to use `gep` to retrieve the buffer before calling `target.assert_failure`.

PR #1044, opened 2022-10-19
Disallow try-catch statements on Solana

LucasSte:
Try-catch statements are not supported on Solana. This PR updates the documentation to inform so and adds an error message when someone tries to use such a construct on Solana.

PS: The intention for this PR is to avoid using ABI decoding/encoding for try-catches on Solana, a construction that requires a special case of decoding with a basic block to handle a decoding failure.

### DIFF
--- a/src/sema/statements.rs
+++ b/src/sema/statements.rs
@@ -15,6 +15,7 @@ use crate::sema::symtable::{VariableInitializer, VariableUsage};
 use crate::sema::unused_variable::{assigned_variable, check_function_call, used_variable};
 use crate::sema::yul::resolve_inline_assembly;
 use crate::sema::Recurse;
+use crate::Target;
 use solang_parser::pt;
 use solang_parser::pt::CatchClause;
 use solang_parser::pt::CodeLocation;
@@ -1810,6 +1811,17 @@ fn try_catch(
     ns: &mut Namespace,
     diagnostics: &mut Diagnostics,
 ) -> Result<(Statement, bool), ()> {
+    if ns.target == Target::Solana {
+        diagnostics.push(Diagnostic::error(
+            *loc,
+            "The try-catch statement is not supported on Solana. Please, go to \
+             https://solang.readthedocs.io/en/latest/language/statements.html#try-catch-statement \
+             for more information"
+                .to_string(),
+        ));
+        return Err(());
+    }
+
     let mut expr = expr.remove_parenthesis();
     let mut ok = None;
 
--- a/docs/language/statements.rst
+++ b/docs/language/statements.rst
@@ -218,6 +218,10 @@ An internal function cannot be called from a try catch statement. Not all proble
 for example, out of gas cannot be caught. The ``revert()`` and ``require()`` builtins may
 be passed a reason code, which can be inspected using the ``catch Error(string)`` syntax.
 
+.. warning::
+    On Solana, any transaction that fails halts the execution of a contract. The try-catch statement, thus,
+    is not supported for Solana contracts and the compiler will raise an error if it detects its usage.
+
 .. code-block:: solidity
 
     contract aborting {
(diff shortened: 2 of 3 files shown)


xermicus (changes requested):
> The intention for this PR is to avoid using ABI decoding/encoding for try-catches on Solana, a construction that requires a special case of decoding with a basic block to handle a decoding failure.

This does imply we should either implement propperly or disallow try-catch for the substrate target as well, or doesn't it?

src/sema/statements.rs
```diff
@@ -1810,6 +1811,17 @@ fn try_catch(
     ns: &mut Namespace,
     diagnostics: &mut Diagnostics,
 ) -> Result<(Statement, bool), ()> {
+    if ns.target == Target::Solana {
+        diagnostics.push(Diagnostic::error(
+            *loc,
+            "The try-catch statement is not supported on Solana. Please, check \
```
seanyoung, line 1817: I think `check` is too colloquial (here the usage of `check` sounds like `check it out`). Something like `read` or `go to` sounds better to me.

seanyoung:
@xermicus try catch is designed to work with cross contract calls. I thought that if the callee failed on substrate (e.g. `revert` or `divide by zero`) then the entire transaction fails on Substrate too.

Should `try catch` be disallowed for substrate too?

seanyoung:
> > The intention for this PR is to avoid using ABI decoding/encoding for try-catches on Solana, a construction that requires a special case of decoding with a basic block to handle a decoding failure.
> 
> This does imply we should either implement it propperly or disallow try-catch for the substrate target as well, or doesn't it?

The scale decoder does implement this, but our new borsh decoder does not.

xermicus:
> The scale decoder does implement this, but our new borsh decoder does not.

Alright then, I didn't know this, thanks :+1: 

> @xermicus try catch is designed to work with cross contract calls. I thought that if the callee failed on substrate (e.g. revert or divide by zero) then the entire transaction fails on Substrate too.

The `seal_call` API returns a return code, with variants like `CalleeReverted` and `CalleeTrapped`. Maybe they are not yet used? I didn't check this yet in codegen / emit

seanyoung:
> The `seal_call` API returns a return code, with variants like `CalleeReverted` and `CalleeTrapped`. Maybe they are not yet used? I didn't check this yet in codegen / emit

I don't think these return codes are used yet. However, it does mean that `try catch` is needed on Substrate. It may need some wiring up to make well with the new return codes.

xermicus:
It seems to be implemented:

https://github.com/hyperledger/solang/blob/main/src/emit/substrate/target.rs#L1133

xermicus (approved):

src/sema/statements.rs
LucasSte: Fixed

src/sema/statements.rs
```diff
@@ -1810,6 +1811,17 @@ fn try_catch(
     ns: &mut Namespace,
     diagnostics: &mut Diagnostics,
 ) -> Result<(Statement, bool), ()> {
+    if ns.target == Target::Solana {
+        diagnostics.push(Diagnostic::error(
+            *loc,
+            "The try-catch statement is not supported on Solana. Please, go to \
+             https://solang.readthedocs.io/en/latest/language/statements.html#try-catch-statement \
+             for more information"
```
seanyoung, line 1819: Nitpick, but I'd argue that `for more information` is superfluous.

seanyoung (approved):

src/sema/statements.rs
LucasSte: In this case, I believe that if I remove `for more information`, there is not a clear ending to the phrase. Have a look: 
`The try-catch statement is not supported on Solana. Please, go to <site>`

PR #1045, opened 2022-10-20
Add substrate builtins

xermicus:
This adds a "substrate" file to have custom builtins available, analog to the "solang" file.

Right now there is only the "Hash" type but there will be more in the future.

### DIFF
--- a/docs/language/builtins.rst
+++ b/docs/language/builtins.rst
@@ -326,6 +326,25 @@ are cast and checked against the function specified as the first argument.
         function bar(int a, bool b) public {}
     }
 
+Hash
+++++
+
+Only available on Substrate. It represents the ``Hash`` type from ``ink_primitives`` via user type definition.
+Its underlying type is ``bytes32``, but it will be reported correctly as the ``Hash`` type in the metadata.
+
+.. code-block:: solidity
+
+    import 'substrate';
+
+    contract c {
+        bytes32 current;
+
+        function set(Hash h) public returns (Hash) {
+            current = Hash.unwrap(h);
+            return Hash.wrap(current);
+        }
+    }
+
 Cryptography
 ____________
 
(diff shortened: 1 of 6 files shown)


src/sema/builtin.rs
```diff
@@ -1616,4 +1616,30 @@ impl Namespace {
...
+        self.files.push(File {
+            path: PathBuf::from("substrate"),
+            line_starts: Vec::new(),
+            cache_no: None,
+        });
+
+        // The Hash type from ink primitives.
+        let type_no = self.user_types.len();
+        self.user_types.push(UserTypeDecl {
+            tags: vec![],
```
seanyoung, line 1631: It might be useful to add some documentation tags here. I think what you need is:
```rust
tags: vec![Tag { Tag: "notice".into(), no: file_no, Value: "ink-like hash value" }],
```
This would be used by the language server.

docs/language/builtins.rst
```diff
@@ -326,6 +326,25 @@ are cast and checked against the function specified as the first argument.
         function bar(int a, bool b) public {}
     }
 
+Hash
+++++
+
+Only available on Substrate. It represents the ``Hash`` type from ``ink_primitives`` via user type definition.
```
LucasSte, line 332:
```suggestion
Only available on Substrate, it represents the ``Hash`` type from ``ink_primitives`` via user type definition.
```

docs/targets/substrate.rst
```diff
@@ -25,3 +25,26 @@ directory. Write this to flipper.sol and run:
...
+
+Builtin Imports
+________________
+
+Some builtin functionality is only available after importing. The following types
+can be imported via the special import file ``substrate``.
+
+.. code-block:: solidity
+
+    import {Hash} from 'solana';
```
LucasSte, line 37:
```suggestion
    import {Hash} from 'substrate';
```

LucasSte (changes requested):

LucasSte (approved):

PR #1047, opened 2022-10-24
Print the return values of seal API calls into the debug buffer

xermicus:
Currently there is no way for devs to know more about why an API call might fall. A prime example to this is `seal_instantiate`, where contract creation can fail for multiple reasons. 

In `ink!` this is done too.

It is turned off by default. ~~Where can I document this? Closest I found is optimizer passes but this is not really an optimization. Can I add a new page under `Extras`?~~


src/bin/solang.rs
```diff
@@ -177,6 +177,12 @@ fn main() {
                             .short('g')
                             .long("generate-debug-info")
                             .hide(true),
+                    )
+                    .arg(
+                        Arg::new("LOGAPICALLS")
```
seanyoung, line 182: `LOGAPIRETURNS` ?

src/bin/solang.rs
```diff
@@ -177,6 +177,12 @@ fn main() {
...
+                            .help("Log the return codes of runtime API calls in the environment")
```
seanyoung, line 183: How do you feel about `syscalls` rather than `runtime API calls`? I am wondering if that is an solana-ism though, not sure if the `seal_*` calls are called syscalls in the Substrate world.

src/emit/substrate/mod.rs
```diff
@@ -1851,3 +1853,47 @@ fn event_id<'b>(
...
+    emit_context!(binary);
+
+    let fmt = format!("{}=", api);
+    let msg = fmt.as_bytes();
+    let length = i32_const!(msg.len() as u64 + 16);
+    let out_buf = binary.vector_new(length, i32_const!(1), None);
+    let out_buf_ptr = binary.vector_bytes(out_buf.into());
+    let mut out_buf_offset = out_buf_ptr;
+
+    let msg_string = binary.emit_global_string("seal_instantiate", msg, true);
```
seanyoung, line 1872: Why `seal_instantiate`?

seanyoung:
> 
> It is turned off by default. Where can I document this? Closest I found is optimizer passes but this is not really an optimization. Can I add a new page under `Extras`?

I think the "Optimizations" section should be renamed "Code generation options"

src/bin/solang.rs
xermicus: Hmm, I don't really like `syscalls`. I see those more as API calls into the on, off or whatever chain environment of the target?

src/bin/solang.rs
```diff
@@ -177,6 +177,12 @@ fn main() {
                             .short('g')
                             .long("generate-debug-info")
                             .hide(true),
+                    )
+                    .arg(
+                        Arg::new("LOGAPIRETURNS")
+                            .help("Log the return codes of runtime API calls in the environment")
+                            .long("log-api-calls")
```
seanyoung, line 184: How about `--log-api-returns` or `--log-api-return-codes` ?

I imagine at some point we may want another option which logs the api calls with their arguments, which would be called `--log-api-calls`, I suppose.

src/bin/solang.rs
```diff
@@ -177,6 +177,12 @@ fn main() {
...
+                            .action(ArgAction::SetFalse),
```
seanyoung, line 185: This is not right. The options above like `--no-cse` disable CSE, so those have `.action(ArgAction::SetFalse)`. In this case, want to enable something, so it should be `.action(ArgAction::SetTrue)`.

Actually it might be best to just remove the entire `.action(..)` line

src/codegen/mod.rs
```diff
@@ -90,6 +90,7 @@ pub struct Options {
     pub common_subexpression_elimination: bool,
     pub generate_debug_information: bool,
     pub opt_level: OptimizationLevel,
+    pub log_api_return_codes: bool,
 }
 
 impl Default for Options {
```
seanyoung, line 96: `Options` is not a great name. I was thinking `CodegenOptions` but this is confusing because it also affects things in emit.

docs/code_gen_options.rst
```diff
@@ -218,4 +218,26 @@ always true, so the check is omitted.
 This also means that, whenever the length of an array is accessed using '.length', it is replaced with a constant.
 
 Note that this optimization does not cover every case. When an array is passed
-as a function argument, for instance, the length is unknown.
\ No newline at end of file
+as a function argument, for instance, the length is unknown.
```
seanyoung, line 221: Please also change the first line and first paragraph of this file so it talks about code generation options, not just optimizations.

seanyoung (changes requested):
Bunch of nitpicks, sorry :grimacing:

src/bin/solang.rs
xermicus: ACK

src/codegen/mod.rs
xermicus: I agree its confusing also because there are the opt levels in there. But would it be OK for you to have this refactored in another PR?

I can add that of my growing list of things I'd like to refactor probably soon anyways

seanyoung: Of course, not sure what it should be called anyway. That was more of a note than review comment anyway

docs/code_gen_options.rst
```diff
@@ -218,4 +218,26 @@ always true, so the check is omitted.
...
+
+Debugging Options
+=================
+
+It may be desirable to have access to debug information regarding the contract execution. 
+How-ever, this might lead to less ideal code as well as much higher gas usage. Hence it is disabled by default,
```
LucasSte, line 227:
```suggestion
However, this might lead to more instructions as well as higher gas usage. Hence, it is disabled by default,
```
I think `less ideal` sounds awkward, so this suggestion is optional. Please, be sure to remove the dash from `how-ever` and place a comma after `hence`.

docs/code_gen_options.rst
```diff
@@ -218,4 +218,26 @@ always true, so the check is omitted.
...
+but can be enabled using CLI flags. Just make sure to not use them in production deployments.
```
LucasSte, line 228:
```suggestion
but can be enabled using CLI flags. Just make sure not to use them in production deployments.
```

docs/code_gen_options.rst
```diff
@@ -218,4 +218,26 @@ always true, so the check is omitted.
...
+----------------------------
+
+Runtime API calls are not necessarely guaranteed to succeed under all possible circumstances.
+By design, the low revel results of these calls might be abstracted away in Solidity.
+But for development purposes, it might be desirable to observe the low level result of such calls directly. 
+The ``--log-api-calls`` flag will make the contract to print the low-level result of runtime calls (if there is any).
+
+.. note::
+
+    For now, this will only have effect when used with the ``Substrate`` target.
```
LucasSte, line 242:
```suggestion
    For now, this only works for the ``Substrate`` target.
```
I think this sounds better.

LucasSte (changes requested):

docs/code_gen_options.rst
```diff
@@ -218,4 +227,26 @@ always true, so the check is omitted.
...
+However, this might lead to more instructions as well as higher gas usage. Hence, it is disabled by default,
+but can be enabled using CLI flags. Just make sure not to use them in production deployments.
+
+.. _log-api-return-codes:
+
+Log runtime API call results
+++++++++++++++++++++++++++++
+
+Runtime API calls are not necessarely guaranteed to succeed under all possible circumstances.
+By design, the low level results of these calls might be abstracted away in Solidity.
```
seanyoung, line 245: You can't call any vm api in Solidity, they are always abstracted away

```suggestion
By design, the low level results of these calls are abstracted away in Solidity.

File: src/emit/substrate/mod.rs [hunk 12]
```diff
@@ -1851,3 +1853,47 @@ fn event_id<'b>(
 
     Some(binary.context.i8_type().const_int(event_id as u64, false))
 }
+
+/// Print the return code of API calls to the debug buffer.
+fn log_return_code<'b>(binary: &Binary<'b>, api: &'static str, code: &IntValue) {
```
seanyoung, line 1858: `IntValue` is just a wrapper for an LLVM pointer: `LLVMValueRef`, you can pass it by value. Now you're passing a pointer to a pointer.

src/emit/substrate/mod.rs
```diff
@@ -1851,3 +1853,47 @@ fn event_id<'b>(
...
+        return;
+    }
+
+    emit_context!(binary);
+
+    let fmt = format!("{}=", api);
+    let msg = fmt.as_bytes();
+    let length = i32_const!(msg.len() as u64 + 16);
+    let out_buf = binary.vector_new(length, i32_const!(1), None);
+    let out_buf_ptr = binary.vector_bytes(out_buf.into());
```
seanyoung, line 1869: Rather than heap allocating this, just use `builder.build_array_alloca()`. Faster and less wastage.

seanyoung (changes requested):
Sorry for the pedantic review :grimacing:

docs/code_gen_options.rst
```diff
@@ -218,4 +227,26 @@ always true, so the check is omitted.
...
 Note that this optimization does not cover every case. When an array is passed
-as a function argument, for instance, the length is unknown.
\ No newline at end of file
+as a function argument, for instance, the length is unknown.
+
+Debugging Options
+-----------------
+
+It may be desirable to have access to debug information regarding the contract execution. 
+However, this might lead to more instructions as well as higher gas usage. Hence, it is disabled by default,
```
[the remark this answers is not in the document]
LucasSte: `might` is on line 236, not on 235.

xermicus:
Thanks for the clean review!

seanyoung (approved):
Very nice

src/bin/solang.rs
```diff
@@ -171,6 +171,14 @@ fn main() {
                             .long("math-overflow")
                             .display_order(6),
                     )
+                    .arg(
+                        Arg::new("LOGAPIRETURNS")
+                            .help("Log the return codes of runtime API calls in the environment")
+                            .long("log-api-return-codes")
+                            .num_args(0..=1)
+                            .require_equals(true)
+                            .default_missing_value("true"),
```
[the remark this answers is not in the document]
seanyoung: > I disagree with having our CLI helps in the documentation at all. In the docs we should just point out that our CLI is self-documenting. No need for duplicating this and making sure it is kept up-to-date etc.

That is a good point. What do you think @LucasSte

LucasSte (approved):

Issue #797, opened 2022-05-05
Compiler enters unreachable code when building for Solana


LucasSte:
The following contract enters unreachable code when we build it for Solana:

```solidity
contract testing {
    function doThis(int a) public view returns (int) {
        return a+2;
    }

    function IlikeTests() public returns(function (int) external returns (int) retPtr) {
        function (int) external returns (int) fPtr = this.doThis;

        return fPtr;
    }
}
```

The error is the following:
```
thread 'main' panicked at 'internal error: entered unreachable code', src/emit/ethabiencoder.rs:1546:18
```

seanyoung:
The eth abi encoding does not know how to encode external function types.

This is not easy/possible to implement on Solana. Eth abi encoding encodes this into 32 bytes: 20 bytes for address, and 4 bytes for the function selector. On Solana, we already use 32 bytes for address, so we need to use a different encoding, which won't be compatible with the eth abi encoding libraries. I guess we could encode it as a struct and then specify it as a struct in the abi file.

Once we switch to borsh encoding, this problem goes away.

LucasSte:
> Once we switch to borsh encoding, this problem goes away.

@seanyoung Yes, this is why I haven't touched this issue.

LucasSte:
Closed by #1053.

LucasSte:
Closed by #1053

PR #1053, opened 2022-10-27
🚀 Use Borsh encoding everywhere on Solana ⚓

LucasSte:
This PR includes many changes in order to enable Borsh encoding everywhere on Solana. Here is a list of all of them:

1. Borsh encoding is used whenever we need to encode or decode things on Solana.
2. There is now a single interface for AbiEncoding that is used by both Borsh and Scale codecs.
3. CFG expressions the encoder normally utilizes are now wired up in constant folding and strength reduce to optimize the output.
4. All Solana runtime tests now use `BorshToken` for arguments and returns instead of `ethabi::Token`.
5. Integration tests and Anchor tests underwent many changes to accommodate the new encoding scheme.
6. `ethabiencoder.rs` has been deleted.
7. Borsh encoding now saves to a variable all basic types that are read from buffer. This made the code less optimal, but solved the problem with `Expression::ReturnData`. A fix has been suggested int #1054.

Integration tests are failing because the Typescript library needs [this PR](https://github.com/solana-labs/solana-solidity.js/pull/44) merged in order to work.


LucasSte:
This PR fixes #797

integration/solana/simple.spec.ts
```diff
@@ -70,17 +70,17 @@ describe('Deploy solang contract and test', function () {
...
         res = await contract.functions.op_i256(3, 1000, 10, { simulate: true });
         expect(Number(res.result)).toBe(100);
         res = await contract.functions.op_i256(4, 1000, 99, { simulate: true });
         expect(Number(res.result)).toBe(10);
         res = await contract.functions.op_i256(6, - 10000000000000, 8, { simulate: true });
-        expect(Number(res.result)).toBe(-2560000000000000);
+        expect(Number(res.result)).toBe(69551044457136128);
         res = await contract.functions.op_i256(7, - 10000000000000, 8, { simulate: true });
-        expect(Number(res.result)).toBe(-39062500000);
+        expect(Number(res.result)).toBe(1061264716448);
```
LucasSte, line 83: @seanyoung These two results of shift were wrong. When I print the result in Solidity, I also get the new numbers. Have you checked if these numbers were indeed the correct output? If not, I believe the implementation of shift for `int256` is incorrect.

src/codegen/encoding/scale_encoding.rs
```diff
@@ -0,0 +1,119 @@
+use crate::codegen::cfg::{ControlFlowGraph, Instr};
```
xermicus, line 1:
```suggestion
// SPDX-License-Identifier: Apache-2.0

use crate::codegen::cfg::{ControlFlowGraph, Instr};
```

src/codegen/encoding/mod.rs
```diff
@@ -51,13 +53,16 @@ pub(super) trait AbiEncoding {
...
 }
 
 /// This function should return the correct encoder, given the target
-pub(super) fn create_encoder(ns: &Namespace) -> impl AbiEncoding {
+pub(super) fn create_encoder(ns: &Namespace, packed: bool) -> Box<dyn AbiEncoding> {
     match &ns.target {
-        Target::Solana => BorshEncoding::new(),
-        _ => unreachable!("Other types of encoding have not been implemented yet"),
+        Target::Solana => Box::new(BorshEncoding::new(packed)),
+        _ => Box::new(ScaleEncoding::new(packed)),
```
xermicus, line 65:
```suggestion
        Target::Solana => Box::new(BorshEncoding::new(packed)),
        Target::Substrate { .. } => Box::new(ScaleEncoding::new(packed)),
        _ => unimplemented!("ABI encoder is not implemented yet"),
```

Using the scale encoder for everything that is not Solana is technically not correct? It will also make it easier to catch up when implementing a new backend.

xermicus (changes requested):
So far, really nice job! :dart:

(Just going to have lunch now, continue with the review later).

src/codegen/expression.rs
```diff
@@ -1465,31 +1385,16 @@ fn abi_encode_call(
...
-    );
-    Expression::Variable(*loc, Type::DynamicBytes, res)
+        .collect::<Vec<Expression>>();
+
+    let mut encoder_args: Vec<Expression> = Vec::with_capacity(args.len() + 1);
+    encoder_args.push(selector);
+    encoder_args.append(&mut args);
+    let mut encoder = create_encoder(ns, false);
+    let (encoded, _) = encoder.abi_encode(loc, encoder_args, ns, vartab, cfg);
+    encoded
```
xermicus, line 1397: Seems like this code is just repeated 3x and could potentially be loaded off into a helper function or something?

src/codegen/expression.rs
```diff
@@ -2370,75 +2264,47 @@ pub fn emit_function_call(
...
-                            packed: vec![Expression::BytesLiteral(
-                                *loc,
-                                Type::Bytes(selector.len() as u8),
-                                selector,
-                            )],
-                            args,
-                        },
-                        Some(address),
-                    )
+                    let mut encoder_args: Vec<Expression> = Vec::with_capacity(6 + args.len());
```
xermicus, line 2276: Maybe I'm wrong here, but this if is confusing to read. Shouldn't we explicitly match the target? I don't understand why we are differentiating here between  the Solana target vs. Substrate Target OR selector.is_some()... What happens when we add more targets?
```rust
let address = match {
    Target::Solana => { ... },
    Target::Substrate { .. } => { ... },
    _ => unimplemented!()
}
```

src/emit/solana/target.rs
```diff
@@ -1194,72 +1194,40 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
...
-        ns: &ast::Namespace,
+        _binary: &Binary<'b>,
+        _function: FunctionValue<'b>,
+        _packed: &[BasicValueEnum<'b>],
+        _args: &[BasicValueEnum<'b>],
+        _tys: &[ast::Type],
+        _ns: &ast::Namespace,
     ) -> PointerValue<'b> {
-        ethabiencoder::encode_to_vector(binary, function, packed, args, tys, true, ns)
+        unreachable!("Abi encoding is implemented in code generation for Solana")
```
xermicus, line 1204:
```suggestion
        unreachable!("ABI encoding is implemented in code generation for Solana")
```

src/emit/solana/target.rs
```diff
@@ -1194,72 +1194,40 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
...
-            )
-            .try_as_basic_value()
-            .left()
-            .unwrap()
-            .into_pointer_value();
-
-        encoder.finish(binary, function, encoded_data, ns);
-
-        (encoded_data, length)
+        unreachable!("Abi encoding is implemented in code generation for Solana")
```
xermicus, line 1217:
```suggestion
        unreachable!("ABI encoding is implemented in code generation for Solana")
```

src/emit/solana/target.rs
```diff
@@ -1194,72 +1194,40 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
...
+        _function: FunctionValue<'b>,
+        _args: &mut Vec<BasicValueEnum<'b>>,
+        _data: PointerValue<'b>,
+        _length: IntValue<'b>,
+        _spec: &[ast::Parameter],
+        _ns: &ast::Namespace,
     ) {
-        self.abi
-            .decode(binary, function, args, data, length, spec, ns);
+        unreachable!("Abi encoding is implemented in code generation for Solana.")
```
xermicus, line 1230:
```suggestion
        unreachable!("ABI encoding is implemented in code generation for Solana.")
```

xermicus (changes requested):

src/codegen/encoding/mod.rs
LucasSte: I see what you mean, but if I do that, tests will fail because, although we do not support EVM yet, we have code generation tests for a fake EVM target, and it needs the `Instr::AbiEncode` in the CFG.

src/codegen/constructor.rs
```diff
@@ -46,25 +49,24 @@ pub(super) fn call_constructor(
...
-        tys.resize(3, Type::Bool);
-        tys[0] = Type::Uint(64);
-        tys[1] = Type::Uint(32);
-        tys[2] = Type::Bytes(1);
-
         let value_arg = value.clone().unwrap_or_else(|| {
             Expression::NumberLiteral(Loc::Codegen, Type::Uint(64), BigInt::zero())
         });
-        let selector = ns.contracts[*contract_no].selector().to_be();
+        let selector = ns.contracts[*contract_no].selector();
```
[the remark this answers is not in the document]
LucasSte: Ethabiencoding, our old encoding scheme, encodes integers in big endian. The both encoding uses little endian.

src/codegen/constructor.rs
```diff
@@ -81,42 +83,11 @@ pub(super) fn call_constructor(
...
-            res: encoded_buffer,
-            expr: Expression::AbiEncode {
-                loc: *loc,
-                tys,
-                packed,
-                args,
-            },
-        },
-    );
+        args.append(&mut constructor_args);
```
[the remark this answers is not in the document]
LucasSte: Yes, it can.

src/codegen/encoding/scale_encoding.rs
LucasSte: Good catch!

src/codegen/expression.rs
LucasSte: Good point

src/codegen/expression.rs
LucasSte: I agree this piece of code here isn't the best one. The match would work well, but it would need repeated code as there one case that is exactly the same for both Solana and Substrate. According to @seanyoung , Solana's API for external call is about to change, so this ugliness will go away soon.

LucasSte:
@xermicus I will fix your comments after @seanyoung posts his review, so I won't mess up with his reviewing.

seanyoung (changes requested):
There is no packed decoder, so I think the abi_decode() in borsh_encoder.rs should do an `assert!(!self.packed_encoder);`

integration/solana/simple.spec.ts
seanyoung: The answers were correct:

`-10000000000000 << 8 = -2560000000000000`
`-10000000000000 >> 8 = -39062500000`

Somehow your change broke this.

LucasSte: I'll investigate this first.

LucasSte: Sorry, I was running the integration tests with an old version of `@solana/solidity`.

LucasSte: Fixed

src/codegen/expression.rs
LucasSte: Any thought on this @seanyoung ?

integration/solana/builtins.spec.ts
```diff
@@ -38,7 +34,7 @@ describe('Deploy solang contract and test', function () {
         let ts = Number(res.result);
 
         expect(ts).toBeLessThanOrEqual(now);
-        expect(ts).toBeGreaterThan(now - 120);
+        expect(ts).toBeGreaterThan(now - 90000);
```
seanyoung, line 37: The `120` was an error margin for javascript now being called after the solidity `block.timestamp`. Why have you increased this to more than a day?

integration/solana/simple.spec.ts
```diff
@@ -94,45 +96,60 @@ describe('Deploy solang contract and test', function () {
         expect(Number(res.result)).toBe(89);
         res = await contract.functions.op_u256(5, 123456789, 9, { simulate: true });
         expect(Number(res.result)).toBe(6662462759719942007440037531362779472290810125440036903063319585255179509);
-        res = await contract.functions.op_i256(6, 10000000000000, 8, { simulate: true });
+        res = await contract.functions.op_u256(6, 10000000000000, 8, { simulate: true });
```
seanyoung, line 99: Why this change?

integration/solana/simple.spec.ts
```diff
@@ -94,45 +96,60 @@ describe('Deploy solang contract and test', function () {
...
         expect(Number(res.result)).toBe(2560000000000000);
-        res = await contract.functions.op_i256(7, 10000000000000, 8, { simulate: true });
+        res = await contract.functions.op_u256(7, 10000000000000, 8, { simulate: true });
```
seanyoung, line 101: Why this change?

src/codegen/expression.rs
seanyoung: The `if` condition would make more sense like so:
```rust
   if ns.target == Target::Solana && dest_func.selector.is_none() {
       // swap then and else statement block
   } 
``` 
This ugliness will disappear soon when we remove all the extra stuff from the instruction data on Solana.

seanyoung (changes requested):

integration/solana/simple.spec.ts
LucasSte: You already test shift for signed integers [here](https://github.com/hyperledger/solang/blob/4ad568fff31e7dede49918d3eea01bc6d5224451/integration/solana/simple.spec.ts#L79-L83).

As all other tests in this set call the `u256` function, I thought you mistyped the function name. In addition, shift with unsigned integers will have no test without this change.

integration/solana/simple.spec.ts
LucasSte: Same as the comment below.

integration/solana/builtins.spec.ts
LucasSte: The difference between `now` and `ts` on my computer is `84427`.

LucasSte: It looks like the time refers to when the folder `test-ledger` was created by `solana-test-validator`. I have just deleted it, and everything worked fine.

src/codegen/encoding/mod.rs
LucasSte: Any suggestions?
@seanyoung and @xermicus ?

integration/solana/simple.spec.ts
```diff
@@ -70,16 +70,18 @@ describe('Deploy solang contract and test', function () {
...
         expect(Number(res.result)).toBe(100);
         res = await contract.functions.op_i256(4, 1000, 99, { simulate: true });
         expect(Number(res.result)).toBe(10);
-        res = await contract.functions.op_i256(6, - 10000000000000, 8, { simulate: true });
+        res = await contract.functions.op_i256(6, -10000000000000, 8, { simulate: true });
+        console.log(res.logs);
         expect(Number(res.result)).toBe(-2560000000000000);
-        res = await contract.functions.op_i256(7, - 10000000000000, 8, { simulate: true });
+        res = await contract.functions.op_i256(7, -10000000000000, 8, { simulate: true });
+        console.log(res.logs);
```
seanyoung, line 84: Please remove debugging

integration/solana/simple.spec.ts
```diff
@@ -70,16 +70,18 @@ describe('Deploy solang contract and test', function () {
...
+        expect(res.result).toStrictEqual(BigInt(-3100));
         res = await contract.functions.op_i256(2, 1000, 4100, { simulate: true });
         expect(Number(res.result)).toBe(4100000);
         res = await contract.functions.op_i256(3, 1000, 10, { simulate: true });
         expect(Number(res.result)).toBe(100);
         res = await contract.functions.op_i256(4, 1000, 99, { simulate: true });
         expect(Number(res.result)).toBe(10);
-        res = await contract.functions.op_i256(6, - 10000000000000, 8, { simulate: true });
+        res = await contract.functions.op_i256(6, -10000000000000, 8, { simulate: true });
+        console.log(res.logs);
```
seanyoung, line 81: Please remove debugging

src/codegen/encoding/mod.rs
seanyoung: Right now we don't have an eth abi encoder, doesn't matter which one you pick. However it does need a big comment.

src/codegen/expression.rs
```diff
@@ -2532,48 +2361,39 @@ pub fn emit_function_call(
...
-                                Expression::NumberLiteral(*loc, Type::Bytes(4), BigInt::zero()),
-                                Expression::NumberLiteral(*loc, Type::Bytes(1), BigInt::zero()),
-                                selector,
-                            ],
-                            args,
-                        },
-                        None,
-                    )
+                let address = if ns.target == Target::Solana {
+                    let mut encoder_args: Vec<Expression> = Vec::with_capacity(6 + args.len());
```
seanyoung, line 2365: When a vector is allocated, it has a minimum of 8 elements, and the sized is doubled every time it runs out of capacity. 
This means that most of the time using `Vec::with_capacity()` is not really necessary. 

If we have more than 2 args, we may get one extra allocation here (more if we have more than 10 args, unlikely). So, I don't think the  `Vec::with_capacity();` followed by a bunch of pushes is really worth it vs. a single `vec![ ... ];` followed by an `extent()`. 

Anyway, up to you, I thought I would mention it.

seanyoung (approved):
Looks great, just some tiny nitpicks

src/codegen/expression.rs
LucasSte: That is not what happens here. According to [Rust docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity), `with_capacity` guarantees that there is space for at least the number of elements you I requested. This way, subsequent pushes do not need reallocation. If I request `with_capacity(10)`, there will be space for ten elements and no reallocation will happen when I push ten elements into the array.

seanyoung: I know how `with_capacity()` works.

What I am saying is, all this extra code does not safe much but makes the code less readable.

xermicus (approved):

PR #1055, opened 2022-10-29
Add Serialize [derive] using serde crate.

Ethnical:
Hi there, 
I would like to add some derive `Serialize, Deserialize` over most of the objects. 
This will permit us to get the **AST** in **JSON** and then help us interface with **JSON** tools already created.
Thanks!


solang-parser/Cargo.toml
```diff
@@ -18,6 +18,7 @@ lalrpop-util = "0.19"
 phf = { version = "0.10", features = ["macros"] }
 unicode-xid = "0.2.0"
 itertools = "0.10"
+serde = {version = "1.0.145", features = ["derive"]}
```
xermicus, line 21: This is not correctly formatted

xermicus (changes requested):
Thanks for your contribution! Quick first review:

1. Please gate this behind a feature flag (and disable it by default).
2. https://github.com/hyperledger/solang/pull/1055/checks?check_run_id=9179878062

Just out of curiosity, how are you going to use this?

solang-parser/Cargo.toml
Ethnical: Hey @xermicus, 
I am currently building a tool for analyzing smart contracts (Security) => https://github.com/Ethnical/Swek3 
I would like to add this PR to convert AST to JSON and then use `ethers-solc` in the future also to integrate it `solang-parser`. :) 
Anyway, what do you mean by bad formatting for the `Cargo.toml` ? I guess you are speaking about some spacing formatting maybe?

LucasSte:
> This will permit us to get the **AST** in **JSON** and then help us interface with **JSON** tools already created.

You have modified the files for the parse tree, not the AST. The AST is defined in `sema/ast.rs`. Is the parse tree what you are looking for?

Also, sign your commits, using the instructions in the DCO CI test.

solang-parser/Cargo.toml
xermicus: Thanks for sharing! So you are using it for static analysis?

>  I guess you are speaking about some spacing formatting maybe?
Yes sorry I wasn't super clear about that. See also the example [here](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/cargo.md)

seanyoung (approved):

LucasSte (approved):

solang-parser/Cargo.toml
Ethnical: @xermicus Yes actually really nice to learn rust :)

xermicus:
@Ethnical could you please gate this behind a feature flag? It adds compile time for something we do not need in the compiler itself. 

You just need to add a feature flag, exclude it from the default features and gate the derive attributes behind that feature flag. For reference:
https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute
https://stackoverflow.com/questions/42551113/is-it-possible-to-conditionally-enable-an-attribute-like-derive

Then we can merge it.

Ethnical:
Hey @xermicus, 
Sorry to be late.. but I am still learning about **derive features** in Rust. But I don't really get it, what should I do here? 
So just to be sure about what I need to do : 
I would like to push this code: 
```rust
#[cfg_attr(feature = "json_tree", Serialize, Deserialize)]
#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy, Serialize, Deserialize)]
```
I am adding at the **top (one time, not on all the derive)** of the `pt.rs` the feature `json_tree` on the derive `Serialize, Deserialize` then when we import `solang-parser`. we will need to enable the feature `json_tree` to activate the derive `Serialize, Deserialize` ? 
Sorry again still discovering main features of Rust...
Thanks!

xermicus:
No worries @Ethnical ! First you change `solang-parser/Cargo.toml` to something like this:

```toml
serde = { version = "1.0.145", features = ["derive"], optional = true }

[features]
default = []
pt-serde = ["serde"]
```

To make sure it is excluded from the default features. Then, you make use of conditional compilation for the imports and **each** datastructure you want to have this derive for. So, for example, in `solang-parser/src/pt.rs`:

```rust
#[cfg(feature = "pt-serde")]
use serde::{Deserialize, Serialize};

...

#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
#[cfg_attr(feature = "pt-serde", derive(Serialize, Deserialize))]
/// file no, start offset, end offset (in bytes)
pub enum Loc {
    Builtin,
....

```

In your project you can then enable the feature for the dependency, so you probably want something like:

```toml
solang-parser = { git = "https://github.com/hyperledger/solang.git", features = ["pt-serde"] }
```

Ethnical:
@xermicus oh got it! will do it asap thanks!!

solang-parser/src/pt.rs
```diff
@@ -577,8 +610,9 @@ impl Expression {
         }
     }
 }
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "pt-serde", derive(Serialize, Deserialize))]
```
xermicus, line 615:
```suggestion
}

#[derive(Debug, PartialEq, Eq, Clone)]
#[cfg_attr(feature = "pt-serde", derive(Serialize, Deserialize))]
```

solang-parser/src/pt.rs
```diff
@@ -644,8 +680,9 @@ impl OptionalCodeLocation for Visibility {
         }
     }
 }
```
xermicus, line 682:
```suggestion
}

```

solang-parser/src/pt.rs
```diff
@@ -751,15 +793,17 @@ pub enum YulStatement {
     FunctionDefinition(Box<YulFunctionDefinition>),
     FunctionCall(Box<YulFunctionCall>),
 }
```
xermicus, line 795:
```suggestion
}

```

xermicus (changes requested):
LGTM

xermicus (approved):

LucasSte (approved):

seanyoung (approved):

PR #1064, opened 2022-11-11
Substrate: Fix events with topics

### LINKED ISSUE #970: Milestone 3. Events with indexed fields are not encoded correctly
If an event has fields declared indexed, they should be passed into the Substrate topics. This is not implemented correctly today. Ensure that indexed fields are passed into topics and add an integration test for this at

https://github.com/hyperledger-labs/solang/tree/main/integration/substrate

Success means that a pull request has had all its review comments addressed and it is merged.

xermicus:
- Fixes codegen to produce event topics the exact same way how `ink!` does. This is required; otherwise off-chain tools using topics won't work for solidity contracts out of the box.
- Adds a sema check to ensure that indexed fields always have a name on Substrate. In `ink!`,  event fields always do have a name. Which means that indexed fields must have a name, or they will produce incompatible topics.
- Updates and extends existing tests and extends the events integration test to check against topics produced directly by an `ink!` contract.

Closes #970


src/codegen/events/substrate.rs
```diff
@@ -32,68 +45,121 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
-                data_tys.push(arg.ty());
+        let loc = pt::Loc::Builtin;
+        let event = &self.ns.events[self.event_no];
+        let hash_len = || Box::new(Expression::NumberLiteral(loc, Type::Uint(32), 32.into()));
+
+        // Events that are not anonymous always have themselves as a topic.
+        // This is static and can be calculated at compile time.
+        if !event.anonymous {
+            let mut encoded =
+                format!("{}::{}", &self.ns.contracts[contract_no].name, &event.name).encode();
```
seanyoung, line 56: Here you are using the contract name of the emitting contract, not where the event is defined. For example:

```solidity
contract A { event X(bool indexed foo); }

contract B { function f() public { emit A.X(true); } }
```
The topic you are using is `B.X`. This is not what the user might be expecting. However, if you want to use the name of the event definition, bare in mind that an event is not always defined in a contract:
```solidity
event X(bool indexed foo);

contract B { function f() public { emit X(true); } }
```
Whatever you decided needs documenting.

src/codegen/events/substrate.rs
```diff
@@ -32,68 +45,121 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
+            encoded[0] = 0; // Set the prefix (there is no prefix for the event topic)
```
seanyoung, line 57: Is the the same as:
```rust
  let encoded = format!("\0{}::{}", &self.ns.contracts[contract_no].name, &event.name).as_bytes();
``` 
?

src/codegen/events/substrate.rs
```diff
@@ -32,68 +45,121 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
+            .fields
+            .iter()
+            .filter(|field| field.indexed)
+            .map(|field| {
+                format!(
+                    "{}::{}::{}",
+                    &self.ns.contracts[contract_no].name,
+                    &event.name,
+                    &field.name_as_str()
+                )
```
seanyoung, line 78: Same as above - does not describe the event definition

src/codegen/events/substrate.rs
```diff
@@ -32,68 +45,121 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
+        for (ast_exp, field) in self.args.iter().zip(event.fields.iter()) {
+            let value_exp = expression(ast_exp, cfg, contract_no, Some(func), self.ns, vartab, opt);
+            data_tys.push(value_exp.ty());
+            data.push(value_exp);
+
+            if !field.indexed {
+                continue;
             }
+
+            let value_exp = expression(ast_exp, cfg, contract_no, Some(func), self.ns, vartab, opt);
```
seanyoung, line 93: You are calculating `ast_exp` twice - better to evaluate it once and put it into a variable.

src/codegen/events/substrate.rs
```diff
@@ -32,68 +45,121 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
+            let encoded = Expression::AbiEncode {
+                loc,
+                tys: vec![value_exp.ty()],
+                packed: vec![],
+                args: vec![value_exp],
+            };
+            let concatenated = Expression::StringConcat(
+                loc,
+                Type::DynamicBytes,
+                StringLocation::CompileTime(topic_prefixes.remove(0)), // TODO not efficient
```
seanyoung, line 103: How about using `VecDeque`  to make this more efficient (not that with 4 topics it does not make much difference though)

src/codegen/events/substrate.rs
```diff
@@ -32,68 +45,121 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
+                    vec![Type::Uint(32)],
+                    Builtin::ArrayLength,
+                    vec![buffer.clone()],
+                )
+                .into(),
+                hash_len(),
+            );
+
+            let bigger = cfg.new_basic_block("bigger".into());
+            let done = cfg.new_basic_block("done".into());
```
seanyoung, line 131: I called it `bigger`, I know. However, it's not a great name. How about `needs_hashing` or `hash_topic` or something like that.

src/sema/types.rs
```diff
@@ -577,6 +577,12 @@ fn event_decl(
                 loc: name.loc,
             })
         } else {
+            if ns.target.is_substrate() && field.indexed {
+                ns.diagnostics.push(Diagnostic::error(
+                    field.loc,
+                    "indexed event fields must have a name on substrate".into(),
```
seanyoung, line 583: It would be great to include a link into the documentation here which explains how topics are encoded on Substrate.

seanyoung (changes requested):

src/codegen/events/substrate.rs
xermicus: Good catch!

src/codegen/events/substrate.rs
xermicus: Might be, but I'd prefer to have it explicit here.

src/codegen/events/substrate.rs
xermicus: > Another thing to consider is, what happens when you emit an event in a free standing function:
> 
> ```solidity
> function foo() {
>     emit A(true);
> }
> ```

They are not a thing in `ink!`. They need to have a contract associated with them or we won't be compatible. But in that case we can actually just use the contract name and be fine.

src/codegen/events/substrate.rs
LucasSte: Aren't you losing the information saved in the first index of this array? If you wish to add another element, you should have done `encoded.insert(0, 0);`

src/codegen/events/substrate.rs
```diff
@@ -32,68 +47,131 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
-            } else {
-                let e = expression(arg, cfg, contract_no, Some(func), self.ns, vartab, opt);
-
-                data.push(e);
-                data_tys.push(arg.ty());
+        let loc = pt::Loc::Builtin;
+        let event = &self.ns.events[self.event_no];
+        // For freestanding events the name of the emitting contract is used
+        let contract_name = &self.ns.contracts[event.contract.unwrap_or(contract_no)].name;
+        let hash_len = || Box::new(Expression::NumberLiteral(loc, Type::Uint(32), 32.into()));
```
LucasSte, line 54: Why is this a function? Can't it be a variable?

src/codegen/events/substrate.rs
```diff
@@ -32,68 +47,131 @@ impl EventEmitter for SubstrateEventEmitter<'_> {
...
+            cfg.add(
+                vartab,
+                Instr::Set {
+                    loc,
+                    res: value_var,
+                    expr: value_exp,
+                },
+            );
+            data_tys.push(value.ty());
+            data.push(value.clone());
```
LucasSte, line 100: You can preallocate `data` and `data_tys` on lines 45 and 46, using `Vec::with_capacity`, for better performance.

src/sema/types.rs
LucasSte, line 583: There are no tests for this.

LucasSte (changes requested):

src/codegen/events/substrate.rs
seanyoung: There is a limit of 4 topics on Substrate. When a rust `Vec` needs to grow, there is a minimum size which is 4 for elements > 1 byte: https://github.com/rust-lang/rust/blob/b7463e8bdb7b039e75241dcc4ea65bbd24d5c608/library/alloc/src/raw_vec.rs#L111-L117

src/sema/types.rs
seanyoung: So how about writing some documentation for Solang on this then. This would help users.

src/codegen/events/substrate.rs
xermicus: > Aren't you losing the information saved in the first index of this array? If you wish to add another element, you should have done encoded.insert(0, 0);

That's correct, but it was intentional. However I came to the conclusion that @seanyoung suggestion is actually better, as it crafts the topic contents directly instead of abusing an encoded string.

src/sema/types.rs
xermicus: Good catch, added one

xermicus: I tend to disagree: As a user, why should I care at all how the topics are working under the hood? The only thing I'd care about is that they are compatible with `ink!` tooling. I guess that's probably the reason ink doesn't mention it in their docs as well. Furthermore, if this goes into our docs, we are providing and maintaining documentation of random parts of the  `ink!` on-chain engine, which in my opinion is something we should avoid heavily.

src/codegen/events/substrate.rs
LucasSte: I don't think @seanyoung solution does the same thing. He added a new character to the string, didn't he?

xermicus: > I don't think @seanyoung's solution does the same thing. He added a new character to the string, didn't he?

Prior, the first byte was the length, which was not needed and replaced with a null byte. But @seanyoung solution is better because it will not depend on the scale encode implementation of `String` to produce correct results.

LucasSte (approved):

seanyoung (approved):

PR #1066, opened 2022-11-11
Documentation cleanup and rename Ternary to ConditionalOperator

seanyoung:
I read through our documentation - only got halfway through!

Just because a conditional operator takes three parameters does not mean it should be called ternary. It's just not very descriptive. Also, we might invent another operator which also takes three operators.

### DIFF
--- a/docs/extension.rst
+++ b/docs/extension.rst
@@ -23,18 +23,7 @@ Both the Visual Studio Code extension code and the language server were develope
 Solidity Language flavour
 -------------------------
 
-The Solidity language flavour depends on what target you are compiling for. For
-example:
-
-* Ethereum style address literals like ``0xE0f5206BBD039e7b0592d8918820024e2a7437b9`` are
-  not supported on Substrate or Solana, but are supported for EVM.
-* On Substrate and Solana, base58 style encoded address literals like
-  ``address"5GBWmgdFAMqm8ZgAHGobqDqX6tjLxJhv53ygjNtaaAn3sjeZ"`` are supported, but
-  not with EVM.
-* On Solana, there is special builtin import file called ``'solana'`` available.
-* Different blockchains offer different builtins. See the :ref:`builtins documentation <builtins>`.
-* See :ref:`language-status` for compatiblity with Ethereum Solidity (solc)
-* There are many more differences, which are noted throughout the documentation.
+The Solidity language flavour depends on what target you are compiling for, see :ref:`language-status` for a brief overview.
 
 You can choose the between the following targets:
 
@@ -48,7 +37,7 @@ evm
    Solidity for any EVM based chain like Ethereum
 
 Note that the language server has support for EVM, but Hyperledger Solang does
-not support compiling for EVM (yet).
+not support compiling for EVM.
 
 Using the extension
 -------------------
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -4,7 +4,7 @@ Installing Solang
 The Solang compiler is a single binary. It can be installed in different ways, listed below.
 
 1. :ref:`Download from Homebrew <download-brew>` (MacOS only)
-2. :ref:`Download pre-compiled binaries <download-binaries>`
+2. :ref:`Download binaries <download-binaries>`
 3. :ref:`Download from a Docker container <download-docker>`
 4. :ref:`Build using Dockerfile <build-dockerfile>`
 5. :ref:`Build from source <build-source>`
@@ -14,7 +14,7 @@ The Solang compiler is a single binary. It can be installed in different ways, l
 Option 1: Download from Brew
 ----------------------------
 
-Solang is available on Brew via a private tap. Currently, this works only for MacOS systems, both Intel and Apple Silicon.
+Solang is available on Brew via a private tap. This works only for MacOS systems, both Intel and Apple Silicon.
 To install Solang via Brew, run the following command:
 
 .. code-block:: text
@@ -23,8 +23,8 @@ To install Solang via Brew, run the following command:
 
 .. _download-binaries:
 
-Option 2: Download release binaries
------------------------------------
+Option 2: Download binaries
+---------------------------
 
 There are binaries available on github releases:
 
@@ -33,6 +33,9 @@ There are binaries available on github releases:
 - `Windows x64 <https://github.com/hyperledger/solang/releases/download/v0.1.13/solang.exe>`_
 - `MacOS intel <https://github.com/hyperledger/solang/releases/download/v0.1.13/solang-mac-intel>`_
 - `MacOS arm <https://github.com/hyperledger/solang/releases/download/v0.1.13/solang-mac-arm>`_
+
+Download the file and save it somewhere in your ``$PATH``, for example the bin directory in your home directory. If the
+path you use is not already in ``$PATH``, then you need to add it yourself.
 
 On MacOS, remember to give execution permission to the file and remove it from quarantine by executing the following commands:
 
@@ -85,7 +88,7 @@ Option 5: Build Solang from source
 ----------------------------------
 
 In order to build Solang from source, you will need rust 1.63.0 or higher,
-and a build of LLVM based on the Solana LLVM tree. There are a few patches which are not upstream yet.
+and a build of LLVM based on the Solana LLVM tree. There are a few LLVM patches required that are not upstream yet.
 First, follow the steps below for installing LLVM and then proceed from there.
 
 If you do not have the correct version of rust installed, go to `rustup <https://rustup.rs/>`_.
@@ -106,13 +109,13 @@ _____________________________________
 
 Solang needs a build of
 `LLVM with some extra patches <https://github.com/solana-labs/llvm-project/>`_.
-These patches make it possible to generate code for Solana, and fixes some
+These patches make it possible to generate code for Solana, and fixes
 concurrency issues in the lld linker.
 
 You can either download the pre-built libraries from
 `github <https://github.com/hyperledger/solang/releases/tag/v0.1.13>`_
-or :ref:`build your own from source <llvm-from-source>`. After that, you need to add the ``bin`` directory to your
-path, so that the build system of Solang can find the correct version of LLVM to use.
+or :ref:`build your own from source <llvm-from-source>`. After that, you need to add the ``bin`` of your
+LLVM directory to your path, so that the build system of Solang can find the correct version of LLVM to use.
 
 Linux
 ~~~~~
--- a/docs/language/contracts.rst
+++ b/docs/language/contracts.rst
@@ -24,15 +24,15 @@ If a constructor has arguments, they must be supplied when the contract is deplo
 If a contract is expected to receive value on instantiation, the constructor should be declared ``payable``.
 
 .. note::
-    On Substrate, constructors have a name. Solang allows naming constructors in the substrate target:
+    Solang allows naming constructors in the Substrate target:
 
     .. code-block:: solidity
- 
+
         contract Foo {
             constructor my_new_foo() {}
         }
 
-    Unnamed constructors will be called ``new`` in the metadata.
+    Constructors without a name will be called ``new`` in the metadata.
 
     Note that constructor names are only used in the generated metadata. For contract instantiation,
     the correct constructor matching the function signature will be selected automatically.
@@ -93,6 +93,9 @@ The constructor should be declared ``payable`` for this to work.
 .. note::
     If no value is specified, then on Parity Substrate the minimum balance (also know as the
     existential deposit) is sent.
+
+.. note::
+    On Solana, this functionality is not available.
 
 Setting the salt, gas, and space for the new contract
 _____________________________________________________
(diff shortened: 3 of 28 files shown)


docs/running.rst
```diff
@@ -16,7 +16,7 @@ learn whether your favorite shell is supported. If so, evaluate the output of
 
 .. code-block:: bash
 
-	  echo 'source <(solang shell-complete bash)' >> ~/.bashrc
+	  echo 'source < $(solang shell-complete bash)' >> ~/.bashrc
```
xermicus, line 19: Why? This addition breaks the command

docs/targets/solana.rst
```diff
@@ -118,7 +121,7 @@ Solana Cross Program Invocation (CPI) does not support this. This means that:
 
 .. note::
 
-    Hypothetically, this could be implemented like so: the caller could transfer
+    A naive way to implement this might be to let the caller transfer
```
xermicus, line 124:
```suggestion
    A naive way to implement this is to let the caller transfer
```

docs/targets/solana.rst
```diff
@@ -253,7 +256,7 @@ Solana Library
 ______________
 
 In Solang's Github repository, there is a directory called ``solana-library``. It contains libraries for Solidity contracts
-to interact with Solana specific instructions. Currently, there are two libraries there: one for SPL tokens and another
+to interact with Solana specific instructions. There are two libraries there: one for SPL tokens and another
```
xermicus, line 259:
```suggestion
to interact with Solana specific instructions. We provide two libraries: one for SPL tokens and another
```

docs/targets/substrate.rst
```diff
@@ -1,7 +1,7 @@
 Parity Substrate
 ================
 
-Solang works with Parity Substrate 3.0. Note that for recent Substrate Version, cross-contract calls as well as using `address`
+Solang works with Parity Substrate 3.0. Note that for recent Substrate versions, cross-contract calls as well as using `address`
```
xermicus, line 4: Sorry but I need to hijack the topic quickly here:

Given the metadata PR gets merged and I fix events with topics (should be doable next week), we have basically all regressions fixed for `substrate-contracts-node` version 0.21 which supports `ink!` v3. Since `ink!` v4 will be released only early next year, could we have a solang release with the mentioned fixes in? Would be great because then we could say use solang v0.14 together with `substrate-contracts-node` v0.21 and you are good.

xermicus (changes requested):
Thanks for doing this!

docs/extension.rst
```diff
@@ -48,7 +37,7 @@ evm
    Solidity for any EVM based chain like Ethereum
 
 Note that the language server has support for EVM, but Hyperledger Solang does
-not support compiling for EVM (yet).
+not support compiling for EVM.
```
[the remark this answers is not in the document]
seanyoung: I don't think things on our backlog/wishlist belong in the documentation.

docs/language/types.rst
```diff
@@ -125,8 +125,8 @@ Address and Address Payable Type
 ________________________________
 
 The ``address`` type holds the address of an account. The length of an ``address`` type depends on
-the target being compiled for. On ewasm, an address is 20 bytes. Solana and Substrate have an address
-length of 32 bytes. The format of an address literal depends on what target you are building for. On ewasm,
+the target being compiled for. On evm, an address is 20 bytes. Solana and Substrate have an address
```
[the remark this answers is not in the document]
seanyoung: Good point

docs/installing.rst
```diff
@@ -14,7 +14,7 @@ The Solang compiler is a single binary. It can be installed in different ways, l
 Option 1: Download from Brew
 ----------------------------
 
-Solang is available on Brew via a private tap. Currently, this works only for MacOS systems, both Intel and Apple Silicon.
+Solang is available on Brew via a private tap. This works only for MacOS systems, for both Intel and Apple Silicon.
```
LucasSte, line 17:
```suggestion
Solang is available on Brew via a private tap. This works only for MacOS systems, both Intel and Apple Silicon.
```

I don't think `for` is necessary here.

docs/language/contracts.rst
```diff
@@ -24,15 +24,15 @@ If a constructor has arguments, they must be supplied when the contract is deplo
 If a contract is expected to receive value on instantiation, the constructor should be declared ``payable``.
 
 .. note::
-    On Substrate, constructors have a name. Solang allows naming constructors in the substrate target:
+    On Substrate, constructors may have a name. Solang allows naming constructors in the substrate target:
```
LucasSte, line 27: These two phrases say exactly the same thing. They are redundant. I suggest we eliminate one of them.

```suggestion
    Solang allows naming constructors in the Substrate target:
```

docs/language/expressions.rst
```diff
@@ -114,8 +110,8 @@ a bool.
 Increment and Decrement operators
 _________________________________
 
-The post-increment and pre-increment operators are implemented like you would expect. So, ``a++``
-evaluates to the value of ``a`` before incrementing, and ``++a`` evaluates to value of ``a``
+The post-increment and pre-increment operators are implemented like you would expect. So, ``i++``
+evaluates to the value of ``i`` before incrementing, and ``++i`` evaluates to value of ``i``
```
LucasSte, line 114:
```suggestion
The post-increment and pre-increment operators are implemented by reading the variable's value before modifying it. ``i++``
evaluates to the value of ``i`` before incrementing, and ``++i`` evaluates to value of ``i``
```

`like you would expect` is not a clear way to say this. People have different understanding (sometimes wrong) about how things work.

docs/language/expressions.rst
```diff
@@ -177,10 +178,10 @@ an interface at runtime, which can be used to write a `supportsInterface()` func
 in the EIP.
 
 The contract code for a contract, i.e. the binary WebAssembly or BPF, can be retrieved using the
-``type(c).creationCode`` and ``type(c).runtimeCode`` fields, as ``bytes``. In Ethereum,
+``type(c).creationCode`` and ``type(c).runtimeCode`` fields, as ``bytes``. On evm,
 the constructor code is in the ``creationCode`` WebAssembly and all the functions are in
```
LucasSte, line 182: Why `creationCode` WebAssembly is the constructor for EVM, if we does not uses WASM?

docs/language/expressions.rst
```diff
@@ -177,10 +178,10 @@ an interface at runtime, which can be used to write a `supportsInterface()` func
 in the EIP.
 
 The contract code for a contract, i.e. the binary WebAssembly or BPF, can be retrieved using the
-``type(c).creationCode`` and ``type(c).runtimeCode`` fields, as ``bytes``. In Ethereum,
+``type(c).creationCode`` and ``type(c).runtimeCode`` fields, as ``bytes``. On evm,
```
LucasSte, line 181: Aren't we using EVM in upper case everywhere?

docs/language/expressions.rst
```diff
@@ -227,7 +228,8 @@ Casting
 _______
 
 Solidity is very strict about the sign of operations, and whether an assignment can truncate a
-value. You can force the compiler to accept truncations or differences in sign by adding a cast.
+value. You can force the compiler to accept truncations or differences in sign by adding an
```
LucasSte, line 231:
```suggestion
value. You can force the compiler to accept truncations or sign shifts by adding an
```
`Differences in sign` sounds awkward to me.

docs/language/statements.rst
```diff
@@ -3,7 +3,7 @@ Statements
 
 In functions, you can declare variables in code blocks. If the name is the same as
 an existing function, enum type, or another variable, then the compiler will generate a
-warning as the original item is no longer accessible.
+warning as the original item is no longer accessible, it is shadowed.
```
LucasSte, line 6:
```suggestion
warning as the original item is no longer accessible, becoming shadowed.
```
That was a comma splice.

docs/language/statements.rst
```diff
@@ -209,9 +225,12 @@ of elements in both sides of the conditional must match the left hand side of th
 Try Catch Statement
 ___________________
 
+Solidity's try catch statement can only be used with external calls or constructor calls using ``new``. The
```
LucasSte, line 228:
```suggestion
Solidity's try-catch statement can only be used with external calls or constructor calls using ``new``. The
```

docs/targets/solana.rst
```diff
@@ -9,9 +9,12 @@ Solana has the following differences to Ethereum Solidity:
 - An address literal has to be specified using the ``address"36VtvSbE6jVGGQytYWSaDPG7uZphaxEjpJHUUpuUbq4D"`` syntax
 - There is no ``ecrecover()`` builtin function, but there is a ``signatureVerify()`` function which can check ed25519
   signatures.
-- Solana has no concept of gas, so there is no gas functions
+- There is no concept of gas, so there is no gas functions
 - Solana balance is stored in a ``uint64``, so *address* ``.balance``, ``.transfer()`` and ``.send()``
   all use ``uint64`` rather than ``uint256``.
+- Solana uses different accounts for the program code, and the contract data.
+- Programs are upgradable. There is no need to implement upgrades in Solidity.
+- Different builtins are different.
```
LucasSte, line 17: Please, remove this line.

src/bin/languageserver/mod.rs
```diff
@@ -669,7 +669,7 @@ impl SolangServer {
                 SolangServer::construct_expr(expr1, lookup_tbl, symtab, ns);
             }
 
-            ast::Expression::Ternary(_locs, _typ, expr1, expr2, expr3) => {
+            ast::Expression::Conditional(_locs, _typ, expr1, expr2, expr3) => {
```
LucasSte, line 672: If don't like this name, because a `if(condition)` is also a conditional. How about `TernaryConditional`?

LucasSte (changes requested):

docs/running.rst
seanyoung: Fixed

docs/language/expressions.rst
seanyoung: This does not make much sense, I've re-written it.

docs/language/expressions.rst
seanyoung: I don't think `sign shifts` is very clear. I've changed it to `sign changes`

docs/running.rst
xermicus: Been there, done that :grin:

src/bin/languageserver/mod.rs
seanyoung: It is known in many languages as the conditional operator: https://en.wikipedia.org/wiki/Ternary_conditional_operator

How do you feel about `Expression::ConditialOperator()`

docs/targets/solana.rst
seanyoung: I've changes it to:

docs/targets/substrate.rst
seanyoung: Good idea.

docs/language/statements.rst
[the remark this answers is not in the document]
seanyoung: I like it, I've updated it to this. Thanks

src/bin/languageserver/mod.rs
LucasSte: That is a good option

LucasSte (changes requested):

docs/installing.rst
```diff
@@ -85,7 +88,7 @@ Option 5: Build Solang from source
 ----------------------------------
 
 In order to build Solang from source, you will need rust 1.63.0 or higher,
-and a build of LLVM based on the Solana LLVM tree. There are a few patches which are not upstream yet.
+and a build of LLVM based on the Solana LLVM tree. There are a few LLVM patches required which are not upstream yet.
```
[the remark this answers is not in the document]
seanyoung: Very interesting. I did not know that.

docs/language/expressions.rst
```diff
@@ -177,10 +178,10 @@ an interface at runtime, which can be used to write a `supportsInterface()` func
...
 
 The contract code for a contract, i.e. the binary WebAssembly or BPF, can be retrieved using the
-``type(c).creationCode`` and ``type(c).runtimeCode`` fields, as ``bytes``. In Ethereum,
-the constructor code is in the ``creationCode`` WebAssembly and all the functions are in
-the ``runtimeCode`` WebAssembly or BPF. Parity Substrate has a single WebAssembly code for both,
-so both fields will evaluate to the same value.
+``type(c).creationCode`` and ``type(c).runtimeCode`` fields, as ``bytes``. On EVM,
+the constructor code is in the ``creationCode`` and all the functions are in
+the ``runtimeCode``. Both Parity Substrate and Solana use the same
+code for both, so both fields will evaluate to the same value.
```
LucasSte, line 184:
```suggestion
the ``runtimeCode``. Parity Substrate and Solana use the same
code for both, so those fields will evaluate to the same value.
```
`both` was repeated three times.

LucasSte (approved):
Just one nitpick.

PR #1074, opened 2022-11-17
Use IndexMap for enum declaration

LucasSte:
Previously, we had a `HashMap<String, (Loc, usize)>` to save the items of an enumeration. Retrieving the item name having its index is awful, so I changed the structure to `IndexMap<String, Loc>`, from which we can retrieve its name and Loc when we index it with an `usize`.

The context for this PR is that I need to retrieve the item name from its index when generating the Anchor IDL from a Solidity contract.

### DIFF
--- a/src/sema/dotgraphviz.rs
+++ b/src/sema/dotgraphviz.rs
@@ -2070,11 +2070,11 @@ impl Namespace {
             let enums = dot.add_node(Node::new("enums", Vec::new()), None, None);
 
             for decl in &self.enums {
-                let mut labels = vec![String::new(); decl.values.len()];
-
-                for (name, (_, pos)) in &decl.values {
-                    labels[*pos] = format!("value: {}", name);
-                }
+                let mut labels = decl
+                    .values
+                    .iter()
+                    .map(|(name, _)| format!("value: {}", name))
+                    .collect::<Vec<String>>();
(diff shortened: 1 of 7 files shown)


src/sema/dotgraphviz.rs
```diff
@@ -2072,8 +2072,8 @@ impl Namespace {
             for decl in &self.enums {
                 let mut labels = vec![String::new(); decl.values.len()];
 
-                for (name, (_, pos)) in &decl.values {
-                    labels[*pos] = format!("value: {}", name);
+                for (pos, value) in decl.values.iter().enumerate() {
+                    labels[pos] = format!("value: {}", value.0);
```
seanyoung, line 2076: It looks like this could be written without a for loop with `.collect::<Vec<String>>()`.

seanyoung (approved):
Very nice @LucasSte! The `HashMap<>` for enums has bothered me for a long time, and this is an elegant solution.

xermicus (approved):

src/sema/dotgraphviz.rs
LucasSte: Let me check.

LucasSte: Fixed

PR #1077, opened 2022-11-18
Validate doc examples

xermicus:
- Move full contract solidity code blocks into dedicated examples (and fix many syntax errors in the process)
- Check for their successful compilation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,7 @@ pretty_assertions = "1.2"
 byte-slice-cast = "1.2.1"
 borsh = "0.9.3"
 walkdir = "2"
+rayon = "1"
 
 [package.metadata.docs.rs]
 no-default-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ path-slash = "0.2"
 pretty_assertions = "1.2"
 byte-slice-cast = "1.2.1"
 borsh = "0.9.3"
+walkdir = "2"
 
 [package.metadata.docs.rs]
 no-default-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,6 @@ path-slash = "0.2"
 pretty_assertions = "1.2"
 byte-slice-cast = "1.2.1"
 borsh = "0.9.3"
-walkdir = "2"
 rayon = "1"
 
 [package.metadata.docs.rs]
--- a/docs/code_gen_options.rst
+++ b/docs/code_gen_options.rst
@@ -45,20 +45,8 @@ of arithmetic may be replaced:
 - 256 or 128 bit divide maybe replaced by 64 bit divide or shift
 - 256 or 128 bit modulo maybe replaced by 64 bit modulo or bitwise and
 
-.. code-block:: solidity
-
-    contract test {
-        function f() public {
-            for (uint i = 0; i < 10; i++) {
-                // this multiply can be done with a 64 bit instruction
-                g(i * 100));
-            }
-        }
-
-        function g(uint256 v) internal {
-            // ...
-        }
-    }
+.. include:: ../examples/strength_reduce.sol
+  :code: solidity
 
 Solang uses reaching definitions to track the known bits of the variables; here solang knows that i can have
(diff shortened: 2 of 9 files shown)


seanyoung:
@xermicus thank you very much, this looks great. It must have been a lot of work to fix all the examples and give them names. :1st_place_medal: 
I'll have to go over it in detail again.

examples/user_defined_type.sol
```diff
@@ -0,0 +1,10 @@
+
```
seanyoung, line 1: This file is not formatted correctedly unless the first blank line is removed

seanyoung (changes requested):
This looks great, and I did not know about the `rayon` crate, very useful. We could reuse that for `contract_testcases`. 

- I don't think the solidity fragments belong in `examples` though. What do you think of `docs/examples/` or `docs/solidity/`?
- The `.gitkeep` files are no longer needed (I think)

examples/user_defined_type.sol
xermicus: Hmm yes. There might be rare cases where the formatter can't handle the file :shrug: Thanks for pointing out.

xermicus:
I really do like the `rayon` crate, it is my favorite data parallel library :blush:

> I don't think the solidity fragments belong in examples though. What do you think of docs/examples/ or docs/solidity/?

That makes it too compliclated in my opinion. I just updated the fragments where it makes sense to full contract example. So we can have as a general rule of thumb: If it is a full contract example, put it into the examples, otherwise just use a solidity code block. WDYT?

seanyoung (approved):
Very nice indeed.

If you feel like it, there are bunch of examples with `.. code-block:: javascript` that should really be `.. code-block:: solidity`.

xermicus:
>  If you feel like it, there are bunch of examples with .. code-block:: javascript that should really be .. code-block:: solidity

I noticed that too, but when I started to change them the doc generator would complain about the syntax. I suspect this is mainly due to some of our custom language additions, so I just let them be.

LucasSte:
> * I don't think the solidity fragments belong in `examples` though. What do you think of `docs/examples/` or `docs/solidity/`?

I have reviewed your files and found no problems, but I share the same opinion as @seanyoung. These Solidity files should not be under `examples`, because a proper example should have a thorough explanation of what they are doing, including many comments for usage.

Creating a top-level `examples` folder will mislead users, making them think we are providing them with comprehensive examples of how to use Solidity for either Solana or Substrate.

docs/examples/contract_storage_accessor.sol
```diff
@@ -0,0 +1,9 @@
+contract ethereum {
+    // As a public mapping,this creates accessor function called balance, which takes
```
LucasSte, line 2:
```suggestion
    // As a public mapping, this creates accessor function called balance, which takes
```

docs/targets/substrate.rst
```diff
@@ -15,7 +15,7 @@ The Parity Substrate has the following differences to Ethereum Solidity:
 - An `assert()`, `require()`, or `revert()` executes the wasm unreachable instruction. The reason code is lost
 
 There is an solidity example which can be found in the
-`examples <https://github.com/hyperledger/solang/tree/main/examples>`_
+`examples <https://github.com/hyperledger/solang/tree/ma./examples>`_
```
LucasSte, line 18: Why has this changed?

docs/examples.rst
```diff
@@ -10,7 +10,7 @@ General examples
 Flipper
 _______
 
-This is the `ink! flipper example <https://github.com/paritytech/ink/blob/v3.3.0/examples/flipper/lib.rs>`_
+This is the `ink! flipper example <https://github.com/paritytech/ink/blob/v3.3./examples/flipper/lib.rs>`_
```
LucasSte, line 13: The new link does not work.

LucasSte (changes requested):

docs/examples/contract_storage_accessor.sol
LucasSte: Nitpicking

LucasSte (approved):

PR #1078, opened 2022-11-21
Use named fields for casts in ast::Expression

seanyoung:

### DIFF
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -496,63 +496,69 @@ pub fn expression(
         ast::Expression::And(loc, left, right) => {
             and(left, cfg, contract_no, func, ns, vartab, loc, right, opt)
         }
-        ast::Expression::CheckingTrunc(loc, ty, e) => {
-            checking_trunc(loc, e, ty, cfg, contract_no, func, ns, vartab, opt)
+        ast::Expression::CheckingTrunc { loc, to, expr } => {
+            checking_trunc(loc, expr, to, cfg, contract_no, func, ns, vartab, opt)
         }
-        ast::Expression::Trunc(loc, ty, e) => Expression::Trunc(
+        ast::Expression::Trunc { loc, to, expr } => Expression::Trunc(
             *loc,
-            ty.clone(),
-            Box::new(expression(e, cfg, contract_no, func, ns, vartab, opt)),
+            to.clone(),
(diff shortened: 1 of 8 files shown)


seanyoung:
I think I'd like to convert this piecemeal, otherwise it's too much work in one go.

xermicus (approved):
Nice, thanks! Totally fine for me to do this in smaller steps

src/codegen/expression.rs
```diff
@@ -496,63 +496,69 @@ pub fn expression(
...
             *loc,
-            ty.clone(),
-            Box::new(expression(e, cfg, contract_no, func, ns, vartab, opt)),
+            to.clone(),
+            Box::new(expression(expr, cfg, contract_no, func, ns, vartab, opt)),
         ),
-        ast::Expression::Cast(loc, ty @ Type::Address(_), e) => {
-            if let Ok((_, address)) = eval_const_number(e, ns) {
-                Expression::NumberLiteral(*loc, ty.clone(), address)
+        ast::Expression::Cast { loc, to, expr } if matches!(to, Type::Address(_)) => {
```
LucasSte, line 517: I think you can still write `ast::Cast{loc, to @ Type::Address(_), expr)` here.

src/sema/ast.rs
```diff
@@ -686,13 +686,37 @@ pub enum Expression {
...
+    },
+    Cast {
+        loc: pt::Loc,
+        to: Type,
+        expr: Box<Expression>,
+    },
+    BytesCast {
+        loc: pt::Loc,
+        to: Type,
+        from: Type,
```
LucasSte, line 717:
```suggestion
        from: Type,
        to: Type,
```

I would switch the order here for better understanding, as we always go from a type to another one.

LucasSte (changes requested):

src/codegen/expression.rs
seanyoung: I thought that too, but:
```
error: expected `,`
   --> src/codegen/expression.rs:517:40
    |
517 |         ast::Expression::Cast { loc, to@ Type::Address(_), expr } => {
    |         ---------------------          ^
    |         |
    |         while parsing the fields for this pattern

```
Surprising, isn't it?

LucasSte (approved):

PR #1080, opened 2022-11-24
Substrate: Support contracts node v0.22

xermicus:
- Change the linker to use stabilized APIs for any remaining `__unstable__`
- Update integration tests:
  - Introduce a helper for the (currently) broken contracts `query` API
  - Use `WeightV2` for gas limits
  - Use estimated gas from dry run instead of hardcoded value
  - Compile test contracts in parallel
- Updates to the `latest` contracts CI image for pallet contracts v0.22.1. I'm aware that this is not ideal but it's what's available right now. There shouldn't be a breaking release soon, we can switch to `production` as soon as its here.

### DIFF
--- a/integration/substrate/package.json
+++ b/integration/substrate/package.json
@@ -5,7 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "tsc; ts-mocha -t 20000 *.spec.ts",
-    "build": "solang compile *.sol test/*.sol --target substrate -v"
+    "evt": "tsc; ts-mocha -t 20000 store.spec.ts",
+    "build": "parallel solang compile -v --target substrate ::: *.sol test/*.sol"
   },
   "author": "Sean Young <sean@mess.org>",
   "license": "Apache-2.0",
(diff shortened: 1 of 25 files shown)


integration/substrate/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "tsc; ts-mocha -t 20000 *.spec.ts",
-    "build": "solang compile *.sol test/*.sol --target substrate -v"
+    "build": "parallel solang compile -v --target substrate ::: *.sol test/*.sol"
```
seanyoung, line 8: We should make solang parallel at some point

integration/substrate/package.json
```diff
@@ -5,7 +5,7 @@
...
   },
   "author": "Sean Young <sean@mess.org>",
```
seanyoung, line 10: This really needs updating, you could add your name here (I think it has to be `contributors` rather than `author`.

seanyoung (approved):

integration/substrate/package.json
xermicus: Good idea, I imagine compiling all contracts in parallel should be easy to do

LucasSte (approved):

PR #1081, opened 2022-11-24
Encode custom width integers using the next power of two width

LucasSte:
The Anchor IDL does not support custom width integers. Instead of creating a user-defined type for them, we will encode them using the smaller integer width greater than the given one.

### DIFF
--- a/src/codegen/encoding/borsh_encoding.rs
+++ b/src/codegen/encoding/borsh_encoding.rs
@@ -148,13 +148,13 @@ impl AbiEncoding for BorshEncoding {
 
     fn get_encoding_size(&self, expr: &Expression, ty: &Type, ns: &Namespace) -> Expression {
         match ty {
-            Type::Enum(_)
-            | Type::Uint(_)
-            | Type::Int(_)
-            | Type::Contract(_)
-            | Type::Bool
-            | Type::Address(_)
-            | Type::Bytes(_) => {
+            Type::Uint(n) | Type::Int(n) => Expression::NumberLiteral(
+                Loc::Codegen,
+                Type::Uint(32),
+                BigInt::from(n.next_power_of_two() / 8),
+            ),
+
+            Type::Enum(_) | Type::Contract(_) | Type::Bool | Type::Address(_) | Type::Bytes(_) => {
                 let size = ty.memory_size_of(ns);
                 Expression::NumberLiteral(Loc::Codegen, Type::Uint(32), size)
             }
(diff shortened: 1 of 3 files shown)


src/codegen/encoding/borsh_encoding.rs
```diff
@@ -235,17 +235,40 @@ impl BorshEncoding {
                 Expression::NumberLiteral(Loc::Codegen, Type::Uint(32), BigInt::from(1u8))
             }
 
-            Type::Uint(length) | Type::Int(length) => {
+            Type::Uint(width) | Type::Int(width) => {
+                let encoding_size = width.next_power_of_two();
+                let expr = if encoding_size != *width {
+                    if matches!(expr_ty, Type::Uint(_)) {
```
seanyoung, line 241: We have `expr_ty.is_signed_int()` for this

seanyoung (approved):
How is this tested?

tests/borsh_encoding.rs
```diff
@@ -106,96 +106,83 @@ impl BorshToken {
...
-        }
-        n if n <= 256 => {
-            let mut val = value.to_signed_bytes_le();
-            let byte_width = (width / 8) as usize;
-            match val.len().cmp(&byte_width) {
-                Ordering::Greater => {
-                    while val.len() > byte_width {
-                        val.pop();
-                    }
+    if width <= 8 {
```
xermicus, line 109: Just out of curiosity, why did you change this to be a `else if` chain?

xermicus (approved):

LucasSte:
> How is this tested?

In `primitives.rs` we have tests for all bit-widths that are passed as arguments to functions, so the encoding and decoding is implicitly tested there: 

https://github.com/hyperledger/solang/blob/cfbc7a9f6bbd13fce0b5bee62313177e4fe61037/tests/solana_tests/primitives.rs#L354

tests/borsh_encoding.rs
LucasSte: In the `match` statement, I wanted to match the width to an exact number, and now I have to deal with a range. If you know how to do this using `match`, please, let me know and I will change my code.

xermicus: [You can try something like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=986405b4429057f9e2af1a275ae283cc)

LucasSte: That looks nice! I'm going to change my code to match your suggestion.

PR #1085, opened 2022-11-28
Improve CI and other fixes

seanyoung:
- Solang CI already has python3, no need to install
- Docs needs git
- solang.io domain no longer registered (expired)
- Do not use ^ for crates

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,8 +25,6 @@ jobs:
         uses: actions/checkout@v3
       - name: Rust stable
         run: rustup default stable
-      - name: Install python3
-        run: apt-get -y install python3
       - uses: enarx/spdx@master
         with:
           licenses: Apache-2.0 MIT
@@ -44,11 +42,11 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
-      - name: Install Python
+      - name: Install Python and git
         run: |
           apt-get update
-          apt-get install -y python3-pip
-      - name: Install Docs requiremets
+          apt-get install -y python3-pip git
+      - name: Install Docs requirements.txt
         run : |
           pip install -r requirements.txt
       - name: Build docs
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -44,11 +42,11 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
-      - name: Install Python
+      - name: Install Python/git
```
LucasSte, line 45:
```suggestion
      - name: Install Python and git
```

LucasSte (approved):
Just a nitpick 😇

seanyoung:
I've been re-running this job, and it can still fail: https://github.com/hyperledger/solang/actions/runs/3563333271/jobs/5991349498

integration/substrate/array_struct_mapping_storage.spec.ts
```diff
@@ -4,7 +4,7 @@ import { ContractPromise } from '@polkadot/api-contract';
 
 describe('Deploy array_struct_mapping_storage contract and test', () => {
     it('array_struct_mapping_storage', async function () {
-        this.timeout(200000);
+        this.timeout(1000000);
```
xermicus, line 7: Unfortunately I don't think this will fix it (but instead give the spurious runs more time until they finally fail). I got these sporadic timeouts on my localhost as well, and tried to adjust the timeouts already without success :confused:

xermicus (changes requested):
Appreciate the effot! Unfortunately this will fix the root cause... Is it getting to annoying? I can try to find the cause later this week if you like

integration/substrate/array_struct_mapping_storage.spec.ts
seanyoung: ok, I've removed this part from the PR

xermicus (approved):
Thanks!

PR #1087, opened 2022-12-01
Add basic `ink!` to solidity call integration test

xermicus:
My original plan was to hack the substrate emit phase to always return a `Result` type (`Ok` variant which is just a single null byte), so that the `LanguageError` from `ink!` version 4 is respected when calling a solidity contract from `ink!`. But while doing so I realized: Currently the only way to call a solidity contract from `ink!` is to use the low level `CallBuilder` interface anyways. And since this interface does not care at all about the `LanguageError`, everything still works. So now, instead of hacking around in emit and holding up the next solang release, I'd rather implement the `LanguageError` story cleanly after the release (IIRC it means refactoring the scale encoder into codegen and then implementing the `Result` type).


integration/substrate/ink_cross_calls.spec.ts
```diff
@@ -0,0 +1,44 @@
+import expect from 'expect';
+import { weight, createConnection, deploy, transaction, aliceKeypair, query, } from './index';
+import { ContractPromise } from '@polkadot/api-contract';
+import { ApiPromise } from '@polkadot/api';
+import { KeyringPair } from '@polkadot/keyring/types';
+
+describe('Deploy flipper contract and test', () => {
```
LucasSte, line 7: This must not be the name of your test. The name ought to explain what you are testing.

LucasSte (changes requested):

integration/substrate/ink_cross_calls.spec.ts
xermicus: Yes this is a copy-pasta error, good catch

integration/substrate/ink/caller/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "caller"
+version = "0.1.0"
+authors = ["[your_name] <[your_email]>"]
```
seanyoung, line 4: Maybe either remove the line or fill it in. It's a test so not terribly important

seanyoung (approved):

LucasSte (approved):

PR #1089, opened 2022-12-01
Use discriminators on Solana instead of selectors ⚓

LucasSte:
In order to integrate with Anchor, Solang needs to support discriminators. This PR refactors the Solana target so that it works with eight byte discriminators instead of four byte selectors.

The discriminator depends only on the name of a function, so function name mangling now works for Solana too when overloading functions.

The Anchor and Solana integration tests are failing, because we need to merge [this PR](https://github.com/solana-labs/solana-solidity.js/pull/47) in the `@solana/solidity.js` repository and release a new version of the library.


src/sema/ast.rs
```diff
@@ -357,9 +360,15 @@ impl Function {
...
 
     /// Generate selector for this function
-    pub fn selector(&self) -> Vec<u8> {
+    pub fn selector(&self, ns: &Namespace) -> Vec<u8> {
         if let Some(selector) = &self.selector {
             selector.clone()
+        } else if ns.target == Target::Solana {
+            match self.ty {
+                FunctionTy::Constructor => discriminator("global", "new"),
+                FunctionTy::Receive => discriminator("global", "receive()"),
```
seanyoung, line 369: Receive functions do not have discriminators/selectors

seanyoung (changes requested):
I only remember when reviewing this: @xermicus added a bunch of code to mangle overloaded function names. Would it not be better to re-use that rather than not permitting function overloaded. This also makes Solang more consistent between Substrate and Solana.

LucasSte:
> I only remember when reviewing this: @xermicus added a bunch of code to mangle overloaded function names. Would it not be better to re-use that rather than not permitting function overloaded. This also makes Solang more consistent between Substrate and Solana.

I'm making this PR a draft again, because doing that will need many changes.

docs/language/builtins.rst
```diff
@@ -237,15 +237,15 @@ ABI encodes the arguments to bytes. Any number of arguments can be provided.
...
+abi.encodeWithSelector(selector, ...)
 ++++++++++++++++++++++++++++++++++++++++++++
 
 ABI encodes the arguments with the function selector first. After the selector, any number of arguments
 can be provided.
 
 .. code-block:: solidity
 
-    bytes foo = abi.encodeWithSelector(hex"01020304", uint16(0xff00), "ABCD");
+    bytes foo = abi.encodeWithSelector(hex"0102030405060708", uint16(0xff00), "ABCD");
```
seanyoung, line 248: This should be marked as an Solana example, just like the one below is marked Substrate

docs/language/builtins.rst
```diff
@@ -49,8 +49,8 @@ uint128 ``msg.value``
 bytes ``msg.data``
     The raw ABI encoded arguments passed to the current call.
 
-bytes4 ``msg.sig``
-    Function selector from the ABI encoded calldata, e.g. the first four bytes. This
+bytes4 (Substrate) or bytes8 (Solana) ``msg.sig``
+    Function selector from the encoded calldata, e.g. the first four or eight bytes. This
```
seanyoung, line 53: This should have a note saying that the selector is know as a discriminator on Solana.

docs/language/builtins.rst
```diff
@@ -237,15 +237,15 @@ ABI encodes the arguments to bytes. Any number of arguments can be provided.
 
 On Substrate, foo will be ``hex"f100"``. On Ethereum this will be ``hex"00000000000000000000000000000000000000000000000000000000000000f1"``.
 
-abi.encodeWithSelector(bytes4 selector, ...)
+abi.encodeWithSelector(selector, ...)
```
seanyoung, line 240: This should have a note saying that the selector is know as a discriminator on Solana.

tests/contract_testcases/solana/functions/selectors-must-be-different.sol
```diff
@@ -10,23 +10,23 @@ contract d {
 	int public c;
 
 	// error: selector is the same as c
-	function f1() selector=hex"c3da42b8" public {}
+	function f1() selector=hex"13fbd725feff6e10" public {}
```
seanyoung, line 13: Since the IDL does not specify selectors/discriminators yet, so how can we support overriding selectors on Solana? This would make contracts which no longer work, unless I'm mistaken.

seanyoung (changes requested):

xermicus (changes requested):
LGTM overall!

tests/contract_testcases/solana/functions/selectors-must-be-different.sol
LucasSte: That's a good point. Should we disable overriding selectors on Solana?

src/sema/contracts.rs
```diff
@@ -805,14 +775,18 @@ fn substrate_requires_public_functions(contract_no: usize, ns: &mut ast::Namespa
...
-    if !ns.target.is_substrate() || ns.diagnostics.any_errors() {
+fn unique_constructor_names(contract_no: usize, ns: &mut ast::Namespace) {
+    if ns.diagnostics.any_errors() {
         return;
     }
 
     let mut functions = HashMap::new();
     for f in &ns.contracts[contract_no].functions {
         let func = &ns.functions[*f];
+        if !func.is_public() {
```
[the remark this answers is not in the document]
LucasSte: The problem with conflicting names only exist for the IDL file. When you call an internal or private function, the compiler handles this case.

src/sema/contracts.rs
```diff
@@ -1223,3 +1197,72 @@ fn compatible_visibility(left: &pt::Visibility, right: &pt::Visibility) -> bool
             | (pt::Visibility::Private(_), pt::Visibility::Private(_))
     )
 }
+
+/// This function checks which function names must be mangled given a contract.
+/// Afterwards, it guarantees that each function has a unique selector.
+fn mangle_function_names_and_verify_selector(contract_no: usize, ns: &mut Namespace) {
```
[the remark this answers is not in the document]
LucasSte: Yes. That seems like a good idea.

tests/contract_testcases/solana/functions/selectors-must-be-different.sol
LucasSte: I added a warning about this.

src/sema/ast.rs
```diff
@@ -268,13 +272,15 @@ pub struct Function {
...
     pub body: Vec<Statement>,
     pub symtable: Symtable,
-    // What events are emitted by the body of this function
+    /// What events are emitted by the body of this function
     pub emits_events: Vec<usize>,
-    // For overloaded functions this is the mangled (unique) name.
+    /// For overloaded functions this is the mangled (unique) name.
     pub mangled_name: String,
+    /// Which contracts should we use the mangled name in?
+    pub use_mangle_name_on: HashSet<usize>,
```
seanyoung, line 280: The name does not include `contracts` which is essential. How about `mangled_name_contracts`.

src/sema/expression.rs
```diff
@@ -1238,6 +1246,55 @@ impl Expression {
...
+                        loc,
+                        &Type::Bytes(selector_length),
+                        to,
+                        implicit,
+                        ns,
+                        diagnostics,
+                    )
+                }
+            }
+            (Type::FunctionSelector, Type::Uint(n) | Type::Int(n)) => {
```
seanyoung, line 1277: Why do we need conversion to `uint64` and even `int64`? Seems wrong.

src/sema/functions.rs
```diff
@@ -192,6 +192,14 @@ pub fn contract_function(
                     ));
                     success = false;
                 } else {
+                    if ns.target == Target::Solana {
+                        ns.diagnostics.push(Diagnostic::warning(
+                            *loc,
+                            "Selector (or discriminator) overriding does not work on Solana. \
+                              The IDL spec needs updating to make this feature available."
```
seanyoung, line 199: This error is not helpful. What needs updating? What is the user supposed to do? 

How about `selector overriding may not have the desired affect. The metadata file does not contain the overriden selector (known as the discriminator) so the contract will be unusable.`

seanyoung (changes requested):

src/sema/expression.rs
LucasSte: `Bytes(8)` can be converted to `uint64` and `int64`.

src/sema/functions.rs
LucasSte: Are we turning this into an error instead of a warning?

seanyoung: I should have said warning. The `solang idl` command produces selector overrides.

src/sema/expression.rs
LucasSte: I changed the error I wrote there to a warning to be more consistent with the error we get when converting `bytes8` to `uint64` (implicit cast now allowed).

src/emit/expression.rs
```diff
@@ -1349,10 +1349,15 @@ pub(super) fn expression<'a, T: TargetRuntime<'a> + ?Sized>(
 
             let start = unsafe { bin.builder.build_gep(data, &[offset], "start") };
 
-            if let Type::Bytes(n) = &returns[0] {
+            if matches!(returns[0], Type::Bytes(_) | Type::FunctionSelector) {
+                let n = match &returns[0] {
+                    Type::Bytes(n) => *n,
+                    Type::FunctionSelector => ns.target.selector_length(),
+                    _ => unreachable!(),
+                };
```
seanyoung, line 1357: How about `returns[0].bits()`?

seanyoung (changes requested):

xermicus (approved):

src/emit/expression.rs
seanyoung: Still better than open-coding it. You could also also add an `fn bytes(self, ns: &Namespace)` to `Type`.

LucasSte: Fixed

docs/targets/solana.rst
```diff
@@ -15,6 +15,7 @@ Solana has the following differences to Ethereum Solidity:
 - Solana uses different accounts for the program code, and the contract data.
 - Programs are upgradable. There is no need to implement upgrades in Solidity.
 - Solana provides different builtins, e.g. ``tx.program_id`` and ``tx.accounts``.
+- Selectors are eight bytes wide and often called "discriminators".
```
seanyoung, line 18:
```suggestion
- Function selectors are eight bytes and known as *discriminators*.
```

src/sema/functions.rs
```diff
@@ -192,6 +192,15 @@ pub fn contract_function(
                     ));
                     success = false;
                 } else {
+                    if ns.target == Target::Solana {
+                        ns.diagnostics.push(Diagnostic::warning(
+                            *loc,
+                            "Selector (or discriminator) overriding may not have the \
+                              desired effect. The metadata file does not contains the overridden \
+                              selector, so the contract will be unused."
```
seanyoung, line 200: Needs condition "if function has body"

```suggestion
                              selector, so the contract will not function."
```

tests/solana_tests/call.rs
```diff
@@ -115,11 +115,11 @@ fn external_raw_call_with_returns() {
     let mut vm = build_solidity(
         r#"
         contract bar0 {
-            bytes4 private constant SELECTOR = bytes4(keccak256(bytes('test_bar(int64)')));
+            bytes8 private constant SELECTOR = bytes8(sha256(bytes('global:test_bar')));
 
             function test_other(bar1 x) public returns (int64) {
                 bytes select = abi.encodeWithSelector(SELECTOR, int64(7));
-                bytes signature = abi.encodeWithSignature("test_bar(int64)", int64(7));
+                bytes signature = abi.encodeWithSignature("global:test_bar", int64(7));
```
seanyoung, line 122: This is not right. We can't suddenly require a `global:` prefix.

seanyoung (changes requested):

src/sema/functions.rs
LucasSte: Please, be more attentive in your suggestions. You have given me this message as a suggestion:

https://github.com/hyperledger/solang/pull/1089#discussion_r1044441794

seanyoung: I originally suggested `unusable`. You changed this to `unused`, which is wrong.

tests/solana_tests/call.rs
seanyoung: I had not fully considered `abi.encodeWithSignature()`. We can prefix `global:` at runtime, but we can't do the camel-case conversion. I guess we have no choice but to leave this up to the solidity dev. In that case, we might as well let the solidity dev prefix `global:`.

Maybe you are right, we should document the discriminator including camel-case conversion.

Hopefully this will all go away once discriminators are in the idl.

src/sema/functions.rs
LucasSte: Sorry, that was a typing mistake of mine.

tests/solana_tests/call.rs
LucasSte: I've updated the documentation. Let me know what you think of it.

seanyoung: Where are the changes?

LucasSte: `docs/language/builtins.rst`. I described there how to use `abi.encodeWithSelector` and `abi.encodeWithSignature`.

docs/language/builtins.rst
```diff
@@ -237,30 +238,41 @@ ABI encodes the arguments to bytes. Any number of arguments can be provided.
...
+the the signature is known as the discriminator image. It is the function name converted to camel case and
+preceded by the prefix ``global:``, so this builtin is equivalent to
+``abi.encodeWithSelector(bytes8(sha256(bytes("global:my_function_name"))), ...)`` for Solana.
 
 .. code-block:: solidity
 
+    // On Solana, the call should be abi.encodeWithSignature("global:test2", uint64(257))
     bytes foo = abi.encodeWithSignature("test2(uint64)", uint64(257));
 
-On Substrate, foo will be ``hex"296dacf0_0101_0000__0000_0000"``. On Ethereum this will be ``hex"296dacf0_0000000000000000000000000000000000000000000000000000000000000101"``.
```
seanyoung, line 262: This is very messy.

```
On Substrate, the signature is the name of the function followed by its arguments, for example:

.. code-block:: solidity
    bytes foo = abi.encodeWithSignature("foo_bar(uint64)", uint64(257));
    
``foo`` will be ``hex"e934aa71_0101_0000__0000_0000"``.  This is equivalent to ``abi.encodeWithSelector(bytes4(keccak256("test2(uint64)")), ...)``.  

On Solana, the signature is known as the discriminator image. It is the function name without any arguments,  converted to camel case, and preceded by the prefix ``global:``. For example, if you had the function ``foo_bar(uint64)``, the discriminator image would be ``global:fooBar``.

.. code-block:: solidity
    bytes foo = abi.encodeWithSignature("global:fooBar", uint64(257));
 
This builtin is equivalent to
``abi.encodeWithSelector(bytes8(sha256(bytes("global:fooBar"))), ...)`` for Solana.
```

seanyoung: Please can you check and insert, thank you.

seanyoung (approved):

docs/language/builtins.rst
LucasSte: Fixed

PR #1091, opened 2022-12-05
Ensure that rational compares are not permitted

seanyoung:
This solidity was not rejected and causes a panic in emit, because rationals should be compile-time constants.

```
function foo1(uint64 a, uint64 b) returns (bool) {
	return (a/b) >= 0.05;
}
```


xermicus (changes requested):
Thanks @seanyoung for so quickly tackling this!

But aren't there more operations? The following Solidity does crash the comiler as well:

```solidity
function foo1(uint64 a, uint64 b) public returns (bool) {
	return (a << b) >= 0.05;
}
```

xermicus (approved):

src/sema/expression.rs
```diff
@@ -3905,11 +3937,19 @@ fn equal(
...
-    ))
+    );
+
+    if ty.is_rational() {
+        if let Err(diag) = eval_const_rational(&expr, ns) {
+            diagnostics.push(diag);
+        }
+    }
+
+    Ok(expr)
```
LucasSte, line 3952: What about `NotEqual`?

tests/contract.rs
```diff
@@ -33,18 +34,24 @@ fn contract_tests(file_path: &str, target: Target) -> io::Result<()> {
...
+
     for entry in read_dir(path)? {
         let path = entry?.path();
 
         if path.is_dir() {
             recurse_directory(path, target)?;
         } else if let Some(ext) = path.extension() {
             if ext.to_string_lossy() == "sol" {
-                parse_file(path, target)?;
+                entries.push(path);
```
LucasSte, line 46: Why is this change necessary?

seanyoung: I changed the test code so it uses `rayon` to use all available cores. This makes the test much faster

src/sema/expression.rs
seanyoung: `NotEqual` is implement with `fn equal()` , so this works fine too. I've added a test for this.

tests/contract.rs
LucasSte: We already run Solana contracts, Substrate contracts and EVM contracts tests in parallel as they are different `#[test]` functions functions. Having another layer of parallelism here may slow down the tests (I have not tested though).

LucasSte (approved):

PR #1092, opened 2022-12-05
Fix github deprecated warnings

seanyoung:
Every CI run we get tons of warnings like:

The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.

It looks like actions-rs/toolchain is unmaintained:

https://github.com/actions-rs/toolchain/issues/221

Many projects are switching over to dtolnay/rust-toolchain, let's do the same.


xermicus (approved):

.github/workflows/build-llvm.yml
```diff
@@ -105,7 +105,7 @@ jobs:
   mac-intel:
     name: Mac Intel
     needs: create
-    runs-on: macos-latest
+    runs-on: macos-11
```
LucasSte, line 108: Why has this changed?

LucasSte: Sorry for being pedantic. You did not say anything about this in your PR summary.

seanyoung: The CI output gives a whole bunch of warnings, see https://github.com/hyperledger/solang/actions/runs/3628240205

One of those warnings is:
```
macOS-latest pipelines will use macOS-12 soon. For more details, see https://github.com/actions/runner-images/issues/6384
 ```
So this fixes the warning by ensuring we continue to use macOS-11 for building. The warning goes away and we remain compatible with macOS 11.

LucasSte: I'm already using MacOS 13, so we should change the pipeline to 12 soon!

LucasSte (approved):

.github/workflows/build-llvm.yml
seanyoung: We should ensure that we are compatible with the oldest possible version we can support (without much effort). Why move to 12? Apple have good binary compatibility.

PR #1094, opened 2022-12-08
Add voting ballot example

xermicus:
While I was searching for some good example to use in the blog post, I came across [this ballot example](https://docs.soliditylang.org/en/v0.8.17/solidity-by-example.html?highlight=comments#voting). It compiles without any warnings for Substrate and Solana. How do you feel about adding it?

### DIFF
--- a/examples/ballot.sol
+++ b/examples/ballot.sol
@@ -1,0 +1,156 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.7.0 <0.9.0;
+/// @title Voting with delegation.
+contract Ballot {
+    // This declares a new complex type which will
+    // be used for variables later.
+    // It will represent a single voter.
+    struct Voter {
+        uint weight; // weight is accumulated by delegation
+        bool voted;  // if true, that person already voted
+        address delegate; // person delegated to
+        uint vote;   // index of the voted proposal
+    }
+
+    // This is a type for a single proposal.
+    struct Proposal {
+        bytes32 name;   // short name (up to 32 bytes)
+        uint voteCount; // number of accumulated votes
+    }
+
+    address public chairperson;
+
+    // This declares a state variable that
+    // stores a `Voter` struct for each possible address.
+    mapping(address => Voter) public voters;
+
+    // A dynamically-sized array of `Proposal` structs.
+    Proposal[] public proposals;
+
+    /// Create a new ballot to choose one of `proposalNames`.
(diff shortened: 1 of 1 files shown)


seanyoung (approved):

examples/ballot.sol
```diff
@@ -0,0 +1,156 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.7.0 <0.9.0;
```
LucasSte, line 2: This is not necessary for Solang.

examples/ballot.sol
```diff
@@ -0,0 +1,156 @@
...
+            }));
+        }
+    }
+
+    // Give `voter` the right to vote on this ballot.
+    // May only be called by `chairperson`.
+    function giveRightToVote(address voter) external {
+        // If the first argument of `require` evaluates
+        // to `false`, execution terminates and all
+        // changes to the state and to Ether balances
```
LucasSte, line 54: Isn't it a good idea to change references of Ethereum to Substrate or Solana here?

examples/ballot.sol
xermicus: This is an unmodified version, directly from the solidity documentation. I thought it would be a neat showcase for us.

examples/ballot.sol
xermicus: See my comment above. Do you think we should slightly modify the example?

examples/ballot.sol
LucasSte: The pragma can remain here. It has not effect for us. Maybe a comment to explain that is useful.

examples/ballot.sol
LucasSte: I don't think the code needs any changes. It is just useful to remove references to Ethereum from the comments, because developers will use these comments to understand how the compiler/environment works.

examples/ballot.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    // Give `voter` the right to vote on this ballot.
+    // May only be called by `chairperson`.
+    function giveRightToVote(address voter) external {
+        // If the first argument of `require` evaluates
+        // to `false`, execution terminates and all
+        // changes to the state and to account balances
+        // are reverted.
+        // This used to consume all gas in old EVM versions, but
+        // not anymore.
```
LucasSte, line 58: I'd remove this too, because we don't have an EVM.

LucasSte (approved):

PR #1100, opened 2022-12-14
Only emit in codegen when needed

seanyoung:
The contract binary is always generated in codegen, but never used in codegen. We can replace `type(contract).runtimeCode` with the binary code in codegen, which removes the need  for 
 `Expression::CodeLiteral` in codegen. We only generate this if needed.

This has an added bonus for debugging issues: if there is an issue in emit, then `--emit cfg` will still 
work and not crash before emitting the cfg (unless `type(foo).runtimeCode` is present).

### DIFF
--- a/src/emit/mod.rs
+++ b/src/emit/mod.rs
@@ -25,7 +25,7 @@ mod storage;
 mod strings;
 pub mod substrate;
 
-use crate::codegen::cfg::HashTy;
+use crate::codegen::{cfg::HashTy, Options};
 use crate::emit::binary::Binary;
 use crate::sema::ast;
 
@@ -383,5 +383,35 @@ impl Target {
         } else {
             ""
         }
+    }
+}
+
+impl ast::Contract {
+    /// Generate the binary. This can be used to generate llvm text, object file
+    /// or final linked binary.
+    pub fn binary<'a>(
+        &'a self,
+        ns: &'a ast::Namespace,
+        context: &'a inkwell::context::Context,
+        filename: &'a str,
+        opt: &'a Options,
+    ) -> binary::Binary {
+        binary::Binary::build(context, self, ns, filename, opt)
+    }
+
+    /// Generate the final program code for the contract
+    pub fn emit(&self, ns: &ast::Namespace, opt: &Options) -> Vec<u8> {
+        if ns.target == Target::EVM {
+            return vec![];
+        }
+
+        self.code
+            .get_or_init(move || {
+                let context = inkwell::context::Context::create();
+                let filename = ns.files[self.loc.file_no()].path.to_string_lossy();
+                let binary = self.binary(ns, &context, &filename, opt);
+                binary.code(Generate::Linked).expect("llvm build")
+            })
+            .to_vec()
     }
 }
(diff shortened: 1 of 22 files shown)


src/sema/ast.rs
```diff
@@ -649,7 +650,8 @@ pub struct Contract {
     pub initializer: Option<usize>,
     pub default_constructor: Option<(Function, usize)>,
     pub cfg: Vec<ControlFlowGraph>,
-    pub code: Vec<u8>,
+    /// Compiled program. Only available after emit.
+    pub code: OnceCell<Vec<u8>>,
```
LucasSte, line 654: Why do we need an `OnceCell` here?

src/sema/contracts.rs
```diff
@@ -46,36 +48,48 @@ impl ast::Contract {
...
             program_id: None,
         }
     }
 
-    /// Generate contract code for this contract
+    /// Generate the binary. This can be used to generate llvm text, object file
+    /// or final linked binary.
     #[cfg(feature = "llvm")]
-    pub fn emit<'a>(
+    pub fn binary<'a>(
```
LucasSte, line 62: This function does not belong in sema.

src/sema/contracts.rs
```diff
@@ -46,36 +48,48 @@ impl ast::Contract {
...
-            opt,
-            math_overflow_check,
-            generate_debug_info,
-            log_api_return_codes,
-        )
+        emit::binary::Binary::build(context, self, ns, filename, opt)
+    }
+
+    /// Generate the final program code for the contract
+    pub fn emit(&self, ns: &ast::Namespace, opt: &Options) -> Vec<u8> {
```
LucasSte, line 73: This function does not belong in sema.

src/sema/ast.rs
seanyoung: In codegen, we don't have `mut` access to the Namespace. We generate the cfg and assume the namespace is read-only. However, we are storing the compiled code there to avoid double compilation.

src/sema/contracts.rs
seanyoung: Good point.

src/emit/mod.rs
```diff
@@ -385,3 +385,39 @@ impl Target {
...
+                    let filename = ns.files[self.loc.file_no()].path.to_string_lossy();
+
+                    let binary = self.binary(ns, &context, &filename, opt);
+
+                    binary.code(Generate::Linked).expect("llvm build")
+                })
+                .clone();
+        }
+
+        code
```
xermicus, line 421: Nit: early return

```suggestion
        if ns.target == Target::EVM {
            return vec![];
        }

        self.code
            .get_or_init(move || {
                let context = inkwell::context::Context::create();

                let filename = ns.files[self.loc.file_no()].path.to_string_lossy();

                let binary = self.binary(ns, &context, &filename, opt);

                binary.code(Generate::Linked).expect("llvm build")
            })
            .to_vec()
```

xermicus (approved):

LucasSte (approved):
The lints test is failing. Please, fix it before merging.

PR #1101, opened 2022-12-14
Update roadmap

LucasSte:
Version 0.2 has already been released. It can be removed from the roadmap.


seanyoung (approved):
The first three items from v0.3 are also completed already

xermicus (approved):

LucasSte:
> The first three items from v0.3 are also completed already

I'm going to fix this before merging.

PR #1102, opened 2022-12-16
Fix rust 1.66 clippy warnings and run crate doc tests

seanyoung:

### DIFF
--- a/src/file_resolver.rs
+++ b/src/file_resolver.rs
@@ -134,13 +134,12 @@ impl FileResolver {
         // first check maps
         let mut iter = path.iter();
         if let Some(first_part) = iter.next() {
-            let relpath: PathBuf = iter.collect();
+            let relpath: &PathBuf = &iter.collect();
 
             for (import_no, import) in self.import_paths.iter().enumerate() {
                 if let (Some(mapping), import_path) = import {
                     if first_part == mapping {
-                        // match!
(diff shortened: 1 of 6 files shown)


src/file_resolver.rs
```diff
@@ -139,7 +139,8 @@ impl FileResolver {
             for (import_no, import) in self.import_paths.iter().enumerate() {
                 if let (Some(mapping), import_path) = import {
                     if first_part == mapping {
-                        // match!
+                        // clippy 1.66 complains that & in &relpath is unnecessary, but we can't remove it due to the loop
+                        #[allow(clippy::needless_borrow)]
```
xermicus, line 143: Unnecessary as can do the following instead:

```rust
            let relpath: &PathBuf = &iter.collect();

            for (import_no, import) in self.import_paths.iter().enumerate() {
                if let (Some(mapping), import_path) = import {
                    if first_part == mapping {
                        if let Ok(full_path) = import_path.join(relpath).canonicalize() {
```

xermicus (changes requested):

src/file_resolver.rs
seanyoung: Brilliant thanks @xermicus

xermicus (approved):
Thanks

LucasSte (approved):

PR #1103, opened 2022-12-16
Ensure no data is returned when a function has no return values

seanyoung:

### DIFF
--- a/src/codegen/dispatch.rs
+++ b/src/codegen/dispatch.rs
@@ -257,8 +257,8 @@ fn add_dispatch_case(
             },
         );
     } else {
-        // TODO: On Solana, We could elide setting the return data if this function calls no external functions
-        // are called and replace this with a simple Instr::Return, which does not set any return data.
+        // TODO: On Solana, we could elide setting the return data if this function calls no external functions
+        // and replace this with a simple Instr::Return, which does not set any return data.
         //
         // The return data buffer is empty when Solana VM first executes a program, but if another program is
         // called via CPI then that program may set return data. We must clear this buffer, else return data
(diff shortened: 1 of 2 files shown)


src/codegen/dispatch.rs
```diff
@@ -256,10 +256,23 @@ fn add_dispatch_case(
                 data_len: zext_len,
             },
         );
+    } else {
+        // TODO: On Solana, We could elide setting the return data if this function calls no external functions
+        // are called and replace this with a simple Instr::Return.
```
LucasSte, line 261: This comment isn't clear.

LucasSte (changes requested):

src/codegen/dispatch.rs
seanyoung: Is this better?

src/codegen/dispatch.rs
```diff
@@ -256,10 +256,27 @@ fn add_dispatch_case(
                 data_len: zext_len,
             },
         );
+    } else {
+        // TODO: On Solana, We could elide setting the return data if this function calls no external functions
+        // are called and replace this with a simple Instr::Return, which does not set any return data.
```
LucasSte, line 261:
```suggestion
        // TODO: On Solana, We could elide setting the return data if this function calls no external functions
        // and replace this with a simple Instr::Return, which does not set any return data.
```

LucasSte (approved):

src/codegen/dispatch.rs
LucasSte: Yep

xermicus (approved):

PR #1104, opened 2022-12-16
Generate Anchor IDL from Solidity contracts ⚓

LucasSte:
This PR generates Anchor ⚓ IDL files for Solidity contracts. There are unit tests for the functions that construct the `struct IDL`. All Solana runtime tests have been updated to leverage Anchor IDL instead of Ethereum ABI.

The final json file is not generated when we compile a Solidity contract to avoid confusing users with files that are not yet useful for them.

### DIFF
--- a/src/abi/anchor.rs
+++ b/src/abi/anchor.rs
@@ -1,5 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::sema::ast::{
+    ArrayLength, Contract, Function, Mutability, Namespace, StructDecl, StructType, Tag, Type,
+};
+use anchor_syn::idl::{
+    Idl, IdlAccount, IdlAccountItem, IdlEnumVariant, IdlEvent, IdlEventField, IdlField,
+    IdlInstruction, IdlType, IdlTypeDefinition, IdlTypeDefinitionTy,
+};
+use num_traits::ToPrimitive;
+use semver::Version;
+use std::collections::HashSet;
+
 use convert_case::{Boundary, Case, Casing};
 use sha2::{Digest, Sha256};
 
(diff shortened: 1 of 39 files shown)


src/abi/anchor.rs
```diff
@@ -17,3 +29,343 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+            types: Vec::new(),
+        }
+    }
+
+    /// Retrieve all the custom types to fill the IDL struct
+    fn move_custom_types(&mut self) -> Vec<IdlTypeDefinition> {
+        let mut custom_types: Vec<IdlTypeDefinition> = Vec::new();
+        std::mem::swap(&mut custom_types, &mut self.types);
+        custom_types
+    }
```
seanyoung, line 219: Could this take `self` rather than `&mut self` and then return `self.types`?

src/abi/anchor.rs
```diff
@@ -17,3 +29,343 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+
+    /// Functions with multiple returns must return a struct in Anchor, so we build a return
+    /// struct containing all the returned types.
+    fn build_struct_for_return(&mut self, func: &Function) -> IdlType {
+        let mut fields: Vec<IdlField> = Vec::with_capacity(func.returns.len());
+        for (item_no, item) in func.returns.iter().enumerate() {
+            fields.push(IdlField {
+                name: format!("return_{}", item_no),
```
seanyoung, line 227: functions can have named return values, might as well use those

src/abi/anchor.rs
```diff
@@ -17,3 +29,343 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+
+        let name = format!("{}_returns", func.name);
+        self.types.push(IdlTypeDefinition {
+            name: name.clone(),
+            docs: Some(vec![format!(
+                "Data structure to hold the multiple returns of function {}",
+                func.name
+            )]),
+            ty: IdlTypeDefinitionTy::Struct { fields },
+        });
```
seanyoung, line 241: Are we checking for multiple definitions of the same name?

src/sema/ast.rs
```diff
@@ -102,6 +102,11 @@ pub trait RetrieveType {
     fn ty(&self) -> Type;
 }
 
+/// This trait helps retrieving common members from Function, StructDecl and EnumDecl
+pub trait DeclarationInfo {
```
seanyoung, line 106: This trait only provides `fn tags()`. 
 - I think should be called what they _are_, not what want them to be in the future (bc you can't predict how the code will change in future). How about calling it `Tags` and in future it can be renamed *when* it acquires more methods.
 - I don't know what this trait is supposed to add. You could implement `fn tags()` for all the types, and not have a trait. You're introducing more lines of code but what is the benefit?

seanyoung (changes requested):

src/abi/anchor.rs
LucasSte: We cannot overload structs in Solidity, can we?

src/abi/anchor.rs
LucasSte: No. Rustc does not allow that.

src/abi/anchor.rs
LucasSte: No. It is a good thing to do.

src/sema/ast.rs
LucasSte: That's a remainder of an older version of my code. I should have removed it.

src/abi/anchor.rs
xermicus: This confuses me as well, please consume `self` here. But would you use this function instead of simply accessing `type_manager.types`?

seanyoung: I just tried it and it works fine.

seanyoung: The function probably needs a better name, or just removed completely.

src/abi/anchor.rs
```diff
@@ -17,3 +29,343 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+            Some(type_manager.convert(&func.returns[0].ty))
+        } else {
+            Some(type_manager.build_struct_for_return(func))
+        };
+
+        instructions.push(IdlInstruction {
+            name: if func.is_constructor() {
+                "new".to_string()
+            } else {
+                func.name.clone()
```
xermicus, line 185: Might use mangled name for overloaded functions? I see later down in the tests you use `.instructions.iter().find(|item| item.name == name)` to match against a function

src/abi/anchor.rs
LucasSte: Fixed

src/abi/anchor.rs
LucasSte: Fixed

src/abi/anchor.rs
LucasSte: Fixed

src/abi/anchor.rs
```diff
@@ -17,3 +28,375 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+    if contract.emits_events.is_empty() {
+        None
+    } else {
+        let mut events: Vec<IdlEvent> = Vec::with_capacity(contract.emits_events.len());
+        for event_no in &contract.emits_events {
+            let def = &ns.events[*event_no];
+            let mut fields: Vec<IdlEventField> = Vec::with_capacity(def.fields.len());
+            for (item_no, item) in def.fields.iter().enumerate() {
+                let name = if item.id.is_none() {
+                    format!("field_{}", item_no)
```
seanyoung, line 74: no check for dups

src/abi/anchor.rs
```diff
@@ -17,3 +28,375 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+            args: vec![],
+            returns: None,
+        })
+    }
+
+    for func_no in contract.all_functions.keys() {
+        if !ns.functions[*func_no].is_public() {
+            continue;
+        }
+
```
seanyoung, line 127: Receive/Fallback/Modifiers should not have an entry in the instructions

src/abi/anchor.rs
```diff
@@ -17,3 +28,375 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+                    pda: None,
+                    relations: vec![],
+                })]
+            }
+        };
+
+        let mut args: Vec<IdlField> = Vec::with_capacity(func.params.len());
+        for (item_no, item) in func.params.iter().enumerate() {
+            let name = if item.id.is_none() || item.id.as_ref().unwrap().name.is_empty() {
+                format!("arg_{}", item_no)
```
seanyoung, line 162: This is a parameter, right? Also no check for dups

src/abi/anchor.rs
```diff
@@ -17,3 +28,375 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+
+    /// Functions with multiple returns must return a struct in Anchor, so we build a return
+    /// struct containing all the returned types.
+    fn build_struct_for_return(&mut self, func: &Function, effective_name: &String) -> IdlType {
+        let mut fields: Vec<IdlField> = Vec::with_capacity(func.returns.len());
+        for (item_no, item) in func.returns.iter().enumerate() {
+            let name = if let Some(id) = &item.id {
+                id.name.clone()
+            } else {
+                format!("return_{}", item_no)
```
seanyoung, line 228: no dup check

src/abi/anchor.rs
```diff
@@ -17,3 +28,375 @@ pub fn discriminator(namespace: &'static str, name: &str) -> Vec<u8> {
...
+                docs: None,
+                ty: self.convert(&item.ty),
+            });
+        }
+
+        self.types.push(IdlTypeDefinition {
+            name: def.name.clone(),
+            docs,
+            ty: IdlTypeDefinitionTy::Struct { fields },
+        });
```
seanyoung, line 273: I just realized there is an issue.

```solidity
contract D {
	struct Foo { int64 f1; }
}
contract C {
	enum Foo { b1, b2, b3 }
}
```
Now we have two types with the same name `Foo`.

tests/solana.rs
```diff
@@ -1661,11 +1681,18 @@ impl VirtualMachine {
...
-        let func = &self.stack[0].abi.as_ref().unwrap().functions[name][0];
-        if func.outputs.is_empty() {
+        if let Some(ret) = &instruction.returns {
+            let mut offset: usize = 0;
+            Some(decode_at_offset(
+                return_data,
+                &mut offset,
+                ret,
+                &self.stack[0].idl.as_ref().unwrap().types,
+            ))
```
seanyoung, line 1691: we should check there is no trailing data, iow `assert_eq!(return_data.len(), offset);`

seanyoung (changes requested):

src/abi/anchor.rs
LucasSte: There is a different IDL file for each contract.

LucasSte: In addition, custom types only appear in the IDL if they part of a function argument or return.

seanyoung:
```solidity
contract D {
	struct Foo { int64 f1; }
}
contract C {
	enum Foo { b1, b2, b3 }
        function f(D.Foo x) public returns (int64) { return x.f1; }
}

File: src/abi/anchor.rs (hunk 8 above)
Comment (LucasSte) [reply src/abi/anchor.rs:162]: Function arguments must be named. I should add a comment about that.

Comment (seanyoung) [reply src/abi/anchor.rs:162]: Function argument don't have to be named

File: src/abi/anchor.rs (hunk 7 above)
Comment (LucasSte) [reply src/abi/anchor.rs:127]:
```
target solana does not support receive() functions, see https://solang.readthedocs.io/en/latest/language/functions.html#fallback-and-receive-function
```

Comment (LucasSte) [reply src/abi/anchor.rs:127]: I'll deal with fallback and modifier only.

**Review by xermicus — CHANGES_REQUESTED** [review]
In the docs there is already a section about generating solidity interfaces from IDL, so this might go into docs as well

**Comment by LucasSte** [comment]
> In the docs there is already a section about generating solidity interfaces from IDL, so this might go into docs as well

We are not generating any IDL files for the user. Adding an IDL section to the docs will confuse developers, because they still need the old Ethereum ABI we generate. As soon as we integrate with Anchor, we'll update the documentation.

**Review by seanyoung — APPROVED** [review]
Looks great!

File: src/abi/tests.rs [hunk 12]
```diff
@@ -0,0 +1,1478 @@
...
+    assert_eq!(
+        idl.instructions[2].returns,
+        Some(IdlType::Defined("getNum_int32_int32_returns".to_string()))
+    );
+
+    assert_eq!(idl.types.len(), 2);
+
+    assert_eq!(
+        idl.types[0],
+        IdlTypeDefinition {
```
xermicus, line 852: Nit: That way these tests are hard to audit. Introducing `let expected_... = ` bindings would help

xermicus (approved):
LGTM

PR #1105, opened 2022-12-19
Make instruction data compatible with anchor

seanyoung:
*Solana* breaking change.For constructors, a function selector is now required.
    
Remove the following from the instruction data:
  - Sender
  - Destination
  - Value
  - Seeds
  - Contract hash


seanyoung:
@xermicus the dispatch in codegen now looks similar to what Substrate needs, once the scale encoder moves to codegen

tests/solana.rs
```diff
@@ -1594,17 +1594,19 @@ impl VirtualMachine {
         res
     }
 
-    fn constructor(&mut self, name: &str, args: &[BorshToken]) {
-        self.constructor_expected(0, name, args)
+    /// FIXME: remove name argument
+    fn constructor(&mut self, _name: &str, args: &[BorshToken]) {
```
seanyoung, line 1598: The name argument should be removed but I am worried about 1000s of conflicts with @LucasSte PR https://github.com/hyperledger/solang/pull/1104

We can resolve this later.

LucasSte: What happened with the Sean Young that told me he can solve all conflicts in 4 minutes? 🤣🤣🤣

In addition, I have not changed the constructor function.

src/codegen/dispatch.rs
```diff
@@ -13,21 +13,18 @@ use crate::{
...
     all_cfg: &[ControlFlowGraph],
     ns: &mut Namespace,
+    opt: &Options,
 ) -> ControlFlowGraph {
     let mut vartab = Vartable::new(ns.next_id);
-    let mut cfg = ControlFlowGraph::new(
-        format!("dispatch_{}", ns.contracts[contract_no].name),
-        ASTFunction::None,
-    );
+    let mut cfg = ControlFlowGraph::new("solang_dispatch".into(), ASTFunction::None);
```
LucasSte, line 27: Why are you removing the name of the contract from the dispatch function?

LucasSte, line 27: Why are you removing the contract's name from the CFG name?

src/codegen/dispatch.rs
```diff
@@ -106,22 +103,53 @@ pub(super) fn function_dispatch(
...
+                argslen.clone(),
+                &mut cases,
+                contract_no,
+                ns,
+                &mut vartab,
+                &mut cfg,
+            );
+        }
+
+        if func_cfg.ty == pt::FunctionTy::Constructor {
```
LucasSte, line 139: This should be an `else if`

src/codegen/dispatch.rs
```diff
@@ -195,18 +223,53 @@ pub(super) fn function_dispatch(
...
     cfg: &mut ControlFlowGraph,
 ) {
-    let bb = cfg.new_basic_block(format!("function_cfg_{}", cfg_no));
-    cfg.set_basic_block(bb);
+    let entry = cfg.new_basic_block(format!("function_cfg_{}", cfg_no));
+    cfg.set_basic_block(entry);
+
+    // check for magic
+    let magic_ok = cfg.new_basic_block("magic_ok".into());
+    let magic_bad = cfg.new_basic_block("magic_bad".into());
```
LucasSte, line 243: What kind of "magic" is this?

tests/solana.rs
seanyoung: Well, you've changed a lot of lines near the constructor function callsites, so you'll get conflicts.

seanyoung: I am happy to fix this now *if* my PR goes in first

src/codegen/dispatch.rs
LucasSte: Formally, the question should be "What does this magic number do?"

src/emit/solana/mod.rs
```diff
@@ -327,169 +288,6 @@ impl SolanaTarget {
...
-            ));
-
-            // There is only one possible constructor
-            let ret = if let Some(constructor_function) = contract.constructor {
-                binary
-                    .builder
-                    .build_call(
-                        constructor_function,
-                        &[input.into(), input_len.into(), sol_params.into()],
-                        "constructor_dispatch_call",
```
LucasSte, line 462: Constructors are now called as if they were a normal function now, right?

src/emit/solana/target.rs
```diff
@@ -1679,17 +1642,22 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
 
             let external_call = binary.module.get_function("external_call").unwrap();
 
-            let arg2 = binary.builder.build_pointer_cast(
-                parameters,
-                external_call.get_type().get_param_types()[2].into_pointer_type(),
+            let arg0 = binary.builder.build_pointer_cast(
```
LucasSte, line 1645: What does `arg0` represent?

stdlib/solana.c
```diff
@@ -50,14 +43,9 @@ entrypoint(const uint8_t *input)
...
 
-    if (params.ka_cur != 0)
-    {
-        return ERROR_INVALID_INSTRUCTION_DATA;
-    }
-
     __init_heap();
 
-    return solang_dispatch(&params);
+    return solang_dispatch(params.input, params.input_len, &params);
```
LucasSte, line 48: You are already receiving a pointer to `params`. Why are you passing `input` and `input_len` as separate arguments?

LucasSte (changes requested):

tests/solana.rs
LucasSte: >  _if_ my PR goes in first

This is an exacting requirement.

src/codegen/dispatch.rs
seanyoung: emit no longer emits `solang_dispatch`, it is generated in codegen now

src/codegen/dispatch.rs
seanyoung: I've updated the comment

src/emit/solana/mod.rs
seanyoung: Yes

src/emit/solana/target.rs
seanyoung: I've made the code clearer

src/codegen/dispatch.rs
```diff
@@ -376,25 +412,28 @@ pub(super) fn constructor_dispatch(
...
+        vartab,
         Instr::ReturnCode {
             code: ReturnCode::Success,
         },
     );
 
-    vartab.finalize(ns, &mut cfg);
-
-    cfg
+    cases.push((
```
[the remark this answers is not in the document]
seanyoung: Good idea, thanks

stdlib/solana.c
```diff
@@ -5,7 +5,7 @@
 #include "stdlib.h"
 #include "solana_sdk.h"
 
-extern uint64_t solang_dispatch(const SolParameters *param);
+extern uint64_t solang_dispatch(const uint8_t *input, uint64_t input_len, SolParameters *param);
```
LucasSte, line 8: Did you run `clang-tidy` in this file?

stdlib/solana.c
```diff
@@ -50,14 +43,12 @@ entrypoint(const uint8_t *input)
...
-    {
-        return ERROR_INVALID_INSTRUCTION_DATA;
-    }
-
     __init_heap();
 
-    return solang_dispatch(&params);
+    // solang_dispatch() is generated in codegen. In codegen, we do not have access
+    // the params. We could use Builtin::Calldata for the input but this copies,
+    // so we simply pass in the input
```
LucasSte, line 50:
```suggestion
    // solang_dispatch() is generated in codegen, where we do not have access
    // the params. We could use Builtin::Calldata for the input but this copies,
    // so we simply pass in the input
```

LucasSte: Isn't it possible to use `Expression::MemberAccess`?

stdlib/solana_sdk.h
```diff
@@ -212,19 +212,11 @@ typedef struct
...
   uint64_t input_len;    /** Length in bytes of the instruction data */
   SolPubkey *program_id; /** program_id of the currently executing program */
   const SolAccountInfo *ka_clock;
-  uint32_t contract;
-  const SolPubkey *sender;
-  SolSignerSeed seeds[10];
-  int seeds_len;
   const SolAccountInfo *ka_instructions;
-  uint64_t value;
 } SolParameters;
```
LucasSte, line 220: Did you run `clang-tidy` in this file?

LucasSte (changes requested):

stdlib/solana.c
LucasSte: You can do `Expression::StructMember(Loc::Codegen, Type::Struct(StructType::SolParamenters), Type::Ref(Type::BufferPointer.into), 3)`

stdlib/solana.c
seanyoung: clang-tidy has nothing useful to say

stdlib/solana_sdk.h
seanyoung: clang-tidy has nothing useful to say

stdlib/solana.c
seanyoung: Very nice, I had missed that we have `StructType::SolParameters` . Fixed now.

xermicus (approved):
LGTM

src/codegen/dispatch.rs
```diff
@@ -8,59 +8,74 @@ use crate::codegen::{
...
-            id: None,
-            ty: Type::Uint(64),
-            ty_loc: None,
-            indexed: false,
-            readonly: false,
-            recursive: false,
+            expr: Expression::Load(
+                Loc::Codegen,
+                Type::Uint(64),
+                Expression::StructMember(Loc::Codegen, Type::Uint(64), sol_params.into(), 3).into(),
```
LucasSte, line 68: Shouldn't the type here be a `Type::Ref(Type::Uint(64)`? If it is not a reference, there is no need to load it.

src/codegen/dispatch.rs
```diff
@@ -8,59 +8,74 @@ use crate::codegen::{
...
-            ty_loc: None,
-            indexed: false,
-            readonly: false,
-            recursive: false,
+            expr: Expression::Load(
+                Loc::Codegen,
+                Type::BufferPointer,
+                Expression::StructMember(
+                    Loc::Codegen,
+                    Type::BufferPointer,
```
LucasSte, line 48: Shouldn't the type here be a `Type::Ref(Type::BufferPointer)`? If it is not a reference, there is no need to load it.

seanyoung: Good point, fixed. 

We really need CFG validation. :disappointed:

LucasSte (approved):

PR #1107, opened 2022-12-20
Give nice diagnostics when old style call arguments are used

seanyoung:
C c = (new C).value(1).gas(2)(1, 2, 3);

Fixes https://github.com/xermicus/fuzzy-sol/issues/141


src/sema/expression.rs
```diff
@@ -7861,16 +7874,22 @@ pub fn call_expr(
...
-            ty,
-            args,
-            context,
-            ns,
-            symtable,
-            diagnostics,
-            resolve_to,
-        )?,
+        _ => {
+            if old_style_constructor_arguments(ty, diagnostics) {
```
xermicus, line 7878: It would be more idiomatic to return an `Err(())` directly from the `old_style_constructor_arguments()` function. Which would allow for doing `old_style_constructor_arguments(ty, diagnostics)?;` instead.

src/sema/expression.rs
```diff
@@ -7885,6 +7904,34 @@ pub fn call_expr(
...
+            if let pt::Expression::MemberAccess(_, ty, call_arg) = ty.as_ref() {
+                if old_style_constructor_arguments(ty, diagnostics) {
+                    // location should be the identifier and the arguments
+                    let mut loc = call_arg.loc;
+                    if let pt::Loc::File(_, _, end) = &mut loc {
+                        *end = func_loc.end();
+                    }
+                    diagnostics.push(Diagnostic::error(
+                        loc,
+                        format!("old style call argument using '.{}(...)' is not supported, use '{{{}: ...}}' instead", call_arg.name, call_arg.name)
```
xermicus, line 7920: Nit: "Old style call arguments" sounds a bit.. lax? What about `deprecated` or omitting it all together?

```suggestion
                        format!("deprecated call argument syntax '.{}(...)' is not supported, use '{{{}: ...}}' instead", call_arg.name, call_arg.name)
```

xermicus (changes requested):
Thanks a lot! Just some nits

src/sema/expression.rs
```diff
@@ -7885,6 +7904,34 @@ pub fn call_expr(
     Ok(expr)
 }
 
+/// is it an (new C).value(1).gas(2)(1, 2, 3) style constructor (not supported)
```
LucasSte, line 7907:
```suggestion
/// Is it an (new C).value(1).gas(2)(1, 2, 3) style constructor (not supported)?
```

LucasSte (approved):

src/sema/expression.rs
seanyoung: Yes, you're right, that would be more idiomatic.

xermicus (approved):

PR #1109, opened 2022-12-21
super should not consider parents without a body

seanyoung:
If the parent is an interface, don't use a function from that for super, as it is uncallable and will generate bogus code.

Fixes: https://github.com/xermicus/fuzzy-sol/issues/128


xermicus (approved):

tests/substrate_tests/inheritance.rs
```diff
@@ -645,6 +645,35 @@ fn test_super() {
...
+            function foo(uint64 x) public virtual returns (uint64) {
+                return 202 + x;
+            }
+        }"##,
+    );
+
+    runtime.constructor(0, Vec::new());
+    runtime.function("bar", Vec::new());
+
+    assert_eq!(runtime.vm.output, 212u64.encode());
```
LucasSte, line 676: Why are you not using `BorshToken` here?

LucasSte: That means checking the returns of `runtime.function`;

seanyoung: These are the Substrate tests

LucasSte: I totally missed that.

LucasSte (approved):

PR #1110, opened 2022-12-21
Fix assignment chains

seanyoung:
Fix assigning to both a storage variable and non-storage in assignment chain.

Fixes https://github.com/xermicus/fuzzy-sol/issues/142


src/sema/expression.rs
```diff
@@ -4260,11 +4260,13 @@ fn assign_single(
...
             }
 
+            let ty = ty.deref_any();
+
             Ok(Expression::Assign(
                 *loc,
                 ty.clone(),
                 Box::new(var.clone()),
-                Box::new(val.cast(&right.loc(), ty.deref_any(), true, ns, diagnostics)?),
+                Box::new(val.cast(&right.loc(), ty, true, ns, diagnostics)?),
```
xermicus, line 4269: Out of curiosity, why is this changed?

xermicus (approved):
LGTM

src/sema/expression.rs
seanyoung: The problem is that the assignment expression `a = c` should have the same type as the value being assigned to `a`. I first changed it to:

```
Ok(Expression::Assign(
                *loc,
                ty.deref_any().clone(),
                Box::new(var.clone()),
                Box::new(val.cast(&right.loc(), ty.deref_any(), true, ns, diagnostics)?),
                Box::new(val.cast(&right.loc(), ty, true, ns, diagnostics)?),
```
But this means `deref_any()` is being called twice, so avoided that with a let binding.

seanyoung: In other words, this changes the second field of Expression::Assign to the right value.

xermicus: Ah yes I get it now :facepalm: thanks

LucasSte (approved):

PR #1111, opened 2022-12-22
Increase timeout on Solana integration tests

seanyoung:
The Solana tests sometimes fail with timeouts, maybe this will work.

### DIFF
--- a/integration/anchor/tests/call_anchor.spec.ts
+++ b/integration/anchor/tests/call_anchor.spec.ts
@@ -19,13 +19,12 @@ async function newAccountWithLamports(connection: Connection): Promise<Keypair> 
 }
 
 describe('Call Anchor program from Solidity via IDL', () => {
-
     it('call_anchor', async function () {
         // This program instantiates an anchor program, calls various functions on it and checks the return values
 
         const connection = new Connection("http://localhost:8899", {
             commitment: "confirmed",
-            confirmTransactionInitialTimeout: 100000,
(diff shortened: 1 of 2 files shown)


seanyoung:
I've run the CI tests 11 times, no failures.

integration/anchor/tests/call_anchor.spec.ts
```diff
@@ -19,13 +19,12 @@ async function newAccountWithLamports(connection: Connection): Promise<Keypair>
...
 
 describe('Call Anchor program from Solidity via IDL', () => {
-
     it('call_anchor', async function () {
         // This program instantiates an anchor program, calls various functions on it and checks the return values
 
         const connection = new Connection("http://localhost:8899", {
             commitment: "confirmed",
-            confirmTransactionInitialTimeout: 100000,
+            confirmTransactionInitialTimeout: 1000000,
```
xermicus, line 27: In the `newConnection` helper in `solana/setup.ts`, this value is 10x. Is this intentional? Also can't we use the same helper here (making this timeout configurable if needed)?

LucasSte (approved):

integration/anchor/tests/call_anchor.spec.ts
seanyoung: Good catch, they are different values, I've replaced them both with `1e6`.

There is no shared typescript code with between the (bare) solana tests and the anchor tests. :cry:

xermicus (approved):

PR #1112, opened 2022-12-23
Fix storage array deref bug

xermicus:
Changes the function `Type::is_storage_bytes()` to account for fixed size bytes


src/sema/types.rs
```diff
@@ -1089,6 +1089,7 @@ impl Type {
                 Box::new(Type::Array(ty.clone(), dim[..dim.len() - 1].to_vec())),
             ),
             Type::Array(ty, dim) if dim.len() == 1 => Type::StorageRef(false, ty.clone()),
+            Type::Bytes(n) => Type::Bytes(*n),
```
LucasSte, line 1092: Bytes is not an array type.

xermicus: The code path leading to the crash is due to an array subscript, and this is valid in `solc` as well, so why is this not an array?

LucasSte: In emit, `Type::Bytes` is an integer. The function you modified is also used in `emit` at the Solana target. If this is the solution for the bug, the emit cannot call this function.

xermicus: > The function you modified is also used in `emit` at the Solana target. If this is the solution for the bug, the emit cannot call this function.

I just noticed this as well, gotta find another approach

src/sema/types.rs
```diff
@@ -1539,7 +1539,7 @@ impl Type {
     /// Is this a storage bytes string
     pub fn is_storage_bytes(&self) -> bool {
         if let Type::StorageRef(_, ty) = self {
-            if let Type::DynamicBytes = ty.as_ref() {
+            if let Type::DynamicBytes | Type::Bytes(_) = ty.as_ref() {
```
LucasSte, line 1542: I am not sure this is the right solution. I think @seanyoung can analyze this better than me.

seanyoung: I agree, I don't think this is the right solution. Having said that, I don't know what is! Let me get back to you on this @xermicus  (I had thought this before but never got round to finding the right solution)

seanyoung: This problem is the same as https://github.com/hyperledger/solang/pull/1115 with one additional detail: `fn storage_array_elem()` should return `Type::Bytes(1)`  for `Type::StorageRef(Type::Bytes(_))`.

xermicus: This kind of was my first attempt we scrapped

seanyoung: I think that attempt was correct, a fixed bytes is an array type. Sorry only noticed that conversation now.

The other fix that is neeeded is that sema should load the `bytes32` from storage with an `Expression::Load()`.

seanyoung:
Fixed by #1123

PR #1115, opened 2022-12-23
Arrays indexes must be dereferenced when needed

LucasSte:
When we index an array, we have a reference to an element of that array. If the result of an indexation is used to index another array, we must dereference it first, because indexes must be integers.

This PR fixes https://github.com/xermicus/fuzzy-sol/issues/132

### DIFF
--- a/src/sema/expression.rs
+++ b/src/sema/expression.rs
@@ -5294,10 +5294,8 @@ fn array_subscript(
         ));
     }
 
-    if index_ty.is_contract_storage() {
-        // make sure we load the index value from storage
-        index = index.cast(&index.loc(), index_ty.deref_any(), true, ns, diagnostics)?;
-    }
+    // make sure we load the index value if needed
+    index = index.cast(&index.loc(), index_ty.deref_any(), true, ns, diagnostics)?;
 
     match array_ty.deref_any() {
         Type::Bytes(_) | Type::Array(..) | Type::DynamicBytes => {
--- a/tests/solana_tests/arrays.rs
+++ b/tests/solana_tests/arrays.rs
@@ -1882,3 +1882,88 @@ fn dynamic_array_push_pop_loop() {
     runtime.constructor(&[]);
     runtime.function("test", &[]);
 }
+
+#[test]
+fn double_index() {
+    let src = r#"
+contract RH {
+
+  function calc(uint256[] memory separators, int256[] memory params) public pure returns (int256[4] memory) {
+    int256 stopLimit = params[separators[4]];
+    int256 contractedValueRatio = params[separators[6]];
+
+    return [stopLimit, contractedValueRatio, 3, 4];
+  }
+
+}
+    "#;
+
+    let mut vm = build_solidity(src);
+    vm.constructor(&[]);
+
+    let separators = BorshToken::Array(vec![
+        BorshToken::Int {
+            width: 256,
+            value: BigInt::from(25u8),
+        },
+        BorshToken::Int {
+            width: 256,
+            value: BigInt::from(25u8),
+        },
+        BorshToken::Int {
+            width: 256,
+            value: BigInt::from(25u8),
+        },
+        BorshToken::Int {
+            width: 256,
+            value: BigInt::from(25u8),
+        },
+        BorshToken::Int {
+            width: 256,
+            value: BigInt::from(1u8),
+        },
+        BorshToken::Int {
+            width: 256,
+            value: BigInt::from(25u8),
+        },
+        BorshToken::Int {
+            width: 256,
+            value: BigInt::from(0u8),
+        },
+    ]);
+
+    let params = BorshToken::Array(vec![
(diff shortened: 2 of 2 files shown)


src/codegen/mod.rs
```diff
@@ -645,6 +645,14 @@ impl RetrieveType for Expression {
 }
 
 impl Expression {
+    /// This function loads a reference if necessary
+    pub(crate) fn load(self) -> Expression {
```
xermicus, line 649: While I endorse this change, this could probably be used in a lot more places? Given that is true, for the sake of this bugfix, I'd not include it here, but rather do a follow up PR using it everywhere possible.

xermicus (approved):

src/codegen/mod.rs
seanyoung: @xermicus I think you are very right! All unary operators can be simple function on `cfg::Expression` with no arguments. We could have a `fn neg_loc(&self, loc: pt::Loc) -> Expression` for a version with location. Binary operators like add be like `fn add(&self, other: Expression) -> Expression` or `fn add_checked(&self, other: Expression)`. These functions can `assert!()` that the arguments are of the same `Type`. We need more functions for variables and literals etc. I think it's a great idea that would make the CFG construction builder interface much, much better.

@LucasSte having said all that, I do have some review comments on this function. I would expect `fn load()` to:
- Always load. How about calling this function `fn load_if_needed()` (can't think of a better name right now)
- Load from storage or memory

seanyoung (changes requested):
See comment above

src/codegen/mod.rs
LucasSte: > While I endorse this change, this could probably be used in a lot more places? Given that is true, for the sake of this bugfix, I'd not include it here, but rather do a follow up PR using it everywhere possible.

I can double check the right places to use this function on this PR already.

LucasSte: > @LucasSte having said all that, I do have some review comments on this function. I would expect `fn load()` to:
> 
> * Always load. How about calling this function `fn load_if_needed()` (can't think of a better name right now)
> * Load from storage or memory

I don't think this function should load both from storage and memory, because they are different operations. Load from storage requires an instruction `Instr::LoadStorage`, while the load from memory does not need it. I can clarify this in the function name.

xermicus: Ok fine, but Sean also made some good points that should be addressed in a later PR anyways

src/codegen/expression.rs
```diff
@@ -2425,7 +2425,13 @@ fn array_subscript(
 
     let mut array = expression(array, cfg, contract_no, func, ns, vartab, opt);
     let index_ty = index.ty();
+    // Array indexes must be integers, so if the index is a reference to a type, we must dereference it.
     let index = expression(index, cfg, contract_no, func, ns, vartab, opt);
+    let index = if let Type::Ref(ty) = index.ty() {
+        Expression::Load(index.loc(), *ty, Box::new(index))
+    } else {
+        index
+    };
```
seanyoung, line 2434: The way we usually do this is using `let index = index.cast(loc, index_ty.deref_any(), true, ns, diagnostics)?;`. This also solved the case where `index` is storage variable and need loading from storage.

seanyoung (changes requested):

src/codegen/expression.rs
LucasSte: > This also solved the case where `index` is storage variable and need loading from storage.

This is not true.

LucasSte: When I refactored codegen last year, I removed loads from the cast function.

seanyoung: The `Expression::Load` should be generated in sema.

seanyoung: This problem is very similar to https://github.com/hyperledger/solang/pull/1112

LucasSte: > The `Expression::Load` should be generated in sema.

This is what I've done now. IMO, loads should be done in codegen, but that requires a large refactoring and is something for another PR.

src/sema/expression.rs
```diff
@@ -5294,8 +5294,8 @@ fn array_subscript(
         ));
     }
 
-    if index_ty.is_contract_storage() {
-        // make sure we load the index value from storage
+    if index_ty.is_contract_storage() || index_ty.is_memory_reference() {
+        // make sure we load the index value
         index = index.cast(&index.loc(), index_ty.deref_any(), true, ns, diagnostics)?;
     }
```
seanyoung, line 5300:
```suggestion
    index = index.cast(&index.loc(), index_ty.deref_any(), true, ns, diagnostics)?;
```

seanyoung: This is more idiomatic and simpler

seanyoung (changes requested):

seanyoung (approved):

PR #1123, opened 2023-01-07
Fix storage deref bug

seanyoung:
Alternate implementation to #1112 

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
 
-extern crate core;
-
 pub mod abi;
 pub mod codegen;
 #[cfg(feature = "llvm")]
@@ -14,7 +12,6 @@ pub mod standard_json;
 // In Sema, we use result unit for returning early
 // when code-misparses. The error will be added to the namespace diagnostics, no need to have anything but unit
 // as error.
-#[allow(clippy::result_unit_err)]
 pub mod sema;
 
 use file_resolver::FileResolver;
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -28,7 +28,7 @@ use num_bigint::BigInt;
 use num_traits::{FromPrimitive, One, ToPrimitive, Zero};
 use solang_parser::pt;
 use solang_parser::pt::{CodeLocation, Loc};
-use std::ops::Mul;
+use std::{cmp::Ordering, ops::Mul};
 
 pub fn expression(
     expr: &ast::Expression,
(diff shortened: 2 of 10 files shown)


xermicus (approved):
LGTM

src/lib.rs
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
-extern crate core;
+#![allow(clippy::comparison_chain)]
+#![allow(clippy::result_unit_err)]
```
LucasSte, line 4: I don't think these should be on the top of the file. It misses other possible clippy warnings one may introduce.

seanyoung: By placing these here (note the `#!`), these clippy warnings are disabled for the crate. I don't think these clippy warnings are ever useful. What do you think is good about these clippies?

LucasSte: The comparison chain warning is good because it tells us if we can simplify our code using a `match` statement instead of an if-else chain.
The second encounters `Result<>` returns in which the `Err` is not implemented. These can be replaced by an `Option` construct. 
I don't think they should be disabled for the crate.

seanyoung: > The comparison chain warning is good because it tells us if we can simplify our code using a `match` statement instead of an if-else chain.

The code in the example https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain is awful, and I refuse to make the change the clippy suggests. Each time this clippy warns we have an individual `#[allow(clippy::comparison_chain)]`. This change simply put a single top level `#![allow(clippy::comparison_chain)]` and removes all the individual `#[allow(clippy::comparison_chain)]` littered in the code. I made this change to avoid having to add another `#[allow(clippy::comparison_chain)]` to the code.

> The second encounters `Result<>` returns in which the `Err` is not implemented. These can be replaced by an `Option` construct. I don't think they should be disabled for the crate.

Have a look at https://rust-lang.github.io/rust-clippy/master/index.html#result_unit_err. This clippy complains is the Error is unit `()`, and does not suggest `Option`, and it only complains about public functions. The idea of this clippy is that if an error is returned, it should contain more information than `()`. We pass diagnostics around as parameter, so we do not use the return value except for `?` in sema.

I've removed `#![allow(clippy::result_unit_err)]` and made the functions `pub(super)` so they are only public to sema, and the clippy no longer complains.

xermicus:
Closes #1112

src/lib.rs
LucasSte: > The code in the example https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain is awful, and I don't think change the clippy suggests improves anything, it makes the code worse. Each time this clippy warns we have an individual `#[allow(clippy::comparison_chain)]`. This change simply put a single top level `#![allow(clippy::comparison_chain)]` and removes all the individual `#[allow(clippy::comparison_chain)]` littered in the code. I made this change to avoid having to add another `#[allow(clippy::comparison_chain)]` to the code.

I'm afraid that disabling the warning for the whole crate will scrap useful warnings to improve code. Sometimes I miss some Rustisms :/

> I don't think change the clippy suggests improves anything, it makes the code worse.

It makes your code look like it is written in Rust, not in C.

seanyoung: > I'm afraid that disabling the warning for the whole crate will scrap useful warnings to improve code. Sometimes I miss some Rustisms :/
> 
> > I don't think change the clippy suggests improves anything, it makes the code worse.
> 
> It makes your code look like it is written in Rust, not in C.

I'd like to see some examples to back up your claim please. You could take the existing code with `#[allow(clippy::comparison_chain)]` attribute, re-write it, and show it is more readable in the re-written form.

LucasSte: > In my book `a < b` will always be easier to read than `a.cmp(b) == Ordering::Less`.

This example won't raise a clippy warning because it is not a comparison chain.

LucasSte: > I'd like to see some examples to back up your claim please. You could take the existing code with `#[allow(clippy::comparison_chain)]` attribute, re-write it, and show it is more readable in the re-written form.

I refactored all occurrences of `#[allow(clippy::comparison_chain)]` [here](https://github.com/hyperledger/solang/pull/1124). It looks nicer to me, specially because we can do `Ordering::Greater => if to_len.is_signed_int()`, which decreases indentation.

LucasSte: What do you think?

It's up to you now. My arguments are depleted.

seanyoung: I have relented and removed the `clippy::comparison_chain` changes. I still think `clippy::comparison_chain` is neutral at best and mostly down to taste.

LucasSte (approved):

PR #1127, opened 2023-01-13
Update Rust version from workflows

LucasSte:
The test workflows are failing because clap does not support Rust 1.63 anymore, so I updated it to 1.64.

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -29,7 +29,7 @@ RUN apt-get clean
 RUN apt-get autoclean
 
 # Get Rust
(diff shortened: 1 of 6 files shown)


xermicus (commented):
Thanks!

The release workflow would benefit from higher rust versions, as they might contain performance improvements. However I also don't like the Idea of using different versions on the release workflow vs. test. What about we always use latest stable for tests and release, but have regression tests against older rust versions, down to our MSRP?

seanyoung (approved):

xermicus (approved):

PR #1128, opened 2023-01-13
Refactor SCALE encoder into codegen

xermicus:
This PR makes the `AbiEncoding` trait more granular. This allowed me to move most of the logic into default implementations, leaving only a few `Type`s that needed to be handled differently in SCALE and Borsh. This means the Borsh and SCALE encoders can share most of their code :partying_face: 

The decoding side is untouched by this PR as it can be done independent. I will do it in a follow up PR, which will in turn allow to move the dispatcher out of emit into codegen as well. Which will finally allow to nuke the emit implementation of the SCALE encoder, as well as `substrace.c` from our stdlib.


src/codegen/events/solana.rs
```diff
@@ -46,18 +46,15 @@ impl EventEmitter for SolanaEventEmitter<'_> {
...
-        let mut encoder = create_encoder(self.ns, false);
         let (abi_encoded, abi_encoded_size) =
-            encoder.abi_encode(&self.loc, to_be_encoded, self.ns, vartab, cfg);
+            abi_encode(&self.loc, to_be_encoded, self.ns, vartab, cfg, false);
 
         cfg.add(
             vartab,
             Instr::EmitEvent {
                 event_no: self.event_no,
                 data: vec![abi_encoded, abi_encoded_size],
```
LucasSte, line 56: Now that you've removed the unused fields from this instructions, there should be different fields for data and size.

src/codegen/expression.rs
```diff
@@ -1488,11 +1493,10 @@ fn abi_encode(
         .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt))
         .collect::<Vec<Expression>>();
 
-    let mut encoder = create_encoder(ns, false);
-    encoder.abi_encode(loc, args, ns, vartab, cfg).0
+    abi_encode(loc, args, ns, vartab, cfg, false).0
 }
 
-fn abi_encode_packed(
+fn abi_encode_packed_expression(
```
LucasSte, line 1499: Why did you change this name?

src/codegen/expression.rs
```diff
@@ -1523,11 +1526,10 @@ fn encode_with_selector(
     let mut encoder_args: Vec<Expression> = Vec::with_capacity(args.len() + 1);
     encoder_args.push(selector);
     encoder_args.append(&mut args);
-    let mut encoder = create_encoder(ns, false);
-    encoder.abi_encode(loc, encoder_args, ns, vartab, cfg).0
+    abi_encode(loc, encoder_args, ns, vartab, cfg, false).0
 }
 
-fn abi_encode_with_selector(
+fn abi_encode_expression_with_selector(
```
LucasSte, line 1532: Why did you change this name?

src/codegen/expression.rs
```diff
@@ -1550,10 +1552,10 @@ fn abi_encode_with_selector(
     let args = args_iter
         .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt))
         .collect::<Vec<Expression>>();
-    encode_with_selector(loc, selector, args, ns, vartab, cfg)
+    encode_many_with_selector(loc, selector, args, ns, vartab, cfg)
 }
 
-fn abi_encode_with_signature(
+fn abi_encode_expression_with_signature(
```
LucasSte, line 1558: Why did you change this name?

src/codegen/events/solana.rs
LucasSte: In addition, `data` does not need to be a vector anymore.

tests/codegen_testcases/solidity/borsh_encoding_complex_types.sol
```diff
@@ -77,7 +77,6 @@ contract EncodingTest {
 	    // CHECK: branch block2
 
         // CHECK: block4: # end_for
-	    // CHECK: ty:bytes %abi_encoded.temp.17 = (alloc bytes len (uint32 8 + %array_bytes_size_0.temp.15))
```
LucasSte, line 80: This instruction should not be gone. We are not allocating memory for the ecoding anymore. In addition, it should be `uint32 12 + %array_bytes_size_0.temp.15`, because we need 8 bytes for the `uint64` and more four bytes for the `string[]` length.

LucasSte (changes requested):

src/codegen/encoding/scale_encoding.rs
```diff
@@ -1,66 +1,196 @@
...
+/// Encode `expr` into `buffer` as a compact integer. More information can found in the
+/// [SCALE documentation](https://docs.substrate.io/reference/scale-codec/).
+fn encode_compact(
+    expr: &Expression,
+    buffer: Option<&Expression>,
+    offset: Option<&Expression>,
+    vartab: &mut Vartable,
+    cfg: &mut ControlFlowGraph,
+) -> Expression {
+    let small = cfg.new_basic_block("small".into());
```
seanyoung, line 36: This might be better in a function, to avoid the code size from bloating

seanyoung (changes requested):
Lucas beat me to it with his comments. My only comment is a suggestion

src/codegen/encoding/borsh_encoding.rs
```diff
@@ -19,63 +18,67 @@ use std::ops::{Add, AddAssign, MulAssign};
...
 
 impl AbiEncoding for BorshEncoding {
-    fn abi_encode(
+    fn size_width(
+        &self,
+        _size: &Expression,
+        _vartab: &mut Vartable,
+        _cfg: &mut ControlFlowGraph,
+    ) -> Expression {
+        Expression::NumberLiteral(Loc::Codegen, U32, 4.into())
```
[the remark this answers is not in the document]
xermicus: I don't see how anyone can confuse this with the rust `u32` type. The syntax around rusts type system does not work like that.

I choose to do this to improve readability and working with our AST. This specific custom type is used everywhere in and is super annoying to type out or read `Type::Uint(32)` all over the place. Your suggestions would make it worse, constants in rust are all capital.

src/codegen/encoding/mod.rs
```diff
@@ -10,27 +10,526 @@ use crate::codegen::encoding::scale_encoding::ScaleEncoding;
...
+use crate::sema::ast::{ArrayLength, Namespace, RetrieveType, StructType, Type, U32};
 use crate::Target;
 use num_bigint::BigInt;
 use num_integer::Integer;
 use num_traits::{One, Zero};
 use solang_parser::pt::Loc;
 use std::ops::{AddAssign, MulAssign, Sub};
 
-/// This trait should be implemented by all encoding methods (ethabi, Scale and Borsh), so that
+pub(super) const LOC: Loc = Loc::Codegen;
```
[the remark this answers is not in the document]
xermicus: `LOC_CODEGEN` is barely shorter but even more awkward to read (to me at least) than `Loc::Codegen`.  The only point of this change is to make everything a bit less verbose.

src/codegen/encoding/mod.rs
```diff
@@ -10,27 +10,526 @@ use crate::codegen::encoding::scale_encoding::ScaleEncoding;
...
+        offset = Expression::Add(*loc, U32, false, offset.into(), advance.into());
+    }
+    (buffer, size)
+}
+
+/// This trait should be implemented by all encoding methods (ethabi, SCALE and Borsh), so that
 /// we have the same interface for creating encode and decode functions.
+///
+/// Note: This trait mostly reflects the situation around SCALE and Borsh encoding schemas.
+/// Any provided default implementation works fine for both of them.
```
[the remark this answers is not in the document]
xermicus: Do you have more details what bothers you? I can rephrase it and add more details. What I am trying to say here is that the trait was designed to cope well (e.g. reflect the situation) with both SCALE and Borsh encoding schemas.

src/codegen/encoding/mod.rs
```diff
@@ -10,27 +10,526 @@ use crate::codegen::encoding::scale_encoding::ScaleEncoding;
...
+            }
+            Type::InternalFunction { .. }
+            | Type::Void
+            | Type::BufferPointer
+            | Type::Mapping(..) => unreachable!("This type cannot be encoded"),
+        }
+    }
+
+    /// Write whatever is inside the given `expr` into `buffer` without any modification.
+    fn encode_linear(
```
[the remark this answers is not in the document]
xermicus: Because it encodes the data as is, as I explained in the comment... Open for suggestions if you have a better name

src/codegen/encoding/mod.rs
```diff
@@ -10,27 +10,526 @@ use crate::codegen::encoding::scale_encoding::ScaleEncoding;
...
+                        buffer,
+                        offset.clone(),
+                        ty,
+                        arg_no,
+                        ns,
+                        vartab,
+                        cfg,
+                    );
+                }
+                let loaded = Expression::Load(LOC, *r.clone(), expr.clone().into());
```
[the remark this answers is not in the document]
LucasSte: This is easily mistaken because the name is the same. If someone from outside our codebase reads this, they will be confused.

src/codegen/encoding/mod.rs
```diff
@@ -10,27 +10,526 @@ use crate::codegen::encoding::scale_encoding::ScaleEncoding;
...
+    /// Encode `expr` into `buffer` as bytes, omitting the size.
+    fn encode_bytes_packed(
+        &mut self,
+        expr: &Expression,
+        len: Expression,
+        buffer: &Expression,
+        offset: &Expression,
+        vartab: &mut Vartable,
+        cfg: &mut ControlFlowGraph,
+    ) -> Expression {
```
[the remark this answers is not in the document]
xermicus: Good catch. This function is no longer needed and I forgot to remove

src/codegen/expression.rs
```diff
@@ -1473,7 +1478,7 @@ fn payable_transfer(
     Expression::Poison
 }
 
-fn abi_encode(
+fn abi_encode_many(
```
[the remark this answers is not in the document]
xermicus: Because it encodes many expressions at once and `abi_encode` and there were like 3 `fn encode` in various modules already.

src/codegen/expression.rs
xermicus: I'm not sure on this one anymore.

src/codegen/expression.rs
xermicus: Same

tests/codegen_testcases/solidity/borsh_encoding_complex_types.sol
xermicus: I'll look into this

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: I don't buy this argument. If someone will be already confused with this, how is he/she going to survive working with our `sema::Expression` vs. `codegen::Expression`, just to give an example.

This is more an argument of personal taste than anything. @seanyoung can you please decide on this. If you oppose naming it the way I did as well, I'm just gonna revert it and leave it as `Loc::Codegen`. We can think of making everything less verbose later on as well (even though this small change makes working with the code base easier for me).

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: Ah, when you are writing `LOC`, it is not clear what variant of `Loc` you are referring to.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: `Any provided default implementation` This is the part that is not clear. I don't know what it means.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: I'd say `encode_directly` or just `copy_to_buffer`. The adjective linear does not refer to something that does not perform changes. `f(x) = 2x + 3` is a linear function, but is does change the argument.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: When you are writing LOC, it is not clear what variant of Loc you are referring to.

src/codegen/expression.rs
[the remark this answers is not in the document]
LucasSte: That makes sense.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Makes sense, I didn't see this before. I'm going to change it, thanks!

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Fair enough, I like `encode_directly` or something like that.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: > Here is an idea.
> 
> ```rust
> use solang_parser::pt::Loc::Codegen;
> ```
> 
> Now you can use `Codegen` rather than `Loc::Codegen`, and you're depending on the rust type system for the shorter name.

Oh, very nice, thanks!

src/codegen/encoding/mod.rs
```diff
@@ -212,63 +707,45 @@ fn calculate_array_size<T: AbiEncoding>(
     };
 
     let size_var = if let Some(compile_type_size) = primitive_size {
+        let dimension = &dims.last().unwrap();
```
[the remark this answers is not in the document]
xermicus: Removed

src/codegen/encoding/mod.rs
```diff
@@ -212,63 +707,45 @@ fn calculate_array_size<T: AbiEncoding>(
...
-                Box::new(size),
-                Box::new(local_size),
-            );
+            let local_size =
+                Expression::NumberLiteral(LOC, U32, item.array_length().unwrap().clone());
+            size = Expression::Multiply(LOC, U32, false, size.into(), local_size.clone().into());
+            if !encoder.is_packed() && matches!(item, ArrayLength::Dynamic) {
+                let size_width = encoder.size_width(&local_size, vartab, cfg);
+                size = Expression::Add(LOC, U32, false, size.into(), size_width.into())
+            }
```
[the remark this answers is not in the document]
xermicus: Thanks for pointing this out, I'll have a look again

src/codegen/encoding/mod.rs
```diff
@@ -284,40 +761,21 @@ fn calculate_array_size<T: AbiEncoding>(
...
-                    Box::new(Expression::Variable(Loc::Codegen, Type::Uint(32), size_var)),
-                    Box::new(Expression::NumberLiteral(
-                        Loc::Codegen,
-                        Type::Uint(32),
-                        BigInt::from(4 * dyn_dims),
-                    )),
-                ),
-            },
-        );
+    if !encoder.is_packed() && matches!(&dims.last().unwrap(), ArrayLength::Dynamic) {
```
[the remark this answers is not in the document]
xermicus: I read the code of the function as follows: If we have `direct_assessment`, then
- We have either 0 dynamic dimensions
- Or exactly one, which is the last one
- We don't recursively walk `get_expr_size`, so we won't account for the encoded length size automatically

Thus follows: If we are not in packed encoder *and* the last dimension is dynamic (since direct assessment can be true if we don't have dynamic dimensions at all), we need to account for the size of the encoded length.

Hence, I am only checking for the last one to be dynamic, because only in that case this code path can be reached. Or am I getting this wrong?

This reminds me that it would deserve a comment or two though.

src/codegen/encoding/mod.rs
```diff
@@ -62,8 +574,8 @@ pub(super) trait AbiEncoding {
 pub(super) fn create_encoder(ns: &Namespace, packed: bool) -> Box<dyn AbiEncoding> {
     match &ns.target {
         Target::Solana => Box::new(BorshEncoding::new(packed)),
-        // Solana utilizes Borsh encoding and Substrate, Scale encoding.
-        // All other targets are using the Scale encoding, because we have tests for a
+        // Solana utilizes Borsh encoding and Substrate, SCALE encoding.
+        // All other targets are using the SCALE encoding, because we have tests for a
         // fake Ethereum target that checks the presence of Instr::AbiDecode and
```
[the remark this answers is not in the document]
xermicus: Good point, but first I have to move the SCALE decoder to codegen and then the dispatcher. Then the tests should fail and I can remove them? I was planning to do separate PR for this.

src/codegen/encoding/mod.rs
```diff
@@ -592,35 +997,37 @@ fn finish_array_loop(for_loop: &ForLoop, vartab: &mut Vartable, cfg: &mut Contro
...
         // We should not dereference a struct or fixed array
-        return Expression::StructMember(Loc::Codegen, ty, Box::new(expr), field);
+        return Expression::StructMember(LOC, ty, expr.into(), field);
     }
+    let s = Expression::StructMember(LOC, Type::Ref(ty.clone().into()), expr.into(), field);
+    Expression::Load(LOC, ty, s.into())
+}
+
+/// Get the array length inside a variable.
+fn array_length(arr: &Expression, vartab: &mut Vartable, cfg: &mut ControlFlowGraph) -> Expression {
```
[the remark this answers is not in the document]
xermicus: Agreed, this gets a bit confusing. However, `get_array_length` loads a subarray and returns the `Builtin::Array` expression on that, where as `array_length` directly uses `Builtin::ArrayLength` and stores that inside an `Expression::Variable`. So they are two distinct helper functions serving a different purpose.

I just decided to nuke `get_array_length` instead, because it used only in one place, and can be refactored to smaller code size (the commit removed 14 lines). What do you think?

src/codegen/events/solana.rs
xermicus: Good point!

src/codegen/encoding/scale_encoding.rs
xermicus: Good idea, will look into it

src/codegen/encoding/borsh_encoding.rs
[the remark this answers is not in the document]
LucasSte: Do you think @seanyoung idea can also be useful here? I can import `Uint` directly and write only `Uint(32)` everywhere.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: `get_array_length` works for any of the array dimensions. It is a general implementation. Let's say your array is a `uint8[3][][4][]`. `get_array_length` works for any of the dimensions you want to fetch, this is why the load is necessary. If you are retrieving the last dimension, no load will be added.

As it was only used in one place, there is no problem in removing it. The function you added only considers the last dimension of an array. Its name should reflect what it does. Alternatively, you can just call `get_array_length` for the last dimension. Storing the length on a variable is only necessary if the length is changing throughout the code, which is not the case for the encoding arguments.

You can choose whether to rename `array_length` to something in the lines of `outer_array_length` or to only use `get_array_length`.

src/codegen/encoding/borsh_encoding.rs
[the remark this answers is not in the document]
xermicus: Yes that makes an acceptable compromise for me.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Doesn't generate `get_array_length` an unnecessary `Subscript` expression if I'm only interested the outer dimension?

>  Storing the length on a variable is only necessary if the length is changing throughout the code
I'm aware of that and this is why I created the helper function, because I needed to this in more than one location.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: This condition is executed regardless of `direct_assessment`'s value, so you need to consider all cases in which dynamic dimensions are available.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: > Doesn't generate `get_array_length` an unnecessary `Subscript` expression if I'm only interested the outer dimension? If this gets optimized away anyways later on in emit, I'm all in for only using this method.

No, there is no subscript generated. Let's say you want dimension 2, and the `dims` vector has a length of 3.

The slice passed to `load_subarray` would be `dims[3..3]`, which has a length of zero.
https://github.com/hyperledger/solang/blob/468e4d6a59794e0ac4a74001986f600bec0a6d68/src/codegen/encoding/mod.rs#L385

Later, in function `load_subarray` the loop does not execute, because the slice has a length of zero. This means the function only returns the array itself without any subscript.

https://github.com/hyperledger/solang/blob/468e4d6a59794e0ac4a74001986f600bec0a6d68/src/codegen/encoding/mod.rs#L462

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: > This condition is executed regardless of direct_assessment's value

I am aware of that. However, `calculate_complex_array_size` calls `get_expr_size` recursively. Any size length information should be handled there. Except when we are in direct assessment. Then we need to handle the size length information manually. But in direct assessment, only the last dimension can be dynamic. Hence this check. Can you provide me with a test case that fails in the situation you describe?

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Sorry I overlooked this. Makes sense, thanks

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte:
```solidity
contract PTokenFactory {
    function strConcat(uint16[][3][] memory vec) public pure returns (bytes memory){
        bytes b = abi.encode(vec);
        return b;
    }
}
```

If you emit the CFG for this case, you'll see that the space necessary to save both dynamic dimensions is not accounted for while we calculate the size of the array.

Later, when the actual buffer is allocated, we have this instruction:

`ty:bytes %abi_encoded.temp.7 = (alloc bytes len (%array_bytes_size_0.temp.3 + uint32 4))`. It accounts only one of the dynamic dimensions. We should have a `+ uint32 8` here.

PS: I have run this command: ` ./target/debug/solang compile --target solana ../solang/examples/test.sol --emit cfg`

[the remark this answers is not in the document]
xermicus: Hmm, this crashes the compiler :neutral_face: It would be nice to have a functioning test for this...

However I'm still convinced the accounting for the encoded size must happen inside `calculate_complex_array_size`, where the actually array size calculation is happening. We need the actual array size value to calculate it's width, and this is the (only) place where it gets calculated?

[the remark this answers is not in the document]
xermicus: I'm confused by the `calculate_complex_array_size` function. Why do we recursively write for loops into our CFG, but then only calculate the array length for the dimension `0`?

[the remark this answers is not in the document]
LucasSte: > Hmm, this crashes the compiler 😐 It would be nice to have a functioning test for this...

If you do `--emit `, you can see the generated code and follow the CFG instructions to see what I'm telling you. There are many bugs with multidimensional arrays that needs fixing, however the encoder was designed to handle all cases already.

> However I'm still convinced the accounting for the encoded size must happen inside `calculate_complex_array_size`, where the actually array size calculation is happening.

No, it does not. The `calculate_complex_array_size ` only calculates the size of each element in the array. It does not account for the space needed to save the dimension. This is how I designed the function.

> I'm confused by the calculate_complex_array_size function. Why do we recursively write loops into our CFG, but then only calculate the array length for the dimension 0? 

if your vector is like `my_struct[][3][]`,  `calculate_complex_array_size ` will implement this only:

```
uint32 size = 0;
for (int i=0; i<my_struct.length; i++) {
    for(int j=0; j<3; j++) {
        for(int k=0; k<my_struct[i][j].length; k++) {
            size += size_of(my_struct[i][j][k]);
        }
    }
}
```

I'm gonna fix the bug in the example and get back to you. Meanwhile, you can check the CFG by doing `--emit cfg`, because that still works.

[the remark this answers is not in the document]
xermicus: > If you do --emit , you can see the generated code and follow the CFG instructions to see what I'm telling you. 

I know. I pointed this out because now I'm working around a problem that we can not even properly test. As a side note, in the future we should avoid having to maintain such code. If we can not test some functionality properly, because it depends on other things that are missing, it is a strong hint that it would be best to avoid implementing such functionality, or fix / add whats needed first.

> I'm gonna fix the bug in the example and get back to you. 

This is a great idea. Thanks. It makes no sense to try to work with such such high-dimensional arrays in encoding, when they are broken anyways to a point it is not even possible to write a test. Note that another option (as an immediate fix) here would be to just forbid array dimensions > 2 in sema (and assert that here). This would allow to simplify this code and fix multidimensional arrays everywhere for good later on. Acutally thinking about it, until we haven't fixed such arrays for good and have good test coverage for them, it would make sense to forbid this in sema anyways.

> The calculate_complex_array_size only calculates the size of each element in the array. It does not account for the space needed to save the dimension. This is how I designed the function.

Fair enough, but I don't see any other solution to get this going for SCALE encoding than to account for the encoded length here. Also this would then be consistent to what is being done in `encode_complex_array_size`. What do you think?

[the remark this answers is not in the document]
xermicus: Ok so apparently, this is not something that is trivial to fix. So we should really to stop worrying about arrays with >2 dimensions here until we have everything to 
1. Properly test this
2. Properly announce the feature ready to use.

I'll wait for #1132 to go in nonetheless, but then just forbid more than 2 array dimensions in sema. I'd like to go forward with this PR.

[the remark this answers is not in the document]
LucasSte: > then just forbid more than 2 array dimensions in sema.

Multidimensional fixed arrays work perfectly. We have tests for them in the encoder/decoder. I would not disable that right now, as we won't be able to detect regressions in the tests we already have.

tests/codegen_testcases/solidity/borsh_encoding_complex_types.sol
LucasSte: > I'll look into this

Did you find why this instruction is missing?

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: https://github.com/hyperledger/solang/blob/468e4d6a59794e0ac4a74001986f600bec0a6d68/tests/solana_tests/abi_encode.rs#L581

https://github.com/hyperledger/solang/blob/468e4d6a59794e0ac4a74001986f600bec0a6d68/tests/solana_tests/abi_encode.rs#L581

tests/codegen_testcases/solidity/borsh_encoding_complex_types.sol
xermicus: I'm gonna update all codegen tests it again after #1132 is in

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: #1132 has been merged.

src/codegen/encoding/scale_encoding.rs
xermicus: Started to work on this, but it will bloat this already quite large PR further. I decided to do this in a follow up.

src/emit/substrate/target.rs
```diff
@@ -1520,23 +1516,13 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
-
         call!(
             "seal_deposit_event",
             &[
                 topic_buf.into(),
                 topic_size.into(),
-                data_ptr.into(),
-                data_len.into(),
+                binary.vector_bytes(data).into(),
+                size.into_int_value().into(),
```
seanyoung, line 1525: This could be `binary.vector_len(data).into()`. I am not sure why the length is passed in separately.

src/codegen/cfg.rs
```diff
@@ -156,10 +156,9 @@ pub enum Instr {
     /// Emit event
     EmitEvent {
         event_no: usize,
-        data: Vec<Expression>,
-        data_tys: Vec<Type>,
+        data: Expression,
+        size: Expression,
```
seanyoung, line 160: Why is the size needed?

src/codegen/encoding/scale_encoding.rs
```diff
@@ -108,12 +242,39 @@ impl AbiEncoding for ScaleEncoding {
...
+                let length = Expression::Builtin(
+                    Codegen,
+                    vec![Uint(32)],
+                    Builtin::ArrayLength,
+                    vec![expr.clone()],
+                );
+                if self.is_packed() {
+                    length
+                } else {
+                    increment_by(encode_compact(&length, None, None, vartab, cfg), length)
```
seanyoung, line 273: It is quite expensive to calculate the length of the compact encoding. 

It is cheaper to just use 4 here, and then adjust the length of the vector after encoding.

seanyoung (changes requested):

src/codegen/encoding/mod.rs
```diff
@@ -10,27 +10,520 @@ use crate::codegen::encoding::scale_encoding::ScaleEncoding;
...
+            }
+            Type::InternalFunction { .. }
+            | Type::Void
+            | Type::BufferPointer
+            | Type::Mapping(..) => unreachable!("This type cannot be encoded"),
+        }
+    }
+
+    /// Write whatever is inside the given `expr` into `buffer` without any modification.
+    fn encode_direct(
```
LucasSte, line 155: This should be `encode_directly`.

src/codegen/encoding/mod.rs
```diff
@@ -10,27 +10,520 @@ use crate::codegen::encoding::scale_encoding::ScaleEncoding;
...
+                        (offset.clone(), None)
+                    } else {
+                        let encoded_size = self.encode_size(&value, buffer, offset, vartab, cfg);
+                        let new_offset = increment_by(offset.clone(), encoded_size.clone());
+                        (new_offset, Some(encoded_size))
+                    };
+
+                    if let Expression::Variable(_, _, size_temp) = value {
+                        let size = calculate_array_bytes_size(size_temp, elem_ty, ns);
+                        (size, new_offset, size_length)
```
LucasSte, line 365: As you already have an `Expression::Variable` here, what do you think of refactoring `calculate_array_bytes_size` to receive an `Expression` instead of the variable number?

src/codegen/encoding/mod.rs
```diff
@@ -614,35 +1047,42 @@ fn finish_array_loop(for_loop: &ForLoop, vartab: &mut Vartable, cfg: &mut Contro
...
         // We should not dereference a struct or fixed array
-        return Expression::StructMember(Loc::Codegen, ty, Box::new(expr), field);
+        return Expression::StructMember(Codegen, ty, expr.into(), field);
     }
+    let s = Expression::StructMember(Codegen, Type::Ref(ty.clone().into()), expr.into(), field);
+    Expression::Load(Codegen, ty, s.into())
+}
+
+/// Get the array length inside a variable.
+fn array_length(arr: &Expression, vartab: &mut Vartable, cfg: &mut ControlFlowGraph) -> Expression {
```
LucasSte, line 1057: This function should be called `array_outer_length`, as it cannot be used for any dimension. In addition, the doc comment should say so.

src/codegen/encoding/scale_encoding.rs
```diff
@@ -1,66 +1,200 @@
...
+        let offset = Expression::Add(
+            Codegen,
+            Uint(32),
+            false,
+            offset.clone().into(),
+            size.clone().into(),
         );
+        let selector = expr.external_function_selector();
+        let selector_size = self.encode_direct(&selector, buffer, &offset, vartab, cfg, 4.into());
+        Expression::Add(Codegen, Uint(32), false, size.into(), selector_size.into())
```
LucasSte, line 186: The selector and the address have sizes that are known during compile time. There is no need to create an addition here.
In addition, this function does exactly the same thing you do on bosh encoding. I believe you can have a single implementation for both cases.

LucasSte (changes requested):
Aren't codegen tests necessary for scale encoding too?

src/emit/substrate/target.rs
xermicus: In the solana emit target, the data and len were passed as different elements in a vector, I think thats why I did it just like that.

src/codegen/encoding/scale_encoding.rs
xermicus: I like the idea, but I need to think about whether this will work for all cases

src/codegen/cfg.rs
xermicus: If solana and substrate emit targets don't need it we can remove it

src/codegen/encoding/mod.rs
xermicus: Ok. Why is this important?

src/codegen/encoding/mod.rs
xermicus: Thanks for checking, I'm gonna leave it as it is then.

src/codegen/encoding/scale_encoding.rs
xermicus: > The selector and the address have sizes that are known during compile time. There is no need to create an addition here.

Well I don't think adding vs. loading a value make a huge difference and this code most likely never ends up inside hot code paths. But I can change it

> In addition, this function does exactly the same thing you do on bosh encoding

For solana the selector comes first

xermicus:
> Aren't codegen tests necessary for scale encoding too?

Since the encoders now share almost all code, I don't see what those test would add? I can implement one for the external functions and the compact encode

src/codegen/cfg.rs
xermicus: Removing this worked just fine :tada:

src/codegen/encoding/scale_encoding.rs
xermicus: Just checked. There are many instances where we call into the `abi_encode` codegen function, but the returned length is ignored. Which means random tests start to blow up if I introduce your proposed optimization.

This in turn means even more changes for this PR that are not directly related to the problem I'm trying to solve, so I'm quite averse implementing this optimization now. However, I think this is a good discovery and we should change every single instance to use the returned length instead, so we have a single source of truth. Then we should also be able to introduce this as an optimization.

src/codegen/cfg.rs
seanyoung: nice!

src/codegen/encoding/scale_encoding.rs
seanyoung: Of course, it was just a suggestion.

seanyoung: It might be worth leaving a comment in the code, so a potential optimization is not forgotten.

src/codegen/encoding/mod.rs
LucasSte: > Ok. Why is this important?

Your functions encodes the type directly in the buffer, so you need an adverb.

seanyoung: I don't think this matter much. It does not improve the function name.

LucasSte: > I don't think this matters much. It does not improve the function name.

It is proper grammar :)

xermicus: [I also don't think it matters much ](https://en.wikipedia.org/wiki/Law_of_triviality) :wink: Changed it, but I won't keep bikeshdding the PR after this.

seanyoung: function naming is not prose, so strict grammar rules don't apply.

src/codegen/encoding/scale_encoding.rs
xermicus: Good idea

seanyoung (approved):

src/codegen/encoding/mod.rs
```diff
@@ -42,17 +535,30 @@ pub(super) trait AbiEncoding {
...
 
-    /// Cache items loaded from storage to reuse them later, so we avoid the expensive operation
-    /// of loading from storage twice. We need the data in two different passes: first to
-    /// calculate its size and then to copy it to the buffer.
+    /// Encoding happens in two steps. First, we look at each argument to calculate its size. If an
+    /// argument is a storage variable, we load it and save it to a local variable.
+    ///
+    /// During a second pass, we copy each argument to a buffer. To copy storage variables properly into
+    /// the buffer, we must load them from storage and save them in a local variable. As we have
+    /// already done this before, we can cache the Expression::Variable, containing the items we loaded before.
```
LucasSte, line 543:
```suggestion
    /// already done this, we can cache the Expression::Variable, containing the items we loaded before.
```
`before` is repeated.

src/codegen/encoding/mod.rs
```diff
@@ -289,16 +765,15 @@ fn calculate_array_size<T: AbiEncoding>(
...
+    }
 }
 
 /// Calculate the size of a complex array.
-/// This function indexes an array from its outer dimension to its inner one
-fn calculate_complex_array_size<T: AbiEncoding>(
-    encoder: &mut T,
+/// This function indexes an array from its outer dimension to its inner one and
+/// accounts for the encoded length size for dynamic dimensions.
+fn calculate_complex_array_size(
```
LucasSte, line 775: Why is `encode_complex_array` a member of `trait AbiEncoding` while `calculate_complex_array_size` is not?
These functions are pretty similar.

tests/codegen_testcases/solidity/scale.sol
```diff
@@ -0,0 +1,57 @@
...
+        // CHECK: ty:bytes %abi_encoded.temp.9 = (alloc bytes len (%temp.8 + (builtin ArrayLength ((arg #0)))))
+        // CHECK: ty:uint32 %temp.10 = (builtin ArrayLength ((arg #0)))
+        // CHECK: branchcond (unsigned more %temp.10 > uint32 1073741824), block13, block14
+        // CHECK: branchcond (unsigned more (builtin ArrayLength ((arg #0))) > uint32 64), block3, block1
+        // CHECK: writebuffer buffer:%abi_encoded.temp.9 offset:uint32 0 value:(%temp.10 * uint32 4)
+        // CHECK: ty:uint32 %temp.11 = uint32 1
+        // CHECK: writebuffer buffer:%abi_encoded.temp.9 offset:uint32 0 value:((%temp.10 * uint32 4) | uint32 1)
+        // CHECK: ty:uint32 %temp.11 = uint32 2
+        // CHECK: writebuffer buffer:%abi_encoded.temp.9 offset:uint32 0 value:((%temp.10 * uint32 4) | uint32 2)
+        // CHECK: ty:uint32 %temp.11 = uint32 4
```
LucasSte, line 52: Are you supposed to write different values to the same offset?
I guess these instructions are in different branches. It would be nice to demarcate the instructions of each branch.

LucasSte (changes requested):

src/codegen/encoding/mod.rs
LucasSte: When I implemented this, I envisioned that the functions that could be common for multiple encoding schemes are the standalone ones, which received the encoder as a generic type argument. Those that should be specific to an encoding algorithm are implemented in its corresponding file.

A standalone function that receives a generic argument and a default implementation for the trait are basically the same. Now we have both methods in the file. Perhaps, it is better to stick to a single strategy to provide common methods for both encoding schemes.

tests/codegen_testcases/solidity/scale.sol
xermicus: Yes that is in different branches. However I need to touch this test again anyways in a following PR when I make this into a dedicated function.

src/codegen/encoding/mod.rs
xermicus: The array size calculation functions were not in the trait before I touched it (only the encode method), so I left it that way. 

I agree that the interface we have here is currently a bit weird. I discussed this already with Sean another day
- The way it is now might or might not be suitable for other encoding schemes
- Depending on their needs, we might need further improvements to this interface
- XDR (Soroban / Stellar) and ETH ABI encoding would be good candidates to analyze
- Hence, before we implement the next encoding scheme, we reshape this API if necessary.

This is way beyond the scope of the PR, so I'm gonna leave it as it is.

LucasSte: > The array size calculation functions were not in the trait before I touched it (only the encode method), so I left it that way.

None of the functions you added to the trait were already there, but you added them anyways. `encode_complex_array` was not in the trait. Can you think of a case when someone needs to override `encode_comple_array` but not `calculate_complex_array_size`?

> I discussed this already with Sean another day

You should start providing either a design document or the context of your discussions in your PR, so we don't have this misunderstandings anymore.

LucasSte (approved):

src/codegen/encoding/mod.rs
LucasSte: That makes sense. Thanks!

PR #1130, opened 2023-01-15
Refactor language server

seanyoung:
Make the language server use a builder-style interface. Now, there is no need to pass the list of hovers around as a function argument.

This fixes the hover for destructure statements with variable definition, `a` in this example.

```solidity
contract c {
	function f() public {
		(int64 a) = (2);
	}
}
```


src/bin/languageserver/mod.rs
```diff
@@ -280,65 +240,74 @@ impl SolangServer {
...
+                self.expression(expr, symtab);
             }
-            ast::Statement::Delete(_locs, _typ, expr) => {
-                SolangServer::construct_expr(expr, lookup_tbl, symtab, ns);
+            ast::Statement::Delete(_, _, expr) => {
+                self.expression(expr, symtab);
             }
-            ast::Statement::Destructure(_locs, _vecdestrfield, expr) => {
-                SolangServer::construct_expr(expr, lookup_tbl, symtab, ns);
+            ast::Statement::Destructure(_, _vecdestrfield, expr) => {
```
xermicus, line 267: Why are these variable names prefixed with an underscore?

xermicus (approved):

src/bin/languageserver/mod.rs
seanyoung: This is what the hyperledger mentee wrote many years ago. I have just removed a few more of these, I think that's all of them.

PR #1131, opened 2023-01-18
Fix recursive types in sema

xermicus:
Mappings and dynamic arrays can cause recursive struct fields which are sizeable and thus valid. `fn find_struct_recursion` is refactored to account for the possibility of recursive (but sizeable) structs. New approach: Structs are parsed into a graph with the edges resemble a connection to another struct. Now, within this graph,

1. we can find all Strongly Connected Components (SCC)
2. If there is a cycle within any SCC, every struct within the SCC is recursive.
3. If no path in this cycle contains any mapping or dynamic array (which break infinite sizes), every struct within the SCC is of infinite size.
4. For any struct in the graph, if there is a path leading into a cyclic SCC, we must flag that struct as recursive and (possibly) infinite size as well. This is to match solc and to avoid infinite recursion loops when recursively walking over types.

Now there is an additional struct field (`Paramater`) attribute `infinite_size`, so we can distinguish between `recursive` and `infinite size`.

But this leads to a problem with recursive type analysis. Recursive methods for `Type` relied on the `recursive` attribute to guard against stack overflows when recursively walking a type. But now, the `recursive` flag does no longer work  to guard against stack overflows. However, if a type is recursive but not of infinite size, we must still walk it recursively to collect all needed information. To fix this:

* The PR introduces `fn Type::guarded_recursion()` as a generic wrapper over some recursive closure. It guards the execution of the closure with a check to make sure we never enter infinite recursion. Instead, in that case, it just returns the bailout value supplied by the user.
* The PR contains several drive-by fixes for various recursive `Type` methods. These fixes also break their the infinite recursion loop, which reduces the amount of function needing the `guarded_recursion` to a minimum.

Code review hint: The main story happens in `src/sema/types.rs`, everything else is mostly just necessary to reflect those changes.

Closes https://github.com/xermicus/fuzzy-sol/issues/31

Contains drive-by fixes:
- For arrays, determining whether the outermost dimension was fixed or dynamic, the last dimension (`dims.last()`) was used. This is not correct as it should be `.first()`.
- Recursive function arguments are not allowed on public or external contract functions
- `fn solana_storage_size` should only be taken into account for the solana target
- `fn solana_storage_size` was not correct for dynamic arrays
- `fn memory_size_of` was not correct for dynamic arrays
- `fn storage_slots` was not correct for substrate


seanyoung (changes requested):
I am sorry, I am not convinced this is the right solution. When you have this struct:

```solidity
struct S {
   S[2] f1;
   int f2;
}
```
Then you have an impossibility: a struct that contains itself recursively. When you try to lay this out in memory, how large is `f1`? It will go on forever.

However, this example:
```solidity
contract c {
     struct S {
          mapping (in64, S) tree;
          int count;
     }
     S s;

    function add(uint64 a) public {
          s.tree[a].count = 1;
     }   

    function add(uint64 a, uint64 b) public {
          s.tree[a].tree[b].count = 1;
     }   

    function add(uint64 a, uint64 b, uint64  c) public {
          s.tree[a].tree[b].tree[c].count = 1;
     }   
}
```
This is a tree of trees (a forest?). Each struct S contains a another tree, and a count. So, if you want to lay out struct S in memory, then you can. `S` does not have infinite size. so

The original problem is that `fn contains_builtins()` does not deal with a mapping that contains itself. Maybe that problem should be solved instead.

src/sema/types.rs
```diff
@@ -203,27 +203,44 @@ fn find_struct_recursion(struct_no: usize, structs_visited: &mut Vec<usize>, ns:
...
+        let field_struct_no = match &field.ty {
+            Type::Struct(StructType::UserDefined(n)) => *n,
+            Type::Mapping(_, v) => {
+                if let Type::Struct(StructType::UserDefined(n)) = **v {
+                    n
+                } else {
+                    continue;
+                }
             }
+            Type::Array(v, _) => {
```
seanyoung, line 215: I think this is only true for arrays of fixed length (outer dimension).

```
struct S {
      S[] a;
      int b'
}
```
Is permitted (although a little weird).

Just like you can do with rust:

```rust
struct Expression {
     expr: Vec<Expression>,
}
```
but not 
```rust
struct Expression {
     expr: [Expression; 2],
}
```

xermicus:
@seanyoung I think you are right

> The original problem is that fn contains_builtins() does not deal with a mapping that contains itself. 

Thanks for the pointer, I'll go over it again

xermicus:
@seanyoung I think we need change the way how `fn find_struct_recursion` works. Right now, we always error out if a recursive field occurs. Instead, we should have a another (dedicated) check for infinite size? If so, I think my original fix could be correct (but not complete).

seanyoung:
On this branch solang goes into a infinite loop with this input:

```solidity
contract C {
	struct S {
		int f1;
		S[] f2;
	}
}
```

xermicus:
Found another case which still does not work:

```
contract C {
        struct S {
                int f1;
                S[] f2;
        }

        function foo(S s) public {}
}
```

xermicus:
More findings:

```solidity
contract Test {
	struct S {
		int foo;
		S[] s;
	}
	function test(S calldata s) public {}
}
```

We compile this. But we should not, `solc` does not allow it:

```
Error: Recursive type not allowed for public or external contract functions.
 --> /tmp/test.sol:5:16:
  |
5 | 	function test(S calldata s) public {}
  | 	             ^^^^^^^^^^^^

```

LucasSte:
```solidity
contract Test {
	struct S {
		int foo;
		S[] s;
	}

	function test(int foo) public returns (S) {
        S ss = S(foo, new S[](2));
        return ss;
    }
}
```

I have a stack overflow in your branch when I try this example.

xermicus:
@LucasSte your example produces a stack overflow in [emit](tests/contract_testcases/substrate/function_types/recursive_02.sol). In order to not bloat this PR and move on (it already contains a bunch of fixes), I'd rather fix this another time. Anyways thanks for the example, I added it to the [issue](https://github.com/hyperledger/solang/issues/1215) for this problem :+1:

LucasSte:
> ```solidity
> contract Test {
> 	struct S {
> 		int foo;
> 		S[] s;
> 	}
> 
> 	function test(int foo) public returns (S) {
>         S ss = S(foo, new S[](2));
>         return ss;
>     }
> }
> ```
> I have a stack overflow in your branch when I try this example.

The overflow here is in the encoder. It enters an infinite loop when calculating the struct size. Maybe this is something for other PR.

LucasSte:
```solidity
contract Test {
	struct S {
		int foo;
		S[] s;
	}

	function test(int f, S[] ss) public returns (S) {
		
		return S(f, ss);
	}
}
```
This code causes a stack overflow in sema: `fn to_signature_string` is called recursively.

src/sema/ast.rs
```diff
@@ -356,11 +357,31 @@ impl Function {
...
+            visibility,
+            pt::Visibility::External(_) | pt::Visibility::Public(_)
+        ) && recursive_parameter
+        {
+            ns.diagnostics.push(Diagnostic::error(
+                loc,
+                "Recursive parameter not allowed for public or external functions.".into(),
+            ))
+        }
+
```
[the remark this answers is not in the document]
xermicus: Moved it :+1:

src/sema/ast.rs
```diff
@@ -356,11 +357,31 @@ impl Function {
...
         let signature = match ty {
             pt::FunctionTy::Fallback => String::from("@fallback"),
             pt::FunctionTy::Receive => String::from("@receive"),
+            _ if recursive_parameter => "".into(),
             _ => ns.signature(&name, &params),
```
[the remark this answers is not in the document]
xermicus: I actually fixed it directly in `fn to_signature_string`, WDYT?

src/sema/types.rs
```diff
@@ -787,7 +806,10 @@ fn struct_offsets(ns: &mut Namespace) {
             let mut largest_alignment = 0;
 
             for field in &ns.structs[struct_no].fields {
-                let alignment = field.ty.align_of(ns);
+                let mut alignment = field.ty.align_of(ns);
+                if alignment == 0 {
+                    alignment = 1
+                }
```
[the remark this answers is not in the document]
xermicus: Yeah, fixed `fn align_of` so the recursion guard is no longer necessary

src/codegen/cfg.rs
```diff
@@ -1308,7 +1308,7 @@ impl ControlFlowGraph {
 
 /// Checks whether there is a virtual fallback or receive function
 fn is_there_virtual_function(
-    ns: &Namespace,
+    ns: &mut Namespace,
```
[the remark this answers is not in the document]
xermicus: No longer since the parameter recursion check is in the right place

xermicus:
My fix for `find_struct_recursion` is not working for all cases. Need to go over it again. A good way to do this is constructing a graph

src/sema/ast.rs
```diff
@@ -228,8 +228,9 @@ pub struct Parameter {
     pub indexed: bool,
     /// Some builtin structs have readonly fields
     pub readonly: bool,
-    /// A struct may contain itself which make the struct infinite size in
-    /// memory. This boolean specifies which field introduces the recursion.
+    /// A recursive struct may contain itself which make the struct infinite size in memory.
+    pub infinite_size: bool,
+    /// Is this struct field recursive. Recursive does not mean infinite size in all cases.
```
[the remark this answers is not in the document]
xermicus: Done

src/sema/types.rs
```diff
@@ -1735,6 +1811,22 @@ impl Type {
             _ => false,
         }
     }
+
+    /// Recursively walk over a type, protected against overflows on infinite recursive types.
+    /// `structs_visited` is the set of already visited structs.
+    /// `bail` is the value that should be returned in case an infinite recursion occured.
+    /// `f` is the closure applied on each iteration. `f` can be recursive.
+    fn recurse<F, O>(&self, structs_visited: &mut HashSet<usize>, bail: O, f: F) -> O
```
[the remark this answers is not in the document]
xermicus: Renamed the function and vastly extended the documentation

src/sema/types.rs
```diff
@@ -1580,47 +1624,72 @@ impl Type {
...
             Type::Mapping(..) => true,
-            Type::Array(ty, _) => ty.contains_mapping(ns),
+            Type::Array(ty, _) => ty.contains_mapping_internal(ns, structs_visited),
             Type::Struct(str_ty) => str_ty
                 .definition(ns)
                 .fields
                 .iter()
-                .any(|f| !f.recursive && f.ty.contains_mapping(ns)),
-            Type::StorageRef(_, r) | Type::Ref(r) => r.contains_mapping(ns),
+                .any(|f| f.ty.contains_mapping_internal(ns, structs_visited)),
```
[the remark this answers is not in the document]
xermicus: Explained in the PR, hopefully everything is clearn now :-)

src/sema/types.rs
```diff
@@ -1328,150 +1356,163 @@ impl Type {
...
-                    if self.is_sparse_solana(ns) {
+        self.storage_align_internal(ns, &mut HashSet::new())
+    }
+
+    fn storage_align_internal(
+        &self,
+        ns: &Namespace,
+        structs_visited: &mut HashSet<usize>,
+    ) -> BigInt {
+        self.recurse(structs_visited, 1.into(), |structs_visited| {
```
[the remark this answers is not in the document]
xermicus: Reduced usage of the recursion guard to a minimum and fixed the other methods, which should resolve this

src/sema/types.rs
```diff
@@ -1261,9 +1378,10 @@ impl Type {
                 .definition(ns)
                 .fields
                 .iter()
-                .map(|f| if f.recursive { 1 } else { f.ty.align_of(ns) })
+                .filter(|f| !f.infinite_size) // Can't calculate alignment for structs with infinite recursion
+                .map(|f| f.ty.align_of(ns))
                 .max()
-                .unwrap(),
+                .unwrap_or(1), // All fields were of infinite size, so we just pretend this
```
LucasSte, line 1384:
```suggestion
                .unwrap_or(1), // All fields had infinite size, so we just pretend the alignment is one
```

src/sema/types.rs
```diff
@@ -1393,17 +1508,12 @@ impl Type {
...
-                        }
-                    })
+                    .filter(|f| !f.infinite_size)
+                    .map(|f| f.ty.storage_slots(ns))
                     .sum(),
+                Type::Array(_, dims) if dims.last() == Some(&ArrayLength::Dynamic) => 1.into(),
                 Type::Array(ty, dims) => {
-                    let one = BigInt::one();
-
+                    let one = 1.into();
```
LucasSte, line 1516: I still prefer the old notation. It makes it clear we have a BigInt.

src/sema/types.rs
```diff
@@ -1446,15 +1555,10 @@ impl Type {
...
-                            BigInt::one()
-                        } else {
-                            field.ty.storage_align(ns)
-                        }
-                    })
+                    .filter(|field| !field.infinite_size)
+                    .map(|field| field.ty.storage_align(ns))
                     .max()
-                    .unwrap(),
+                    .unwrap_or_else(|| 1.into()), // All fields were infinite size, we pretend one storage slot.
```
LucasSte, line 1561:
```suggestion
                    .unwrap_or_else(|| 1.into()), // All fields had infinite size, we pretend one storage slot.
```

src/sema/types.rs
```diff
@@ -197,37 +203,113 @@ fn type_decl(
...
+/// A struct field is not of infinite size, if there are any 2 neighboring nodes in the path,
+/// where all connecting edges between any two nodes in the `path` are mappings or dynamic arrays.
+fn check_infinite_struct_size(graph: &Graph, path: Vec<usize>, ns: &mut Namespace) {
+    let mut infinite_size = true;
+    let mut offenders = HashSet::new();
+    for (a, b) in path.windows(2).map(|w| (w[0], w[1])) {
+        let mut infinite_edge = false;
+        for edge in graph.edges_connecting(a.into(), b.into()) {
+            match &ns.structs[a].fields[*edge.weight()].ty {
+                Type::Array(_, dim) if dim.last() != Some(&ArrayLength::Dynamic) => {}
```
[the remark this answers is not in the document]
LucasSte: This case still leads to a stack overflow in sema:

```solidity
contract Test {
	struct S {
		int foo;
		S[][2] s;
	}

	function test(int f, S[] ss) public returns (S) {
		return S(f, ss);
	}
}
```

src/sema/types.rs
```diff
@@ -1503,25 +1607,28 @@ impl Type {
 
     /// Does this type contain any types which are variable-length
     pub fn is_dynamic(&self, ns: &Namespace) -> bool {
-        match self {
+        self.is_dynamic_internal(ns, &mut HashSet::new())
+    }
+
+    fn is_dynamic_internal(&self, ns: &Namespace, structs_visited: &mut HashSet<usize>) -> bool {
+        self.guarded_recursion(structs_visited, false, |structs_visited| match self {
```
LucasSte, line 1614: Why do you need the recursion guard if you already identify the recursive fields beforehand?

src/sema/types.rs
```diff
@@ -1735,6 +1867,79 @@ impl Type {
...
+    /// `F` is expected to be a closure that recursively walks the `Type`.
+    /// `O` is the output type of the closure.
+    ///
+    /// `structs_visited` is the set of already visited structs. It is automatically updated for each struct already seen.
+    /// `bail` is the value that should be returned in case an infinite recursion occured.
+    /// `f` is the closure being called by this function.
+    ///
+    /// This function is useful in the various scenarios.
+    ///
+    /// Naturally, it can be used to detect recursive types (see `fn Type::is_recrsive()`).
```
LucasSte, line 1915:
```suggestion
    /// Naturally, it can be used to detect recursive types (see `fn Type::is_recursive()`).
```

tests/contract_testcases/substrate/function_types/recursive_01.sol
```diff
@@ -0,0 +1,8 @@
+contract C {
+	struct S {
+		int f1;
+		S[] f2;
```
LucasSte, line 4: `S[][2] foo` leads to a stack overflow.

tests/contract_testcases/substrate/structs/parse_structs_10.dot
```diff
@@ -12,6 +12,12 @@ strict digraph "tests/contract_testcases/substrate/structs/parse_structs_10.sol"
 	note_13 [label="recursive field 'c'\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:9:17-20"]
 	diagnostic_14 [label="struct 'B' has infinite size\nlevel Error\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:12:20-21"]
 	note_15 [label="recursive field 'd'\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:13:17-20"]
+	diagnostic_16 [label="struct 'A' has infinite size\nlevel Error\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:16:20-21"]
+	note_17 [label="recursive field 'd'\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:17:17-20"]
+	diagnostic_18 [label="struct 'A' has infinite size\nlevel Error\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:16:20-21"]
+	note_19 [label="recursive field 'b'\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:18:17-20"]
+	diagnostic_20 [label="struct 'A' has infinite size\nlevel Error\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:16:20-21"]
+	note_21 [label="recursive field 'c'\ntests/contract_testcases/substrate/structs/parse_structs_10.sol:19:17-20"]
```
LucasSte, line 20: We shouldn't be raising repeated errors.

LucasSte: I think you can add multiple notes to the same error.

xermicus: They are different errors. There are multiple erroneous fields on struct A so we expect multiple errors. The details are not shown here. But we still expect them.

tests/contract_testcases/substrate/function_types/recursive_01.sol
xermicus: Fixed, thanks

src/sema/types.rs
```diff
@@ -1735,6 +1867,79 @@ impl Type {
             _ => false,
         }
     }
+
+    // Does this type contain itself
+    pub fn is_recursive(&self, ns: &Namespace) -> bool {
+        self.is_recursive_internal(ns, &mut HashSet::new())
+    }
```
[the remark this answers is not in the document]
xermicus: Good catch, thanks. This is a leftover from before the analysis was fixed.

src/sema/ast.rs
```diff
@@ -228,8 +228,10 @@ pub struct Parameter {
     pub indexed: bool,
     /// Some builtin structs have readonly fields
     pub readonly: bool,
-    /// A struct may contain itself which make the struct infinite size in
-    /// memory. This boolean specifies which field introduces the recursion.
+    /// A recursive struct may contain itself which make the struct infinite size in memory.
+    pub infinite_size: bool,
+    /// Is this struct field recursive. Recursive does not mean infinite size in all cases:
+    /// `struct S { S[] s }` is recursive but not of infinite size.
```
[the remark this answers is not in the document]
xermicus: I don't get it. Why would implementing it in solidity fix it?

However, it is a good discovery. Recursive types are not allowed in public interfaces anyways and it looks like solc forbids encoding / decoding them:

```solidity
contract Test {
        struct S { int foo; S[] s; }

        function test(bytes memory b) public {
                S memory bar = S({foo: 0, s: new S[](0)});
                abi.encode(bar);
                S memory s = abi.decode(b, (S));
        }
}
```

```
solc --bin /tmp/test.sol

Error: This type cannot be encoded.
 --> /tmp/test.sol:9:14:
  |
9 |             abi.encode(bar);
  |                        ^^^

Error: Decoding type struct Test.S memory not supported.
  --> /tmp/test.sol:10:31:
   |
10 |            S memory s = abi.decode(b, (S));
   |              
```

I'll open another PR once we have this one merged

tests/contract_testcases/substrate/structs/parse_structs_10.dot
LucasSte: The error message is exactly the same for all of them. You should have only one error message with multiple notes.

src/sema/types.rs
```diff
@@ -197,37 +203,113 @@ fn type_decl(
...
+///   1. All structs in the namespace are parsed into a graph.
+///      Nodes in the graph represent the structs.
+///      Edges develop when a struct encapsulates another struct.
+///      Edges have the originating struct field as their weight.
+///      So we known from which struct field the connection originated later on.
+///   2. Find all Strongly Connected Components (SCC) in the graph.
+///   3. For any node inside in any SCC, if there is a path from the node to itself, we detected a cycle.
+///   4. For every cycle, check if it is of infinite size and flag involved struct fields accordingly.
+///   5. For any struct in the namespace, check if there are any paths leading into a cycle.
+///      If there are, flag the corresponding struct field as `recursive`.
```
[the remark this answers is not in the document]
xermicus: I think the meaning here is correct, even for formal definition.

src/sema/types.rs
```diff
@@ -862,6 +947,33 @@ fn struct_offsets(ns: &mut Namespace) {
...
+            Type::Array(ty, _) | Type::Ref(ty) | Type::Slice(ty) | Type::StorageRef(_, ty) => {
+                ty.user_struct_no(ns)
+            }
+            Type::UserType(no) => ns.user_types[*no].ty.user_struct_no(ns),
+            Type::InternalFunction {
+                params, returns, ..
+            }
+            | Type::ExternalFunction {
+                params, returns, ..
+            } => params
```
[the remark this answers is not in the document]
xermicus: Because they hold a list of `Type`s which then can really encapsulate anything, including structs.

src/sema/types.rs
xermicus: I do not. It is already clear we have a BigInt from the function signature. My version is less to type. I find my version easier to read. So in my opinion, it is better like that. Bike shedding :smile:

src/sema/types.rs
```diff
@@ -1421,7 +1531,6 @@ impl Type {
         }
     }
 
-    /// Alignment of elements in storage
     pub fn storage_align(&self, ns: &Namespace) -> BigInt {
```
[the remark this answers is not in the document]
xermicus: Restored it

src/sema/types.rs
xermicus: A recursive field does not necessarily imply  a dynamic field.

tests/contract_testcases/substrate/structs/parse_structs_10.dot
LucasSte: Your function `flag_as_infinite` is adding multiple times the same error message to Diagnostics. Instead of doing that, you can create a single diagnostic for `struct 'A' has infinite size` with multiple notes. `struct Diagnostic` has a `notes: Vec<Note>` field. The problem is not the test.

xermicus: `solc` doesn't even report the offending struct fields, nor does it report about all recursive structs (if there are multiple). But I changed it anyways, looks like this now:

```bash
error: struct 'A' has infinite size
   ┌─ /home/glow/code/solang/tests/contract_testcases/substrate/structs/parse_structs_10.sol:16:20
   │
16 │             struct A {
   │                    ^
17 │                 D d;
   │                 --- recursive field 'd'
18 │                 B b;
   │                 --- recursive field 'b'
19 │                 C c;
   │                 --- recursive field 'c'
```

src/sema/ast.rs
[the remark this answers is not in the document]
LucasSte: > I don't get it. Why would implementing it in solidity fix it?

In a struct like:
```solidity
struct S {
    int foo;
    S[] s;
}
```

The encoder we have in our compiler would recursively (and indefinitely) be building a for-loop to encode all the vectors within the struct. If we could do this using solidity, we would only enter the recursion if `s.length > 0`. During compile time, we have no information about the size of the vector.

src/sema/types.rs
LucasSte: No problem to leave it there. I left a comment because this was a random change in the codebase - it was not related to this PR.

src/sema/types.rs
LucasSte: I mean, can you avoid the recursion guard if you check the `recursive` boolean?

src/sema/types.rs
xermicus: True, looks random, I must have touched this at one point while working on something else without even noticing

src/sema/types.rs
```diff
@@ -197,37 +204,117 @@ fn type_decl(
...
-            if types_seen.contains(&field_struct_no) {
-                continue;
+/// A struct field is considered to be of infinite size, if it contains itself infinite times.
+///
+/// This function sets the `infinitie_size` flag accordingly for all connection between `nodes`.
+/// `nodes` is assumed to be a set of strongly connected nodes from within the `graph`.
+///
+/// Any node (struct) can have one or more edges (types) to some other node (struct).
+/// A struct field is not of infinite size, if there are any 2 neighboring nodes in the path,
+/// where all connecting edges between any two nodes in the `path` are mappings or dynamic arrays.
```
LucasSte, line 214: Path is the sequence of distinct edges between two vertices. The semantic of the word is not correct here, because you do not specify which vertices you are accounting for. The set of all edges in a graph is not necessarily a path.

In addition, you first say `there are any two neighboring nodes` and later you say `any two nodes`. It looks like the former reference is disconnected with the latter reference.

I suggest this sentence be reformulated.

src/sema/types.rs
xermicus: I don't think so.  Given the following:

```solidity
struct A {  A[] a; }
```

Now, `is_dynamic` should return true for struct `A`. But if we avoid recursion because field `A.a` is recursive, won't find out.

src/sema/types.rs
```diff
@@ -197,37 +204,117 @@ fn type_decl(
...
+///
+/// The algorithm consists of these steps:
+///   1. All structs in the namespace are parsed into a graph.
+///      Nodes in the graph represent the structs.
+///      Edges develop when a struct encapsulates another struct.
+///      Edges have the originating struct field number as their weight.
+///      So we known from which struct field the connection originated later on.
+///   2. Find all Strongly Connected Components (SCC) in the graph.
+///   3. For any node inside in any SCC, if there is a path from the node to itself, we've detected a cycle.
+///   4. For every cycle, check if it is of infinite size and flag involved struct fields accordingly.
```
LucasSte, line 284: What do you understand by a cycle of infinite size? Is A->B->C->A an infinite sized cycle? I guess you meant the memory size of the structs forming the cycle, right? The wording here looks like you are talking about the size of the cycle (the quantity of vertices you have in it).

src/sema/types.rs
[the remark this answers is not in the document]
LucasSte: `Type::InternalFunction` and `Type::ExternalFunction` are only function pointers. The types they encapsulate have no impact on memory.

Check this example that fails on your branch, but works on solc:
```solidity
contract foo {
    struct S {
        int ff;
        function (S memory) external fptr;
    }

    function func(S memory) public {
        
    }

}
```
S is not a recursive parameter.

src/sema/types.rs
```diff
@@ -1250,7 +1373,6 @@ impl Type {
         self.memory_size_of(ns) < BigInt::from(u16::MAX)
     }
 
-    /// Calculate the alignment
     pub fn align_of(&self, ns: &Namespace) -> usize {
```
LucasSte, line 1376: This function still needs a doc comment.

LucasSte (changes requested):

src/sema/types.rs
LucasSte: In the example you've given, you can detect the dynamic array in `Type::Array` before recursing for A. In this case, the recursion can be avoided.

src/sema/ast.rs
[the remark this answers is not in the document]
xermicus: I still don't get it, feel free to enlighten me. Everything written in Solidity ends up in our codegen graph, so why is it not possible to write the same code into our CFG manually?

src/sema/types.rs
```diff
@@ -197,37 +204,117 @@ fn type_decl(
...
-            types_seen.insert(field_struct_no);
-
-            if structs_visited.contains(&field_struct_no) {
-                ns.diagnostics.push(Diagnostic::error_with_note(
-                    def.loc,
-                    format!("struct '{}' has infinite size", def.name),
-                    field.loc,
-                    format!("recursive field '{}'", field.name_as_str()),
-                ));
+/// A struct field is recursive, if it is connected to a cyclic path.
```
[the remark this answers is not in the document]
xermicus: Yes. We use the `Parameter.recursive` struct field as a recursion guard in some places.

src/sema/types.rs
xermicus: Yes of course I meant memory size. I updated the comment so it says memory size. However, completely alien to me how anyone could confuse it like this. It makes no sense. It's impossible to write a contract with structs having an infinite "quantity of vertices" first hand.

src/sema/types.rs
```diff
@@ -787,7 +874,10 @@ fn struct_offsets(ns: &mut Namespace) {
             let mut largest_alignment = 0;
 
             for field in &ns.structs[struct_no].fields {
-                let alignment = field.ty.align_of(ns);
+                let mut alignment = field.ty.align_of(ns);
+                if alignment == 0 {
+                    alignment = 1;
+                }
```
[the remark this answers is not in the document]
xermicus: No, good catch, delete this code

src/sema/types.rs
xermicus: Ok, exchanged `path` with `connection`. Please keep in mind we are not trying to write a formal book about graph theory here. This is just a doc comment providing additional context

src/sema/types.rs
xermicus: It's not possible.

```solidity
struct A {  B b; }
struct B { C[] c; }
struct C { C c; }
```

 `is_dynamic` should stilll return true for struct A. We need to do the recursion, regardless whether a struct field is flagged as recursive

src/sema/types.rs
```diff
@@ -1735,6 +1861,63 @@ impl Type {
...
+    ///
+    /// `structs_visited` is the set of already visited structs. It is automatically updated for each struct already seen.
+    /// `bail` is the value that should be returned in case an infinite recursion occured.
+    /// `f` is the closure being called by this function.
+    ///
+    /// This function is useful in the various scenarios.
+    ///
+    /// Naturally, it can be used to detect recursive types (see `fn Type::is_recursive()`).
+    ///
+    /// Moreover,functions like `fn Type::contains_mapping()` need to recursively check the type to contain mappings.
```
LucasSte, line 1895:
```suggestion
    /// Moreover, functions like `fn Type::contains_mapping()` need to recursively check if the type contains mappings.
```

src/sema/types.rs
```diff
@@ -1735,6 +1861,63 @@ impl Type {
...
+    /// Consider the following valid type:
+    ///
+    /// ```solidity
+    /// struct A { B b; }
+    /// struct B { A[] a; }
+    /// ```
+    ///
+    /// Looking at nested or referential types individually does not work here, this can only be done recursively.
+    /// However, naive recursion will lead to infinity here.
```
LucasSte, line 1904:
```suggestion
    /// Looking at nested or referential types individually does not work here. This can only be done recursively;
    /// however, a naive recursion will run indefinitely.
```

tests/contract_testcases/substrate/structs/parse_structs_22.sol
```diff
@@ -0,0 +1,11 @@
...
+	struct S {
+		int foo;
+		S[][2] s;
+	}
+
+	function test(int f, S[][2] ss) public returns (S) {
+		return S(f, ss);
+	}
+}
+
```
LucasSte, line 11: Just for completeness, you can add a test for a struct that holds function pointers.

LucasSte (approved):

src/sema/ast.rs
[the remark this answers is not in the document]
LucasSte: I think you're right. If, instead of generating code recursively, we used function calls for encoding arrays after that we have check for their size, we would have no problem.

src/sema/types.rs
```diff
@@ -197,37 +204,117 @@ fn type_decl(
...
-            if types_seen.contains(&field_struct_no) {
-                continue;
+/// A struct field is considered to be of infinite size, if it contains itself infinite times (not in a vector).
+///
+/// This function sets the `infinitie_size` flag accordingly for all connections between `nodes`.
+/// `nodes` is assumed to be a set of strongly connected nodes from within the `graph`.
+///
+/// Any node (struct) can have one or more edges (types) to some other node (struct).
+/// A struct field is not of infinite size, if there are any 2 connecting nodes,
+/// where all edges between any of two connecting nodes are mappings or dynamic arrays.
```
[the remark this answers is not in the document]
xermicus: It is not correct, a struct is not of infinite memory size if there are any (not all) two neighboring nodes in the connection, where all edges are mappings or dyn. arrays. I updated this sentence and added an example to make that more clear.

tests/contract_testcases/substrate/structs/parse_structs_22.sol
xermicus: Good idea, added it

src/sema/types.rs
seanyoung: I have no problem parsing/understanding the comment, it's perfectly clear.

seanyoung (approved):

PR #1132, opened 2023-01-18
Fix encoder for multidimensional dynamic arrays

LucasSte:
The example in the tests is the one of the cases I found in which we can use two dynamic dimensions in an array. The example helped me find various bugs in the way I was treating them in the encoder. This is also going to help @xermicus with his #1128 PR.

@xermicus You can decided whether to merge this PR or incorporate the change into your PR.

### DIFF
--- a/src/codegen/encoding/mod.rs
+++ b/src/codegen/encoding/mod.rs
@@ -240,13 +240,18 @@ fn calculate_array_size<T: AbiEncoding>(
         }
 
         let type_size = Expression::NumberLiteral(Loc::Codegen, Type::Uint(32), compile_type_size);
-        let size = Expression::Multiply(
+        let mut size = Expression::Multiply(
             Loc::Codegen,
             Type::Uint(32),
             false,
             Box::new(size),
             Box::new(type_size),
         );
+
+        if !encoder.is_packed() && dims.last().unwrap() == &ArrayLength::Dynamic {
+            size = increment_four(size);
+        }
+
         let size_var = vartab.temp_anonymous(&Type::Uint(32));
         cfg.add(
             vartab,
@@ -287,30 +292,6 @@ fn calculate_array_size<T: AbiEncoding>(
         size_var
     };
 
-    // Each dynamic dimension size occupies 4 bytes in the buffer
-    let dyn_dims = dims.iter().filter(|d| **d == ArrayLength::Dynamic).count();
-
-    if dyn_dims > 0 && !encoder.is_packed() {
-        cfg.add(
-            vartab,
-            Instr::Set {
-                loc: Loc::Codegen,
-                res: size_var,
-                expr: Expression::Add(
-                    Loc::Codegen,
-                    Type::Uint(32),
-                    false,
-                    Box::new(Expression::Variable(Loc::Codegen, Type::Uint(32), size_var)),
-                    Box::new(Expression::NumberLiteral(
-                        Loc::Codegen,
-                        Type::Uint(32),
-                        BigInt::from(4 * dyn_dims),
-                    )),
-                ),
-            },
-        );
-    }
-
     Expression::Variable(Loc::Codegen, Type::Uint(32), size_var)
 }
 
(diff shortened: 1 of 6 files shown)


xermicus:
Cool, thanks! We can merge it into master as its own PR, it makes reviewing easier. I can solve the resulting merge conflicts for my branch. On a quick first glance you now do the size calculation for vector lengths the same way as I planned to, and the rest is related to the sub array loading which I didn't touch.

src/codegen/encoding/mod.rs
```diff
@@ -428,39 +418,28 @@ fn calculate_struct_size<T: AbiEncoding>(
...
-}
-
-/// Dereferences a subarray. If we have 'int[3][][4] vec' and we need 'int[3][]',
-/// this function returns so.
-/// 'dims' should contain only the dimensions we want to index
-/// 'index' is the list of indexes to use
-/// 'index_first_dim' chooses whether to index the first dimension in dims
-fn load_sub_array(
+/// Indexes an array. If we have 'int[3][][4] vec' and we need 'int[3][]',
+/// 'int[3]' or 'int' this function returns so.
```
xermicus, line 422: What is `int`, you mean just the whole thing?

src/codegen/encoding/mod.rs
```diff
@@ -428,39 +418,28 @@ fn calculate_struct_size<T: AbiEncoding>(
...
+/// 'dims' is the vector containing the arrays dimensions
+/// 'index' is the list of indexes to use for each dimension
+/// 'return_pointer' tells us if we should return a pointer
```
xermicus, line 425: This comment is not helpful, as that is already obvious by looking at the code. It would be better to explain why and/or when a user would want that.

src/codegen/encoding/mod.rs
```diff
@@ -469,13 +448,28 @@ fn load_sub_array(
             Box::new(Expression::Variable(
                 Loc::Codegen,
                 Type::Uint(32),
-                indexes[indexes.len() - i - 1],
+                indexes[dims.len() - i - 1],
             )),
         );
+
+        // Multidimensional arrays have an inconsistent indexing method. We should only load if the
+        // dimension is dynamic.
```
xermicus, line 456: A comment on why fixed size dimension do not need a load would be helpful. Also this comment would serve better as doc-comments to the function.

xermicus (changes requested):
Was this supposed to fix the example from #1128?

```solidity
contract PTokenFactory {
    function strConcat(uint16[][3][] memory vec) public pure returns (bytes memory){
        bytes b = abi.encode(vec);
        return b;
    }
}
``` 

Still crashes with the same failed assertion:

```bash
❯ cargo run -- compile --target substrate ptok.sol 
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/solang compile --target substrate ptok.sol`
solang: /home/runner/work/solang/solang/llvm-project/llvm/include/llvm/IR/Instructions.h:923: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
```

LucasSte:
> Was this supposed to fix the example from #1128?
> 
> ```solidity
> contract PTokenFactory {
>     function strConcat(uint16[][3][] memory vec) public pure returns (bytes memory){
>         bytes b = abi.encode(vec);
>         return b;
>     }
> }
> ```

No, this PR does not fix the example. It contains problems that are far broader than just the context of your PR. Even if you remove the `abi.encode`, indexing `vec[0][1][2]` will crash the compiler.

> Still crashes with the same failed assertion:

You should be more attentive Cyrill. The compiler still crashes, but the error is not the same one from the main branch. This is the error from main:

`thread 'main' panicked at 'deref on non-array', src/sema/types.rs:1028:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace`

src/codegen/encoding/mod.rs
LucasSte: I'll improve the comment.

> as it is already obvious what return_pointer does by looking at the code.

Well, if you analyze well, you'll see that even if `return_pointer` is false, you can have a pointer returned, so it is not as obvious as you said.

xermicus:
Well, in #1128, where this example comes from, you said

> I'm gonna fix the bug in the example 

Which is why I didn't actually double check the error message because I thought it should be fixed. Besides that,  doesn't matter. But fair enough, it's just another a fix for another bug.

xermicus (approved):

src/codegen/encoding/mod.rs
LucasSte: I've updated all the comments you pointed. Can you please re-check?

LucasSte:
> Well, in #1128, where this example comes from, you said
> 
> > I'm gonna fix the bug in the example
> 
> Which is why I didn't actually double check the error message because I thought it should be fixed. Besides that, doesn't matter. But fair enough, it's just another a fix for another bug.

That was my intention, but I found that to not be worth the time, so I came up with a simpler example that tested the same code path during encoding and decoding.

seanyoung (approved):
Good explanation

src/codegen/encoding/mod.rs
xermicus: Looks good, thanks

PR #1134, opened 2023-01-20
Add casts for yul pointer to integer conversion and vice versa

seanyoung:
yul allows implicit conversion of pointers to integers and vice versa.

Fixes https://github.com/xermicus/fuzzy-sol/issues/79


seanyoung:
I tried avoiding the repeated `.cast(to_ty, ns)` but ended up with uglier code.

src/codegen/mod.rs
```diff
@@ -912,6 +912,40 @@ impl Expression {
...
+            }
+
+            _ if !from.is_contract_storage()
+                && !to.is_contract_storage()
+                && !from.is_reference_type(ns)
+                && !to.is_reference_type(ns) =>
+            {
+                // cast pointer to different pointer
+                Expression::Cast(self.loc(), to.clone(), self.clone().into())
+            }
```
LucasSte, line 947: I understand this is the easiest solution to fix the bug, but it is complex to understand. 
Is it time for us to redesign `ast::Type` and perhaps include a pointer type (at another time, of course)?
I was thinking that we might also need a `codegen::Type`, because eliminating `Type::ExternalFunction` from codegen would clean up some messes there.

xermicus (approved):
LGTM

src/codegen/mod.rs
seanyoung: I was thinking that when we introduce SSA, this becomes a new intermediate format.  So, CFG -> mid-level-IR and SSA becomes new low-level-IR.

Ideally the mid-level-IR does not have pointers (like our CFG now), but the low-level-IR does have pointers. So, I am not keen on introducing pointers in our CFG.

PR #1137, opened 2023-01-24
Refactor `sema/expression.rs`

LucasSte:
The file `sema/expression.rs` has eight thousand lines. It is difficult to navigate and to keep track of changes. This PR splits the original file into many small archives, under `sema/expression`.

This PR only moves code. There is no change in the code itself.

@seanyoung and @xermicus, please, let me know if I should rename a file or if there is any misplaced function in this PR.


src/sema/expression/expression_impl.rs
```diff
@@ -0,0 +1,1203 @@
...
+use num_bigint::{BigInt, Sign};
+use num_rational::BigRational;
+use num_traits::ToPrimitive;
+use num_traits::{FromPrimitive, Zero};
+use solang_parser::diagnostics::Diagnostic;
+use solang_parser::pt;
+use solang_parser::pt::CodeLocation;
+use std::cmp::Ordering;
+
+impl Expression {
```
xermicus, line 17: It feels weird to me having that in not in the same module where the `Expression` enum is defined (however, the `ast` module would blow up in size if we do this). How about putting this inside `mod.rs` instead?

xermicus (approved):
LGTM

src/sema/expression/arithmetics.rs
```diff
@@ -0,0 +1,864 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::sema::ast::{Expression, Namespace, RetrieveType, StringLocation, Type};
```
seanyoung, line 3: To me arithmetics sounds strange, I don't really know of this being used in the plural. `arithmetic.rs` might be better.

There is a pedantic point about bitwise operations not being arithmetic, but I don't think that matters.

src/sema/expression/expression_impl.rs
seanyoung: I concur, this also means we get rid of the filename `expression_impl.rs` which is a bit strange. Names should describe what things do, not what syntax they use. 

This reminds me of C++ pimpl which surely an abomination :see_no_evil:

seanyoung (approved):

src/sema/expression/arithmetics.rs
LucasSte: You're right. The word has no `s` in the end. I can call this file `mathematics.rs`. Does this improve anything?

src/sema/expression/expression_impl.rs
LucasSte: Moving `impl Expression` to `mod.rs` will enlarge the file to more than three thousand lines. I really would like to avoid that. Another option is to create a new file for `fn expression` and leave the `impl Expression` in `mod.rs`. Is this better?

I agree with @xermicus that `impl Expression` should be in the same file as `enum Expression`, but I think @seanyoung is not keen on the idea to split the AST definition in multiple files.

seanyoung: The AST is already defined in multiple files: yul is defined elsewhere.

Also, we have `impl Type` in various places, certainly not just in `ast.rs`. I don't think this is necessarily bad, but then again I don't think large files are so bad either.

src/sema/expression/arithmetics.rs
seanyoung: To be pedantic, it's not math, it's simple arithmetic.

LucasSte: > To be pedantic, it's not math, it's simple arithmetic.

Arithmetic is also math.

src/sema/expression/expression_impl.rs
LucasSte: > Another option is to create a new file for `fn expression` and leave the `impl Expression` in `mod.rs`. Is this better?

What about this?

Issue #1129, opened 2023-01-14
Ad support for Soroban as a target


tomerweller:
[Soroban](https://soroban.stellar.org/docs) is a new smart contracts platform built by and for the Stellar ecosystem. The [Soroban environment](https://github.com/stellar/rs-soroban-env) utilizes Parity's [WASMI](https://github.com/paritytech/wasmi) runtime and is written in rust. The main [contract developer SDK](https://github.com/stellar/rs-soroban-sdk) is also written in Rust.

Stellar is one of the most active L1 blockchains by TPS (according to [realtps.net](https://realtps.net/)) and is connected to a global network of on and off ramps, including hundreds of thousands of cash agents (through moneygram). These factors make Soroban especially appealing for smart contract developers.

seanyoung:
Very interesting. I think it's entirely possible to add support to Solang for Soroban contracts runtime. 

This would require the following steps:
 - Add new target Soroban
 - Solang sema should have special rules for Soroban builtins
 - Write XDR encoder/decoder for Solang (at least from my cursory reading, it looks like Soroban uses the XDR encoding)
 - Solang emit should implement new target SorobanTarget and handle low level code like host function calls
 - Add whole bunch of tests and documentation

Now right now I don't think any of the maintainers have time to work on this, but we certainly are always available to help anyone with this. We have a call every weekday at 13:00 UTC on Hyperledger discord server `#solang-meetup`

leighmcculloch:
@seanyoung Thanks for an outline into what work would be required.

@JakeUrban @graydon and I spent a small amount of time last week exploring what would be involved. We wrote (hacked together) some changes to get the most basic contract executing on Soroban, to help us understand how compatible Soroban would be with Solang, and Solidity, and identify blind spots.

I tried to capture that in comments on:
- #1138

There are some problems that we'd have to come up with a story for:
- Solidity supports function overloading, Soroban does not. Soroban functions need to be max 10 characters long in name, so appending parameter lists wouldn't work. We might need to hash function names, or suffix them.
- Solidity has guaranteed to execute constructors, but Soroban currently does not. We'd probably need to map them to some regular functions named `init...`
- Solidity has multiple return values. Soroban doesn't support multiple return values in WASM. We would probably address this by placing the return values into a Soroban `Vec`.
- Solidity has u256 values, but Soroban only supports up to 128-bit values. We could map them to Soroban's `Bytes` but that might be inefficient.
[...]

PR #1138, opened 2023-01-24
Explore adding Soroban as a target

leighmcculloch:
This pull request contains an exploration of adding Soroban as a target to Solang.

The primary goal was to learn and fill gaps in our knowledge, to determine how feasible it would be to add Soroban as a target, and to identify what problems would need to be solved to make adding the target possible.

The implementation in this pull request is rudimentary and take shortcuts. i.e. It's a hack. Hopefully though this might lead someone contributing an implementation in the future.

The pull request has comments noting problems to solve, and areas where Soroban doesn't fit into existing assumptions, or Soroban is conceptually incompatible.

This pull request isn't intended for merging and I'll close it immediately, but I've opened this as a pull request on solang because I wanted to store this exploration near the related issue for future reference.

The changes in this pull request successfully compile a very simple Solidity contract that can run on Soroban:

https://user-images.githubusercontent.com/351529/214360701-8490230a-5bd1-4f03-ac57-f4660744cfaf.mp4

Related to https://github.com/hyperledger/solang/issues/1129

Thanks @JakeUrban @tomerweller @graydon for helping with this.


Cargo.toml
```diff
@@ -57,6 +57,7 @@ parse-display = "0.8.0"
 parity-scale-codec = "3.1"
 ink = "4.0.0-beta"
 scale-info = "2.3"
+stellar-xdr = { version = "0.0.12", default-features = false, features = ["std", "next", "base64"] }
```
leighmcculloch, line 60: Stellar XDR isn't used by contracts and doesn't end up in the WASM code/instructions that get generated, but it is imported because Soroban contracts embed an API specification for the contract functions and custom types, and that is encoded in XDR and embedded in a custom section of the .wasm file.

src/codegen/cfg.rs
```diff
@@ -410,6 +410,7 @@ impl BasicBlock {
 #[derive(Clone)]
 pub struct ControlFlowGraph {
     pub name: String,
+    pub original_name: String,
```
leighmcculloch, line 413: At the moment contract functions end up in the wasm with the actual original name in Soroban. If there's a `max` function, the wasm ends up with a `max` export.

For example, the following Soroban Rust contract:
```
#![no_std]
use soroban_sdk::contractimpl;

pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn max(x: u64, y: u64) -> u64 {
        x.max(y)
    }
}
```

Results in the following wasm (truncated), where the `;3;` function is exported as `max`:
```wat
(module
  (type (;0;) (func (param i64) (result i64)))
  (type (;1;) (func (param i32 i64)))
  (type (;2;) (func (param i64 i64) (result i64)))
  (type (;3;) (func))
  (import "i" "0" (func (;0;) (type 0)))
  (import "i" "_" (func (;1;) (type 0)))
  (func (;2;) (type 1) (param i32 i64)
    ...)
  (func (;3;) (type 2) (param i64 i64) (result i64)
    ...)
  (func (;4;) (type 3))
  (memory (;0;) 16)
  (global (;0;) (mut i32) (i32.const 1048576))
  (global (;1;) i32 (i32.const 1048576))
  (global (;2;) i32 (i32.const 1048576))
  (export "memory" (memory 0))
  (export "max" (func 3))
  (export "_" (func 4))
  (export "__data_end" (global 1))
  (export "__heap_base" (global 2)))
```

In this hack we modified Solang to preserve the original function name and then hacked it to use that name in the emitted functions, just so we could get it minimally working. 

There's a problem here in that Solidity supports function overloading of both regular functions and constructors, and Soroban and WASM requires a function to have a unique name. cc @graydon

src/emit/binary.rs
```diff
@@ -603,40 +594,15 @@ impl<'a> Binary<'a> {
...
-                    .unwrap()
-                    .ptr_type(AddressSpace::Generic)
-                    .into(),
-            );
-        }
-
-        // Solana return type should be 64 bit, 32 bit on wasm
-        self.return_values[&ReturnCode::Success]
-            .get_type()
-            .fn_type(&args, false)
```
leighmcculloch, line 639: Soroban functions always accept and return `i64` values, where those values are a Soroban RawVal, a 64-bit encoded value that Soroban contracts need to unpack. Soroban doesn't have a return code, and doesn't write return values to imported memory, so I commented this code out.

A RawVal is a 4-bit tag space with a 60-bit value space. Small values like 32-bit integers are encoded directly into the 64-bit value. Larger values like byte arrays, vectors, maps, are a 32-bit handle embedded into the RawVal, and there are host functions that can be used to extract elements of the larger value.

This pull request doesn't contain adding RawVal decode/encode for types, so when I was trying it out the uint64 values the contract were receiving were the RawVal encoded value, not the user value.

src/emit/functions.rs
```diff
@@ -33,14 +33,13 @@ pub(super) fn emit_functions<'a, T: TargetRuntime<'a>>(
...
-            let func_decl = if let Some(func) = bin.module.get_function(&cfg.name) {
+            let func_decl = if let Some(func) = bin.module.get_function(&cfg.original_name) {
                 // must not have a body yet
                 assert_eq!(func.get_first_basic_block(), None);
 
                 func
             } else {
-                bin.module
-                    .add_function(&cfg.name, ftype, Some(Linkage::Internal))
+                bin.module.add_function(&cfg.original_name, ftype, None)
```
leighmcculloch, line 42: I removed the internal linkage on emitted functions and instead rely completely on the `internalize` function to mark internal functions as internal.

src/linker/wasm.rs
```diff
@@ -29,14 +29,11 @@ pub fn link(input: &[u8], name: &str) -> Vec<u8> {
         CString::new("--global-base=0").unwrap(),
     ];
 
-    command_line.push(CString::new("--export").unwrap());
-    command_line.push(CString::new("deploy").unwrap());
-    command_line.push(CString::new("--export").unwrap());
-    command_line.push(CString::new("call").unwrap());
+    command_line.push(CString::new("--export-dynamic").unwrap());
```
leighmcculloch, line 32: Since the exported functions are dynamic, I resorted to using `--export-dynamic`, although in a proper implementation maybe it would be better to retain more control over this and iterate over the list of functions with `--export` flags.

src/linker/wasm.rs
```diff
@@ -29,14 +29,11 @@ pub fn link(input: &[u8], name: &str) -> Vec<u8> {
...
 
-    command_line.push(CString::new("--import-memory").unwrap());
-    command_line.push(CString::new("--initial-memory=1048576").unwrap());
-    command_line.push(CString::new("--max-memory=1048576").unwrap());
+    // command_line.push(CString::new("--import-memory").unwrap());
+    // command_line.push(CString::new("--initial-memory=1048576").unwrap());
+    // command_line.push(CString::new("--max-memory=1048576").unwrap());
```
leighmcculloch, line 36: Soroban doesn't provide a `env.memory` import. Soroban does do memory copies, but Soroban contracts export memory instead.

src/linker/wasm.rs
```diff
@@ -64,27 +61,27 @@ pub fn link(input: &[u8], name: &str) -> Vec<u8> {
...
+        //             | "seal_terminate"
+        //             | "seal_random"
+        //             | "seal_call" => "seal1",
+        //             _ => "seal0",
+        //         };
+        //         *imports[ind].module_mut() = module_name.to_owned();
+        //     }
+
+        //     ind += 1;
+        // }
```
leighmcculloch, line 84: This change can be ignored. Because Soroban has no always imported value, if a contract doesn't use any host functions/imports, there is no import section to mutate and line 64 panicked. I'm sure there would be a better way to code this to just check if the import section existed or not.

xermicus:
Awesome that you could pull this off!

src/emit/binary.rs
xermicus: >  I assume we'd inject those instructions into the beginning of functions, or as a bitcode as a form of stdlib.

For Borsh (Solana) and SCALE (Substrate / ink!), we have our own encoder and decoder implementations in `codegen`. Adding Soroban as a target would imply implementing an encoder for your encoding schema as well, which should solve the problem (encoding is needed in multiple places, e.g. events, calls, return values).

src/codegen/cfg.rs
leighmcculloch: Ahh, yes I think that would work for constructors.

For function name mangling, at the moment we still need to stay under 10 characters. Maybe a truncated hash would be best. Although, it will mean other Soroban contracts would be calling functions like `.98a7df9(x, y)` instead of `.max(x, y)`.

@graydon This could be a good argument for supporting a mapping of rendered function names to exported WASM functions in the Soroban contract specification XDR. It would be a simple addition, and would allow code generators to generate nice function names like `max` while the actual WASM function would be named `98a7df9`.

src/emit/binary.rs
leighmcculloch: Great, yes that seems like the best place to implement it. Thanks 👍🏻.

PR #1140, opened 2023-01-26
Do not build stdlib/bigint.c while it cannot be built with llvm 14

seanyoung:
This breaks the container build.

### DIFF
--- a/integration/substrate/asserts.spec.ts
+++ b/integration/substrate/asserts.spec.ts
@@ -29,7 +29,7 @@ describe('Deploy asserts contract and test', () => {
         expect(res0.output?.toJSON()).toEqual(1);
 
         let res1 = await query(conn, alice, contract, "testAssertRpc");
-        expect(res1.result.toHuman()).toEqual({ "Err": { "Module": { "error": "0x0b000000", "index": "7" } } });
+        expect(res1.result.toHuman()).toEqual({ "Err": { "Module": { "error": "0x0b000000", "index": "8" } } });
 
         let gasLimit = await weight(conn, contract, "testAssert");
         let tx = contract.tx.testAssert({ gasLimit });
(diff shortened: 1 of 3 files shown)


integration/substrate/asserts.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('Deploy asserts contract and test', () => {
         expect(res0.output?.toJSON()).toEqual(1);
 
         let res1 = await query(conn, alice, contract, "testAssertRpc");
-        expect(res1.result.toHuman()).toEqual({ "Err": { "Module": { "error": "0x0b000000", "index": "7" } } });
+        expect(res1.result.toHuman()).toEqual({ "Err": { "Module": { "error": "0x0b000000", "index": "8" } } });
```
LucasSte, line 32: Why did you change this file?
It seems unrelated to this PR.

seanyoung: The substrate tests don't work anymore unless I change this.

seanyoung: @xermicus do you know anything about this?

xermicus: Our CI uses some  `latest` docker image for the substrate contracts node. There was an update published a couple hours ago which is causing the error.

Before the update, the `contracts` pallet was the 7th module in the pallet, but now it is the 8th. The but the actual error is still the same (`ContractTrapped`, which is granted, not very obvious here), so this change is fine. Unless you want me to open a dedicated PR to fix this

xermicus (approved):

integration/substrate/asserts.spec.ts
xermicus: #1146 will fix this, sorry

LucasSte (approved):

PR #1146, opened 2023-01-26
pin substrate node on CI

xermicus:
It is annoying to use `:latest` images for our CI (I still use todays `latest` image here but pinned it).


seanyoung:
> It is annoying to use `:latest` images for our CI (I still use todays `latest` image here but pinned it).

We could also make the test not depend on the exact value of index in this test; `error` is really only what matters here, right?

seanyoung (approved):

xermicus:
> We could also make the test not depend on the exact value of index in this test; `error` is really only what matters here, right?

We might get an error from a different pallet or a different error from the contracts pallet (although unlikely, it could happen...)

LucasSte (approved):

PR #1147, opened 2023-01-26
Bump solana_rbpf

seanyoung:
- Bump solana_rbpf to 0.2.35. This makes it possible to use rust 1.67
- Fix sporadic test failure

### DIFF
--- a/tests/solana.rs
+++ b/tests/solana.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::borsh_encoding::{decode_at_offset, encode_arguments, BorshToken};
+use anchor_syn::idl::Idl;
 use base58::{FromBase58, ToBase58};
 use byteorder::{ByteOrder, LittleEndian, WriteBytesExt};
 use libc::c_char;
@@ -12,19 +13,17 @@ use solana_rbpf::{
     elf::Executable,
     error::EbpfError,
     memory_region::{AccessType, MemoryMapping, MemoryRegion},
-    question_mark,
-    syscalls::BpfSyscallContext,
-    user_error::UserError,
     verifier::RequisiteVerifier,
     vm::{
-        Config, EbpfVm, SyscallObject, SyscallRegistry, TestInstructionMeter, VerifiedExecutable,
+        Config, EbpfVm, ProgramResult, SyscallRegistry, TestInstructionMeter, VerifiedExecutable,
     },
 };
-use solang::compile;
-
-use anchor_syn::idl::Idl;
-use solang::abi::anchor::{discriminator, generate_anchor_idl};
-use solang::{file_resolver::FileResolver, Target};
+use solang::{
+    abi::anchor::{discriminator, generate_anchor_idl},
+    compile,
+    file_resolver::FileResolver,
+    Target,
+};
 use std::{
     cell::{RefCell, RefMut},
     collections::HashMap,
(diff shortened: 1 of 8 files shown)


tests/solana.rs
```diff
@@ -39,6 +36,20 @@ use tiny_keccak::{Hasher, Keccak};
 mod borsh_encoding;
 mod solana_tests;
 
+/// Error handling for syscall methods
+macro_rules! question_mark {
```
xermicus, line 40: I think a much more idiomatic way to do this would be to return an `Error` instead of passing a mutable pointer to some value representing an error. Then you can either use `map_err` with the actual question mark `?` syntax sugar

```rust
computation = whatever().map_err(|e| ProgramResult::Err(e.into()))?;
```

Or, even better, the casting can be directly implement, something like `impl From<WhatEverError> for ProgramResult`. Which would no longer require `map_err` and the `?` sugaring should just work.

Completely unrelated to what the PR is trying, but I couldn't resist sorry :monkey: So feel free to just take this as an improvement / suggestion that can be at a later time anyways.

tests/solana.rs
```diff
@@ -1438,26 +1366,29 @@ impl<'a> SyscallObject<UserError> for SyscallInvokeSignedC<'a> {
                 vm.stack.insert(0, p);
 
                 let res = vm.execute(&instruction.accounts, &instruction.data);
-                assert_eq!(res, Ok(0));
+                assert!(matches!(res, ProgramResult::Ok(0)));
```
xermicus, line 1369: Nitpick: uses one less macro and makes it less verbose.

```suggestion
                assert_eq!(res.unwrap(), 0);
```

xermicus (approved):

tests/solana_tests/yul.rs
```diff
@@ -502,7 +502,10 @@ contract testing  {
     runtime.constructor(&[]);
     let returns = runtime.function("test_address", &[]).unwrap();
     let addr = returns.into_bigint().unwrap();
-    let b_vec = addr.to_bytes_be().1;
+    let mut b_vec = addr.to_bytes_be().1;
+    while b_vec.len() < 32 {
+        b_vec.insert(0, 0);
+    }
```
LucasSte, line 508: Why is this necessary?

seanyoung:
```rust
    let returns = runtime.function("test_address", &[]).unwrap();
    let addr = returns.into_bigint().unwrap();
    let b_vec = addr.to_bytes_be().1;
    let mut b_vec = addr.to_bytes_be().1;
```
The function returns `address` as `uint256`. If the highest bits are 0, then `addr.to_bytes_be().1` may not return 32 bytes.

seanyoung: I'll add a comment

tests/solana.rs
seanyoung: > I think a much more idiomatic way to do this would be to return an `Error` instead of passing a mutable pointer to some value representing an error.

Yes, I totally agree. Unfortunately the system call interface of the `solana_rbpf` crate takes `result` as a mutable pointer and returns unit, so there is nothing we can do about it,

I don't really know why that is done. If I were to hazard a guess, it is to avoid an allocation of a `Result<>` in the system call. I can't find an explanation in the source code.

> Completely unrelated to what the PR is trying to achieve, but I couldn't resist, sorry monkey So feel free to just take this as an improvement / suggestion that can be at a later time anyways.

It's a good point, I'll ask why it has been done this way.

xermicus: I see. Assuming they have their own fair reason, would be interesting to know :blush:

tests/solana.rs
seanyoung: This is good point, `unwrap()` is available on `ProgramResult` so I've updated these.

LucasSte (approved):

tests/solana.rs
seanyoung: I got an answer: That has to do with the calling convention and wether arguments and return values are passed as registers or through the stack. By using a reference we can explicitly control that, thus the ABI is more stable.

Issue #1083, opened 2022-11-26
Print runtime erros into the debug buffer.


xermicus:
Runtime errors could be mentioned in the debug buffer, if they occur. Possible examples:
- `assert(...)`,  `require(...)`, `revert(...)` statements
- Math overflow checks
- Things already mentioned in #971:
  - Invalid encodings in the input buffer
  - Invalid function selector
  - Div by zero
  - OOB access and empty array pop
 
Depends on: #1048

xermicus:
Closed by #1150

Issue #972, opened 2022-08-18
Milestone 5. Errors should pass source file and line number


xermicus:
This milestone is an improvement on #971 . When an error occurs, we would like to know what line number, column number and file name that error occurred at.

If Solidity is compiled with `-g` then each of the errors above should be amended with the correct line number, column number and source file name. This will give developers a good indication of where the problem occurred. This information should either be encoded in the `seal_debug_message()` or added to the `Err<>` metadata as described above.

Success is here that whenever an error occurs in Solidity code, the developer has easy access to where the error occurred in the code, to help debug the issue.

This feature should have tests at

https://github.com/hyperledger-labs/solang/tree/main/integration/substrate

Success means that a pull request has had all its review comments addressed and it is merged.

xermicus:
This information can go into the debug buffer. Would be nice to have #1048 first.

xermicus:
#1150

PR #1150, opened 2023-02-01
Reporting runtime errors

salaheldinsoliman:
runtime errors are reported alongside with file name and line:column number. this is enabled by ``--log-runtime-errors`


src/bin/solang.rs
```diff
@@ -384,6 +391,8 @@ fn compile(matches: &ArgMatches) {
 
     let log_api_return_codes = *matches.get_one::<bool>("LOGAPIRETURNS").unwrap();
 
+    let report_errors = matches.contains_id("REPORTERRORS");
```
seanyoung, line 394: This does not work. `--report-errors=false` will result in `true`. You need:

```rust
let report_errors = *matches.get_one::<bool>("REPORTERRORS").unwrap();
```

src/codegen/expression.rs
```diff
@@ -3060,3 +3121,21 @@ fn code(loc: &Loc, contract_no: usize, ns: &Namespace, opt: &Options) -> Express
 
     Expression::AllocDynamicBytes(*loc, Type::DynamicBytes, size.into(), Some(code))
 }
+
+pub(crate) fn report_error(
```
seanyoung, line 3125: This function adds a `Instr::DebugPrint {}`, so it should be called the same, e.g. `fn debug_print()` or so. Naming is in consistent - is it debug print or report error? There should be one name and the solang command line, `Instr` and this function should use the same term.

Also this needs to be called for `revert()`, `revert('this is my reason')`, and `require(expr, reason)` too.

seanyoung (changes requested):
Basic idea looks good, some comments though.

I am concerned that `revert('bla');` should report the `bla` reason as well. This expression can be  is a runtime defined value.

src/codegen/expression.rs
salaheldinsoliman: I think the name `Instr::ReportError` should be used instead of  `DebugPrint`, and the name of the function as well. thanks

for `revert`, there is this line: https://github.com/salaheldinsoliman/solang/blob/0353dbd3162174799b44c9fd6a3d81c7027b75d8/src/codegen/expression.rs#L1364 that if there is no expression within revert, it omits `revert encountered in line x`. So I will add the same for `require` as well

xermicus:
Closes #1083

src/bin/solang.rs
```diff
@@ -181,6 +181,13 @@ fn main() {
                             .short('g')
                             .long("generate-debug-info")
                             .hide(true),
+                    )
+                    .arg(
+                        Arg::new("REPORTERRORS")
+                            .help("Enable printing runtime errors")
+                            .short('d')
+                            .long("report-errors")
```
xermicus, line 189: Can we name this `--log-runtime-errors` instead? We already have `--log-api-return-codes`, and this would make it consistent. @seanyoung WDYT?

integration/substrate/asserts.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('Deploy asserts contract and test', () => {
         expect(res0.output?.toJSON()).toEqual(1);
 
         let res1 = await query(conn, alice, contract, "testAssertRpc");
-        expect(res1.result.toHuman()).toEqual({ "Err": { "Module": { "error": "0x0b000000", "index": "8" } } });
+        expect(res1.result.toHuman()).toEqual({ "Err": { "Module": { "error": "0x0b000000", "index": "7" } } });
```
xermicus, line 32: I pinned the contract node image recently, so this should not change out of nowhere?

xermicus (changes requested):
Amazing, looking forward to getting this merged!

I will give it a proper review tomorrow.

xermicus:
@salaheldinsoliman #1154 will fix substrate on the CI

integration/substrate/asserts.spec.ts
salaheldinsoliman: I was running tests on a different image locally, sorry for that

src/bin/solang.rs
salaheldinsoliman: Nice, ``--log-runtime-errors`` is more descriptive

src/codegen/cfg.rs
```diff
@@ -184,7 +207,14 @@ pub enum Instr {
         data_len: Expression,
     },
     /// Return a code at the end of a function
-    ReturnCode { code: ReturnCode },
+    ReturnCode {
+        code: ReturnCode,
+    },
+
+    DebugPrint {
```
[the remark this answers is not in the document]
salaheldinsoliman: will do that

src/codegen/expression.rs
salaheldinsoliman: Did you mean replacing Instr::Print with Instr::DebugPrint All together? @seanyoung

seanyoung: > Did you mean replacing Instr::Print with Instr::DebugPrint All together? @seanyoung

I think that is an even better idea!

seanyoung: Well, or replacing `Instr::DebugPrint` with `Instr::Print`. @xermicus  WDYT

src/bin/solang.rs
seanyoung: Yes, very nice.

integration/substrate/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "tsc; ts-mocha -t 20000 *.spec.ts",
-    "build": "parallel solang compile -v --target substrate ::: *.sol test/*.sol",
+    "build": "parallel solang compile -v --target substrate --report-errors=true --math-overflow=true::: *.sol test/*.sol",
```
xermicus, line 8: Please always make sure it works locally first

```suggestion
    "build": "parallel solang compile -v --target substrate --report-errors=true --math-overflow=true ::: *.sol test/*.sol",
```

src/emit/expression.rs
```diff
@@ -1919,3 +1941,49 @@ fn runtime_cast<'a>(
...
+    bin: &Binary<'a>,
+    ns: &Namespace,
+    error: &String,
+    loc: Option<Loc>,
+) -> BasicValueEnum<'a> {
+    let error_with_line = if let Some(loc) = loc {
+        if let Loc::Codegen = loc {
+            format!("{error} in  codegen")
+        } else {
+            let file_no = loc.file_no();
```
xermicus, line 1973: This doesn't cover all cases, see `UniswapV2Pair.sol` for an example where the location is neither `Codegen` nor `File`, which crashes the compiler.

xermicus (changes requested):
I was trying out your branch locally but found a couple problems. The new `--report-errors` flag crashes the compiler for one of the substrate integration test (`UniswapV2Pair.sol`), see my other comment below.

Can you please merge main to fix the CI as well?

integration/substrate/asserts.spec.ts
xermicus: No problem

integration/solana/simple.spec.ts
```diff
@@ -50,8 +50,8 @@ describe('Simple solang tests', function () {
 
         res = await contract.functions.op_u64(0, 1000, 4100, { simulate: true });
         expect(Number(res.result)).toBe(5100);
-        res = await contract.functions.op_u64(1, 1000, 4100, { simulate: true });
-        expect(Number(res.result)).toBe(18446744073709548516); // (2^64)-18446744073709548516 = 3100
+        res = await contract.functions.op_u64(1, 4100, 1000, { simulate: true });
+        expect(Number(res.result)).toBe(3100);
```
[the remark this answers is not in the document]
salaheldinsoliman: Those conditions fail when math overflow flag is enabled

src/codegen/cfg.rs
```diff
@@ -185,6 +192,9 @@ pub enum Instr {
     },
     /// Return a code at the end of a function
     ReturnCode { code: ReturnCode },
+    /// Report the runtime error before reaching aborting execution (AssertFailure).
+    /// If Instr::Print is used, the error string and location must be turned to an Expression first.
+    ReportError { string: String, loc: Loc },
```
xermicus, line 197: Now we have `Instr::Print` and `Instr::ReportError`, as well as `fn print()` and `fn debug_print()` in emit. But they do the same thing, it is unlikely we ever going to have two possible locations where something could be printed into (at least for the substrate target(. Looking at the implementation, it is not clear to me why `Instr::ReportError` is needed, the two isntructions do almost exactly the same. Also I feel like this will make it awkward for formatting the debug buffer. Can you share your thoughts behind having a dedicated instruction for this?

src/codegen/expression.rs
xermicus: @seanyoung sorry somehow I missed this comment earlier on :facepalm: I agree, ideally this can be replaced with `Instr::Print`. Just glanced at the code, unless I missed something this should be easy. @salaheldinsoliman see also the [comment below](https://github.com/hyperledger/solang/pull/1150#discussion_r1099048822).

src/codegen/cfg.rs
xermicus: Other than that, this doesn't look to bad @salaheldinsoliman! I really like the many test cases you added :medal_sports:

xermicus:
Closes #1083 
Closes #972

tests/substrate_tests/errors.rs
```diff
@@ -0,0 +1,286 @@
...
+    runtime.vm.value = 1;
+    runtime.function_expect_failure("dont_pay_me", Vec::new());
+    assert_eq!(runtime.printbuf, "non payable function recieved value");
+}
+
+/*fn debug_buff () {
+
+
+    let mut runtime = build_solidity_with_options(src, math_overflow_flag, log_api_return_codes, log_runtime_errors)
+}*/
```
xermicus, line 286:
```suggestion
```

src/codegen/cfg.rs
salaheldinsoliman: My rationale is more readability: If Instr::Print is used, either an extra parameter should be added (Option<String>) maybe or convert its parameter to an enum having an expression and string variations. Also ``loc`` has to be passed along to be concatenated to the error. I think maintenance in this case would be harder... 
I have thought about this before starting, and read this nice [thread](https://softwareengineering.stackexchange.com/questions/194504/should-my-code-be-dry-or-readable-if-it-cant-be-both) and reached my final verdict of making an extra instruction

xermicus: I see, however I don't think an extra instruction with code added to `emit` makes the code more readable here. If you move the logic of `fn print_runtime_error()` into codegen, you don't have to deal with passing `String`s and `Loc`s around at all.

salaheldinsoliman: I considered making a codegen expression from the error string and loc, but this would make the CFG look a bit complicated if someone wants to debug his code using ``--emit cfg``

salaheldinsoliman: Other than that CFG matter, I think removing Instr::ReportError all together is much better.

xermicus: How is this going to add substantially to the CFG? And if this should ever be an issue, just don't use `--log-runtime-errors` to debug the CFG, problem solved.

The `print_runtime_error` can exist in emit without having a  `ReportError` instruction. It could also be made into a callable function, then you should be able to refactor this into codegen and just call it in emit. 

I'm still not convinced introducing an additional `ReportError` instruction is justified. @seanyoung WDYT?

salaheldinsoliman: I had another look today and you are correct. repetition here can be avoided easily if a string is converted to an expression. a programmer debugging the CFG will now that this is inserted by the compiler because of (``Loc::Codegen``). Thanks!
I will push a commit removing ReportError all together and move ``print_runtime_error`` to codegen.

integration/solana/debug_buffer.sol
```diff
@@ -0,0 +1,154 @@
+import 'solana';
+
+@program_id("PriZbiyLjDtLWb2AcExcE7pxCkesSJzb7Z1x52tUYXt")
```
seanyoung, line 3: You can remove this line and remove the file `Printer.key`. The file will be generated for each run, no need to check the file in.

integration/solana/debug_buffer.sol
```diff
@@ -0,0 +1,154 @@
...
+        bytes smol_buf = new bytes(num);
+
+        //bytes32 b32 = new bytes(num);
+        bytes32 b32 = bytes32(smol_buf);
+        return b32;
+    }
+}
+
+
+@program_id("Cre7AzxtwSxXwU2jekYtCAQ57DkBhY9SjGDLdcrwhAo6")
```
seanyoung, line 127: You can remove this line and `Creature.key`

integration/solana/debug_buffer.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    it('Prints runtime errors', async function () {
+  
+        try {
+            let res = await program.methods.setStorageBytes().accounts({ dataAccount: storage.publicKey }).simulate();
+        }
+        catch (e: any) {
+            const logs = e.simulationResponse.logs;
+            expect(logs).toContain('Program log: set storage index out of bounds in  file: 1, line: 42,10');
```
seanyoung, line 27: The formatting can be better. We need the file name here. I think `filename:line-number:column-number` would be much better. WDYT @xermicus

src/codegen/expression.rs
```diff
@@ -1330,7 +1344,39 @@ fn require(
...
-                cfg.add(vartab, Instr::Print { expr });
+                let expr_vec = args.get(1);
+                let final_string = if let ast::Expression::AllocDynamicBytes {
+                    init: Some(vec),
+                    ..
+                } = expr_vec.as_ref().unwrap()
+                {
+                    String::from_utf8(vec.to_vec()).unwrap()
+                } else {
+                    "".to_string()
```
seanyoung, line 1355: This needs fixing. If the revert argument is an expression, then the print should be constructed using the same method as `"{} require condition fail".format(expr)`. 

```rust
let expr = Expression::FormatString(loc, vec![
(FormatString::StringLiteral, Expression::StringLiteral(loc, "foo.sol:102:4: revert failed with reason: ")), 
(FormatString::Default, expr))
]);

Instr::Print { loc, expr }
```

src/codegen/expression.rs
```diff
@@ -1348,10 +1394,46 @@ fn revert(
...
         .map(|s| expression(s, cfg, contract_no, func, ns, vartab, opt));
+    if expr.is_some() {
+        let expr_vec = args.get(0);
+        let final_string = if let ast::Expression::AllocDynamicBytes {
+            init: Some(vec), ..
+        } = expr_vec.as_ref().unwrap()
+        {
+            String::from_utf8(vec.to_vec()).unwrap()
+        } else {
+            "".to_string()
```
seanyoung, line 1410: Same here

src/codegen/expression.rs
```diff
@@ -3060,3 +3197,45 @@ fn code(loc: &Loc, contract_no: usize, ns: &Namespace, opt: &Options) -> Express
...
+    let expr = ast::Expression::AllocDynamicBytes {
+        loc: Loc::Codegen,
+        ty: Type::String,
+        length: Box::new(len_expr),
+        init: Some(string.into_bytes()),
+    };
+
+    expression(&expr, cfg, contract_no, func, ns, vartab, opt)
+}
+pub(crate) fn debug_print(
```
seanyoung, line 3225: This is not a great name. How about `log_runtime_error`?

seanyoung (changes requested):
Good stuff so far!

integration/solana/debug_buffer.spec.ts
xermicus: ACK

src/codegen/expression.rs
```diff
@@ -3060,3 +3197,45 @@ fn code(loc: &Loc, contract_no: usize, ns: &Namespace, opt: &Options) -> Express
...
+    opt: &Options,
+    func: Option<&Function>,
+) -> Expression {
+    let len_expr = ast::Expression::NumberLiteral {
+        loc: Loc::Codegen,
+        ty: Type::Uint(32),
+        value: BigInt::from(string.len()),
+    };
+
+    let expr = ast::Expression::AllocDynamicBytes {
```
xermicus, line 3216: Since the string is known at compile time, can't we use `Expression::BytesLiteral` here?

src/codegen/mod.rs
```diff
@@ -1342,3 +1344,28 @@ impl From<&ast::Builtin> for Builtin {
         }
     }
 }
+
+pub(super) fn print_runtime_error(ns: &Namespace, error: &String, loc: Option<Loc>) -> String {
```
xermicus, line 1348: Nit: better to use  a `&str` ref instead, otherwise I don't think this will work with static str refs for example.

seanyoung:
Fixes  #1152

integration/solana/debug_buffer.sol
salaheldinsoliman: This contract is being created by another contract using ``new()``, so it should have a program_id I think

src/codegen/expression.rs
salaheldinsoliman: I think Bytes literal cannot be passed to a print instruction, as bytes literal is converted to an int and printed that way. Instead, it should be wrapped in a Expression::FormatString

xermicus: @seanyoung probably knows best

seanyoung: `Expression::BytesLiteral()` is for `bytes4 foo = hex"41424344";`, not for strings. Emit will replace them with integers.

src/codegen/mod.rs
```diff
@@ -1342,3 +1344,23 @@ impl From<&ast::Builtin> for Builtin {
...
     }
 }
+
+pub(super) fn print_runtime_error(ns: &Namespace, error: &str, loc: Option<Loc>) -> String {
+    if let Some(loc) = loc {
+        match loc {
+            Loc::File(..) => {
+                let file_no = loc.file_no();
+                let curr_file = &ns.files[file_no];
+                let filename = &curr_file.path.file_name().unwrap();
```
seanyoung, line 1354: `.display()`

src/codegen/mod.rs
```diff
@@ -1342,3 +1344,23 @@ impl From<&ast::Builtin> for Builtin {
...
+                let (line_no, offset) = curr_file.offset_to_line_column(loc.start());
+                format!("{} in {:?}:{}:{}", error, filename, line_no + 1, offset)
```
seanyoung, line 1356: do not use `{:?}`

src/codegen/expression.rs
```diff
@@ -3060,3 +3165,31 @@ fn code(loc: &Loc, contract_no: usize, ns: &Namespace, opt: &Options) -> Express
...
+                Loc::Codegen,
+                Type::Bytes(string.as_bytes().len() as u8),
+                string.as_bytes().to_vec(),
+            ),
+        )],
+    )
+}
+pub(crate) fn log_runtime_error(
+    report_error: bool,
+    reason: String,
```
seanyoung, line 3184: `reason: &str`

src/emit/functions.rs
```diff
@@ -105,6 +105,12 @@ pub(super) fn abort_if_value_transfer<'a, T: TargetRuntime<'a> + ?Sized>(
 
     binary.builder.position_at_end(abort_value_transfer);
 
+    target.log_runtime_error(
+        binary,
+        "non payable function recieved value".to_string(),
```
seanyoung, line 110: received

might as well include the function name

src/emit/substrate/target.rs
```diff
@@ -1183,6 +1209,12 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
 
             binary.builder.position_at_end(bail_block);
 
+            self.log_runtime_error(
+                binary,
+                "contract creation failed".to_string(),
```
seanyoung, line 1214: we should probably include the error code. What do you think @xermicus

integration/solana/debug_buffer.sol
seanyoung: Good point, you are right

src/bin/solang.rs
```diff
@@ -185,6 +185,12 @@ fn main() {
                             .long("generate-debug-info")
                             .action(ArgAction::SetTrue)
                             .hide(true),
+                    )
+                    .arg(
+                        Arg::new("LOGRUNTIMEERRORS")
+                            .help("Log runtime errors in the environment")
+                            .long("log-runtime-errors")
+                            .action(ArgAction::SetTrue),
```
seanyoung, line 193: This should probably be on by default, and a cli option to turn it off. What do you think @xermicus

seanyoung (changes requested):

src/bin/solang.rs
salaheldinsoliman: I think it should be on by default for ``revert(reason)`` and ``require(bool, reason)``

src/emit/substrate/target.rs
salaheldinsoliman: Isn't the error code logged by ``log_api_return_values`` ?

src/codegen/mod.rs
```diff
@@ -1342,3 +1344,22 @@ impl From<&ast::Builtin> for Builtin {
...
+    if let Some(loc) = loc {
+        match loc {
+            Loc::File(..) => {
+                let file_no = loc.file_no();
+                let curr_file = &ns.files[file_no];
+                let (line_no, offset) = curr_file.offset_to_line_column(loc.start());
+                format!("{} in {}:{}:{}", error, curr_file, line_no + 1, offset)
+            }
+            Loc::Codegen => {
+                format!("{error} in  codegen")
```
LucasSte, line 1358: Reporting this is a little strange. Every construct in codegen can be traced to something developers have types, except for the dispatch. A work for the future is to ensure the condegen expression are traced back to a location in sema.

src/emit/expression.rs
```diff
@@ -1930,3 +1951,21 @@ fn runtime_cast<'a>(
...
+    let size = bin.context.i32_type().const_int(input.len() as u64, false);
+
+    let elem_size = bin
+        .llvm_type(&elem, ns)
+        .size_of()
+        .unwrap()
+        .const_cast(bin.context.i32_type(), false);
+
+    let init = &Option::Some(input.as_bytes().to_vec());
+    bin.vector_new(size, elem_size, init.as_ref()).into()
```
LucasSte, line 1970:
```suggestion
    let init = Option::Some(input.as_bytes().to_vec());
    bin.vector_new(size, elem_size, init.as_ref()).into()
```
I think the ampersand is not needed. You are already calling `as_ref` for init.

src/emit/mod.rs
```diff
@@ -253,6 +267,7 @@ pub trait TargetRuntime<'a> {
     ) -> BasicValueEnum<'a>;
 
     /// Calls constructor
+    #[allow(clippy::too_many_arguments)]
```
LucasSte, line 270: Clippy is right.

src/emit/solana/target.rs
```diff
@@ -187,6 +199,12 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
             .build_conditional_branch(in_range, set_block, bang_block);
 
         binary.builder.position_at_end(bang_block);
+        self.log_runtime_error(
+            binary,
+            "set storage index out of bounds".to_string(),
```
LucasSte, line 204:
```suggestion
            "storage index out of bounds".to_string(),
```
The message was hard to understand for someone who does not know `Instr::SetStorageBytes`.

src/emit/solana/target.rs
```diff
@@ -2399,4 +2433,24 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
...
+        &self,
+        bin: &Binary,
+        reason_string: String,
+        reason_loc: Option<Loc>,
+        ns: &Namespace,
+    ) {
+        if !bin.options.log_runtime_errors {
+            return;
+        }
+        emit_context!(bin);
```
LucasSte, line 2447: This macro should only be used in Substrate. It is not defined for Solana.

src/emit/substrate/target.rs
```diff
@@ -431,6 +443,12 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
             .build_conditional_branch(in_range, retrieve_block, bang_block);
 
         binary.builder.position_at_end(bang_block);
+        self.log_runtime_error(
+            binary,
+            "set storage index out of bounds".to_string(),
```
LucasSte, line 448:
```suggestion
            "storage index out of bounds".to_string(),
```

LucasSte (changes requested):

src/codegen/expression.rs
```diff
@@ -2756,6 +2853,14 @@ fn array_subscript(
     );
 
     cfg.set_basic_block(out_of_bounds);
+    log_runtime_error(
+        opt.log_runtime_errors,
+        "array out of bounds",
```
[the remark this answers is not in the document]
LucasSte: I couldn't find a test for this.

src/emit/mod.rs
LucasSte: You don't need to fix this. This should go in a larger refactoring.

src/bin/solang.rs
xermicus: As discussed in todays daily:
- Should behave the same as `--log-api-return-codes` (disabled by default)
- Should enable or disable everything related to runtime error reporting introduced in this PR, so it can be omitted completely for production environments

src/codegen/expression.rs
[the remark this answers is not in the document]
salaheldinsoliman: https://github.com/salaheldinsoliman/solang/blob/3aeb145c787f96d118fc529870589653f61efc29/tests/substrate_tests/errors.rs#L117
there is similar test in ``tests/solana_tests/runtime_errors.rs``, as well as Solana and Substrate integration tests

src/codegen/mod.rs
salaheldinsoliman: In that case, do you think only the error should be reported ?

src/emit/expression.rs
salaheldinsoliman: you are right, good catch

src/emit/solana/target.rs
salaheldinsoliman: sorry, silly copy and paste mistake

src/codegen/mod.rs
xermicus: `codegen` won't mean anything to a regular user of the compiler. Reporting only the error would be the best option here. Lucas is right, ideally this should point out where in the source the problem is, we should fix this at some later point

xermicus (changes requested):
After this we should be pretty close :+1:

integration/solana/debug_buffer.sol
```diff
@@ -0,0 +1,150 @@
+import 'solana';
```
seanyoung, line 1: The debug buffer is a substrate term. It does not make any sense for Solana. How about calling this file runtime_errors (both sol and .spec.ts file)?

integration/solana/debug_buffer.spec.ts
```diff
@@ -0,0 +1,148 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { Connection, Keypair, PublicKey, sendAndConfirmTransaction, SystemProgram, Transaction } from '@solana/web3.js';
+import expect from 'expect';
+import { loadContract } from './setup';
+import { Program, Provider, BN } from '@project-serum/anchor';
+import path = require("path")
+
+describe('DebugBuffer', function () {
```
seanyoung, line 9: Runtime Errors

integration/solana/debug_buffer.sol
```diff
@@ -0,0 +1,150 @@
...
+
+
+contract Printer {
```
seanyoung, line 4: RuntimeErrors

integration/solana/debug_buffer.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    let storage: Keypair;
+    let payer: Keypair;
+    let provider: Provider;
+
+    before(async function () {
+        ({ program, storage, payer, provider } = await loadContract('Printer'));
+    });
+
+    it('Prints runtime errors', async function () {
+        let absolutePath = path.resolve("debug_buffer.sol");
```
seanyoung, line 22: I'm not sure we need the whole path. Maybe just the file name. What do you think @xermicus?

We need reproducable builds. If the path depends on this then that's a pain

src/emit/mod.rs
seanyoung: The salt/seed/value create contract arguments should be collected in a struct and passed as one

seanyoung (changes requested):

integration/solana/debug_buffer.sol
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    // non payable function dont_pay_me received value
+    function dont_pay_me() public {}
+
+    function pay_me() public payable {}
+
+    function i_will_revert() public {
+        revert();
+    }
```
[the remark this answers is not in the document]
salaheldinsoliman: makes sense, this is a copy paste error

integration/solana/debug_buffer.spec.ts
xermicus: Good point, I had the same thought (alas the full path is not necessary), but reproducible builds didn't came to my mind. But with that in mind, I think it would better to just use only the file name. Otherwise we have a CLI flag that breaks reproducible builds, which is something we should be keen to avoid if possible.

src/emit/mod.rs
seanyoung: @salaheldinsoliman thanks for doing that, I actually meant to say this can be done in another PR, but  looks great - thank you! :pray:

tests/substrate_tests/mod.rs
```diff
@@ -10,6 +10,7 @@ mod arrays;
 mod builtins;
 mod calls;
 mod contracts;
+pub mod errors;
```
seanyoung, line 13: does not need to be `pub`

seanyoung (approved):
Great stuff, thanks @salaheldinsoliman

src/emit/mod.rs
salaheldinsoliman: @seanyoung no worries

src/codegen/mod.rs
```diff
@@ -1342,3 +1344,25 @@ impl From<&ast::Builtin> for Builtin {
         }
     }
 }
+
+pub(super) fn print_runtime_error(ns: &Namespace, error: &str, loc: Option<Loc>) -> String {
```
xermicus, line 1348: The function name is confusing, as it doesn't have anything to do with "runtime error printing"; additionally there is now already a `fn log_runtime_error` which adds up confusion. Can you please  rename it to `error_msg_with_loc` or something like that?

src/codegen/expression.rs
```diff
@@ -3064,3 +3174,31 @@ fn code(loc: &Loc, contract_no: usize, ns: &Namespace, opt: &Options) -> Express
...
+            FormatArg::StringLiteral,
+            Expression::BytesLiteral(
+                Loc::Codegen,
+                Type::Bytes(string.as_bytes().len() as u8),
+                string.as_bytes().to_vec(),
+            ),
+        )],
+    )
+}
+pub(crate) fn log_runtime_error(
```
xermicus, line 3191:
```suggestion

pub(crate) fn log_runtime_error(
```

src/emit/substrate/mod.rs
```diff
@@ -219,7 +219,7 @@ impl SubstrateTarget {
 
         // after copying stratch, first thing to do is abort value transfers if constructors not payable
         if abort_value_transfers {
-            abort_if_value_transfer(self, binary, function, ns);
+            abort_if_value_transfer(self, binary, function, ns, "constructor");
```
xermicus, line 222: I doubt this is correct. `public_function_prelude` is used during emitting the `deploy` (for constructors) and the `call` (for messages) function. The perfect solution would be to have the function name here. But for me for now it doesn't have to be perfect, just not wrong. Can you please just distinguish between "deploy" and "call" here  (can be added as a &'static str argument to `fn public_function_prelude`).

src/emit/substrate/target.rs
xermicus: Ideally this would include the error code, but doesn't have to be perfect for me. @salaheldinsoliman is also right, while not super convenient, it is already available.

xermicus (changes requested):
Sorry @salaheldinsoliman I found a few more things. But they are super minor, would be great if you could fix them.

src/codegen/mod.rs
salaheldinsoliman: Good point, I think ``format_error_msg_with_loc`` is nice

src/emit/math.rs
```diff
@@ -590,7 +607,10 @@ pub(super) fn build_binary_op_with_overflow_check<'a, T: TargetRuntime<'a> + ?Si
     right: IntValue<'a>,
     op: BinaryOp,
     signed: bool,
+    ns: &Namespace,
+    loc: Loc,
 ) -> IntValue<'a> {
+    emit_context!(bin);
```
LucasSte, line 613: This macro exists only for Substrate and it is not necessary here.

src/emit/mod.rs
```diff
@@ -263,13 +283,13 @@ pub trait TargetRuntime<'a> {
...
-        salt: Option<IntValue<'b>>,
-        seeds: Option<(PointerValue<'b>, IntValue<'b>)>,
+        contract_args: ContractArgs<'b>,
         ns: &Namespace,
+        loc: Loc,
     );
 
     /// call external function
+    #[allow(clippy::too_many_arguments)]
     fn external_call<'b>(
```
LucasSte, line 293: I think we can use `struct ContractArgs` here too. What do you think @seanyoung?

LucasSte (changes requested):

src/emit/mod.rs
seanyoung: Ideally yes, however we're trying to get the PR over the line so it can be merged, so let's leave this up to @salaheldinsoliman

xermicus (approved):
Nice, thanks!

We probably should mention this in our docs too, but that can go into a follow-up PR as well.

LucasSte (approved):

seanyoung:
:tada:

PR #1156, opened 2023-02-03
Retry uploading of program code after failure

seanyoung:
Uploading of program code intermittently fails, which causes CI tests to fail. Improve this a little bit.

### DIFF
--- a/integration/solana/setup.ts
+++ b/integration/solana/setup.ts
@@ -117,7 +117,7 @@ async function setup() {
 
             console.log(`Loading ${name} at ${program.publicKey}...`);
             const program_so = fs.readFileSync(file);
-            for (; ;) {
+            for (let retries = 5; retries > 0; retries -= 1) {
                 try {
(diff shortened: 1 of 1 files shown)


integration/solana/setup.ts
```diff
@@ -117,12 +117,29 @@ async function setup() {
 
             console.log(`Loading ${name} at ${program.publicKey}...`);
             const program_so = fs.readFileSync(file);
-            await BpfLoader.load(connection, payer, program, program_so, BPF_LOADER_PROGRAM_ID);
+            for (; ;) {
```
xermicus, line 120: Could set a fixed number of attempts? It feels to me like if after like 5 consecutive attempts it is still failing, something else will be the problem, and then we don't just run into a timeout for the test.

seanyoung: good point

seanyoung: fixed

xermicus (approved):

Issue #1159, opened 2023-02-05
Cannot parse named parameter for mappings (introduced in 0.8.18)


leekt:
**Describe the bug**

As of solidity 0.8.18, named parameter for mapping has been introduced, but solang-parser cannot parse is properly

https://blog.soliditylang.org/2023/02/01/solidity-0.8.18-release-announcement/

```solidity
contract C {
    mapping(address user => uint256) public balances;
    mapping(address => mapping(address spender => uint256)) public allowance;
}
```

**To Reproduce**

paste the code below to `solang-parser/src/test.rs` and run test

```rust
#[test]
fn parse_mapping_with_named_parameter() {
    let src = r#"
contract C {
    mapping(address user => uint256) public balances;
    mapping(address => mapping(address spender => uint256)) public allowance;
}
    "#;
    let (actual_parse_tree, _) = crate::parse(src, 0).unwrap();
    assert_eq!(actual_parse_tree.0.len(), 1);
}
```

**Expected behavior**

should be able to parse it correctly

**Screenshots**

**Hyperledger Solang version**
v0.2.1

**Include the complete solidity source code**
```solidity
contract C {
    mapping(address user => uint256) public balances;
    mapping(address => mapping(address spender => uint256)) public allowance;
}
```

**Additional context**

I'm working on my version of fix but I'm not lexer genius so i haven't made a PR but here is my version for reference : https://github.com/leekt/solang

seanyoung:
What is this name used for? I'm looking the solidity source code and I can't see why they added this.

leekt:
I think this is more of an dev-friendly thing
This will generate friendly abi file that describes the parameter of mapping variables

Ex these two will generate same abi
![image](https://user-images.githubusercontent.com/15259621/217320404-4aec9374-3b07-4434-8302-83f168107380.jpeg)
Source : https://twitter.com/jtriley_eth/status/1620842058364878850

PR #1160, opened 2023-02-07
Allow mappings to be named (solc-0.8.18 feature)

seanyoung:
Fixes https://github.com/hyperledger/solang/issues/1159


leekt:
I think it should also consider named key value

```
mapping(address user => uint256 timestamp) public lastUpdated
```

something like this

seanyoung:
> I think it should also consider named value
> 
> ```
> mapping(address user => uint256 timestamp) public lastUpdated
> ```

Thanks for pointing that out - that was an oversight. Fixed now.

seanyoung:
Note that the type contract variables was not included in the ast dot file. That is fixed now, but it does make the diff a lot bigger.

docs/language/types.rst
```diff
@@ -391,6 +391,11 @@ Mappings are declared with ``mapping(keytype => valuetype)``, for example:
 .. include:: ../examples/mapping_type.sol
   :code: solidity
 
+Mappings may have a name for the key or the value, for example:
+``mapping(address owner => uint64 balance)``. The names are used in the metadata of
+the contract. If the mapping is public, then the accessor function will have
```
LucasSte, line 396:
```suggestion
the contract. If the mapping is public, the accessor function will have
```

LucasSte (approved):

docs/language/types.rst
seanyoung: Thanks

xermicus (approved):

PR #1164, opened 2023-02-08
Add --contract and --output-meta CLI flags

seanyoung:
These are needed for the anchor cli tools


docs/running.rst
```diff
@@ -47,8 +47,15 @@ Options:
   is ignored for any other target.
 
 -o, \-\-output *directory*
-  This option takes one argument, which is the directory where output should
-  be saved. The default is the current directory.
+  Sets the directory where the output should be saved. This defaults to the current directory if not specifed.
```
LucasSte, line 50: What is the current directory? That of the Solidity source file or that of the Solang binary?

docs/running.rst
```diff
@@ -47,8 +47,15 @@ Options:
...
+
+\-\-output\-meta *directory*
+  Sets the directory where metadata should be saved. For Solana, the metadata is the Anchor IDL file,
+  and for Substrate this is the .contract file. If not specified, the directory specified by ``--output``
+  is used, and if that is not set either, the current directory is used.
```
LucasSte, line 55:
```suggestion
  and, for Substrate, the .contract file. If this option is not set, the directory specified by ``--output``
  is used, and if that is not set either, the current directory is used.
```
I think you should clarify which directory is the the current.

tests/cli.rs
```diff
@@ -32,4 +62,23 @@ fn create_output_dir() {
...
+        "--contract",
+        "flapper,flipper", // not flipper
+        "--output",
+        "tests/create_me",
+    ])
+    .assert()
+    .failure();
+
+    remove_file("tests/create_me/flipper.json").unwrap();
+    remove_file("tests/create_me/flipper.so").unwrap();
```
LucasSte, line 82: Are we only building the contracts we found? I though we would ignore everything if any of the contracts specified could not be found.

LucasSte (changes requested):

docs/running.rst
seanyoung: Neither, a file does not have current directory, only a process. How about calling it the current working directory?

src/bin/solang.rs
```diff
@@ -416,33 +431,87 @@ fn compile(matches: &ArgMatches) {
...
+            errors = true;
+        }
+    }
+
+    if let Some("ast-dot") = matches.get_one::<String>("EMIT").map(|v| v.as_str()) {
+        exit(0);
+    }
+
+    // ensure we have any contracts
+    if !errors && !namespaces.iter().any(|(ns, _)| !ns.contracts.is_empty()) {
```
xermicus, line 472:
```suggestion
    if !errors && namespaces.iter().all(|(ns, _)| ns.contracts.is_empty()) {
```

This is written off my head (didn't test), but it would make it much easier to grasp

tests/cli.rs
```diff
@@ -32,4 +65,22 @@ fn create_output_dir() {
...
+
+    let mut cmd = Command::cargo_bin("solang").unwrap();
+
+    cmd.args([
+        "compile",
+        "examples/flipper.sol",
+        "--target",
+        "solana",
+        "--contract",
+        "flapper,flipper", // not flipper
```
xermicus, line 77: I'm confused by this comment

tests/cli.rs
```diff
@@ -19,6 +22,36 @@ fn create_output_dir() {
     .success();
 
     File::open("tests/create_me/flipper.json").expect("should exist");
```
xermicus, line 24: Can we please use something like the `tempdir` from [tempfile](https://crates.io/crates/tempfile) instead? If the compiler crashes during the test, e.g. because of tinkering in codegen or emit. Then these files just sit there and show up in git status and are easily committed by accident (`git add tests` :eyes:). Where as with using `tempfile` worst case the files just end up hang around in some location I don't see and don't care about.

As a bonus, with `tempfile` we don't have to deal with deletion of these files manually.

tests/cli.rs
seanyoung: Fixed

src/bin/solang.rs
```diff
@@ -416,33 +431,87 @@ fn compile(matches: &ArgMatches) {
...
+    }
+
+    // ensure we all the requested contracts
+    let not_found: Vec<_> = contract_names
+        .iter()
+        .filter(|name| {
+            !namespaces
+                .iter()
+                .flat_map(|(ns, _)| ns.contracts.iter())
+                .any(|contract| **name == contract.name)
```
xermicus, line 484: This freely mixes contracts from different files. What if I compile multiple files with oversecting contract names, but only want some contract that from a specific file. If this is fine or something you intended, this should go into the CLI help. Otherwise, I think the correct algorithm to do this would be to build a HashSet where we insert keys as follows:
- `"filename=contractname"` 
- `"contractname"

Then the user can do `--contracts foo,bar.sol=bar` if he want's the `foo` contract from every namespace but the `bar` contract only from `bar.sol`

xermicus (changes requested):

tests/cli.rs
seanyoung: We are testing directory creation as well as file creation. The output is ignored by `.gitignore` so nothing should go into `git status` or `git add tests/`.

tests/cli.rs
seanyoung: :rofl: yes, that's is not very clear. How about "not just flipper"

src/bin/solang.rs
seanyoung: If the user has multiple files with the same contract name, then that is just a recipe for total disaster. How will this work in a dev environment like vscode?

I think the right solution is to detect this problem and error about it.

tests/cli.rs
xermicus: Ok this is siightly defused, the reason why I saw this before is that the `.json` files were not in `.gitignore`.

However I still think it would the right approach to use a `tempdir` as the directory creation can be tested inside there as well. Not a fan of abusing our `tests` dir as a `/tmp`

src/bin/solang.rs
xermicus: > If the user has multiple files with the same contract name, then that is just a recipe for total disaster

I fully agree with you on this.

> How will this work in a dev environment like vscode?

What do you mean with this?

> I think the right solution is to detect this problem and error about it.

Do you mean in general, or just when the `--contract` flag is set? Here, `solc` is happily compiling the exact same contract with the same name but in different files:

```bash
solc --bin /tmp/*.sol 2>/dev/null

======= /tmp/a.sol:A =======
Binary:
6080604052348015600f57600080fd5b50606d80601d6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c80630dbe671f14602d575b600080fd5b60336035565b005b56fea2646970667358221220a816946390647c30b12c5d3b89636422c282d0132869c61c082513a7dfa981fa64736f6c63430008120033

======= /tmp/b.sol:A =======
Binary:
6080604052348015600f57600080fd5b50606d80601d6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c80630dbe671f14602d575b600080fd5b60336035565b005b56fea2646970667358221220a433fe66d6602c3827ae7e0060d4539266d520fd0f1a61cc952ab89aabf7d1bf64736f6c63430008120033
```

I think if we have to rely on contract names being unique per file name, we have a problem.

tests/cli.rs
seanyoung: I see your point, I don't mind adding tempdir for this.

My personal issue with tempdir is that it not really a panacea, if your program exits with e.g. `std::process::exit(1)` you still have a temp directory with garbage, but now its in /tmp. There are some tricks you can do with creating a single files without a name but that does not help for our case, because we need to test a directory.

seanyoung: ok, the test now uses the `tempfile` crate (the tempdir crate docs says to use tempfile instead). It's actually pretty nice and now the test is shorter in code, because the cleanup is much simpler, and now in nearly all cases the temp directory is gone. Definitely an improvement, thanks.

src/bin/solang.rs
seanyoung: > > How will this work in a dev environment like vscode?
> 
> What do you mean with this?

I was thinking of anchor because that's what I've been working on. So you have a bunch of solidity files on one hand, 
and then you have some client code/test code in written in typescript. Where do you specify that you want contract `A` from file X.sol but not Y.sol? Would it not be better if the environment said "error: contract A defined in X.sol and Y.sol".
 
> > I think the right solution is to detect this problem and error about it.
> 
> Do you mean in general, or just when the `--contract` flag is set?

I was thinking in general. This can always be a problem. Having said that, I don't know what can be done.

> Here, `solc` is happily compiling the exact same contract with the same name but in different files:
> 
> ```shell
> solc --bin /tmp/*.sol 2>/dev/null
> 
> ======= /tmp/a.sol:A =======
> Binary:
> 6080604052348015600f57600080fd5b50606d80601d6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c80630dbe671f14602d575b600080fd5b60336035565b005b56fea2646970667358221220a816946390647c30b12c5d3b89636422c282d0132869c61c082513a7dfa981fa64736f6c63430008120033
> 
> ======= /tmp/b.sol:A =======
> Binary:
> 6080604052348015600f57600080fd5b50606d80601d6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c80630dbe671f14602d575b600080fd5b60336035565b005b56fea2646970667358221220a433fe66d6602c3827ae7e0060d4539266d520fd0f1a61cc952ab89aabf7d1bf64736f6c63430008120033
> ```
> 
> I think if we have to rely on contract names being unique per file name, we have a problem. I'm not aware on any reason why the file names should affect compilation in any way. We could make an exception for the `--contract` flag but this would be really odd.

I don't know what to do here, if solc permits this. We could do some semantic comparison of the contracts, but this is hard to do because 
- the locs will definitely be different 
- the struct definitions etc will have different numbers because they are in different namespaces 
- we will have to recurse down any base contracts/interfaces used/libraries used, globals used

As a cop-out, I'd like to point out this is not a new problem for this PR :sweat_smile:

xermicus: I see, didn't even have the front end tooling in mind. Looks like we have uncovered something here requiring more thought. Let's move on and turn this into a documentation problem for now.

> As a cop-out, I'd like to point out this is not a new problem for this PR sweat_smile

:100:  But I'd appreciate if we could point this out in the CLI help for `--contract`

seanyoung: >  But I'd appreciate if we could point this out in the CLI help for `--contract`

Absolutely right, that belongs in the docs. I've added a warning.

docs/running.rst
```diff
@@ -115,6 +122,16 @@ Options:
...
    Disable the :ref:`common-subexpression-elimination` optimization
 
+.. warning::
+
+    If multiple Solidity source files define the same contract name, you will get a single
+    compiled contract file for this contract name. As a result, you will only have a single
+    contract with the duplicate name without knowing from which Solidity file it originated.
+
+    Solang will not give a warning about this problem. This is the same behavior as
+    Ethereum Solidity.
```
xermicus, line 132:
```suggestion
    Solang will not give a warning about this problem.
```

It is not the same behaviour as `solc`: 

```bash
❯ echo "contract A { function a() public pure {} }" > /tmp/a.sol && cp /tmp/a.sol /tmp/b.sol
❯ solc --bin -o /tmp/out /tmp/a.sol /tmp/b.sol 2>/dev/null
❯ ls /tmp/out/
_tmp_a_sol_A.bin  _tmp_b_sol_A.bin
```

It looks like `solc` just mangles the path into the contract artifacts. We don't need to fix that in this PR, but should line up with `solc` eventually later on.

docs/running.rst
LucasSte: That sounds good.

src/bin/solang.rs
```diff
@@ -416,33 +431,86 @@ fn compile(matches: &ArgMatches) {
 
     let mut errors = false;
 
+    // build a map of requested contract names, and a flag specifying whether it was found or not
```
LucasSte, line 434:
```suggestion
    // Build a map of requested contract names, and a flag specifying whether it was found or not
```

src/bin/solang.rs
```diff
@@ -416,33 +431,86 @@ fn compile(matches: &ArgMatches) {
...
+            errors = true;
         }
     }
 
     if let Some("ast-dot") = matches.get_one::<String>("EMIT").map(|v| v.as_str()) {
         exit(0);
     }
 
-    let std_json = *matches.get_one("STD-JSON").unwrap();
+    // ensure we have at least one contracts
```
LucasSte, line 470:
```suggestion
    // Ensure we have at least one contract
```

src/bin/solang.rs
```diff
@@ -416,33 +431,86 @@ fn compile(matches: &ArgMatches) {
...
+    }
 
-    if errors {
-        if std_json {
-            println!("{}", serde_json::to_string(&json).unwrap());
-            exit(0);
-        } else {
-            eprintln!("error: not all contracts are valid");
-            exit(1);
+    // ensure we all the requested contracts
```
LucasSte, line 476:
```suggestion
    // Ensure we have all the requested contracts
```

src/bin/solang.rs
```diff
@@ -507,106 +566,106 @@ fn process_file(
...
-                    "info: Saving binary {} for contract {}",
-                    bin_filename.display(),
-                    binary.name
-                );
-            }
+    if std_json {
+        json_contracts.insert(
+            binary.name,
+            JsonContract {
+                abi: abi::ethereum::gen_abi(contract_no, ns),
```
LucasSte, line 625: Why are we still generating Ethereum ABI?

LucasSte (approved):

xermicus (approved):

src/bin/solang.rs
seanyoung: This is for the solidity standard-json format, which is not really used anywhere. It was used by burrow and ewasm, both of which we don't support any more.

We could remove standard-json but I think it might be useful in the future, for example compiling over http. It needs to be redefined to be something more useful, as you can see below it still refers to ewasm.

PR #1165, opened 2023-02-08
LLVM 15

LucasSte:
This PR adapts emit to opaque pointers and removes all the pointer casts, which are not necessary anymore. These changes are essential for us to migrate to LLVM 15.

LLVM builds have been tested [here](https://github.com/LucasSte/solang/actions/runs/4128649851/jobs/7133349350).


CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
+
+### Added
+- Solidity mappings can now have a named key and a named value types. [seanyoung](https://github.com/seanyoung)
+
+### Changed
+- Solang now uses LLVM 15. [LucasSte](https://github.com/LucasSte)
+- The Subtrate Scale encoding now happens during the compiler's code generation phase, instead of during the code 
+emission phase. [xermicus](https://github.com/xermicus)
+- Solidity on Solana now required the Anchor framework for the client code, and the `@solana/solidity.js`
+  Typescript library has been deprecated.
```
seanyoung, line 15: It's more than deprecated: it no longer works because we don't generate the abi file.

Actually I was thinking of removing the npm library and marking the repo readonly.

src/sema/types.rs
```diff
@@ -1490,7 +1490,7 @@ impl Type {
             Type::Ref(r) => r.is_reference_type(ns),
             Type::StorageRef(_, r) => r.is_reference_type(ns),
             Type::InternalFunction { .. } => false,
-            Type::ExternalFunction { .. } => false,
+            Type::ExternalFunction { .. } => true,
```
seanyoung, line 1493: On Ethereum, an external function fits into a 256 bit register, so it's not a reference type. Ideally this should be `ns.target != Target::EVM` with a comment.

src/sema/types.rs
```diff
@@ -1068,7 +1068,7 @@ impl Type {
             Type::Ref(_) => false,
             Type::StorageRef(..) => false,
             Type::InternalFunction { .. } => false,
-            Type::ExternalFunction { .. } => false,
+            Type::ExternalFunction { .. } => true,
```
seanyoung, line 1071: On Ethereum, an external function fits into a 256 bit register, so it's not a reference type. Ideally this should be `ns.target != Target::EVM` with a comment.

seanyoung (commented):
Very nice stuff. Thanks @LucasSte 

I'd like to go over the changes one more time

CHANGELOG.md
LucasSte: >  it no longer works.
This is the meaning of 'deprecate', according to the [dictionary](https://www.dictionary.com/browse/deprecate):

> Computers. to cease supporting or recommending the use of (older elements, features, or versions of software)

LucasSte: None of the links you provided me explain the difference between deprecated and unavailable.

LucasSte: I'll change to `phased out`.

LucasSte: > Actually I was thinking of removing the npm library and marking the repo readonly.

That is a great idea. I was also thinking about that, but I would do it after we do a new release. Solang v0.2.1 still needs `@solana/solidity.js`.

LucasSte:
Tests that need the container image are failing, but those that do not, are working: The Mac Intel test, the Mac Arm test (always the first one to finish!) and the Windows test.

.github/workflows/build-llvm.yml
```diff
@@ -13,8 +13,8 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
-        tag_name: llvm14.0-1
-        release_name: LLVM Libraries v14.0 (v1)
+        tag_name: llvm15.0-1
+        release_name: LLVM Libraries v15.0 (v1)
```
seanyoung, line 17: If we're using solang-llvm, then this file can go

.github/workflows/release.yml
```diff
@@ -14,27 +14,27 @@ jobs:
       uses: actions/checkout@v2
       with:
         submodules: recursive
-    - run: git clone --depth 1 --branch solana-rustc/14.0-2022-03-22 https://github.com/solana-labs/llvm-project.git
+    - run: git clone --depth 1 --branch solana-rustc/15.0-2022-08-09 https://github.com/solana-labs/llvm-project.git
```
seanyoung, line 17: If we're using solang-llvm when the release workflow should not build llvm any more

seanyoung (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
+### Added
+- Solidity mappings can now have named key and named value types. [seanyoung](https://github.com/seanyoung)
+
+### Changed
+- Solang now uses LLVM 15. [LucasSte](https://github.com/LucasSte)
+- The Substrate Scale encoding now happens during the compiler's code generation phase, instead of during the code 
+emission phase. [xermicus](https://github.com/xermicus)
+- Solidity on Solana now required the Anchor framework for the client code, and the `@solana/solidity.js`
+  Typescript library is no longer compatible with Solidity.
+
```
xermicus, line 16:
```suggestion
- When casting hex literal numbers into the `bytesN` type, the hex literal may use leading zeros to match the size with the according `bytesN`, which aligns solang with `solc`. [xermicus](https://github.com/xermicus)
```

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
 
+## v0.2.2 (unreleased)
+
+### Added
+- Solidity mappings can now have named key and named value types. [seanyoung](https://github.com/seanyoung)
+
+### Changed
+- Solang now uses LLVM 15. [LucasSte](https://github.com/LucasSte)
+- The Substrate Scale encoding now happens during the compiler's code generation phase, instead of during the code 
+emission phase. [xermicus](https://github.com/xermicus)
```
xermicus, line 13:
```suggestion
```

We will spare this for later :blush:

xermicus (commented):
Thanks! :rocket:

.github/workflows/test.yml
```diff
@@ -59,7 +59,7 @@ jobs:
   linux-x86-64:
     name: Linux x86-64
     runs-on: ubuntu-20.04
-    container: ghcr.io/hyperledger/solang:ci
+    container: ghcr.io/hyperledger/solang-llvm@sha256:8ee994dfad1b3fc8aadb9c6747c765087b2b2bcbad516f55ba575d86086d2f54
```
seanyoung, line 62: You need to use a tagged version

seanyoung (approved):
Just one thing to fix. That was a massive amount of work!

xermicus (approved):

PR #1166, opened 2023-02-09
Fix --math-overflow flag and add tests

seanyoung:
- fix `--math-overflow` cli option
- add test for `--math-overflow`
- adds unchecked to the cfg printer

### DIFF
--- a/integration/solana/overflow.sol
+++ b/integration/solana/overflow.sol
@@ -1,0 +1,17 @@
+contract overflow {
+	function addu32(uint32 a, uint32 b) public pure returns (uint32 c) {
+		c = a + b;
+	}
+
+	function subu32(uint32 a, uint32 b) public pure returns (uint32 c) {
+		c = a - b;
+	}
+
+	function mulu32(uint32 a, uint32 b) public pure returns (uint32 c) {
+		c = a * b;
+	}
+
+	function powu32(uint32 a, uint32 b) public pure returns (uint32 c) {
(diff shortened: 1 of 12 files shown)


src/codegen/yul/builtin.rs
```diff
@@ -229,13 +229,13 @@ fn process_arithmetic(
 
     match builtin_ty {
         YulBuiltInFunction::Add => {
-            Expression::Add(*loc, left.ty(), true, Box::new(left), Box::new(right))
+            Expression::Add(*loc, left.ty(), false, Box::new(left), Box::new(right))
```
LucasSte, line 232: `true` means the operation is unchecked. `false` means the operations is NOT unchecked.

Yul operation should be unchecked, so you need to leave the `true` here.

Check it here:

https://github.com/hyperledger/solang/blob/143da73c8b12b7a1ac1ed3b78280da1ae6f27273/src/emit/expression.rs#L127

LucasSte: Another idea is to change to nomenclature to something less confusing.

seanyoung: Thanks for catching that, like you say I got confused with the *un*checked. That what I get for working late!

LucasSte (approved):

LucasSte:
Please, fix the failing tests before merging.

xermicus (approved):

PR #1167, opened 2023-02-09
Fix CSE unchecked operations bug

LucasSte:
@seanyoung found that the CSE pass was not differentiating checked and unchecked arithmetic expressions. This PR fixes this issue.


seanyoung (approved):
Thanks!

PR #1168, opened 2023-02-09
Fix CSE bug

LucasSte:
@xermicus found a bug in common subexpression elimination while working on the SCALE decoder. This PR fixes the bug, but the fix is only palliative for it might not cover all possible cases.

The problem would be best solved implementing a correct version of partial redundancy elimination, which leverages the available expressions analysis (already available) and an anticipated expression analysis not only to detect common subexpressions, but to know where exactly to evaluate them. This solution, nevertheless, is too time consuming to code and will be easier to implement when we have a new IR with SSA.

More information about partial redundancy elimination can be found in the Chapter 9.5 of the book `Compilers: Principles, Techniques & tools`, second edition, from Monica S. Lam.

### DIFF
--- a/src/codegen/subexpression_elimination/common_subexpression_tracker.rs
+++ b/src/codegen/subexpression_elimination/common_subexpression_tracker.rs
@@ -238,6 +238,24 @@ impl CommonSubExpressionTracker {
         self.cur_block = block_no;
     }
 
+    /// Find the depth of each control flow graph block
+    /// The definition of depth is the distance (in number o graph edges) from the entry block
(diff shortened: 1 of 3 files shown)


seanyoung (changes requested):
> @xermicus found a bug in common subexpression elimination while working on the SCALE decoder. This PR fixes the bug, but the fix is only palliative for it might not cover all possible cases.

We need to know what the bug is, what do you mean by "it might not cover all possible cases". 

The CSE pass should produce correct code in call cases.

PR #1173, opened 2023-02-12
Account for leading zeros in hex number literals size width

xermicus:
Matching the behavior of `solc`.

Signed-off-by: xermicus <cyrill@parity.io>


seanyoung:
What exact behaviour are you matching? This is not allowed in solc/solang:
```solidity
contract c {
	bytes4 b = 0xfefdfc;
}
```
Note that as an exception, solc does allow 0 to be converted (which solang already allows)
```solidity
contract c {
	bytes4 b = 0x00;
}
```

xermicus:
`solc` allows this, while we don't:

```solidity
bytes4 b = bytes4(0x000000AA)
```

However it now came to my attention, that this is only allowed in `solc` if the constant is prefixed width zeros to match the byte width, so I have to go over it again.

seanyoung:
> `solc` allows this, while we don't:
> 
> ```solidity
> bytes4 b = bytes4(0x000000AA)
> ```
> 
> However it now came to my attention, that this is only allowed in `solc` if the constant is prefixed width zeros to match the byte width, so I have to go over it again.

Ah interesting, I had missed that. That's an annoying change, the prefixed zeros aren't stored in the ast `Expression`. Maybe you can set the type of `Expression::NumberLiteral()` to the length including prefixed bytes.

xermicus:
Yes quite annoying. Changed the PR so that for hex literals (only), the bit width is taken from the literal size which includes leading zeros instead. I also included a drive-by fix in sema where `Type::bytes()` returned the bit width instead of byte width.

seanyoung:
Looks like the right solution, good stuff!

It should probably include a test for  
```solidity
bytes2 c = 0x00bffc; // not allowed
int16 d  = 0x00bffc; // is allowed
```

LucasSte (approved):

seanyoung:
> Error: Type int_const 49148 is not implicitly convertible to expected type int16. Literal is too large to fit in int16.
>  --> match.sol:4:3:
>   |
> 4 |             int16 d  = 0x00bffc; // is allowed
>   |             ^^^^^^^^^^^^^^^^^^^
> 

Sorry, copy and paste bug. The reason is not allowed is because 0xbffc does not fit into *signed* `int16`. However:

```solidity
uint16 d  = 0x00bffc;
```
is allowed (solc-0.8.18)

xermicus:
Ah I see, added your example as a test case :+1:

seanyoung (approved):
Great, thanks. 

Only tiny nitpick is that it would be nice to have a test with `_` in it or an odd number of hex digits.

xermicus:
We have now this exact situation, e.g. need to remember to manually request a review from Lucas

xermicus:
Good catch @seanyoung! This is giving me slight headaches, why DO they allow `uint16 i = 0x0_00AA;` but not `bytes3( 0x00_00AA);`

seanyoung:
> Good catch @seanyoung! This is giving me slight headaches, why DO they allow `uint16 i = 0x0_00AA;` but not `bytes3(0x00_00AA);`

You mean `bytes3(0x0_00AA)`, right? I am guessing that zero-prefixes are only allowed if you end up with an even number of hex digits overall.

> I mean I can bail out on an odd length, or we can permit this. @seanyoung WDYT?

I think we should be at least as permissive as solc, so maybe ignore the `hex_str_len` if it is odd?

`0x` literals for `bytesN` types are a total mess and should really be banned, but we're stuck with being compatible with solc for now.

xermicus:
> I think we should be at least as permissive as solc, so maybe ignore the hex_str_len if it is odd?

Very good idea, thanks!

PR #1175, opened 2023-02-13
Expand list of solang code owners

seanyoung:
When creating a pull request, the users in this files are used to populate the reviewers. If it lists @hyperledger/solang-committers then we can't see who those users are and as soon as one of the members of @hyperledger/solang-committers leaves a review, then the pull request is marked as approved while we really want *all* the maintainers to approve.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Solang Committers
-* @hyperledger/solang-committers
+* @seanyoung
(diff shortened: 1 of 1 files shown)


seanyoung:
I've added @ryjones but left @hyperledger/solang-committers  in place as the reviewers to demonstrate the problem.

xermicus (approved):

xermicus:
We have now this exact situation, e.g. need to remember to manually request a review from Lucas

PR #1179, opened 2023-02-14
Use scale encoder in emit for release

xermicus:
As discussed: #1128 introduced a refactor version of the scale encoder in codegen. However, the emit version of the SCALE encoder could not be phased out yet; the dispatcher is written in emit as well, so we need to refactor the decoder (WIP, currently blocked by #1168 ) and then dispatcher into codegen (WIP after the decoder) first, too.

This resulted in the compiler using the new version of the encoder but only fo `abi.encode` and not in the dispatcher.

We don't want to release a version with a half-baked SCALE encoder/decoder implementation. Minor issue is that it "feels wrong" to use different implementations for the same thing, depending on the circumstance. Major issue is that many tests go through the encoding logic in the dispatcher, but the dispatcher still uses the emit version. It is fine for the `main` branch but a no-go to ship that in a release.


seanyoung:
Thank you @xermicus for doing this

seanyoung (approved):

src/codegen/encoding/mod.rs
```diff
@@ -36,6 +36,10 @@ pub(super) fn abi_encode(
     cfg: &mut ControlFlowGraph,
     packed: bool,
 ) -> (Expression, Expression) {
+    // TODO: Use the codegen version once the substrate dispatcher is refactor into codegen.
```
LucasSte, line 39:
```suggestion
    // TODO: Use the codegen version once the substrate dispatcher is refactored into codegen.
```

tests/codegen_testcases/solidity/common_subexpression_elimination.sol
```diff
@@ -484,57 +475,56 @@ contract c1 {
...
-                p = (a-b)*4;
+                p = (a - b) * 4;
                 b++;
             }
-            bool e2 = p<1;
+            bool e2 = p < 1;
         }
 
-        return k/a + int(uint(a)**uint(g));
+        return k / a + int256(uint256(a)**uint256(g));
```
LucasSte, line 528: Why did you change `int` to `int256`?

LucasSte (approved):

tests/codegen_testcases/solidity/common_subexpression_elimination.sol
xermicus: The solidity formatter did this. It is a lint

PR #1181, opened 2023-02-15
Split `method_call_pos_args` function

LucasSte:
This PR splits the function `method_call_pos_args` into multiple small functions for better readability and maintenance. I had this idea when thinking about the new API we are going to implement for mappings on Solidity.

I am open to other ideas on how to manage the multiple functions, if you don't like the approach I have taken.


src/sema/expression/function_call.rs
```diff
@@ -547,20 +547,19 @@ pub(super) fn function_call_named_args(
     Err(())
 }
 
-/// Resolve a method call with positional arguments
-pub(super) fn method_call_pos_args(
+/// Check if the function is a method of a variable
+fn try_variable(
```
seanyoung, line 551: This is not a good name. We are handling the case of `foo.bar(..)` here, e.g. `super.func()`, `library.libfunc()`, `abi.encode()`. How about `fn try_namespaced()`? It's not great either but a little better.

src/sema/expression/function_call.rs
```diff
@@ -1007,371 +892,623 @@ pub(super) fn method_call_pos_args(
...
-                    let val_expr = expression(
-                        &args[0],
-                        context,
-                        ns,
-                        symtable,
-                        diagnostics,
-                        ResolveTo::Type(&elem_ty),
-                    )?;
+/// Check if we can resolve the call with ns.resolve_type
+fn try_resolve_type(
```
seanyoung, line 911: try_user_type is better, I think

src/sema/expression/function_call.rs
```diff
@@ -1007,371 +892,623 @@ pub(super) fn method_call_pos_args(
...
-                || ns.functions[*function_no].ty != pt::FunctionTy::Function
-            {
-                continue;
-            }
+    Ok(None)
+}
 
-            name_matches.push(*function_no);
-        }
+/// Check if the function call is to a type's builtin method
```
seanyoung, line 994: This also handles calls on contracts, not builtin

seanyoung (changes requested):

src/sema/expression/function_call.rs
```diff
@@ -547,20 +547,19 @@ pub(super) fn function_call_named_args(
     Err(())
 }
 
-/// Resolve a method call with positional arguments
-pub(super) fn method_call_pos_args(
+/// Check if the function is a method of a variable
```
xermicus, line 550: Please explain what all the different variants of the return type mean

src/sema/expression/function_call.rs
```diff
@@ -688,137 +687,23 @@ pub(super) fn method_call_pos_args(
...
-                "format only allowed on string literals".to_string(),
-            ));
-            Err(())
-        };
-    }
+    Ok(None)
+}
 
-    if let Type::StorageRef(immutable, ty) = &var_ty {
+/// Check if the function is a method of a storage reference
```
xermicus, line 693: Again, return type is not self-explanatory and deserves a comment

src/sema/expression/function_call.rs
```diff
@@ -1007,371 +892,623 @@ pub(super) fn method_call_pos_args(
...
-                1 => {
-                    let val_expr = expression(
-                        &args[0],
-                        context,
-                        ns,
-                        symtable,
-                        diagnostics,
-                        ResolveTo::Type(&elem_ty),
-                    )?;
+/// Check if we can resolve the call with ns.resolve_type
```
xermicus, line 910: Same

src/sema/expression/function_call.rs
xermicus, line 994: Same

xermicus (approved):
LGTM

src/sema/expression/function_call.rs
LucasSte: Do you mean `Result<Option<Expression>, ()>`?

seanyoung (approved):

PR #1185, opened 2023-02-20
Fix bugs in codegen compact encoder

xermicus:
Fixes an off-by-one, incorrect dirty tracking and missing cast before buf writes


seanyoung (approved):

src/codegen/encoding/scale_encoding.rs
```diff
@@ -166,13 +166,12 @@ fn encode_compact(
...
     cfg.set_basic_block(big);
     if let (Some(buffer), Some(offset)) = (buffer, offset) {
-        let mul2 = Expression::BitwiseOr(Codegen, Uint(32), mul.into(), two.into());
         cfg.add(
             vartab,
             Instr::WriteBuffer {
                 buf: buffer.clone(),
                 offset: offset.clone(),
-                value: mul2,
+                value: Expression::BitwiseOr(Codegen, Uint(32), mul.into(), two.into()),
```
LucasSte, line 174: I see you have casted the value in the two cases above. Is this supposed not be casted as well?

LucasSte (approved):

src/codegen/encoding/scale_encoding.rs
xermicus: In this case the vector length is supposed to be 32 bit wide so no cast is needed.

PR #1188, opened 2023-02-20
Debug buffer formatting for Substrate

salaheldinsoliman:
the debug buffer for Substrate target contains three sources of strings: programmer debug prints, api return code logs and runtime errors. This PR formats the debug buffer, making it more readable with the option to be parsed by some external tool.


src/emit/substrate/target.rs
```diff
@@ -1837,7 +1922,7 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
         emit_context!(bin);
         let error_with_loc = error_msg_with_loc(ns, &reason_string, reason_loc);
         let custom_error = string_to_basic_value(bin, ns, error_with_loc);
-        call!(
+        /*call!(
```
xermicus, line 1925: Please delete if it is no longer needed

src/emit/substrate/mod.rs
```diff
@@ -1859,6 +1861,21 @@ fn log_return_code(binary: &Binary, api: &'static str, code: IntValue) {
         .unwrap()
         .into_pointer_value();
 
+    let limiter_string = binary.emit_global_string("limiter", limiter, true);
```
xermicus, line 1864: Nitpick, it is called a "delimiter"

xermicus (commented):
It is going into the right direction. However I don't see escaping of the argument for the  `print` builtin (should be implemented in codegen), which does not allow tools to further parse it. @seanyoung WDYT regarding this?

src/emit/substrate/target.rs
```diff
@@ -1051,19 +1051,104 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
         (data, length)
     }
 
-    fn print(&self, binary: &Binary, string_ptr: PointerValue, string_len: IntValue) {
+    fn print(
+        &self,
+        binary: &Binary,
+        string_ptr: PointerValue,
+        string_len: IntValue,
+        source: PrintSource,
```
xermicus, line 1059: This might be abusing the `print` instruction facility. I thought it quite nice when print basically just called into seal_debug and that's it. Also this way we have code duplication for other targets?

On a high level, it seems to be the better solution to leave the formatting responsibility to the callers inside codegen. @seanyoung WDYT?

salaheldinsoliman: I also think this is a mistake. I have another version of formatting print in codegen instead of emit and it looks much more better

salaheldinsoliman: With that said, I will push a commit soon with print formatting done in codegen

src/emit/solana/target.rs
```diff
@@ -2383,6 +2389,7 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
             bin,
             bin.vector_bytes(custom_error),
             bin.vector_len(custom_error),
+            true,
```
xermicus, line 2392: Why is `is_error` ignored here?

src/emit/substrate/target.rs
```diff
@@ -1062,8 +1068,9 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
         .left()
         .unwrap()
         .into_int_value();
-
-        log_return_code(binary, "seal_debug_message", ret);
+        if !is_error {
```
xermicus, line 1071: Why did you add this guard?

We must assume that `seal_debug_message` can fail too. But this guard hides it from the user.

xermicus (changes requested):
Looks good overall

src/codegen/mod.rs
```diff
@@ -1349,20 +1349,24 @@ pub(super) fn error_msg_with_loc(ns: &Namespace, error: &str, loc: Option<Loc>)
...
+    let file_no = loc.file_no();
+    let curr_file = &ns.files[file_no];
+    let (line_no, offset) = curr_file.offset_to_line_column(loc.start());
+    format!(
+        "{}:{}:{}",
+        curr_file.path.file_name().unwrap().to_str().unwrap(),
+        line_no + 1,
+        offset
+    )
+}
```
seanyoung, line 1372: This function has two problems: -
 
- It only takes the start location into account, not the end. This means that user is missing same information
 - It is very similar to `ns.loc_to_string()`, however that uses the full path. 
 
 How about adding a `ns.loc_to_filename_string()` which does the same as `ns.loc_to_string()` but with only the filename and not the full path.

src/codegen/cfg.rs
```diff
@@ -63,7 +63,7 @@ pub enum Instr {
     /// Abort execution
     AssertFailure { encoded_args: Option<Expression> },
     /// Print to log message
-    Print { expr: Expression },
+    Print { expr: Expression, is_error: bool },
```
seanyoung, line 66: It is not clear to me what `is_error` does. It should either have a comment explaining it or be removed.

seanyoung (changes requested):

LucasSte (approved):

src/codegen/cfg.rs
salaheldinsoliman: is_error is used in the print function to decide on whether we should log seal_debug_message or not. I though it would make sense to mask the user from logging a seal_debug_message after a runtime error has been reported.

salaheldinsoliman: @xermicus What do you think?

xermicus: I vouch for removal. A print instruction can always behave the same, no need to hide it from the user IMO. It only makes the code more complicated

src/codegen/expression.rs
```diff
@@ -812,7 +812,13 @@ pub fn expression(
         } => {
             let expr = expression(&args[0], cfg, contract_no, func, ns, vartab, opt);
 
-            cfg.add(vartab, Instr::Print { expr });
+            let to_print = if ns.target.is_substrate() {
+                add_prefix_and_dilimiter_to_print(expr)
```
xermicus, line 816: It is called `delimiter`

```suggestion
                add_prefix_and_delimiter_to_print(expr)
```

src/emit/substrate/mod.rs
```diff
@@ -1800,9 +1800,11 @@ fn log_return_code(binary: &Binary, api: &'static str, code: IntValue) {
 
     emit_context!(binary);
 
-    let fmt = format!("{api}=");
+    let fmt = format!("api_return_code: {api}=");
     let msg = fmt.as_bytes();
-    let length = i32_const!(msg.len() as u64 + 16);
+    let dilimiter = ",\n".as_bytes();
```
xermicus, line 1805: `delimiter` please, "dilimiter" doesn't exist :upside_down_face:

src/emit/substrate/mod.rs
```diff
@@ -1800,9 +1800,11 @@ fn log_return_code(binary: &Binary, api: &'static str, code: IntValue) {
 
     emit_context!(binary);
 
-    let fmt = format!("{api}=");
+    let fmt = format!("api_return_code: {api}=");
```
xermicus, line 1803: `api_return_code` is a bit awkward to read. What about using just `call`?

xermicus (commented):
Looks quite nice, only found some details

src/emit/substrate/mod.rs
salaheldinsoliman: silly spelling mistake, sorry

src/codegen/expression.rs
```diff
@@ -1383,6 +1400,14 @@ fn require(
...
                                 ),
                             ),
+                            (
+                                FormatArg::StringLiteral,
+                                Expression::BytesLiteral(
+                                    Loc::Codegen,
+                                    Type::Bytes(",\n".as_bytes().len() as u8),
+                                    ",\n".as_bytes().to_vec(),
+                                ),
+                            ),
```
seanyoung, line 1410: Might as well concatenate the last two string literals at compile time rather than runtime.

src/codegen/expression.rs
```diff
@@ -1434,6 +1469,14 @@ fn revert(
...
                         ),
                     ),
+                    (
+                        FormatArg::StringLiteral,
+                        Expression::BytesLiteral(
+                            Loc::Codegen,
+                            Type::Bytes(",\n".as_bytes().len() as u8),
+                            ",\n".as_bytes().to_vec(),
+                        ),
+                    ),
```
seanyoung, line 1479: Might as well concatenate the last two string literals at compile time rather than runtime.

src/codegen/expression.rs
```diff
@@ -3200,17 +3243,27 @@ fn code(loc: &Loc, contract_no: usize, ns: &Namespace, opt: &Options) -> Express
...
+                    string.as_bytes().to_vec(),
+                ),
+            ),
+            (
+                FormatArg::StringLiteral,
+                Expression::BytesLiteral(
+                    Loc::Codegen,
+                    Type::Bytes(",\n".as_bytes().len() as u8),
+                    ",\n".as_bytes().to_vec(),
+                ),
```
seanyoung, line 3264: Might as well concatenate the last two string literals at compile time rather than runtime.

src/sema/file.rs
```diff
@@ -48,6 +48,33 @@ impl File {
...
+    pub fn loc_to_filename_string(&self, start: usize, end: usize) -> String {
+        let (from_line, from_column) = self.offset_to_line_column(start);
+        let (to_line, to_column) = self.offset_to_line_column(end);
+
+        if from_line == to_line && from_column == to_column {
+            format!("{}:{}:{}", self, from_line + 1, from_column + 1)
+        } else if from_line == to_line {
+            format!(
+                "{}:{}:{}-{}",
+                self.path.file_name().unwrap().to_str().unwrap(),
```
seanyoung, line 61: self.path.file_name().unwrap().display()

src/sema/file.rs
```diff
@@ -48,6 +48,33 @@ impl File {
         }
     }
 
+    /// Give a position as a human readable position, with only the filename not the full path.
+    pub fn loc_to_filename_string(&self, start: usize, end: usize) -> String {
+        let (from_line, from_column) = self.offset_to_line_column(start);
+        let (to_line, to_column) = self.offset_to_line_column(end);
```
seanyoung, line 54: This function is a copy & paste of `fn loc_to_string()` with modifications. Please factor out the common code in a single function, e.g. pass in the filename/path as `&str` and then render the full path or filename, then call the common function.

Also use `self.path.file_name().unwrap().display()` rather than `self.path.file_name().unwrap().to_str().unwrap()`; paths are not necessarily valid utf8 on unix (although we don't have tests for this).

seanyoung (changes requested):

src/sema/file.rs
salaheldinsoliman: file_name.unrwap returns an OsStr, which should be converted to an str first. 
I think a good solution here is:
get `fullpath: String = path.display()` first, and the parse it by hand. (this would deal with returning a valid utf8, since display is implented)

src/codegen/expression.rs
```diff
@@ -1369,11 +1375,22 @@ fn require(
...
+                    );
                     let print_expr = Expression::FormatString(
                         Loc::Codegen,
                         vec![
+                            (
+                                FormatArg::StringLiteral,
+                                Expression::BytesLiteral(
+                                    Loc::Codegen,
+                                    Type::Bytes(prefix.len() as u8),
+                                    prefix.to_vec(),
```
seanyoung, line 1391: Here you should create a new string which is `prefix` and `error_string` concatenated, and then create a single `FormatArg::StringLiteral` rather than two.

src/codegen/expression.rs
```diff
@@ -1421,10 +1438,22 @@ fn revert(
...
+            );
             let print_expr = Expression::FormatString(
                 Loc::Codegen,
                 vec![
+                    (
+                        FormatArg::StringLiteral,
+                        Expression::BytesLiteral(
+                            Loc::Codegen,
+                            Type::Bytes(prefix.len() as u8),
+                            prefix.to_vec(),
```
seanyoung, line 1454: Here you should create a new string which is `prefix` and `error_string` concatenated, and then create a single `FormatArg::StringLiteral` rather than two.

src/codegen/expression.rs
```diff
@@ -3224,7 +3253,58 @@ pub(crate) fn log_runtime_error(
...
+                prefix.to_vec(),
+            ),
+        ));
+        new_vec.append(args);
+        new_vec.push((
+            FormatArg::StringLiteral,
+            Expression::BytesLiteral(
+                Loc::Codegen,
+                Type::Bytes(delimiter.len() as u8),
+                delimiter.to_vec(),
```
seanyoung, line 3281: Here you should create a new string which is `prefix` and `delimiter` concatenated, and then create a single `FormatArg::StringLiteral` rather than two.

src/codegen/expression.rs
```diff
@@ -3224,7 +3253,58 @@ pub(crate) fn log_runtime_error(
...
+                        prefix.to_vec(),
+                    ),
+                ),
+                (FormatArg::Default, expr),
+                (
+                    FormatArg::StringLiteral,
+                    Expression::BytesLiteral(
+                        Loc::Codegen,
+                        Type::Bytes(delimiter.len() as u8),
+                        delimiter.to_vec(),
```
seanyoung, line 3304: Here you should create a new string which is `prefix` and `delimiter` concatenated, and then create a single `FormatArg::StringLiteral` rather than two.

src/sema/file.rs
```diff
@@ -31,15 +31,23 @@ impl File {
...
             )
         } else {
             format!(
                 "{}:{}:{}-{}:{}",
-                self,
+                if full_path {
+                    format!("{self}")
+                } else {
+                    self.file_name()
+                },
```
seanyoung, line 50: You are duplicating code which you are also adding to line 34-38

src/sema/file.rs
```diff
@@ -71,6 +79,16 @@ impl File {
...
     }
+
+    pub fn file_name(&self) -> String {
+        #[cfg(not(windows))]
+        let res = format!("{self}").split('/').last().unwrap().to_string();
+
+        #[cfg(windows)]
+        let res = format!("{self}").split('\\').last().unwrap().to_string();
+
+        res
```
seanyoung, line 90: Don't do your own file name handling, better to the standard library for that:
 ```rust
format!("{}", self.path.filename().unwrap().to_path_buf())
```

**Review by seanyoung — CHANGES_REQUESTED** [review]

File: src/codegen/expression.rs (hunk 16 above)
Comment (salaheldinsoliman) [reply src/codegen/expression.rs:1391]: I cannot understand why that should be done to be honest...
`StringConcat` calls `stdlib/ void concat()` which allocates a new buffer for the concatenated string and makes two copies, then another copy for the format. `FormatString` would allocate a buffer and do three copies total.
Is this concerning readability?

Comment (salaheldinsoliman) [reply src/codegen/expression.rs:1391]: I want to reach something like this: `format ("runtime_error: {require_expression} require condition failed,"). E.g wrapping the the require expression with a prefix and delimiter.
Why should I concatenate the prefix and the error string?

Comment (salaheldinsoliman) [reply src/codegen/expression.rs:1391]: given `require (1>2, "SESA!")`
I want to reach: `runtime_error: SESA! require condition failed,`

File: src/sema/file.rs (hunk 21 above)
Comment (salaheldinsoliman) [reply src/sema/file.rs:90]: `path.file_name()` is an `OsStr`, which should be converted to an `str` first as `OsStr` doesn't implement `display` or `to_path_buf`. This what I was doing in `self.path.file_name().unwrap().to_str().unwrap()`. 

I think doing it by hand is better. (`display()` is first implemented for windows and other Os', then it's parsed accordingly)

File: src/codegen/expression.rs (hunk 16 above)
Comment (seanyoung) [reply src/codegen/expression.rs:1391]: Sorry, reading the code again I was confused.

File: src/codegen/expression.rs (hunk 17 above)
Comment (seanyoung) [reply src/codegen/expression.rs:1454]: Please ignore, sorry

File: src/codegen/expression.rs (hunk 18 above)
Comment (seanyoung) [reply src/codegen/expression.rs:3281]: Please ignore, sorry

File: src/codegen/expression.rs (hunk 19 above)
Comment (seanyoung) [reply src/codegen/expression.rs:3304]: Please ignore, sorry

File: src/sema/file.rs (hunk 21 above)
Comment (seanyoung) [reply src/sema/file.rs:90]: You're right, `.to_path_buf()` is not implemented and it's not great either to convert to another struct.

I still think you can avoid doing the parsing by hand. This time I've tested it in on branch.

```rust
    pub fn file_name(&self) -> String {
        self.path.file_name().unwrap().to_string_lossy().into()
    }
```

salaheldinsoliman: Thanks!

src/codegen/expression.rs
```diff
@@ -1369,11 +1375,22 @@ fn require(
         Target::Solana | Target::Substrate { .. } => {
             if opt.log_runtime_errors {
                 if let Some(expr) = expr {
-                    let error_string =
-                        error_msg_with_loc(ns, " require condition failed", Some(expr.loc()));
+                    let prefix = "runtime_error: ".as_bytes();
```
seanyoung, line 1378: Could be replaced with `b"runtime_error: "`

src/sema/file.rs
```diff
@@ -71,6 +76,10 @@ impl File {
             self.line_starts[line_no - 1] + column_no
         }
     }
+
+    pub fn file_name(&self) -> String {
+        self.path.file_name().unwrap().to_string_lossy().into()
```
seanyoung, line 81: Doesn't need to be a function, can be inlined in `fn loc_to_string()`

seanyoung (approved):

src/sema/file.rs
salaheldinsoliman: I think this could be useful if the file name is needed at other pieces of the code

xermicus (approved):
This is a nice feature, thanks!

PR #1189, opened 2023-02-20
Use anticipated expressions to fix CSE bug

LucasSte:
This PR fixes a bug @xermicus has found in the common subexpression elimination pass and introduces the following changes to our codebase:

1. `InstrOrigin` has been removed, as we don't use it for anything.
2. I have added an anticipated expression analysis to find the correct place to evaluate a common expression.
3. The dominator of two vertices (not for all the graph) is found using a flow analysis.


xermicus (approved):
As discussed in the standup, in the broad scheme we want to use `Lengauer-Tarjan` to find dominators, as other use cases require finding dominators in general and not just in between two distinct nodes.

But I need this fix to unblock some emit->codegen refactorings, so can be done later.

src/codegen/subexpression_elimination/anticipated_expressions.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::collections::HashMap;
+
+/// The AnticipatedExpression struct manages everything related to traversing the CFG backwards, so
+/// we can perform an anticipated expression analysis.
+///
+/// "An expression is anticipated at a point if it is certain to be evaluated along any
+/// path before this expression's value is changed (any variables its evaluation depends
+/// on are reassigned)."
+///
+/// The anticipated expression analysis helps us find where a common subexpression can be evaluated.
```
seanyoung, line 17: This doesn't describe anticipated expression. How about:

Available expressions shows us where sub expressions are already evaluated. Using anticipated expression we can do one more step: determine when an expression can be evaluated, because all the definitions it depends on are available. 

This is can be used for moving a common subexpression to another block.

src/codegen/subexpression_elimination/anticipated_expressions.rs
```diff
@@ -0,0 +1,173 @@
...
+#[derive(Default, Clone)]
+pub(super) struct AnticipatedExpressions<'a> {
+    /// The AvailableExpressionSet for each CFG block, when the graph is evaluated in reverse
+    reverse_sets: HashMap<usize, AvailableExpressionSet<'a>>,
+    /// The CFG represented ad a DAG, but with each edge reversed
```
seanyoung, line 22: as a dag

src/codegen/subexpression_elimination/anticipated_expressions.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        let mut local_level: u16 = u16::MAX;
+        for child in &dag[block] {
+            local_level = std::cmp::min(
+                local_level,
+                AnticipatedExpressions::blocks_depth(dag, *child, level + 1, depth),
+            );
+        }
+
+        local_level
```
seanyoung, line 62: might be nicer as 
```rust
dag[block].iter()
   .map(|child| AnticipatedExpressions::blocks_depth(dag, *child, level + 1, depth))
   .min()
   .unwrap_or(u16::MAX)
```

src/codegen/subexpression_elimination/anticipated_expressions.rs
seanyoung: You could include a reference to the compiler book chapter on anticipated expressions

seanyoung (changes requested):
The testing seems limited. There are no unit tests for the reversed dag/flow/anticipated expressions, just a codegen tests with the original problem.

src/codegen/subexpression_elimination/mod.rs
```diff
@@ -48,10 +51,11 @@ pub struct AvailableExpression {
 
 /// Each BasicExpression is a graph node
 #[derive(Clone)]
-pub struct BasicExpression {
+pub struct BasicExpression<'a> {
     expr_type: ExpressionType,
     expression_id: NodeId,
-    children: HashMap<NodeId, Rc<RefCell<BasicExpression>>>,
+    children: HashMap<NodeId, Rc<RefCell<BasicExpression<'a>>>>,
+    pub reference: &'a Expression,
```
seanyoung, line 58: I don't know what this new field is for. All the fields could use with (doc) comments

src/codegen/subexpression_elimination/common_subexpression_tracker.rs
```diff
@@ -34,25 +34,22 @@ bitflags! {
...
     cur_block: usize,
-    new_cfg_instr: Vec<(InstrOrigin, Instr)>,
+    new_cfg_instr: Vec<Instr>,
     parent_block_instr: Vec<(usize, Instr)>,
     /// Map from variable number to common subexpression
     mapped_variables: HashMap<usize, usize>,
-    /// The CFG is a cyclic graph. In order properly find the lowest common block,
-    /// we transformed it in a DAG, removing cycles from loops.
-    cfg_dag: Vec<Vec<usize>>,
+    anticipated_expressions: AnticipatedExpressions<'a>,
```
seanyoung, line 47: All the fields could do with comments (well some more than others)

src/codegen/subexpression_elimination/common_subexpression_tracker.rs
```diff
@@ -34,25 +34,22 @@ bitflags! {
 }
```
seanyoung, line 34: Is `Color` used anywhere?

src/codegen/subexpression_elimination/mod.rs
```diff
@@ -161,21 +209,24 @@ fn kill_loop_variables(block: &BasicBlock, cur_set: &mut AvailableExpressionSet,
...
+        cur_set.remove_mapped(*var_no);
         cur_set.kill(*var_no);
     }
 }
 
 /// Find the correct visiting order for the CFG traversal, using topological sorting. The visiting
 /// order should be the same as the execution order. This function also returns a DAG for the
 /// execution graph. This helps us find the lowest common ancestor later.
-fn find_visiting_order(cfg: &ControlFlowGraph) -> (Vec<(usize, bool)>, Vec<Vec<usize>>) {
+fn find_visiting_order(cfg: &ControlFlowGraph) -> CfgAsDag {
```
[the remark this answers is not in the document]
seanyoung: I did think the DAG could be created using BFS and then you would not need the stack, however if I remember correctly BFS requires a stack for its implementation so you it probably doesn't win you anything. :laughing:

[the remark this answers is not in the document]
LucasSte: > The removal of cycles depends on the back edge being in the stack, so I guess that would have to be a dominating node. That is always true for code we generate, but surely it's possible to create code with gotos which make this fail (if we had gotos). Could do with a comment.

The algorithm works with gotos as well. A cycle happens when a vertex is in the stack but it hasn't been visited, when this happens the returning edge is removed. We can have gotos that do not create cycles (which would work as well) and those that create. If a go-to create a cycle, the detection would remove the returning edge too.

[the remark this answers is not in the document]
LucasSte: > Also I am unsure why `degrees: Vec<i32>` uses a signed type. Can it go negative? If so, that should be explained.

While analyzing the graph, all degrees will be positive, because we I am not calculating out-degrees, only in-degrees. When calculating the topological sorting, I subtract 1 from the degree to determine which node I should be visiting next. This subtraction can make the value on the vector negative (which is the correct behavior).

https://github.com/hyperledger/solang/blob/12177ba59bca0e05113019e6a085d05d85ad9d32/src/codegen/subexpression_elimination/mod.rs#L196

src/codegen/subexpression_elimination/common_subexpression_tracker.rs
LucasSte: That is a remainder from my old code and should be removed. Thanks for catching this.

src/codegen/subexpression_elimination/mod.rs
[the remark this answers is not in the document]
seanyoung: > The algorithm works with gotos as well. A cycle happens when a vertex is in the stack but it hasn't been visited. When this happens the returning edge is removed. We can have gotos that do not create cycles (which would work as well) and those that create. If a go-to creates a cycle, the detection would remove the returning edge too.

Yes, you are right. Nevermind

PR #1196, opened 2023-02-22
error should not be a keyword

seanyoung:
This makes it possible to use error as a variable name, contract name etc.

Note that we do not support the `error Foo(..);` in solang - however it is supported in the parser.

In solc error is not a keyword.


solang-parser/src/lexer.rs
```diff
@@ -33,7 +33,6 @@ pub enum Token<'input> {
 
     Struct,
     Event,
-    Error,
```
xermicus, line 36: There is also this weird `Error(string)` type, will this still work?

LucasSte (approved):

solang-parser/src/lexer.rs
seanyoung: I'm not sure what you mean. Are you thinking of `error Error(string);`? We don't support that yet, but the parser still supports it.

xermicus (approved):

rkrasiuk:
@seanyoung thanks! ❤️

PR #1198, opened 2023-02-23
pda does not require a signer in metadata

seanyoung:
If we have a seed, we don't need a signer.

### DIFF
--- a/src/abi/anchor.rs
+++ b/src/abi/anchor.rs
@@ -161,7 +161,7 @@ fn idl_instructions(
                     /// provides an address that does not exist yet, so SystemProgram.CreateAccount is called
                     /// on-chain.
                     ///
-                    /// However, if a @seed is also provided, then program can sign for the account
+                    /// However, if a @seed is also provided, the program can sign for the account
                     /// with the seed using program derived address (pda) when SystemProgram.CreateAccount is called,
                     /// so no signer is required from the client.
                     is_signer: func.has_payer_annotation() && !func.has_seed_annotation(),
--- a/src/abi/anchor.rs
+++ b/src/abi/anchor.rs
@@ -157,7 +157,14 @@ fn idl_instructions(
                 vec![IdlAccountItem::IdlAccount(IdlAccount {
                     name: "dataAccount".to_string(),
                     is_mut: true,
-                    is_signer: func.has_payer_annotation(),
+                    /// With a @payer annotation, the account is created on-chain and needs a signer. The client
+                    /// provides an address that does not exist yet, so SystemProgram.CreateAccount is called
+                    /// on-chain.
+                    ///
+                    /// However, if a @seed is also provided, then program can sign for the account
+                    /// with the seed using program derived address (pda) when SystemProgram.CreateAccount is called,
+                    /// so no signer is required from the client.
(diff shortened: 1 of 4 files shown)


src/abi/anchor.rs
```diff
@@ -157,7 +157,9 @@ fn idl_instructions(
                 vec![IdlAccountItem::IdlAccount(IdlAccount {
                     name: "dataAccount".to_string(),
                     is_mut: true,
-                    is_signer: func.has_payer_annotation(),
+                    /// With @payer annotation, the account is created on-chain; however, if a @seed
+                    /// is provided, then th
```
LucasSte, line 161: This comment is incomplete.

LucasSte (changes requested):

src/abi/anchor.rs
seanyoung: I don't know what happened, must have been distracted half-way through typing that! Fixed

src/abi/anchor.rs
```diff
@@ -157,7 +157,14 @@ fn idl_instructions(
                 vec![IdlAccountItem::IdlAccount(IdlAccount {
                     name: "dataAccount".to_string(),
                     is_mut: true,
-                    is_signer: func.has_payer_annotation(),
+                    /// With @payer annotation, the account is created on-chain and needs a signer. The client
+                    /// provides an address which does not exist yet, so SystemProgram.CreateAccount is called
```
LucasSte, line 161:
```suggestion
                    /// provides an address that does not exist yet, so SystemProgram.CreateAccount is called
```
This is a restrictive clause.

src/abi/anchor.rs
```diff
@@ -157,7 +157,14 @@ fn idl_instructions(
...
+                    /// on-chain.
+                    ///
+                    /// However, if a @seed is provided, then the program can sign for the account
```
LucasSte, line 164:
```suggestion
                    /// However, if a @seed is provided, the program can sign for the account
```
"Omit needless words"

src/abi/anchor.rs
```diff
@@ -157,7 +157,14 @@ fn idl_instructions(
...
+                    /// using the seed using program derived address (pda) when SystemProgram.CreateAccount is called,
```
LucasSte, line 165: These two `using` here seem a bit awkward. Is that what you wanted to say?

xermicus (approved):

src/abi/anchor.rs
seanyoung: I agree, improved the wording

src/abi/anchor.rs
```diff
@@ -157,7 +157,14 @@ fn idl_instructions(
                 vec![IdlAccountItem::IdlAccount(IdlAccount {
                     name: "dataAccount".to_string(),
                     is_mut: true,
-                    is_signer: func.has_payer_annotation(),
+                    /// With a @payer annotation, the account is created on-chain and needs a signer. The client
+                    /// provides an address that does not exist yet, so SystemProgram.CreateAccount is called
+                    /// on-chain.
+                    ///
+                    /// However, if a @seed is also provided, then program can sign for the account
```
LucasSte, line 164:
```suggestion
                    /// However, if a @seed is also provided, the program can sign for the account
```

LucasSte (approved):

PR #1199, opened 2023-02-23
Remove usage of `seal_random` in favor of `instantiation_nonce`

xermicus:
Switch to use the new `instantiation_nonce` for missing salts. `seal_random` will only be provided work for already existing contracts.


src/emit/substrate/target.rs
```diff
@@ -1092,19 +1092,14 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
-            );
-        }
+        let salt = contract_args.salt.unwrap_or_else(|| {
+            call!("instantiation_nonce", &[], "instantiation_nonce")
+                .try_as_basic_value()
+                .left()
+                .unwrap()
+                .into_int_value()
+        });
+        binary.builder.build_store(salt_buf, salt);
```
seanyoung, line 1102: The salt_len is defined as 32 above, but we are saving a i64 into it if the salt is not set.

seanyoung:
deprecate => remove?

LucasSte:
How are you testing this new function?

xermicus:
> deprecate => remove?

I wrote this with the perspective of pallet contracts deprecating it, but for solang it also makes sense to name it a removal

xermicus:
> How are you testing this new function?

This is used in several integration tests. Basically, this code path is taken whenever no `salt` is in the contract instantiation args.

src/emit/substrate/target.rs
seanyoung: The type of salt when specified in the contract is `uint256`, see `fn parse_call_args()`. It might be nicer to make this consistent (although it should remain `uint256` for the evm target, still hope to get that working one day).

LucasSte (approved):

seanyoung (approved):
Those 256 bit values are annoying

PR #1202, opened 2023-02-24
Update LLVM compilation steps

LucasSte:
We must disable ZSTD to build LLVM, so the docs must contain the correct cmake command.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -175,7 +175,8 @@ Now run cmake to create the makefiles. Replace the *installdir* argument to ``CM
 
 	cmake -G Ninja -DLLVM_ENABLE_ASSERTIONS=On '-DLLVM_ENABLE_PROJECTS=clang;lld'  \
 		-DLLVM_ENABLE_TERMINFO=Off -DCMAKE_BUILD_TYPE=Release \
-		-DCMAKE_INSTALL_PREFIX=installdir -B build llvm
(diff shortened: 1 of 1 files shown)


docs/installing.rst
```diff
@@ -175,6 +175,7 @@ Now run cmake to create the makefiles. Replace the *installdir* argument to ``CM
 
 	cmake -G Ninja -DLLVM_ENABLE_ASSERTIONS=On '-DLLVM_ENABLE_PROJECTS=clang;lld'  \
 		-DLLVM_ENABLE_TERMINFO=Off -DCMAKE_BUILD_TYPE=Release \
+        -DLLVM_ENABLE_ZSTD=Off \
```
xermicus, line 178: Indentation seems wrong, also seems to confuse tabs vs. spaces here

LucasSte: The diff seems wrong, but I swear I cannot see this in my computer.

Screenshot 2023-02-24 at 12 36 41

LucasSte: I replaced tabs by spaces, but nothing changed.

xermicus: That's super odd, in the github diff view there a two tabs but the line you introduced just has some spaces

xermicus (approved):
I'm not pedantic about indentation or tabs vs. spaces, please just make sure it looks consistent on the read-the-docs :+1:

PR #1205, opened 2023-02-28
Fix readme example

LucasSte:
The Javascript from the example we have on our README is not working, because:

1. The data account is never created.
2. The RPC to `new` has incorrect parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,40 +86,54 @@ Save the following to `flipper.js`:
 ```javascript
 const { readFileSync } = require('fs');
 const anchor = require('@project-serum/anchor');
+const solana =  require('@solana/web3.js');
 
 const IDL = JSON.parse(readFileSync('./flipper.json', 'utf8'));
 const PROGRAM_SO = readFileSync('./flipper.so');
 
 (async function () {
-	const provider = anchor.AnchorProvider.env();
+    const provider = anchor.AnchorProvider.env();
 
-	const dataAccount = anchor.web3.Keypair.generate();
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -86,40 +86,54 @@ Save the following to `flipper.js`:
...
+    const transaction = new solana.Transaction();
+    transaction.add(
+        solana.SystemProgram.createAccount({
+            fromPubkey: provider.wallet.publicKey,
+            newAccountPubkey: dataAccount.publicKey,
+            lamports,
+            space,
+            programId,
+        }));
+    await provider.sendAndConfirm(transaction, [dataAccount]);
```
seanyoung, line 112: This is not correct. The account is created on-chain

seanyoung (changes requested):

seanyoung:
I've tested it again, and it works for me - as long as `examples/solana/flipper.sol` is used, not `examples/flipper.sol`. 

We should probably rename the latter to `examples/substrate/flipper.sol` and make it clearer in the `README.md`.

@LucasSte would you mind trying it again and making sure it works? Thanks

LucasSte:
> I've tested it again, and it works for me - as long as `examples/solana/flipper.sol` is used, not `examples/flipper.sol`.
> 
> We should probably rename the latter to `examples/substrate/flipper.sol` and make it clearer in the `README.md`.
> 
> @LucasSte would you mind trying it again and making sure it works? Thanks

Yes, it works with `examples/solana/flipper.sol`

PR #1206, opened 2023-02-28
Fix flipper.sol example

LucasSte:
This PR moves `example/flipper.sol` to the substrate examples folder and fixes the references in the docs. It also updates the README example with the correct contract.


README.md
```diff
@@ -32,6 +32,7 @@ For other operating systems, please check the [installation guide](https://solan
 After installing the compiler, write the following to flipper.sol:
 
 ```solidity
+@program_id("F1ipperKF9EfD821ZbbYjS319LXYiBmjhzkkf5a26rC") // For Substrate, this line must be removed.
```
seanyoung, line 35: This is confusing. 

For Solana, you also need a `@payer(..)` tag for the constructor. 

Maybe we should delete the Simple Example

xermicus, line 35: I don't like this either. What about providing two different versions of the file for both targets (e.g. move it to `Build for Solana` and `Build for Substrate` accordingly)?

xermicus (changes requested):

README.md
LucasSte: @xermicus I'm replying your comment here as well.

I believe we can have a separate MD files for Solana and Substrate. These Solang 101 tutorial is good to have in our main repository instead of the documentation.

seanyoung: We have readthedocs documentation already. I think it's better to keep the readme short and to just delete the flipper source code

xermicus: Good idea. The build commands can then use the `flipper.sol` files from the correct directory

xermicus (approved):

seanyoung (approved):

PR #1208, opened 2023-03-02
Simplify switch in constant folding

LucasSte:
I was checking if `fn block_edges` could be pre-calculated during CFG generation, for we call it six times (one more after we collect all Solana accounts). The only problem would be constant folding, which simplifies branches. While I was checking that, I realized we could simplify switches before trying to remove `fn block_edges`.


src/codegen/constant_folding.rs
```diff
@@ -329,6 +329,22 @@ pub fn constant_folding(cfg: &mut ControlFlowGraph, ns: &mut Namespace) {
...
+                        for (match_item, block) in &cases {
+                            if let Expression::NumberLiteral(_, _, match_num) = match_item {
+                                if match_num == num {
+                                    simplified_branch = Some(*block);
+                                }
+                            }
+                        }
+                        cfg.blocks[block_no].instr[instr_no] = Instr::Branch {
+                            block: simplified_branch.unwrap_or(*default),
+                        };
```
seanyoung, line 344: This could be written more succinctly with an rust  iterator.

seanyoung, line 344: This could be written more succinctly with an rust iterator.

seanyoung (approved):

xermicus (approved):

PR #1211, opened 2023-03-04
Refactor scale decoder into codegen

xermicus:
Decoding for `abi.decode` is now done in codegen. Like with the encoder, most of the code can be shared with the borsh decoder.


seanyoung (approved):
LGTM

tests/codegen_testcases/solidity/borsh_decoding_complex_types.sol
```diff
@@ -12,10 +12,10 @@ contract Testing {
 	    
         // CHECK: ty:uint32 %temp.10 = (builtin ArrayLength ((arg #0)))
 	    // CHECK: ty:uint32 %temp.12 = uint32 0
+	    // CHECK: ty:uint32 %temp.13 = (builtin ReadFromBuffer ((arg #0), uint32 0))
```
LucasSte, line 15: The instruction is misplaced because you are not validating the offset to read the `string[]` length.

tests/codegen_testcases/solidity/borsh_decoding_complex_types.sol
```diff
@@ -33,34 +33,28 @@ contract Testing {
 	    // CHECK: branch block3
         
         // CHECK: block5: # body
-	    // CHECK: ty:uint32 %1.cse_temp = (%temp.12 + uint32 4)
+	    // CHECK: ty:uint32 %temp.16 = (builtin ReadFromBuffer ((arg #0), %temp.12))
```
LucasSte, line 36: Same as above, but now we are not validating the offset to read the `string` length.

tests/codegen_testcases/solidity/borsh_decoding_complex_types.sol
```diff
@@ -70,67 +64,61 @@ contract Testing {
...
-        // CHECK: assert-failure
+		// CHECK: block5: # cond 
+		// CHECK: branchcond (unsigned less %for_i_0.temp.25 < (builtin ArrayLength (%temp.21))), block7, block8
+
+		// CHECK: block6: # next 
+		// CHECK: ty:uint32 %for_i_0.temp.25 = (%for_i_0.temp.25 + uint32 1)
+		// CHECK: branch block5
+		
+		// CHECK: block7: # body
+		// CHECK: ty:uint32 %temp.26 = (builtin ReadFromBuffer ((arg #0), %temp.22))
```
LucasSte, line 98: No validation was performed to read the string length.

tests/codegen_testcases/solidity/borsh_decoding_complex_types.sol
```diff
@@ -139,99 +127,95 @@ contract Testing {
...
-        // CHECK: block2: # out_of_bounds
-	    // CHECK: assert-failure
-        
-        // CHECK: block3: # cond
-	    // CHECK: branchcond (unsigned less %for_i_0.temp.37 < (builtin ArrayLength (%temp.33))), block5, block6
-        
-        // CHECK: block4: # next
-	    // CHECK: ty:uint32 %for_i_0.temp.37 = (%for_i_0.temp.37 + uint32 1)
-	    // CHECK: branch block3
+		// CHECK: block0: # entry
```
LucasSte, line 130: This is the validation thing again. Please, do not just copy and paste new codegen code assuming it is all correct. This test is here to ensure the correct flow path exist when we build the contract.

tests/codegen_testcases/solidity/borsh_decoding_complex_types.sol
```diff
@@ -139,99 +127,95 @@ contract Testing {
...
-	    // CHECK: ty:uint32 %temp.44 = (builtin ReadFromBuffer ((arg #0), %temp.40))
-	    // CHECK: ty:uint32 %4.cse_temp = ((%temp.44 + uint32 4) + %temp.40)
-	    // CHECK: branchcond (%4.cse_temp <= %temp.32), block17, block18
-    
-        // CHECK: block16: # out_of_bounds
-	    // CHECK: assert-failure
+		// CHECK: block7: # inbounds
+		// CHECK: ty:uint64 %temp.36 = (builtin ReadFromBuffer ((arg #0), %temp.32))
+		// CHECK: ty:uint32 %temp.38 = %1.cse_temp
+		// CHECK: ty:uint32 %temp.39 = (builtin ReadFromBuffer ((arg #0), %temp.38))
```
LucasSte, line 165: No validation before read.

tests/codegen_testcases/solidity/borsh_decoding_simple_types.sol
```diff
@@ -130,49 +130,38 @@ contract Testing {
...
-	    // CHECK: branchcond (uint32 4 <= %temp.86), block1, block2
-
-        // CHECK: block1: # inbounds
-	    // CHECK: ty:uint32 %temp.87 = (builtin ReadFromBuffer ((arg #0), uint32 0))
-	    // CHECK: ty:uint32 %1.cse_temp = ((%temp.87 + uint32 4) + uint32 0)
-	    // CHECK: branchcond (%1.cse_temp <= %temp.86), block3, block4
+		// CHECK: block0: # entry
+		// CHECK: ty:bytes %buffer = (arg #0)
+		// CHECK: ty:uint32 %temp.86 = (builtin ArrayLength ((arg #0)))
+		// CHECK: ty:uint32 %temp.87 = (builtin ReadFromBuffer ((arg #0), uint32 0))
```
LucasSte, line 136: Validation is missing.

tests/codegen_testcases/solidity/borsh_decoding_simple_types.sol
```diff
@@ -184,21 +173,26 @@ contract Testing {
     // BEGIN-CHECK: Testing::Testing::function::decodeEnum__bytes
     function decodeEnum(bytes memory buffer) public pure returns (WeekDays) {
         WeekDays a = abi.decode(buffer, (WeekDays));
-	    // CHECK: ty:uint32 %temp.94 = (builtin ArrayLength ((arg #0)))
-	    // CHECK: branchcond (uint32 1 <= %temp.94), block1, block2
 
-        // CHECK: block1: # inbounds
-	    // CHECK: ty:enum Testing.WeekDays %temp.95 = (builtin ReadFromBuffer ((arg #0), uint32 0))
-	    // CHECK: branchcond (unsigned less uint32 1 < %temp.94), block3, block4
+		// CHECK: block0: # entry
```
LucasSte, line 177: This line wasn't in the original test.

tests/codegen_testcases/solidity/borsh_decoding_complex_types.sol
```diff
@@ -70,67 +64,61 @@ contract Testing {
...
         NonConstantStruct memory cte = abi.decode(buffer, (NonConstantStruct));
 
-	    // CHECK: ty:uint32 %temp.20 = (builtin ArrayLength ((arg #0)))
-	    // CHECK: branchcond (uint32 8 <= %temp.20), block1, block2
-        
-        // CHECK: block1: # inbounds
-		// CHECK: ty:uint64 %temp.21 = (builtin ReadFromBuffer ((arg #0), uint32 0))
-	    // CHECK: ty:uint32 %temp.23 = uint32 8
-	    // CHECK: branchcond (uint32 12 <= %temp.20), block3, block4
+		// CHECK: ty:bytes %buffer = (arg #0)
```
LucasSte, line 67: This line wasn't in the original test, but it is ok to add it.

tests/codegen_testcases/solidity/borsh_decoding_simple_types.sol
```diff
@@ -217,27 +211,30 @@ contract Testing {
...
-
-        // CHECK: block1: # inbounds
-	    // CHECK: ty:struct Testing.noPadStruct %temp.97 = struct {  }
-	    // CHECK: memcpy src: %buffer, dest: %temp.97, bytes_len: uint32 8
-	    // CHECK: ty:uint128 %temp.98 = (builtin ReadFromBuffer ((arg #0), uint32 8))
-	    // CHECK: ty:uint8 %temp.99 = (builtin ReadFromBuffer ((arg #0), uint32 24))
-	    // CHECK: ty:bytes32 %temp.100 = (builtin ReadFromBuffer ((arg #0), uint32 25))
-	    // CHECK: ty:struct Testing.PaddedStruct %temp.101 = struct { %temp.98, %temp.99, %temp.100 }
-	    // CHECK: branchcond (unsigned less uint32 57 < %temp.96), block3, block4
+		// CHECK: block0: # entry
```
LucasSte, line 214: This line wasn't in the test.

tests/codegen_testcases/solidity/borsh_decoding_simple_types.sol
```diff
@@ -217,27 +211,30 @@ contract Testing {
...
+		// CHECK: ty:bytes %buffer = (arg #0)
```
LucasSte, line 215: Neither this one.

tests/codegen_testcases/solidity/borsh_decoding_simple_types.sol
```diff
@@ -246,42 +243,38 @@ contract Testing {
...
+		// CHECK: ty:uint32 %temp.101 = (builtin ArrayLength ((arg #0)))
+		// CHECK: branchcond (uint32 32 <= %temp.101), block1, block2
+
+		// CHECK: block1: # inbounds
+		// CHECK: ty:uint32[4] %temp.102 =  [  ]
+		// CHECK: memcpy src: %buffer, dest: %temp.102, bytes_len: uint32 16
+		// CHECK: ty:struct Testing.noPadStruct[2] %temp.103 =  [  ]
+		// CHECK: memcpy src: (advance ptr: %buffer, by: uint32 16), dest: %temp.103, bytes_len: uint32 16
+		// CHECK: ty:uint32 %temp.104 = (builtin ReadFromBuffer ((arg #0), uint32 32))
```
LucasSte, line 256: Read from buffer not validated.

tests/codegen_testcases/solidity/borsh_decoding_simple_types.sol
```diff
@@ -246,42 +243,38 @@ contract Testing {
...
-
-        // CHECK: block7: # not_all_bytes_read
-	    // CHECK: assert-failure
-
-        // CHECK: block8: # buffer_read
-	    // CHECK: ty:uint32[4] %a = %temp.103
-	    // CHECK: ty:struct Testing.noPadStruct[2] %b = %temp.104
-	    // CHECK: ty:struct Testing.noPadStruct[] %c = %temp.106
+		// CHECK: block0: # entry
+		// CHECK: ty:bytes %buffer = (arg #0)
```
LucasSte, line 247: These lines were not originally in the test.

tests/codegen_testcases/solidity/encode_decode_double_dynamic_array.sol
```diff
@@ -89,10 +89,10 @@ contract Testing {
 
 	    // CHECK: ty:uint32 %temp.14 = (builtin ArrayLength ((arg #0)))
 	    // CHECK: ty:uint32 %temp.16 = uint32 0
+	    // CHECK: ty:uint32 %temp.17 = (builtin ReadFromBuffer ((arg #0), uint32 0))
```
LucasSte, line 92: Not validated.

tests/codegen_testcases/solidity/encode_decode_double_dynamic_array.sol
```diff
@@ -110,6 +110,7 @@ contract Testing {
         // CHECK: branch block3
 
         // CHECK: block5: # body
+	    // CHECK: ty:uint32 %temp.20 = (builtin ReadFromBuffer ((arg #0), %temp.16))
```
LucasSte, line 113: Not validated.

tests/substrate_tests/function_types.rs
```diff
@@ -330,3 +330,49 @@ fn ext() {
...
+        }
+        "##,
+    );
+
+    runtime.function("it", vec![]);
+    let mut a = runtime.vm.output.clone();
+
+    runtime.set_program(1);
+    runtime.function("encode_decode_call", a.clone());
+    assert_eq!(runtime.vm.output, 127u8.encode());
```
LucasSte, line 373: Shouldn't `a` appear in the output too?

tests/substrate_tests/strings.rs
```diff
@@ -613,6 +623,29 @@ fn string_escape() {
...
+    // "Call me Ishmael. Some years ago—never mind how long precisely—having little or no money in my purse, and nothing particular to interest me on shore, I thought I would sail about a little and see the watery part of the world. It is a way I have of driving off the spleen and regulating the circulation. Whenever I find myself growing grim about the mouth; whenever it is a damp, drizzly November in my soul; whenever I find myself involuntarily pausing before coffin warehouses, and bringing up the rear of every funeral I meet; and especially whenever my hypos get such an upper hand of me, that it requires a strong moral principle to prevent me from deliberately stepping into the street, and methodically knocking people’s hats off—then, I account it high time to get to sea as soon as I can. This is my substitute for pistol and ball. With a philosophical flourish Cato throws himself upon his sword; I quietly take to the ship. There is nothing surprising in this. If they but knew it, almost all men in their degree, some time or other, cherish very nearly the same feelings towards the ocean with me."
```
LucasSte, line 629: You can save this to a variable and test if the string contents are properly decoded.

LucasSte (changes requested):

src/codegen/encoding/mod.rs
```diff
@@ -546,6 +607,567 @@ pub(super) trait AbiEncoding {
...
+                vartab,
+                Instr::MemCopy {
+                    source: source_address,
+                    destination: array_expr.clone(),
+                    bytes: array_length.clone(),
+                },
+            );
+
+            let bytes_size = if matches!(dims.last(), Some(ArrayLength::Dynamic)) {
+                increment_by(array_length, size_length)
```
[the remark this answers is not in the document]
xermicus: Why not, I was thinking about this too

src/codegen/encoding/mod.rs
```diff
@@ -546,6 +607,567 @@ pub(super) trait AbiEncoding {
...
+            }
+
+            Type::DynamicBytes | Type::String => {
+                // String and Dynamic bytes are encoded as size + elements
+                let (array_length_var, size_length) =
+                    self.retrieve_array_length(buffer, offset, vartab, cfg);
+                let array_length = Expression::Variable(Codegen, Uint(32), array_length_var);
+                let size = increment_by(array_length.clone(), size_length.clone());
+                validator.validate_offset(
+                    increment_by(offset.clone(), size.clone()),
```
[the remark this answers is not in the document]
xermicus: The first time I need a pointer which is (offset + total size) for the OOB check, but the second time I need a pointer to the start of the encoded data (without the vector size), so they are not the same?

src/codegen/encoding/mod.rs
```diff
@@ -546,6 +607,567 @@ pub(super) trait AbiEncoding {
...
+                    ns,
+                    vartab,
+                    cfg,
+                );
+
+                let allocated_array = allocate_array(ty, array_length_var, vartab, cfg);
+                let advanced_pointer = Expression::AdvancePointer {
+                    pointer: buffer.clone().into(),
+                    bytes_offset: increment_by(offset.clone(), size_length).into(),
```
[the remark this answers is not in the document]
xermicus: See above comment

src/codegen/encoding/scale_encoding.rs
```diff
@@ -22,6 +23,158 @@ impl ScaleEncoding {
...
+            cond,
+            cases: cases.to_vec(),
+            default,
+        },
+    );
+
+    let done = cfg.new_basic_block("done".into());
+    // sizes of 2**30 (1GB) or larger are not allowed
+    cfg.set_basic_block(default);
+    cfg.add(vartab, Instr::AssertFailure { encoded_args: None });
```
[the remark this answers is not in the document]
xermicus: It means that it will just trap for sizes too big. I updated the comment to make it more clear.

src/codegen/encoding/mod.rs
```diff
@@ -546,6 +607,567 @@ pub(super) trait AbiEncoding {
...
+
+                let read_expr = Expression::Variable(Codegen, ty.clone(), read_var);
+
+                (read_expr, size)
+            }
+
+            Type::DynamicBytes | Type::String => {
+                // String and Dynamic bytes are encoded as size + elements
+                let (array_length_var, size_length) =
+                    self.retrieve_array_length(buffer, offset, vartab, cfg);
```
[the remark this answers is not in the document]
LucasSte: LLVM does not differentiate between signed and unsigned integers. So, let's say that we don't have the last four bytes to read when retrieving the length, but only two (offset + 2). We can read a -3 (remember, in LLVM everything is signed), so we would validate an offset of 4 (size width) + (-3) + offset = offset + 1. This is within bounds. Now we have an access violation when trying to read the string.

Check the example below. Imagine the buffer has only two bytes (an int16), but you are going to read four bytes, because no validation happens. If the extra two bytes you read happen to be `[255, 255]`, the number you read is -4. Your validator will try to validate 4 + (-4) + offset = offset, which is within bounds. When we try to read the string, we are going to have a segmentation fault (Access violation on Solana VM).

```Rust
#[test]
fn validation() {
    #[derive(Debug, BorshSerialize)]
    struct Input {
        item_1: i16,
        item_2: i16, // Rubbish: imagine this as the extra two bytes you are reading.
    }

    let mut vm = build_solidity(
        r#"
    contract Testing {
        function test(bytes memory buffer) public view returns (string) {
            string ss = abi.decode(buffer, string);

            return ss;
        }
    }
        "#,
    );

    vm.constructor(&[]);

    let input = Input {
        item_1: -4,
        item_2: -1,
    };
    let encoded = input.try_to_vec().unwrap();

    let res = vm.function("test", &[BorshToken::Bytes(encoded)]);
    std::println!("{:?}\n", res);
}

```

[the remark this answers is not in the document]
LucasSte: > 2. I was told that the buffer always has some margin at the end anyways.

I think there might be space between heap allocated memory, but we cannot always assume that. The allocate functions we have allocate the exact number of bytes we request.

src/codegen/encoding/mod.rs
```diff
@@ -546,6 +607,567 @@ pub(super) trait AbiEncoding {
...
+                    );
+                    (
+                        Expression::NumberLiteral(Codegen, Uint(32), elem_no),
+                        Expression::NumberLiteral(Codegen, Uint(32), 0.into()),
+                        offset.clone(),
+                        allocated_vector,
+                    )
+                } else {
+                    let (array_length, size_length) =
+                        self.retrieve_array_length(buffer, offset, vartab, cfg);
```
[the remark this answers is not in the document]
LucasSte: The validation should also appear here, as we commented.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
seanyoung: > > 2. I was told that the buffer always has some margin at the end anyways.
> 
> I think there might be space between heap allocated memory, but we cannot always assume that. The allocate functions we have allocate the exact number of bytes we request.

The current heap implementation does always have some bytes at the end, for house keeping.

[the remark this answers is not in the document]
seanyoung: > > LLVM does not differentiate between signed and unsigned integers
> 
> Are numbers in LLVM always signed? I don't get it, surely you can do unsigned operations? Why are we using a signed number for the offset then, while it doesn't make sense for something like a buffer offset to be negative first place?

That's not strictly true. Variables / values are neither signed nor unsigned. Operations are signed or unsigned like signed div/unsigned div. Add/Sub/Mul are not signed either, but the overflow checking versions are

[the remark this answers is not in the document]
xermicus: Thanks, it is how I expect this to work then, and so the scale version then should also work even for bogus data (given #1216 first):  
- Parsing a smaller value should trap the contract latest when the decoding is finished, because not all bytes were read
- Parsing a larger value should trap the contract because of the following OOB read

@LucasSte WDYT?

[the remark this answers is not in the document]
LucasSte: I think that's right Cyril. We still need to validate if we haven't read rubbish data. After reading the `array_length`, we should check if `offset + size_width` (please, don't call this `size_length`) is within bounds. This check should happen before the validation of `offset + size_width + array_length`.

src/codegen/encoding/mod.rs
```diff
@@ -546,6 +606,563 @@ pub(super) trait AbiEncoding {
...
+        dims: &[ArrayLength],
+        validator: &mut BufferValidator,
+        ns: &Namespace,
+        vartab: &mut Vartable,
+        cfg: &mut ControlFlowGraph,
+    ) -> (Expression, Expression) {
+        // Checks if we can memcpy the elements from the buffer directly to the allocated array
+        if allow_direct_copy(array_ty, elem_ty, dims, ns) {
+            // Calculate number of elements
+            let (array_size, size_width, offset, var_no) =
```
LucasSte, line 784: I suggest the first variable to be called `array_bytes_size`.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: We should validate that the bytes read for `array_length` are within the array bounds.

src/codegen/encoding/scale_encoding.rs
```diff
@@ -198,49 +352,50 @@ impl AbiEncoding for ScaleEncoding {
...
-    fn abi_decode(
+    fn decode_external_function(
         &self,
-        loc: &Loc,
         buffer: &Expression,
-        types: &[Type],
-        _ns: &Namespace,
+        offset: &Expression,
+        ty: &Type,
+        validator: &mut super::buffer_validator::BufferValidator,
```
LucasSte, line 360: Do you need to specify the complete path?

src/codegen/mod.rs
```diff
@@ -646,7 +646,20 @@ impl RetrieveType for Expression {
     }
 }
 
+/// Increment an expression by some value.
+///
+/// Assumes
 impl Expression {
+    pub(crate) fn add_u32(&self, other: &Expression) -> Self {
```
LucasSte, line 653:
```suggestion
    pub(crate) fn add_u32(self, other: Expression) -> Self {
```
We end up consuming the expressions we add, so a suggestion would be to move the values instead of receiving the reference. We can leave the choice wether to clone the expression to where this function is called.

LucasSte: If you implement this, don't forget to remove the clones.

src/codegen/mod.rs
```diff
@@ -646,7 +646,20 @@ impl RetrieveType for Expression {
     }
 }
 
+/// Increment an expression by some value.
+///
+/// Assumes
```
LucasSte, line 651: What does it assume?

tests/codegen_testcases/solidity/borsh_decoding_complex_types.sol
LucasSte: The validation here is for `offset + size_width + array_length`. I was talking about validating `offset + size_width`. As we don't have overflow detection turned on by default, `array_length` can be negative, so the former validation is going to be wrong.

tests/codegen_testcases/solidity/scale.sol
```diff
@@ -20,38 +20,76 @@ contract ExternalFunctions {
...
     function storage_callback() public {
-        // CHECK: %temp.6 = load storage slot(uint256 0) ty:function(int32) external returns (uint64)
-        // CHECK: ty:bytes %abi_encoded.temp.7 = (alloc bytes len uint32 40)
-        // CHECK: writebuffer buffer:%abi_encoded.temp.7 offset:uint32 0 value:hex"f503f5fe"
-        // CHECK: writebuffer buffer:%abi_encoded.temp.7 offset:uint32 4 value:(load (struct function(int32) external returns (uint64)(%temp.6) field 1))
-        // CHECK: writebuffer buffer:%abi_encoded.temp.7 offset:uint32 36 value:(load (struct function(int32) external returns (uint64)(%temp.6) field 0))
-        // CHECK: _ = external call::regular address:(builtin GetAddress ()) payload:%abi_encoded.temp.7 value:uint128 0 gas:uint64 0 accounts: seeds:
+        // CHECK: %temp.7 = load storage slot(uint256 0) ty:function(int32) external returns (uint64)
+        // CHECK: ty:bytes %abi_encoded.temp.8 = (alloc bytes len uint32 40)
+        // CHECK: writebuffer buffer:%abi_encoded.temp.8 offset:uint32 0 value:hex"f503f5fe"
```
LucasSte, line 25: You should add the checks for the branches here.

tests/codegen_testcases/solidity/scale.sol
```diff
@@ -20,38 +20,76 @@ contract ExternalFunctions {
...
+    // CHECK: block12: # done
+    // CHECK: memcpy src: (arg #0), dest: (advance ptr: %abi_encoded.temp.10, by: (uint32 0 + %temp.12)), bytes_len: %temp.11
+    // CHECK: return 
+
+    // CHECK: block13: # fail
+    // CHECK: assert-failure
+
+    // CHECK: block14: # prepare
+    // CHECK: branchcond (unsigned more %temp.11 > uint32 63), block10, block8
+
```
LucasSte, line 92: No tests for the decoder?
At least a test for the part that is not common with bosh is welcome.

xermicus: It is only decoding of compact integers and external functions which are covered elsewhere. But yeah it doesn't hurt to add a codegen test case for them too :+1:

LucasSte (approved):

PR #1220, opened 2023-03-14
Remove non-idiomatic use of LinkedList

seanyoung:
See the note here:

https://doc.rust-lang.org/std/collections/struct.LinkedList.html

### DIFF
--- a/src/codegen/statements.rs
+++ b/src/codegen/statements.rs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use num_bigint::BigInt;
-use std::collections::LinkedList;
 
 use super::encoding::abi_encode;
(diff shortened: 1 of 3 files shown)


LucasSte (approved):
Have you run any benchmark?

seanyoung:
> Have you run any benchmark?

What's your concern?

xermicus (approved):

LucasSte:
> > Have you run any benchmark?
> 
> What's your concern?

I was curious if these changes provided any speedup in the compiler.

seanyoung:
> > > Have you run any benchmark?
> > 
> > 
> > What's your concern?
> 
> I was curious if these changes provided any speedup in the compiler.

I didn't run any benchmarks. I suspect there little difference even in a micro-benchmark, and no difference in the overall compiler speed.

PR #1222, opened 2023-03-14
fix compilation failure with -g flag in substrate integration

salaheldinsoliman:
This PR fixes compilation error when inserting the `-g` flag in Substrate integration tests

### DIFF
--- a/src/emit/binary.rs
+++ b/src/emit/binary.rs
@@ -61,17 +61,14 @@ impl<'a> Binary<'a> {
         context: &'a Context,
         contract: &'a Contract,
         ns: &'a Namespace,
-        filename: &'a str,
         opt: &'a Options,
     ) -> Self {
         let std_lib = load_stdlib(context, &ns.target);
         match ns.target {
             Target::Substrate { .. } => {
-                substrate::SubstrateTarget::build(context, &std_lib, contract, ns, filename, opt)
+                substrate::SubstrateTarget::build(context, &std_lib, contract, ns, opt)
             }
-            Target::Solana => {
-                solana::SolanaTarget::build(context, &std_lib, contract, ns, filename, opt)
-            }
+            Target::Solana => solana::SolanaTarget::build(context, &std_lib, contract, ns, opt),
             Target::EVM => unimplemented!(),
         }
     }
--- a/src/emit/cfg.rs
+++ b/src/emit/cfg.rs
@@ -193,6 +193,18 @@ pub(super) fn emit_cfg<'a, T: TargetRuntime<'a> + ?Sized>(
                         None,
                     );
                     bin.builder.set_current_debug_location(debug_loc);
+                } else {
+                    // For instructions that do not have a location, insert a debug location pointing to line 0.
+                    // If -g flag is enabled, every instruction should have a debug location. This is necessary
+                    // because llvm's inliner pass requires function call instructions to have a debug location.
+                    let debug_loc = dibuilder.create_debug_location(
+                        bin.context,
+                        0_u32,
+                        0_u32,
+                        di_func_scope.unwrap().as_debug_info_scope(),
+                        None,
+                    );
+                    bin.builder.set_current_debug_location(debug_loc);
                 }
             }
 
--- a/integration/substrate/create_contract.sol
+++ b/integration/substrate/create_contract.sol
@@ -1,9 +1,8 @@
-
 contract creator {
-    child public c;
+    child_create_contract public c;
 
     function create_child() public {
-        c = new child{value: 1e15}();
+        c = new child_create_contract{value: 1e15}();
     }
 
     function call_child() public view returns (string memory) {
@@ -11,10 +10,10 @@ contract creator {
     }
 }
 
-contract child {
+contract child_create_contract {
     constructor() payable {}
 
-    function say_my_name() pure public returns (string memory) {
+    function say_my_name() public pure returns (string memory) {
         return "child";
     }
 }
--- a/integration/substrate/create_contract.spec.ts
+++ b/integration/substrate/create_contract.spec.ts
@@ -23,7 +23,7 @@ describe('Deploy create_contract contract and test', () => {
         let deploy_contract = await deploy(conn, alice, 'creator.contract', BigInt(1e16));
 
         // we need to have upload the child code
-        let _ = await deploy(conn, alice, 'child.contract', BigInt(0));
+        let _ = await deploy(conn, alice, 'child_create_contract.contract', BigInt(0));
 
         let contract = new ContractPromise(conn, deploy_contract.abi, deploy_contract.address);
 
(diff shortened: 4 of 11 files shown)


src/emit/cfg.rs
```diff
@@ -193,6 +193,18 @@ pub(super) fn emit_cfg<'a, T: TargetRuntime<'a> + ?Sized>(
                         None,
                     );
                     bin.builder.set_current_debug_location(debug_loc);
+                } else {
+                    // For instructions that are inserted in codegen, insert a debug location pointing to line 0.
```
LucasSte, line 197:
```suggestion
                    // For instructions do not have a location, insert a debug location pointing to line 0.
```

src/emit/cfg.rs
```diff
@@ -193,6 +193,18 @@ pub(super) fn emit_cfg<'a, T: TargetRuntime<'a> + ?Sized>(
...
+                    // If -g flag is enabled, each instruction should have a debug location. This is necessary
```
LucasSte, line 198:
```suggestion
                    // If -g flag is enabled, every instruction should have a debug location. This is necessary
```

LucasSte (approved):

xermicus (approved):
Thanks

src/emit/binary.rs
```diff
@@ -213,7 +213,7 @@ impl<'a> Binary<'a> {
         let (dibuilder, compile_unit) = module.create_debug_info_builder(
             true,
             inkwell::debug_info::DWARFSourceLanguage::C,
-            filename,
+            name,
```
seanyoung, line 216: This argument to `create_debug_info_builder()` is the filename, changing it to the contract is not really the correct solution.

Having said that, this does hint towards the problem. Somehow the filename is not consistent.

Maybe on this line https://github.com/hyperledger/solang/blob/main/src/emit/mod.rs#L427 your new function `.filename()` should be used:


```rust
let filename = ns.files[self.loc.file_no()].filename();
```
That's a good idea anyway because of deterministic builds.

seanyoung: Note that `self.loc` on this line should be the location of the contract, so if that is not producing the right file name, then there is an issue in location tracking.

salaheldinsoliman: I tried this locally and I don't know why it works. In that case, the parent contract filename would be the filename of the child contract. (the child.ll file would change if I compile the contract from the CL, from when it is re-compiled when compiling the parent contract).
So my theory I mentioned on Discord is wrong :), could you explain this please? Thanks!

seanyoung: Actually I think the problem is elsewhere. Both `create_contract.sol` and `runtime_errors.sol` have a contract definition for `child`.

salaheldinsoliman: > Actually I think the problem is elsewhere. Both `create_contract.sol` and `runtime_errors.sol` have a contract definition for `child`.

But the definition for child is in the same file, so that will not cause a problem

seanyoung:
```
$ grep 'contract child' *.sol
create_contract.sol:contract child {
runtime_errors.sol:contract child {
```

salaheldinsoliman: I think the problem lies in that the parent filename is the one being provided to all child contracts. What I did wrong was provide the contract name.
The right solution here would be providing the filename in which the contract is defined, I did that in  my latest commit. 
Thanks :)

seanyoung:
You've left in a comment which nicely demonstrates the problem.

```rust
println!("NEW IS CALLED WITH FILENAME {filename} and NAME: {name}");
```

The output contains:
```
NEW IS CALLED WITH FILENAME create_contract.sol and NAME: child
NEW IS CALLED WITH FILENAME runtime_errors.sol and NAME: child
```
Both these contracts define the contract `child`. One of those should be renamed.

src/emit/binary.rs
```diff
@@ -189,11 +186,12 @@ impl<'a> Binary<'a> {
         context: &'a Context,
         target: Target,
         name: &str,
-        filename: &str,
+        filename: String,
         opt: &'a Options,
         std_lib: &Module<'a>,
         runtime: Option<Box<Binary<'a>>>,
     ) -> Self {
+        println!("NEW IS CALLED WITH FILENAME {filename} and NAME: {name}");
```
xermicus, line 194: I think this slipped in, please remove

src/emit/binary.rs
```diff
@@ -189,11 +186,12 @@ impl<'a> Binary<'a> {
         context: &'a Context,
         target: Target,
         name: &str,
-        filename: &str,
+        filename: String,
```
xermicus, line 189: Nitpick: Why was this changed? You use `filename.as_str()` later on anyways. In general, if you do not need ownership over a value, it is better to ask for a reference.

src/emit/cfg.rs
```diff
@@ -193,6 +193,18 @@ pub(super) fn emit_cfg<'a, T: TargetRuntime<'a> + ?Sized>(
                         None,
                     );
                     bin.builder.set_current_debug_location(debug_loc);
+                } else {
+                    // For instructions that do not have a location, insert a debug location pointing to line 0.
```
xermicus, line 197: Nice comment, thanks

seanyoung:
> You've left in a comment which nicely demonstrates the problem.
> 
> ```rust
> println!("NEW IS CALLED WITH FILENAME {filename} and NAME: {name}");
> ```
> 
> The output contains:
> 
> ```
> NEW IS CALLED WITH FILENAME create_contract.sol and NAME: child
> NEW IS CALLED WITH FILENAME runtime_errors.sol and NAME: child
> ```
> 
> Both these contracts define the contract `child`. One of those should be renamed.

I suspect that the tests pass because it's being compiled with gnu parallel. The test in runtime_errors.sol expect the create_child to fail, however the one in create_contract.sol must be right. So, depending on which gets compiled first this might work or not work.

xermicus:
That is a good finding @seanyoung, I created #1228 to prevent this in the future (can also be enabled for solana integration tests). The CI in this PR should fail now, I'm currently just waiting to check and will then fix the conflicting names in that PR too

xermicus:
> Thanks @xermicus , I will change the conflicting names in this PR myself

Very nice, thanks

salaheldinsoliman:
> @salaheldinsoliman the substrate tests fails with:
> 
> ```
> 
>   33 passing (10s)
>   1 pending
>   1 failing
> 
>   1) Deploy runtime_errors.sol and test the debug buffer
>        logs errors to the debug buffer:
>      Error: ENOENT: no such file or directory, open 'child.contract'
>       at Object.openSync (fs.js:498:3)
>       at Object.readFileSync (fs.js:394:35)
>       at deploy (index.js:30:39)
>       at Context.<anonymous> (runtime_errors.spec.ts:25:42)
>       
> ```
> 
> The CI tests then doesn't exit, so the test gets cancelled or it timed out

Fixed, thanks

PR #1223, opened 2023-03-17
Remove unused argument to parse_number

seanyoung:
2nd argument to `fn parse_number()` in the lexer is not used, remove.

### DIFF
--- a/solang-parser/src/lexer.rs
+++ b/solang-parser/src/lexer.rs
@@ -559,8 +559,7 @@ impl<'input> Lexer<'input> {
 
     fn parse_number(
         &mut self,
-        start: usize,
-        end: usize,
+        mut start: usize,
         ch: char,
     ) -> Result<(usize, Token<'input>, usize), LexicalError> {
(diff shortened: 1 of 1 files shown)


xermicus (approved):

solang-parser/src/lexer.rs
```diff
@@ -605,7 +604,7 @@ impl<'input> Lexer<'input> {
             start -= 1;
```
LucasSte, line 604: You can remove line 601 and make `start` a mutable parameter.

LucasSte (approved):

solang-parser/src/lexer.rs
seanyoung: We use start again on line 607, so this would not work

LucasSte: I don't understand why not. You create a new variable with the same name on line 601.

seanyoung: You're right, fixed

PR #1224, opened 2023-03-17
Treat enums as 8bit uint in constant hashing

xermicus:
Closes https://github.com/xermicus/fuzzy-sol/issues/51
Closes https://github.com/xermicus/fuzzy-sol/issues/67

### DIFF
--- a/tests/codegen_testcases/solidity/constant_folding.sol
+++ b/tests/codegen_testcases/solidity/constant_folding.sol
@@ -18,6 +18,7 @@ contract Enum {
     }
     mapping(Lender => address) lenders;
 
+    // BEGIN-CHECK: Enum::Enum::constructor::7d622c65
     constructor(
         address usdt,
         address usdc,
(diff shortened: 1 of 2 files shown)


LucasSte:
Please, add a test case for this fix.

xermicus:
> Please, add a test case for this fix.

Done

seanyoung (approved):

tests/codegen_testcases/solidity/constant_folding.sol
```diff
@@ -2,9 +2,40 @@
...
+
+contract Enum {
+    enum Lender {
+        USDT,
+        USDC,
+        DAI
+    }
+    mapping(Lender => address) lenders;
+
+    constructor(
```
LucasSte, line 21: I think you need to add the `BEGIN-CHECK` here, so the tester code can split each function test.

LucasSte (approved):

Issue #786, opened 2022-05-01
Be clear on solana support


deepaktrama:
Thanks for adding solana support and keeping it open source. It will help accelerate solana dev.
However it's not clear if this is ready for production on solana or is it really very alpha/work in progress.

https://solang.readthedocs.io/en/latest/targets.html just mentions a few differences (excerpt below), reading this I assumed it was ready for prime time use.
Request you to please correct me if I am wrong and update the docs accordingly.

Target Specific
Solana
The Solana target requires [Solana](https://www.solana.com/) v1.8.1.

Solana has the following differences to Ethereum Solidity:

The address type is 32 bytes, not 20 bytes. This is what Solana calls an “account”
An address literal has to be specified using the address"36VtvSbE6jVGGQytYWSaDPG7uZphaxEjpJHUUpuUbq4D" syntax
There is no ecrecover() builtin function, but there is a signatureVerify() function which can check ed25519 signatures.
Solana has no concept of gas, so there is no gas functions
Solana balance is stored in a uint64, so msg.value, address .balance, .transfer() and .send() all use uint64 rather than uint256.

seanyoung:
@deepaktrama Solana support is very new, and has not been widely tested (as it is new). There are two developers working on Solang full-time, both are employed by Solana Labs, and Solana support is our priority.

Solana support should be working. solang does lack spl-token support and cannot to call non-solidity contracts, we are working on this.

There are some issues that need addressing, we will answer them as soon as possible.

seanyoung:
I think we should add some words to the documentation around:
- what works today
- what it being worked on
- what sort of support can you expect

@deepaktrama would that be what you had in mind?

deepaktrama:
@seanyoung yes, those are good points and I also suggest to add below points with reference to SOLANA which will help increase confidence and help adoption.

- Known issues / things which are not working currently (i.e. which solidity syntax/features/functions/apis do not currently work)
- How developers may have to write code differently for solana than what's already written for ethereum
- A reference/link to a complete and complex program which works on solana
- List of companies/projects currently using this in production 
- Comments from the developers if you recommend using this in production systems as it stands currently or can complex programs be written with this atm or is this right now just to give an idea to the community of what's coming up and it's not complete (i.e. certain solidity syntax and features will not work at all) and maybe it should only be used for trivial programs or research only. All these questions can be answered by your team since they are developing this and know best.

xermicus:
@LucasSte #1225 fixed this?

PR #1225, opened 2023-03-17
Be clear on Solana support

LucasSte:
This PR clarify the support for Solana, as requested in #786 .


docs/language/managing_values.rst
```diff
@@ -89,3 +93,7 @@ Here is an example:
     does not come with gas overhead. This means the ``receive()`` function is not required in the
     receiving contract, and it will not be called if it is present. If you want the ``receive()``
     function to be called, use ``address.call{value: 100}("")`` instead.
+
+.. note::
+    On Solana, these functions manage values from a contract's data account when invoked from a contract variable
+    or from the `this` builtin.
```
seanyoung, line 99: This should mention that it uses native tokens rather than spl tokens

docs/targets/solana.rst
```diff
@@ -374,3 +405,13 @@ The usage of system instructions needs the correct setting of writable and signe
...
 and `system_instruction.spec.ts <https://github.com/hyperledger/solang/blob/main/integration/solana/system_instruction.spec.ts>`_
+
+Known Issues
+____________
+
+1. `Issue #698 <https://github.com/hyperledger/solang/issues/698>`_: Custom errors are not yet supported.
+2. `Issue #887 <https://github.com/hyperledger/solang/issues/877>`_: Solidity *mapping* may consume all the storage capacity due to a bug that keeps inserting new entries, even if we
+   are only checking if a key exists.
+3. `Issue #932 <https://github.com/hyperledger/solang/issues/932>`_: Multidimensional dynamic arrays are not properly implemented and Solang might fail during compilation.
+4. `Issue #1192 <https://github.com/hyperledger/solang/issues/1192>`_: The data account layout is not yet specified in the Anchor IDL.
```
seanyoung, line 416: Remove this, people can find the issues on github themselves, and also we don't want to maintain this list - it will get out of date as we resolve issues.

docs/targets/solana.rst
```diff
@@ -1,23 +1,52 @@
...
-  signatures.
-- There is no concept of gas, so there is no gas functions
-- Solana balance is stored in a ``uint64``, so *address* ``.balance``, ``.transfer()`` and ``.send()``
-  all use ``uint64`` rather than ``uint256``.
-- Solana uses different accounts for the program code, and the contract data.
-- Programs are upgradable. There is no need to implement upgrades in Solidity.
+Solana Overview
+_______________
+
+The Solana blockchain has a unique account model that resembles a key-value database.
```
seanyoung, line 7: What does this mean or what are you trying to say? I don't know how Solana resembles a key-value database.

docs/targets/solana.rst
```diff
@@ -1,23 +1,52 @@
...
+A Solidity contract on Solana utilizes two accounts: a data account and a program account.
+The program account stores the contract's executable binary and owns the data account,
+which holds all the storage variables.
+
+Contract upgrades
++++++++++++++++++
+
+Provided that the data layout from a new contract is compatible with that of an old one, it is possible
+to update the binary in the program account and retain the same data, rendering
+contract upgrades implemented in Solidity unnecessary.
```
seanyoung, line 20: This should say how to do those upgrades

docs/targets/solana.rst
```diff
@@ -1,23 +1,52 @@
...
+
+Data types
+++++++++++
+
+- An account address consists of a 32-bits key, which is represented by the ``address`` type. This data model differs from
+  Ethereum 20-bits addresses.
```
seanyoung, line 26: bytes not bits

docs/targets/solana.rst
```diff
@@ -1,23 +1,52 @@
...
+- Solana's virtual machine registers are 64-bit wide, so 64-bit integers ``uint64`` and ``int64`` are preferable over
+  ``uint256`` and `int256`. Likewise, the builtin functions for *address* ``.balance``, ``.transfer()`` and ``.send()``
+  use 64-bit integers.
+- An address literal has to be specified using the ``address"36VtvSbE6jVGGQytYWSaDPG7uZphaxEjpJHUUpuUbq4D"`` syntax.
+
+Runtime
++++++++
+
+- Each instruction on Solana has a fixed amount of compute units. This concept supersedes the idea of gas from
+  Ethereum, so gas related functions are not available on Solana.
```
seanyoung, line 36: On ethereum, each instruction has a fixed number of gas units. This does not explain the difference between gas and compute units

docs/targets/solana.rst
```diff
@@ -1,23 +1,52 @@
...
+- There is no ``ecrecover()`` builtin function because Solana does not use the ECDSA algorithm, but there
+  is a ``signatureVerify()`` function which can check ed25519 signatures. As a consequence, it is not possible to
+  recover a signer from a signature.
+- The Solana target requires `Solana <https://www.solana.com/>`_ v1.8.1.
+- Function selectors are eight bytes wide and known as *discriminators*.
 - Solana provides different builtins, e.g. ``tx.program_id`` and ``tx.accounts``.
-- Function selectors are eight bytes and known as *discriminators*.
+- Try-catch statements do not work on Solana. If any invocation fails, the runtime will halt the contract.
```
seanyoung, line 43: - invocation is not a term used by ethereum or solana and there it is ambiguous. 
- the execution is halted, not the contract (how is a contract halted)

seanyoung (changes requested):

docs/targets/solana.rst
LucasSte: > * there it is ambiguous.

Which different meanings can you spot in the sentence?

seanyoung: invocation could be be transaction, instruction, internal function call, external function call for example

LucasSte: All those meanings are correct for the sentence, aren't they?
If a transaction, instruction, internal function call or external function call fail, the execution will halt.

docs/targets/solana.rst
LucasSte: I agree people can find this on Github, but they won't read all the open issues we have. I just wanted to highlight things that should be avoided by now.

docs/targets/solana.rst
seanyoung: It's still very ambiguous. It's not clear at all what you mean.

docs/targets/solana.rst
seanyoung: Well that's one way of describing the database. How is that distinct from any other blockchain? Why does it matter?

docs/targets/solana.rst
LucasSte: > * invocation is not a term used by ethereum or solana and there it is ambiguous.

CPI means "cross program invocation".

docs/targets/solana.rst
seanyoung: We don't want to put people off.

LucasSte: > We don't want to put people off.

This won't put people off, but making them spending time trying out things we know won't work might disappoint them.

LucasSte: This was requested in #786

docs/targets/solana.rst
LucasSte: > It's still very ambiguous. It's not clear at all what you mean.

It might be ambiguous, but you could pinpoint all the intended meanings, so it is not unclear.

seanyoung: You're being pedantic. It is not clear to users.

docs/targets/solana.rst
LucasSte: Removed

docs/targets/solana.rst
LucasSte: Added.

docs/targets/solana.rst
LucasSte: I have rephrased the sentence.

docs/targets/solana.rst
LucasSte: I have changed `invocation` (which is a term used by Solana - recall CPI) to `external call or contract creation`, which are the constructions allowed in a try-catch statement.

docs/targets/solana.rst
LucasSte: I have removed the first sentence.

docs/targets/solana.rst
LucasSte: Fixed

docs/targets/solana.rst
seanyoung: I've looked at again at #786 and I think what the ask is a list of things that currently don't work with solang. That is not the same as a list of known issues in github.
What is much more important is a list of things which currently just don't work, for example:
 - `msg.sender` does not work
 - value transfer with function call does not work
 - many yul builtins/opcodes don't work
 - external calls need more information (accounts)
 - erc20 
 - solidity try/catch 
 - solidity error definitions/revert with error definition
 
That's all I can think of right now. 

It might be reasonable to include a link to any issues in github with the label `solana`

docs/targets/solana.rst
seanyoung: This is very terse, it just deploys a program to known key pair (which we suggest anyway). Maybe a link to https://docs.solana.com/cli/deploy-a-program#set-a-programs-upgrade-authority would help.

docs/targets/solana.rst
seanyoung: It does not explain the fundamental difference between solana compute units and gas and why there are no gas or compute unit functions available. 

A developer with a solidity contract with gas functions is supposed to know how to re-write their contract without gas functions.

docs/targets/solana.rst
seanyoung: The terminology which is used by Solana is CPI, not invocation.

docs/targets/solana.rst
LucasSte: > This is very terse, it just deploys a program to known key pair (which we suggest anyway). Maybe a link to https://docs.solana.com/cli/deploy-a-program#set-a-programs-upgrade-authority would help.

The link you added here points to "Set an upgrade authority", but there is a section about redeploying a program (https://docs.solana.com/cli/deploy-a-program#redeploy-a-program). Which one should I add?

docs/targets/solana.rst
LucasSte: All those points are important information to add.

I still have the impression, thought, that #786 wants things that are not working (issues):
`Known issues / things which are not working currently (i.e. which solidity syntax/features/functions/apis do not currently work)`

LucasSte: Updated

docs/targets/solana.rst
LucasSte: I added a link to `Redeploy a program`. Let me know if this is what you were looking for.

seanyoung (changes requested):

docs/targets/solana.rst
```diff
@@ -1,34 +1,90 @@
...
-  all use ``uint64`` rather than ``uint256``.
-- Solana uses different accounts for the program code, and the contract data.
-- Programs are upgradable. There is no need to implement upgrades in Solidity.
+Solana Overview
+_______________
+
+As the underlying Solana environment is different than that of Ethereum, Solidity inner workings
+have been modified to function properly. A Solidity contract on Solana utilizes two accounts: a data account
+and a program account. The program account stores the contract's executable binary and owns the data account,
+which holds all the storage variables.
```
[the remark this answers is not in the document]
LucasSte: It should not be `For example` there, because it is not an example. The paragraph explain what those account are for.

docs/targets/solana.rst
```diff
@@ -1,34 +1,90 @@
...
 - Solana provides different builtins, e.g. ``tx.program_id`` and ``tx.accounts``.
-- Function selectors are eight bytes and known as *discriminators*.
 
-This is how to build your Solidity for Solana:
+Compute budget
+++++++++++++++
+
+On Ethereum, when calling a smart contract function, one needs to specify the amount of gas the operation is allowed
+to use. Gas can be used as a way for giving a contract priority execution on chain, provided that the gas available
+for execution is high enough. Solidity for Ethereum has gas builtins, like ``gasleft`` or the Yul ``gas()`` builtin,
```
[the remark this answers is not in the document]
LucasSte: The execution priority part is not wrong. If you offer more gas, your tip for the validator will be higher, so your priority for execution.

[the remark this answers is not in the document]
LucasSte: I am confused myself. The way I understand the following is also true for Solana. `So, you pay as much as all the instructions in the transaction uses, and there`. On Ethereum, the user specify the gas limit, on Solana the limit is fixed. I still don't get it why Solana does not have a function like `compute_units_left`. The best explanation I could find was the priority.
I could find the reason why one needs the gas functions on Ethereum, besides the prioritization.

docs/targets/solana.rst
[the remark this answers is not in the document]
seanyoung: In the `Solana Overview` you start talking about program accounts and data accounts for some reason. There are tons of differences, I don't know why this one has to be called out in the overview.

`For example,` I wanted to add should go earlier in the paragraph, before `A Solidity contract on Solana utilizes two accounts`

docs/targets/solana.rst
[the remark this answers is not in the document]
seanyoung: > I am confused myself. The way I understand the following is also true for Solana. So, you pay as much as all the instructions in the transaction uses, and there.

That's not how I understand it. Fee calculation is based on number of signatures. I might be mistaken, where did you get  this from?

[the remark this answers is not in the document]
seanyoung: Do you have anything more than a proposal?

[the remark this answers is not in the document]
LucasSte: It is an accepted proposal, so it is going to be implemented someday (ok, not a reliable source). I found this though:
https://docs.solana.com/transaction_fees#calculating-transaction-fees

> Transactions fees are calculated based on two main parts:
> 1. a statically set base fee per signature, and
> 2. the computational resources used during the transaction, measured in "[compute units].  (https://docs.solana.com/terminology#compute-units)"

docs/targets/solana.rst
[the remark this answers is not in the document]
LucasSte: Fixed

docs/targets/solana.rst
```diff
@@ -1,34 +1,90 @@
...
+have been modified to function properly. A Solidity contract on Solana utilizes two accounts: a data account
+and a program account. The program account stores the contract's executable binary and owns the data account,
+which holds all the storage variables.
+
+Contract upgrades
++++++++++++++++++
+
+Provided that the data layout from a new contract is compatible with that of an old one, it is possible
+to update the binary in the program account and retain the same data, rendering
+contract upgrades implemented in Solidity unnecessary. Solana's CLI tool provides a command to update a program:
```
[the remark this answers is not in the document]
LucasSte: Fixed

docs/targets/solana.rst
```diff
@@ -1,34 +1,90 @@
...
+check `Solana's documentation <https://docs.solana.com/cli/deploy-a-program#redeploy-a-program>`_.
+
+Data types
+++++++++++
+
+- An account address consists of a 32-bytes key, which is represented by the ``address`` type. This data model
+  differs from Ethereum 20-bytes addresses.
+- Solana's virtual machine registers are 64-bit wide, so 64-bit integers ``uint64`` and ``int64`` are preferable over
+  ``uint256`` and ``int256``. Likewise, the builtin functions for *address* ``.balance``, ``.transfer()`` and ``.send()``
+  use 64-bit integers.
```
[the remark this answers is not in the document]
LucasSte: Fixed

docs/targets/solana.rst
```diff
@@ -1,34 +1,90 @@
...
+- An address literal has to be specified using the ``address"36VtvSbE6jVGGQytYWSaDPG7uZphaxEjpJHUUpuUbq4D"`` syntax.
```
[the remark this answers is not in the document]
LucasSte: Fixed

docs/targets/solana.rst
```diff
@@ -1,34 +1,90 @@
...
-This is how to build your Solidity for Solana:
+Compute budget
+++++++++++++++
+
+On Ethereum, when calling a smart contract function, one needs to specify the amount of gas the operation is allowed
+to use. Gas can be used as a way for giving a contract priority execution on chain, provided that the gas available
+for execution is high enough. Solidity for Ethereum has gas builtins, like ``gasleft`` or the Yul ``gas()`` builtin,
+which returns the amount of gas left for execution.
+
+On the other hand, Solana utilizes the concept of compute budget. Every smart contract function is allowed the same
```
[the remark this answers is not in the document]
LucasSte: Fixed

docs/targets/solana.rst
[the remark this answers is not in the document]
LucasSte: I have rewritten this section. Let me know if it sounds better now.

xermicus (approved):
It's nice to have Solana explained a bit more in our docs :+1: I left some suggestions for further improvements

docs/targets/solana.rst
```diff
@@ -1,34 +1,95 @@
...
-- Function selectors are eight bytes and known as *discriminators*.
 
-This is how to build your Solidity for Solana:
+Compute budget
+++++++++++++++
+
+On Ethereum, when calling a smart contract function, one needs to specify the amount of gas the operation is allowed
+to use. Gas serves to pay for a contract execution on chain and can be a way for giving a contract priority execution
+when extra gas is offered in a transaction. Each EVM instruction has an associated gas value, which translates to real
+ETH cost. As there is no limit gas for the amount of gas one can provide in a transaction,
```
seanyoung, line 53: > As there is no limit gas for the amount of gas one can provide in a transaction,

Ethereum has a gas limit field in a transaction for exactly this

docs/targets/solana.rst
```diff
@@ -1,34 +1,95 @@
...
+Solidity for Ethereum has gas builtins, like ``gasleft`` or the Yul ``gas()`` builtin,
+which returns the amount of gas left for execution.
+
+On the other hand, Solana is optimized for low latency and high transaction throughput. Every smart contract function
+is allowed the same quantity of compute units (currently that value is 200 thousand), and every instruction of a
```
seanyoung, line 58: This should make clear that compute units are the solana equivalent to gas. It is implied here but not said explicitly, which makes it harder to read

docs/targets/solana.rst
```diff
@@ -1,34 +1,95 @@
...
+
+Solidity for Solana incompatibilities with Solidity for Ethereum
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+- There is no ``msg.sender``, as detailed :ref:`here <msg_sender_solana>`.
+- There is no ``ecrecover()`` builtin function because Solana does not use the ECDSA algorithm, but there
+  is a ``signatureVerify()`` function, which can check ed25519 signatures. As a consequence, it is not possible to
+  recover a signer from a signature.
+- Try-catch statements do not work on Solana. If any external call or contract creation fails, the runtime will
+  halt execution.
```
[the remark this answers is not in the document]
seanyoung: Good point, it should say that the entire transaction is reverted/aborted

docs/targets/solana.rst
LucasSte: What does this mean?

docs/language/functions.rst
```diff
@@ -80,7 +82,7 @@ see the section on :ref:`account_meta`.
 .. include:: ../examples/solana/function_call_external_accounts.sol
   :code: solidity
 
-If ``{accounts}`` is not specified, then all account are passed.
+If ``{accounts}`` is not specified, all accounts passed to the current are forwarded to the call.
```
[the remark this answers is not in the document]
LucasSte: That was a typing mistake. Thanks for catching it.

docs/targets/solana.rst
LucasSte: There is no limit for what you can input int the `gas limit` field.

LucasSte: Can you elaborate more on what you are trying to point here?

docs/targets/solana.rst
```diff
@@ -1,34 +1,95 @@
...
+where ``<KEYPAIR_FILEPATH>`` is the program's keypair json file and ``<PROGRAM_FILEPATH>``
+is the program binary ``.so`` file. For more information about redeploying a program,
+check `Solana's documentation <https://docs.solana.com/cli/deploy-a-program#redeploy-a-program>`_.
+
+Data types
+++++++++++
+
+- An account address consists of a 32-bytes key, which is represented by the ``address`` type. This data model
+  differs from Ethereum 20-bytes addresses.
+- Solana's virtual machine registers are 64-bit wide, so 64-bit integers ``uint64`` and ``int64`` are preferable over
```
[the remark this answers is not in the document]
seanyoung: That could definitely use more words. This is not obvious to a solidity developer.

In addition this should mention that multiply, divide and modulo are a lot more expensive.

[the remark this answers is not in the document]
xermicus: `Solana's virtual machine registers are 64-bit wide` is the _reason_ on why they are preferable, not the explanation

docs/targets/solana.rst
seanyoung: > As there is no limit gas for the amount of gas one can provide in a transaction,

This is not clear. This could be read as: 
- it is not possible to limit the amount of gas that a transaction can use.
- you can specify any value you want as a gas limit (you're limited by your balance after all)

Also: do you mean `no gas limit`?

docs/targets/solana.rst
```diff
@@ -1,34 +1,96 @@
...
-This is how to build your Solidity for Solana:
+Compute budget
+++++++++++++++
+
+On Ethereum, when calling a smart contract function, one needs to specify the amount of gas the operation is allowed
+to use. Gas serves to pay for a contract execution on chain and can be a way for giving a contract priority execution
+when extra gas is offered in a transaction. Each EVM instruction has an associated gas value, which translates to real
+ETH cost. As there is no limit gas for the amount of gas one can provide in a transaction,
+Solidity for Ethereum has gas builtins, like ``gasleft`` or the Yul ``gas()`` builtin,
+which returns the amount of gas left for execution.
```
seanyoung, line 55: also: `block.gaslimit`, `tx.gasprice`

docs/targets/solana.rst
LucasSte: > Also: do you mean `no gas limit`?

`limit gas` was your suggested wording.

`there is no limit gas except for the amount of eth provided in the transaction for gas. `

docs/targets/solana.rst
LucasSte: These are nice to add.

These builtins were missing since I opened the PR. It takes time to go from one version to another, so please make your reviews as complete as possible.

docs/targets/solana.rst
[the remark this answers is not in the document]
LucasSte: Fixed

docs/targets/solana.rst
LucasSte: Fixed.

docs/targets/solana.rst
```diff
@@ -1,34 +1,99 @@
...
+
+Data types
+++++++++++
+
+- An account address consists of a 32-bytes key, which is represented by the ``address`` type. This data model
+  differs from Ethereum 20-bytes addresses.
+- Solana's virtual machine registers are 64-bit wide, so 64-bit integers ``uint64`` and ``int64`` are preferable over
+  ``uint256`` and ``int256``. An operation with types wider than 64-bits is split into multiple operations, making
+  it slower and consuming more compute units. This is the case, for instance, with multiplication, division and modulo
+  using `uint256`.
```
seanyoung, line 37: This makes it appear as though only mul/div/mod cost more and other operations do not. It's just not clear at all.

```suggestion
  ``uint256``, ``int256``, ``uint128``, or ``int128``. An operation with types wider than 64-bits is split into multiple instructions, making
  it slower and consuming more compute units. Multiplication, division and modulo will require even more instructions,
  depending what values are involved. In addition, 256 bit values require more storage, either in account data, on the
  stack, or on the heap.
```
```

File: docs/language/managing_values.rst (hunk 1 above)
Comment (seanyoung) [reply docs/language/managing_values.rst]: It should mention that it does not use spl tokens.

**Review by seanyoung — APPROVED** [review]

File: docs/targets/solana.rst (hunk 18 above)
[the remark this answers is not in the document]
Comment (xermicus) [reply docs/targets/solana.rst:34]: Very nice!

File: docs/targets/solana.rst [hunk 21]
```diff
@@ -1,34 +1,100 @@
...
+value is 200k), and every instruction of a contract consumes exactly one compute unit. There is no need to provide
+an amount of compute units for a transaction and they are not charged, except when one wants priority execution on
+chain, in which case one would pay per compute unit consumed. Therefore, functions for gas are not available on
+Solidity for Solana.
+
+
+Solidity for Solana incompatibilities with Solidity for Ethereum
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+- There is no ``msg.sender``, as detailed :ref:`here <msg_sender_solana>`.
```
seanyoung, line 72: making the text of a link `here` or `this` is bad style but this PR is taking too long

LucasSte:
@seanyoung I have updated the styling you mentioned and added more words in `managing_values.rst`. Please, have a look again.

seanyoung (approved):

PR #1226, opened 2023-03-17
Forbid encoding and decoding of recursive types in sema

xermicus:
This aligns Solang with solc. Also on our `main` branch, `solang compile` crashes on the added test case.

### DIFF
--- a/src/sema/builtin.rs
+++ b/src/sema/builtin.rs
@@ -1106,8 +1106,8 @@ pub(super) fn resolve_namespace_call(
                         if ty.is_mapping() || ty.is_recursive(ns) {
                             diagnostics.push(Diagnostic::error(
                                 *loc,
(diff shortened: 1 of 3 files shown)


LucasSte (approved):

src/sema/builtin.rs
```diff
@@ -1103,10 +1103,10 @@ pub(super) fn resolve_namespace_call(
                             broken = true;
                         }
 
-                        if ty.is_mapping() {
+                        if ty.is_mapping() || ty.is_recursive(ns) {
                             diagnostics.push(Diagnostic::error(
                                 *loc,
-                                "mapping cannot be abi decoded or encoded".to_string(),
+                                format!("'{}' cannot be abi decoded or encoded", ty.to_string(ns)),
```
seanyoung, line 1109: It would be nice if the error message explained that the type cannot be encoded because it's recursive.

Issue #1216, opened 2023-03-09
Always enable math overflow unless in unchecked {} block


seanyoung:
The `--math-overflow` flag should be removed, and should always be enabled.

The `unchecked` fields in codegen should only be set for code in `unchecked {}` block or generated code which does not require overflow checks.

Most likely there are a lot of tests than need fixing.

PR #1227, opened 2023-03-19
Add release flag

salaheldinsoliman:
1- add a `--release` flag, that disables debugging features (logging prints, api return codes and runtime errors).
2- make all debugging features on by default. (make current debugging flags disable debugging)


integration/solana/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Integration tests with Solang and Solana",
   "scripts": {
     "test": "tsc; ts-node setup.ts; mocha --parallel *.spec.ts",
-    "build": "solang compile *.sol --target solana -v --log-runtime-errors --math-overflow"
+    "build": "solang compile *.sol --target solana -v --math-overflow"
```
xermicus, line 7: I feel like we should get rid of the math overflow flag. It should be checked arithmetic everywhere, except inside `unchecked {}` blocks?

However this would fit better into another PR

integration/substrate/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "tsc; ts-mocha -t 20000 *.spec.ts",
-    "build": "parallel solang compile -v -g --target substrate --log-runtime-errors --math-overflow --log-api-return-codes ::: *.sol test/*.sol",
+    "build": "parallel solang compile -v -g --target substrate --no-log-api-return-codes --math-overflow ::: *.sol test/*.sol;  solang compile debug_buffer_format.sol create_contract.sol --target substrate -v",
```
xermicus, line 8: All tests can be compiled with runtime API calls logging enabled

src/bin/solang.rs
```diff
@@ -203,8 +203,18 @@ fn main() {
...
+                            .action(ArgAction::SetFalse),
+                    ).arg(
+                        Arg::new("LOGPRINTS")
+                            .help("Disable logging prints in the environment")
+                            .long("no-log-prints")
+                            .action(ArgAction::SetFalse),
+                    ).arg(
+                        Arg::new("RELEASE")
+                            .help("Disable all debugging features such as prints, logging runtime errors, and logging api return codes")
+                            .long("debug")
```
xermicus, line 217:
```suggestion
                            .long("release")
```

src/codegen/mod.rs
```diff
@@ -90,6 +90,8 @@ pub struct Options {
     pub opt_level: OptimizationLevel,
     pub log_api_return_codes: bool,
     pub log_runtime_errors: bool,
+    pub log_prints: bool,
+    pub release: bool,
```
xermicus, line 94: I'm not sure this is needed ATM? This is just set but then never read out again to change something in the code generation, so might as well get rid of it.

src/bin/solang.rs
```diff
@@ -434,6 +448,8 @@ fn compile(matches: &ArgMatches) {
         opt_level,
         log_api_return_codes,
         log_runtime_errors,
+        log_prints,
+        release,
```
xermicus, line 452: Here we craft an `Options` struct that can have invalid combinations (e.g. log prints enabled but also the release flag set). It is then up to anyone consuming the options struct later to always check for the release Flag. I don't think this is ideal. It would be better to get rid of this release flag and set the options accordingly.

```rust
log_prints &= !release;
log_api_return_codes &= !release;
log_runtime_errors &= !release;
```

xermicus (changes requested):

LucasSte:
These changes to the CLI must be described in the docs. I also believe the example from the READ should tell developers to use the `--release` flag.

src/codegen/mod.rs
salaheldinsoliman: Do you mean `log_prints` ? it is used to disable putting print instructions in the cfg. 
Check: https://github.com/salaheldinsoliman/solang/blob/bc705be1d56fb80758febc4d0a23a2e19faa29c9/src/codegen/expression.rs#L825

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
+It is imperative to disable debugging options in release builds, as debugging would result in increased instructions and greater gas consumption.
+Solang provides three debugging options, namely debug prints, logging API return codes, and logging runtime errors.
 
-.. _log-api-return-codes:
+.. _no-log-prints:
+
+Print Function
+++++++++++++++
+
+Solang provides a ``print()`` functionality mainly for debugging. This feature  is enabled by default.
```
xermicus, line 162:
```suggestion
Solang provides a `print builtin <https://solang.readthedocs.io/en/latest/language/builtins.html#print-string>` which  is enabled by default.
```

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
 -----------------
 
-It may be desirable to have access to debug information regarding the contract execution. 
-However, this will lead to more instructions as well as higher gas usage. Hence, it is disabled by default,
-but can be enabled using CLI flags. Just make sure not to use them in production deployments.
+It is desirable to have access to debug information regarding the contract execution in the testing phase.
+Hence, debugging options are enabled by default, but can be disbaled using CLI flags. 
+Therefore, by default, debugging options are enabled; however, they can be deactivated by utilizing the command-line interface (CLI) flags.
+It is imperative to disable debugging options in release builds, as debugging would result in increased instructions and greater gas consumption.
+Solang provides three debugging options, namely debug prints, logging API return codes, and logging runtime errors.
```
xermicus, line 155: I think this sentence should be moved to the beginning of this chapter (`Debugging Options`). It should also say that these 3 options are exactly the ones being disable by the `--release` flag.

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
 
-.. _log-api-return-codes:
+.. _no-log-prints:
+
+Print Function
+++++++++++++++
+
+Solang provides a ``print()`` functionality mainly for debugging. This feature  is enabled by default.
+The ``no-log-prints`` flag will instruct the compiler not to log debugging prints in the environment.
```
xermicus, line 163:
```suggestion
The ``no-print`` flag will instruct the compiler to omit compiling any print statement into the contract artifact.
```

src/bin/solang.rs
```diff
@@ -202,9 +202,19 @@ fn main() {
...
-                            .help("Log runtime errors in the environment")
-                            .long("log-runtime-errors")
+                        Arg::new("NOLOGRUNTIMEERRORS")
+                            .help("Disable logging runtime errors in the environment")
+                            .long("no-log-runtime-errors")
+                            .action(ArgAction::SetFalse),
+                    ).arg(
+                        Arg::new("NOLOGPRINTS")
+                            .help("Disable logging prints in the environment")
+                            .long("no-log-prints")
```
xermicus, line 212:
```suggestion
                            .long("no-print")
```

src/bin/solang.rs
```diff
@@ -202,9 +202,19 @@ fn main() {
...
                     .arg(
-                        Arg::new("LOGRUNTIMEERRORS")
-                            .help("Log runtime errors in the environment")
-                            .long("log-runtime-errors")
+                        Arg::new("NOLOGRUNTIMEERRORS")
+                            .help("Disable logging runtime errors in the environment")
+                            .long("no-log-runtime-errors")
+                            .action(ArgAction::SetFalse),
+                    ).arg(
+                        Arg::new("NOLOGPRINTS")
```
xermicus, line 210:
```suggestion
                        Arg::new("NOPRINT")
```

src/bin/solang.rs
```diff
@@ -407,9 +417,13 @@ fn compile(matches: &ArgMatches) {
...
 
-    let log_api_return_codes = *matches.get_one("LOGAPIRETURNS").unwrap();
+    let release = *matches.get_one::<bool>("RELEASE").unwrap();
+
+    let log_api_return_codes = *matches.get_one::<bool>("NOLOGAPIRETURNS").unwrap() & !release;
+
+    let log_runtime_errors = *matches.get_one::<bool>("NOLOGRUNTIMEERRORS").unwrap() & !release;
 
-    let log_runtime_errors = *matches.get_one::<bool>("LOGRUNTIMEERRORS").unwrap();
+    let log_prints = *matches.get_one::<bool>("NOLOGPRINTS").unwrap() & !release;
```
xermicus, line 426:
```suggestion
    let log_prints = *matches.get_one::<bool>("NOPRINT").unwrap() & !release;
```

integration/solana/package.json
seanyoung: We have an issue open for that #1216

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
 Debugging Options
 -----------------
 
-It may be desirable to have access to debug information regarding the contract execution. 
-However, this will lead to more instructions as well as higher gas usage. Hence, it is disabled by default,
-but can be enabled using CLI flags. Just make sure not to use them in production deployments.
+It is desirable to have access to debug information regarding the contract execution in the testing phase.
+Hence, debugging options are enabled by default, but can be disbaled using CLI flags. 
+Therefore, by default, debugging options are enabled; however, they can be deactivated by utilizing the command-line interface (CLI) flags.
+It is imperative to disable debugging options in release builds, as debugging would result in increased instructions and greater gas consumption.
```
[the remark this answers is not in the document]
seanyoung: `imperative` has a very strong connotation, which is totally uncalled for here.
```suggestion
Debugging options are disabled in release builds, as debug builds greatly increase contract size and gas consumption.
```

docs/code_gen_options.rst
LucasSte, line 1: The `--release` flag should be mentioned in the CLI guide: https://solang.readthedocs.io/en/latest/running.html
In addition, this information should also be clear in the example.

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
+Solang provides three debugging options, namely debug prints, logging API return codes, and logging runtime errors. For more flexible debugging,
+Solang supports disabling each debugging feature on its own, as well as disabling them all at once with the ``--release`` flag.
 
-.. _log-api-return-codes:
+.. _no-print:
+
+Print Function
+++++++++++++++
+
+Solang provides a `print builtin <https://solang.readthedocs.io/en/latest/language/builtins.html#print-string>` which is enabled by default.
```
LucasSte, line 162: For references inside the same document, we do not use links. We use cross-references, like these ones:

https://github.com/hyperledger/solang/blob/242c187fa981b523102c0147f4ed272836c01f69/docs/language/managing_values.rst?plain=1#L46

https://github.com/hyperledger/solang/blob/242c187fa981b523102c0147f4ed272836c01f69/docs/language/types.rst?plain=1#L163

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
+.. _no-log-api-return-codes:
 
 Log runtime API call results
 ++++++++++++++++++++++++++++
 
 Runtime API calls are not guaranteed to succeed.
 By design, the low level results of these calls are abstracted away in Solidity.
 For development purposes, it can be desirable to observe the low level return code of such calls.
-The ``--log-api-return-codes`` flag will make the contract print the return code of runtime calls, if there are any.
+The contract will print the return code of runtime calls by default, and this feature could be disabled by providing the
```
LucasSte, line 173:
```suggestion
The contract will print the return code of runtime calls by default, and this feature can be disabled by providing the 
```

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
 
+
+.. _no-log-runtime-errors:
+
+Log Runtime Errors
+++++++++++++++++++
+
+In most cases, contract execution will emit a human readable error message in case a runtime error is encountered.
+The error is printed out alongside with the filename and line number that caused the error.
+This feature is enabled by default, and could be disabled by the ``--no-log-runtime-errors`` flag.
```
LucasSte, line 188:
```suggestion
This feature is enabled by default, and can be disabled by the ``--no-log-runtime-errors`` flag.
```

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
+
+.. _release:
+
+Release builds:
++++++++++++++++
+
+Release builds must not contain any debugging related logic. The ``--release`` flag will turn off all debugging features, 
+therefore reducing the required gas and storage.
```
LucasSte, line 196:
```suggestion
thereby reducing the required gas and storage.
```

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
-However, this will lead to more instructions as well as higher gas usage. Hence, it is disabled by default,
-but can be enabled using CLI flags. Just make sure not to use them in production deployments.
+It is desirable to have access to debug information regarding the contract execution in the testing phase.
+Therefore, by default, debugging options are enabled; however, they can be deactivated by utilizing the command-line interface (CLI) flags.
+Debugging options should be disabled in release builds, as debug builds greatly increase contract size and gas consumption.
+Solang provides three debugging options, namely debug prints, logging API return codes, and logging runtime errors. For more flexible debugging,
+Solang supports disabling each debugging feature on its own, as well as disabling them all at once with the ``--release`` flag.
 
-.. _log-api-return-codes:
+.. _no-print:
```
LucasSte, line 157: You have included markers on top of every section you wrote here, but you have not used them anywhere. Please, only create markers if you are going to use them.

src/bin/solang.rs
```diff
@@ -202,9 +202,19 @@ fn main() {
...
+                    ).arg(
+                        Arg::new("NOPRINT")
+                            .help("Disable logging prints in the environment")
+                            .long("no-print")
+                            .action(ArgAction::SetFalse),
+                    ).arg(
+                        Arg::new("RELEASE")
+                            .help("Disable all debugging features such as prints, logging runtime errors, and logging api return codes")
+                            .long("release")
                             .action(ArgAction::SetTrue),
```
LucasSte, line 218: Does `SetTrue` mean it will be disabled by default? 
I am not familiar with these commands.

docs/code_gen_options.rst
LucasSte: This is important to maintain coherence if we someday change the website domain or the organization of the docs. A cross-reference will never brake.

LucasSte (changes requested):

docs/code_gen_options.rst
salaheldinsoliman: Thanks, I will use them in the CLI guide

src/bin/solang.rs
salaheldinsoliman: `SetTrue` means that the variable `release` will be set to true if the argument was passed to the command

xermicus (commented):
Looks good now. However some integration test (or maybe a codegen test would do as well and be much easier to set up) verifying that the `--release` flag works as intended would be great to have.

integration/substrate/release_version.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+            BigInt(0)
+        );
+        let contract = new ContractPromise(
+            conn,
+            deployed_contract.abi,
+            deployed_contract.address
+        );
+
+        // The --release flag should remove all debugging features, making the debug buffer empty
+        let res = await debug_buffer(conn, contract, `print_then_error`, [20])
```
xermicus, line 22: Can you change this so it calls the contract twice:
- First time it does not overflow and we should check the result. The call should succeed, the debug buffer should be empty
- Second time it does overflow, we don't check the result but still expect the debug buffer to be empty.

The problem with the test as it is now is that the call might fail for any reason and leave the debug buffer empty (it doesn't really tell why the debug buffer was empty).

xermicus (approved):

integration/substrate/release_version.spec.ts
salaheldinsoliman: But now the overflow flag is removed, meaning that the output will not change only if there is an `unchecked` block added to the contract. There is also a print statement that will be in the debug buffer if the contract fails

docs/code_gen_options.rst
```diff
@@ -149,21 +148,49 @@ as a function argument, for instance, the length is unknown.
...
+
+Print Function
+++++++++++++++
```
LucasSte, line 160: I think this should be called `Debugging Logs`, not `Print Function`. Print function is an internal builtin and your CLI option does not enable or disable it.

LucasSte (approved):

docs/code_gen_options.rst
seanyoung: @LucasSte it does disable `print()`

docs/code_gen_options.rst
seanyoung: @salaheldinsoliman `--release` is not mentioned in `docs/running.rst`

salaheldinsoliman: I mentioned it, thanks

seanyoung (approved):

PR #1228, opened 2023-03-20
Integration tests: Require contract names to be unique

xermicus:
Waiting for #1222

### DIFF
--- a/integration/substrate/build.sh
+++ b/integration/substrate/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -e 
+
+dup_contracts=$(grep -r '^contract .* {' | awk '{ print $2 }' | sort | uniq -d)
+if [[ $dup_contracts ]]; then
+	echo "Found contract with duplicate names: ${dup_contracts}"
+	/bin/false
+else
+	parallel solang compile -v --target substrate --log-runtime-errors --math-overflow ::: *.sol test/*.sol;  solang compile debug_buffer_format.sol --target substrate -v --log-runtime-errors --log-api-return-codes
+fi
+
--- a/integration/substrate/build.sh
+++ b/integration/substrate/build.sh
@@ -6,6 +6,6 @@ if [[ $dup_contracts ]]; then
 	echo "Found contract with duplicate names: ${dup_contracts}"
 	/bin/false
 else
-	parallel solang compile -v --target substrate --log-runtime-errors --math-overflow ::: *.sol test/*.sol;  solang compile debug_buffer_format.sol --target substrate -v --log-runtime-errors --log-api-return-codes
+	parallel solang compile -v -g --target substrate --log-runtime-errors --math-overflow --log-api-return-codes ::: *.sol test/*.sol
 fi
 
--- a/integration/substrate/package.json
+++ b/integration/substrate/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "tsc; ts-mocha -t 20000 *.spec.ts",
-    "build": "parallel solang compile -v -g --target substrate --log-runtime-errors --math-overflow --log-api-return-codes ::: *.sol test/*.sol",
+    "build": "./build.sh",
     "build-ink": "docker run --rm -v $(pwd)/ink/caller:/opt/contract paritytech/contracts-ci-linux@sha256:2eaa0869c562adabcfc011a2f3ad6b2289cdd0b3a8923e24d29efd4452b397de cargo contract build --manifest-path /opt/contract/Cargo.toml"
   },
   "contributors": [
--- a/integration/substrate/package.json
+++ b/integration/substrate/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "tsc; ts-mocha -t 20000 *.spec.ts",
-    "build": "parallel solang compile -v --target substrate --log-runtime-errors --math-overflow ::: *.sol test/*.sol;  solang compile debug_buffer_format.sol --target substrate -v --log-runtime-errors --log-api-return-codes",
+    "build": "./build.sh",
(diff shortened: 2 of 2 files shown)


seanyoung:
I think this is a good idea. I do wonder if the solang binary itself can detect this problem, and error out.

Problem here is that if you compile `a.sol` and `b.sol`, and `a.sol` imports `b.sol`, then we should detect that we end up with contract code for `b.sol` twice, but defined in the same file location.

This also leads to an interesting question:  say you have a.sol which imports b.sol

```bash
solang compile --target foo a.sol
```
Should we actually generate contract files for `b.sol`, did the user actually want them?

xermicus:
> I do wonder if the solang binary itself can detect this problem, and error out.

I don't think so. These integration test contracts are independent from each others - it is just unfortunate that they live in the same directory, so that their contract blob names will conflict.

> Should we actually generate contract files for b.sol, did the user actually want them?

I'd say no. I'd favor if in that case, we'd not emit any contract binaries other than those defined `a.sol`. If you agree I'll open another PR

seanyoung:
> > I do wonder if the solang binary itself can detect this problem, and error out.
> 
> I don't think so. These integration test contracts are independent from each others - it is just unfortunate that they live in the same directory, so that their contract blob names will conflict.

But solang writes the file `child.contract` multiple times to the same directory. Surely it can detect it is overwriting its own output?

> > Should we actually generate contract files for b.sol, did the user actually want them?
> 
> I'd say no. I'd favor if in that case, we'd not emit any contract binaries other than those defined `a.sol`. If you agree I'll open another PR

I think this would be great - simply only output contracts that are defined in the first file in the namespace (ignoring any builtin file like `solana` or `substrate`).

If we had this, then it should be easier to solve the problem above, _I think_.

xermicus:
> But solang writes the file child.contract multiple times to the same directory. Surely it can detect it is overwriting its own output?

Ah now I get it. Yes but that would annoy me; each time I recompile I'd need to delete the old version of the binary. Not a fan of this idea. 

> simply only output contracts that are defined in the first file in the namespace

Does a namespace strictly map to a file in all cases?

seanyoung:
> Ah now I get it. Yes but that would annoy me; each time I recompile I'd need to delete the old version of the binary. Not a fan of this idea.

However, when solang is invoked, and we compile all the sources files on the command line, we can get determine that we have to different versions of a contract with the same name, and error out. No need to check the filesystem.

> > simply only output contracts that are defined in the first file in the namespace
> 
> Does a namespace strictly map to a file in all cases?

So a namespace (the term is not great, suggestions welcome!) starts empty and then we add stuff to it. There is one namespace object per file on the command line. Anything it imports is added to it, including all the contracts. So, the `files: Vec<File>` in Namespace is:
 - first list of builtin files
 - followed by main file on command line
 - followed by all the files that are imported (including transitively)

Hope that makes sense. I think the namespace started off with that name because I was really thinking about how to resolve symbols per file. 

> > If we had this, then it should be easier to solve the problem above, I think.
> 
> Agreed, it would definitively help to not accidentally overwrite the contract blob with something else down the dependency chain.

Yes, even if solc does this, it does seem like a recipe for disaster

seanyoung:
> Isn't this already covered because contract names must be unique? If I have `a.sol` that imports `b.sol` and both define contract `A`:
> 
> ```
> error: A is already defined as a contract name
>   ┌─ /tmp/a.sol:1:8
>   │  
> 1 │   import "/tmp/b.sol";
>   │          ^^^^^^^^^^^^
> 2 │   
> 3 │ ╭ contract A {
> 4 │ │     function a_in_a() public pure {}
> 5 │ │ }
>   │ ╰─' location of previous definition
> ```

Good point, I think that was a bad example. Say we have `x.sol` which defines `contract x {}`.

Now we have file `a.sol` and `b.sol`, and both do `import 'x.sol';`. Now `x.contract gets written twice.

salaheldinsoliman:
I agree with @seanyoung that this should be done with Solang binary, but I think it should only be an overwriting warning not an explicit error

seanyoung:
This PR introduces a new shell script but does not use it.

xermicus:
@seanyoung sorry my bad, I resolved that merge conflict the wrong way. Fixed

LucasSte (approved):

seanyoung:
I have an alternative implementation at #1238

xermicus:
> I have an alternative implementation at #1238

The problem solved in #1238 is orthogonal to this one? Two unrelated test cases can still define a contract with the same name, and then we have a problem as only one can exist in the same directory.

seanyoung:
> The problem solved in #1238 is orthogonal to this one? Two unrelated test cases can still define a contract with the same name, and then we have a problem as only one can exist in the same directory.

That's exactly the problem #1238 is trying to solve.

xermicus:
If we have 2 _independent_ integration tests, let's say inside `a.sol` and `b.sol`, both defining contract `foo`. They do not import each other, they are completely separate. What happens is (in arbitrary order because we compile with `parallel`, but it doesn't matter as the problem persists even if would do it sequentially):
- `solang compile --target substrate a.sol` -> writes `foo.contract`
- `solang compile --target substrate a.sol` -> writes `foo.contract`

This completely breaks any integration test deploying either `a.sol` or `b.sol`. I just tried it out on the main branch, and as I expected, this problem persists despite #1238 already merged.

seanyoung (approved):

xermicus:
But it is a good point anyways: I think we could get rid of `parallel` and this check once we implement parallel compilation of contracts.

seanyoung:
> But it is a good point anyways: I think we could get rid of `parallel` and this check once we implement parallel compilation of contracts.

Yes, we can implement this in a more efficient way: when two files import the same file, the parse tree can be re-used. Also, it saves the fork/exec overhead for each parallel execution.

PR #1229, opened 2023-03-21
No need for ubuntu-20.04, we can move to ubuntu-latest

seanyoung:
These jobs use containers anyway.

This also contains fixes for running on self-hosted runners, like cleanup the substrate docker container

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   repolinter:
     name: Repolinter
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -236,12 +236,15 @@ jobs:
       run: |
         chmod 755 ./bin/solang
         echo "$(pwd)/bin" >> $GITHUB_PATH
+    - run: |
+        # Ensure nohup writes output to a file, and not the terminal.
+        # If we don't solana-test-validator might not be able to write its output
```
LucasSte, line 241:
```suggestion
        # Ensure nohup writes output to a file, not the terminal.
        # If we don't, solana-test-validator might not be able to write its output
```

.github/workflows/test.yml
```diff
@@ -278,7 +281,10 @@ jobs:
       run: |
         chmod 755 ./bin/solang
         echo "$(pwd)/bin" >> $GITHUB_PATH
-    - run: nohup solana-test-validator -q &
+    - run: |
+        # Ensure nohup writes output to a file, and not the terminal.
+        # If we don't solana-test-validator might not be able to write its output
```
LucasSte, line 286:
```suggestion
        # Ensure nohup writes output to a file, not the terminal.
        # If we don't, solana-test-validator might not be able to write its output
```

LucasSte (approved):

PR #1232, opened 2023-03-22
Refactor accounts collection to maintain proper ordering

LucasSte:
We are going to introduce new syntax of Solidty on Solana, that allows developers to declare account names within `@signer` and `@writer` annotations. These accounts will be accessible through `tx.accounts.account_name`, which relies on the deserialization order of accounts. 

For serialization consistency, we must care about the order in which we place accounts in the IDL file, so a refactor was necessary in `solana_accounts.rs` to account for a proper ordering.

### DIFF
--- a/src/codegen/solana_accounts.rs
+++ b/src/codegen/solana_accounts.rs
@@ -1,35 +1,37 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::codegen::cfg::{ControlFlowGraph, Instr, InternalCallTy};
+use crate::codegen::cfg::{ASTFunction, ControlFlowGraph, Instr, InternalCallTy};
 use crate::codegen::{Builtin, Expression};
-use crate::sema::ast::Namespace;
+use crate::sema::ast::{Contract, Function, Mutability, Namespace, SolanaAccount};
 use crate::sema::Recurse;
 use base58::FromBase58;
 use indexmap::IndexSet;
 use num_bigint::{BigInt, Sign};
 use num_traits::Zero;
 use once_cell::sync::Lazy;
-use std::collections::hash_map::Entry;
+use solang_parser::pt::FunctionTy;
 use std::collections::{HashMap, HashSet, VecDeque};
 
 /// These are the accounts that we can collect from a contract and that Anchor will populate
-/// automatically if their name matches the source code description:
+/// automatically if their names match the source code description:
 /// https://github.com/coral-xyz/anchor/blob/06c42327d4241e5f79c35bc5588ec0a6ad2fedeb/ts/packages/anchor/src/program/accounts-resolver.ts#L54-L60
-#[derive(Hash, Eq, PartialEq, Copy, Clone)]
-pub(super) enum SolanaAccount {
-    ClockAccount,
-    InstructionAccount,
-    SystemAccount,
-    AssociatedProgramId,
-    Rent,
-    TokenProgramId,
-}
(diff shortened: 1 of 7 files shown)


src/codegen/solana_accounts.rs
```diff
@@ -167,7 +225,7 @@ pub(super) fn collect_accounts_from_contract(
         recurse_data.next_queue.clear();
     }
 
-    recurse_data.accounts
+    std::mem::swap(&mut recurse_data.functions, &mut ns.functions);
```
seanyoung, line 228: Can't you just assign rather than swap?

src/codegen/solana_accounts.rs
```diff
@@ -101,44 +87,110 @@ struct RecurseData<'a> {
...
-) -> HashMap<usize, IndexSet<SolanaAccount>> {
-    let mut visiting_queue: IndexSet<(usize, usize)> = IndexSet::from_iter(
-        ns.contracts[contract_no]
-            .functions
-            .iter()
-            .map(|ast_no| (contract_no, ns.contracts[contract_no].all_functions[ast_no])),
-    );
+pub(super) fn collect_accounts_from_contract(contract_no: usize, ns: &mut Namespace) {
+    let mut functions: Vec<Function> = Vec::new();
+    std::mem::swap(&mut functions, &mut ns.functions);
```
seanyoung, line 124: This is not great, but having said that I don't know how to write this any better. @xermicus have you got any ideas?

seanyoung (approved):

src/codegen/solana_accounts.rs
LucasSte: Good point. I can try.

src/codegen/solana_accounts.rs
LucasSte: I need to modify `ns.functions`, but I also need a reference to `ns` inside `RecurseData` and I already borrow `ns` to traverse the CFG, so rust complains about borrowing `ns` again as mutable. Swapping functions twice was the solution I found.

src/codegen/solana_accounts.rs
xermicus: Yeah this function is hard to grasp. Probable an `iter_mut()` would do as well. If it has to be using swap, can you please add a comment why it is necessary?

src/codegen/solana_accounts.rs
xermicus: A I see, maybe collect + assign the result would solve it?

LucasSte: > A I see, maybe collect + assign the result would solve it?

A collect would involve copies, wouldn't it?

xermicus: I doubt this will have a noticeable performance impact. I favor easier code over premature optimizations here.

src/codegen/solana_accounts.rs
LucasSte: I added a comment and used the assignment at the end of the function.

xermicus (approved):

src/codegen/solana_accounts.rs
seanyoung: I agree with @xermicus, don't optimize for execution speed over code clarity.

You're assuming the copy is of `struct Function`, but I was thinking the collect and assign could be the solana accounts, not the entire `Function`.

Either way, I don't think there are perfect solutions here and I am not sure how useful it is to spend more time on this.

PR #1233, opened 2023-03-22
Upgrade Anchor

LucasSte:
This PR upgrades the Anchor version to v0.27.0. The IDL does not have the `state` field anymore.

The container CI image needs to be updated in https://github.com/hyperledger/solang-llvm/pull/9.

### DIFF
--- a/src/bin/idl/mod.rs
+++ b/src/bin/idl/mod.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use anchor_syn::idl::{Idl, IdlInstruction, IdlType, IdlTypeDefinitionTy};
+use anchor_syn::idl::{Idl, IdlAccountItem, IdlInstruction, IdlType, IdlTypeDefinitionTy};
 use clap::ArgMatches;
 use itertools::Itertools;
 use serde_json::Value as JsonValue;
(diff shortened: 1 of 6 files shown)


src/bin/idl/mod.rs
```diff
@@ -310,7 +297,7 @@ fn instruction(
             )?;
         }
 
-        write!(f, ") {}external", if state { "" } else { "view " })?;
+        write!(f, ") external")?;
```
seanyoung, line 300: The ts code makes functions view *if* there are return values and there is no mutable account needed. We could do the same.

LucasSte: Good point. I'm going to add code to handle this.

src/bin/idl/mod.rs
```diff
@@ -332,6 +320,18 @@ fn instruction(
...
+fn mutable_account_exists(accounts: &[IdlAccountItem]) -> bool {
+    let mut exists = false;
+    for item in accounts {
+        exists |= match item {
+            IdlAccountItem::IdlAccount(acc) => acc.is_mut,
+            IdlAccountItem::IdlAccounts(accs) => mutable_account_exists(&accs.accounts),
+        }
+    }
+
+    exists
```
seanyoung, line 332: Do you not like iterators?
```suggestion
    accounts.iter().any(|item| match item {
            IdlAccountItem::IdlAccount(acc) => acc.is_mut,
            IdlAccountItem::IdlAccounts(accs) => mutable_account_exists(&accs.accounts),
        })
```

seanyoung (approved):

src/bin/idl/mod.rs
LucasSte: > Why do you not like iterators?

I am not used to them. I am used to writing `for(int I=0; i<n; i++)` in C, so I often write for-loops in Rust too.

PR #1235, opened 2023-03-23
Substrate: Test for instantiation nonce to increase

xermicus:
Explicitly asserting the `instantiation_nonce` runtime API return value to increase after a subsequent contract deployment.

### DIFF
--- a/integration/substrate/runtime_errors.spec.ts
+++ b/integration/substrate/runtime_errors.spec.ts
@@ -4,103 +4,68 @@ import { ContractPromise } from "@polkadot/api-contract";
 
 describe('Deploy runtime_errors.sol and test the debug buffer', () => {
     it('logs errors to the debug buffer', async function () {
-
         let conn = await createConnection();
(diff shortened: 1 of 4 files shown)


seanyoung (approved):

integration/substrate/runtime_errors.spec.ts
```diff
@@ -4,103 +4,68 @@ import { ContractPromise } from "@polkadot/api-contract";
...
-
         let contract = new ContractPromise(
             conn,
             deployed_contract.abi,
             deployed_contract.address
         );
-
         let child_contract = await deploy(conn, alice, 'child_runtime_errors.contract', BigInt(0));
 
         let res = await debug_buffer(conn, contract, `get_storage_bytes`, [])
```
LucasSte, line 22: In Typescript, we use `let` for mutable variable and `constant` for those whose value does not change.

LucasSte (approved):

PR #1237, opened 2023-03-24
Improve parser

DaniPopes:
This PR does 3 main things to the `solang-parser` crate:
1. adds documentation to all items
2. improves some logic in the parsing functions, like less allocations and some minor refactoring
3. most importantly, adds parse tree helpers (moving existing implementations):
  - `fmt`: implements `std::fmt::Display` for all parse tree types
  - `loc`, expanding on the `CodeLocation` traits:
    - implementations for all possible parse tree types
    - generic implementations for `Vec`, `Option` etc
    - renames the `loc` method to `loc_opt` on `OptionalCodeLocation` to avoid name collision
    - previously there were some standalone (not with the traits) methods on structs, these were moved to the trait implementations
  - `ord`: implements `std::cmp::{Ord, PartialOrd}`
  - helper methods on parse tree types; you can view them by looking for the `impl` blocks in `pt`

The massive diff in the lexer tests is just whitespace and renaming the fully qualified `collect::<Vec<...>>` to just `Vec<_>`

Some of these changes have been upstreamed from [foundry-rs/foundry](https://github.com/foundry-rs/foundry)'s `forge-doc` and `forge-fmt`.

BREAKING CHANGES:
- `pt::OptionalCodeLocation::loc` renamed to `loc_opt`
- `pt::Level::to_string` renamed to `as_str`
- `lexer::Spanned<Token, Loc, Error>` split into `Spanned<'a>` and `Result<'a, T = Spanned<'a>, E = LexicalError>`
- `pt::UsingList`'s `Error` variant changed from tuple to unit (no more `()` at the end)


seanyoung:
The new files need:
```
```
on the first line, else CI will fail. Also note the clippy errors.

solang-parser/src/doccomment.rs
```diff
@@ -1,27 +1,65 @@
 // SPDX-License-Identifier: Apache-2.0
 
+//! Solidity parsed doc comments.
+//!
+//! See also the Solidity documentation on [natspec].
+//!
+//! [natspec]: https://docs.soliditylang.org/en/latest/natspec-format.html
```
xermicus, line 7:
```suggestion
//! See also the Solidity documentation on [natspec](https://docs.soliditylang.org/en/latest/natspec-format.html).
```

Markdown links are supported here.

solang-parser/src/helpers/fmt.rs
```diff
@@ -0,0 +1,2518 @@
...
+        }
+    }
+}
+
+impl Display for pt::ContractTy {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        f.write_str(self.as_str())
+    }
+}
+impl pt::ContractTy {
```
xermicus, line 413:
```suggestion
}

impl pt::ContractTy {
```

solang-parser/src/helpers/fmt.rs
```diff
@@ -0,0 +1,2518 @@
...
+            }
+            Self::Default(_, block) => {
+                f.write_str("default ")?;
+                block.fmt(f)
+            }
+        }
+    }
+}
+
+#[inline]
```
xermicus, line 1211: Can you elaborate here why we need this? It is is almost always unnecessary and deserves  a comment that proves this is needed here.

solang-parser/src/helpers/fmt.rs
```diff
@@ -0,0 +1,2518 @@
...
+fn fmt_parameter_list(list: &pt::ParameterList, f: &mut Formatter<'_>) -> Result {
+    let iter = list.iter().flat_map(|(_, param)| param);
+    write_separated_iter(iter, f, ", ")
+}
+
+#[inline]
```
xermicus, line 1217: Same

solang-parser/src/helpers/fmt.rs
```diff
@@ -0,0 +1,2518 @@
...
+        //     }
+        // };
+        // ( assembly { $($t:tt)* } ) => {
+        //     pt::Statement::Assembly {
+        //         loc: loc!(),
+        //         dialect: None,
+        //         flags: None,
+        //         block: pt::YulBlock,
+        //     }
+        // };
```
xermicus, line 1400:
```suggestion
```

xermicus (commented):
Thanks for your contribution!

solang-parser/src/doccomment.rs
DaniPopes: That is also a valid markdown link, and makes reading easier IMO.

solang-parser/src/helpers/fmt.rs
DaniPopes: very simple functions that only call another function, I don't see why not to add this; same for all the helper functions that do simple matching or just a field access

solang-parser/src/helpers/fmt.rs
DaniPopes: Display `impl` is grouped with its corresponding struct `impl`, but I can change that if you want to

xermicus: Well I thought this slipped by. Anyways it's a nitpick, I'll leave it up to you

solang-parser/src/doccomment.rs
xermicus: Very nice, I didn't know that

solang-parser/src/helpers/fmt.rs
xermicus: This function looks a lot like it should be getting inlined by the compiler automatically, especially considering it's private. If I see this `#[inline]` here I'd assume that the compiler doesn't inline this function when it should
and it is used in a hot path somewhere where removing it introduces significant performance penalties.

If these assumptions are correct, it would be nice to leave a comment explaining that, otherwise the `#[inline]` is probably unnecessary.

seanyoung: The compiler knows perfectly well when things should be inlined or not. It does not sound there is a solid reason for the `#[inline]` attribute.

solang-parser/src/pt.rs
```diff
@@ -1,1003 +1,1688 @@
...
+    /// The list of values.
+    ///
+    /// This field contains `None` only if an error occurred during parsing.
     pub values: Vec<Option<Identifier>>,
 }
 
+/// A variable attribute.
 #[derive(Debug, PartialEq, Eq, Clone)]
 #[cfg_attr(feature = "pt-serde", derive(Serialize, Deserialize))]
+#[repr(u8)] // for cmp; order of variants is important
```
LucasSte, line 833: Why is this necessary?

solang-parser/src/pt.rs
```diff
@@ -1,1003 +1,1688 @@
...
+    /// `payable`
+    Payable(Loc),
 }
 
+/// Function visibility.
+///
+/// Deprecated for [FunctionTy] other than `Function`.
 #[derive(Debug, PartialEq, Eq, Clone)]
 #[cfg_attr(feature = "pt-serde", derive(Serialize, Deserialize))]
+#[repr(u8)] // for cmp; order of variants is important
```
LucasSte, line 1290: Why do you need the representation to be `u8`?

DaniPopes: it's needed to add the Ord implementations by using only the discriminant, see the cmp file. But I realize it might be better off just deriving the implementations

solang-parser/src/pt.rs
```diff
@@ -1,1003 +1,1688 @@
...
 pub enum UserDefinedOperator {
+    /// `&`
     BitwiseAnd,
+    /// `~`
+    ///
+    /// AKA BitwiseNot
     Complement,
+    /// `-`
+    ///
+    /// Note that this is the same as `Subtract`, and that it is currently not being parsed.
```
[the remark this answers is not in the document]
seanyoung: There is no way to distinguish them without semantic analysis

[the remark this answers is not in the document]
seanyoung: sema uses it.

This is not related to this PR.

solang-parser/src/pt.rs
DaniPopes: The problem with using the derive macro is that it imposes the trait bound on all of the fields, which will also be checked during the call.
For example, deriving `PartialOrd` for `FunctionAttribute` transitively requires the trait to be implemented for `Mutability`, `IdentifierPath`, and `Base`; we can achieve this by implementing it manually, which goes against the whole point of deriving in the first place, or deriving it for these types too, which will eventually require deriving for almost the entire `pt` module.

By doing it manually we can make ordering depend only on the discriminant, regardless of the variants' value/fields; I think is fine for the few types (attributes, visibility) that have a recommended order in the Solidity standards.

solang-parser/src/helpers/ord.rs
```diff
@@ -0,0 +1,43 @@
...
+                fn cmp(&self, other: &Self) -> Ordering {
+                    Ord::cmp(self.as_discriminant(), other.as_discriminant())
+                }
+            }
+        )+
+    };
+}
+
+// SAFETY: every type must be `repr(u8)` for this to be safe, see comments in macro implementation.
+impl_with_cast!(pt::Visibility, pt::VariableAttribute, pt::FunctionAttribute);
```
seanyoung, line 43: I do totally get why you've done it this way, but I'd like to avoid `unsafe {}`. What do you think of the following?

```suggestion
impl pt::Visibility {
    fn solidity_order(&self) -> u8 {
        match self {
            pt::Visibility::External(..) => 0,
            pt::Visibility::Internal(..) => 1,
            pt::Visibility::Public(..) => 2,
            pt::Visibility::Private(..) => 3,
        }
    }
}

impl PartialOrd for pt::Visibility {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.solidity_order().cmp(&other.solidity_order()))
    }
}

impl Ord for pt::Visibility {
    fn cmp(&self, other: &Self) -> Ordering {
        self.solidity_order().cmp(&other.solidity_order())
    }
}

impl pt::VariableAttribute {
    fn solidity_order(&self) -> u8 {
        match self {
            pt::VariableAttribute::Visibility(..) => 0,
            pt::VariableAttribute::Constant(..) => 1,
            pt::VariableAttribute::Immutable(..) => 2,
            pt::VariableAttribute::Override(..) => 3,
        }
    }
}

impl PartialOrd for pt::VariableAttribute {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.solidity_order().cmp(&other.solidity_order()))
    }
}

impl Ord for pt::VariableAttribute {
    fn cmp(&self, other: &Self) -> Ordering {
        self.solidity_order().cmp(&other.solidity_order())
    }
}

impl pt::FunctionAttribute {
    fn solidity_order(&self) -> u8 {
        match self {
            pt::FunctionAttribute::Visibility(..) => 0,
            pt::FunctionAttribute::Mutability(..) => 1,
            pt::FunctionAttribute::Virtual(..) => 2,
            pt::FunctionAttribute::Immutable(..) => 3,
            pt::FunctionAttribute::Override(..) => 4,
            pt::FunctionAttribute::BaseOrModifier(..) => 5,
            pt::FunctionAttribute::Error(..) => 6,
        }
    }
}

impl PartialOrd for pt::FunctionAttribute {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.solidity_order().cmp(&other.solidity_order()))
    }
}

impl Ord for pt::FunctionAttribute {
    fn cmp(&self, other: &Self) -> Ordering {
        self.solidity_order().cmp(&other.solidity_order())
    }
}
```

seanyoung (commented):
There is a lot of good stuff in this - very large - pull request, it is great, thank you very much!

So I only have one comment, please us let know what you think. Thanks for some quality code changes. :pray:

solang-parser/src/helpers/ord.rs
DaniPopes: That's how I originally implemented it, though I didn't feel like it was really idiomatic. Seeing as the standard Discriminant doesn't implement Ord, I then tried this method which is recommended in the Rust docs. I believe it's also easier to maintain and has no overhead due to it being just a pointer cast.
But if you would like to avoid unsafe at all costs then I can change this, up to you.

solang-parser/src/helpers/fmt.rs
DaniPopes: > The compiler knows perfectly well when things should be inlined or not. It does not sound there is a solid reason for the `#[inline]` attribute.

I don't believe that is the case. You are probably right in saying that since these functions are private then the compiler is more likely to inline them, but giving extra hints that they should be doesn't hurt, especially because they are private and that they should probably be inlined regardless.

> If these assumptions are correct, it would be nice to leave a comment explaining that (ideally pointing to some benchmarks or tests), otherwise the `#[inline]` is probably unnecessary.

I don't have benchmarks nor do I intend to write any simply to justify a mere compiler hint that may or may not sway performance and/or compilation speed by less than a single percentage point. Since there are almost no downsides to adding these hints I would rather just leave them as is.

This being said, if you still want me to remove the `#[inline]` hints on these private functions, I will do so.

DaniPopes: > They are a maintenance burden, especially without any comment providing context.

Given:
- `fn a<T>(x: T, y, z) { ... }`
- `#[inline] fn b(x: &[u8]) { a(x, <default_y>, <default_z>) }`

Since `b` is just a simple shortcut for a special case of `a`, is the reason to add `#[inline]` not obvious, or at least deductible, without requiring 2 paragraphs of comments detailing why this should be the case?

I could've just not written `b` and instead write manual calls to `a` everywhere, but since I noticed a common pattern for the same type I wrote `b` to make easier and cleaner, and to reduce any possible performance impact I add a hint that it should be the same as the manual `a` calls.

But alas, if you so insist I'll add a comment linking to this discussion.

solang-parser/src/helpers/fmt.rs
```diff
@@ -0,0 +1,2501 @@
...
+            }
+            Self::Default(_, block) => {
+                f.write_str("default ")?;
+                block.fmt(f)
+            }
+        }
+    }
+}
+
+#[inline]
```
xermicus, line 1211: It is obvious why it should be inlined. What is not obvious is why the compiler doesn't do it automatically or why we think it needs this hint (we _don't_ know) and so why the need for this attribute arises.

I left the comments as a suggestion, WDYT? :upside_down_face: 

```suggestion
// This function is private so it should be inlined by the compiler anyways. 
// We provided this as an extra hint regardless because we don't expect compile time penalties or other negative impacts from it.
#[inline]
```

solang-parser/src/helpers/fmt.rs
```diff
@@ -0,0 +1,2501 @@
...
+fn fmt_parameter_list(list: &pt::ParameterList, f: &mut Formatter<'_>) -> Result {
+    let iter = list.iter().flat_map(|(_, param)| param);
+    write_separated_iter(iter, f, ", ")
+}
+
+#[inline]
```
xermicus, line 1217:
```suggestion
// This function is private so it should be inlined by the compiler anyways. 
// We provided this as an extra hint regardless because we don't expect compile time penalties or other negative impacts from it.
#[inline]
```

solang-parser/src/helpers/fmt.rs
xermicus: https://github.com/hyperledger/solang/pull/1237/files#r1151621676

solang-parser/src/helpers/ord.rs
seanyoung: It's a matter of taste - let's go with your version. Thank you for the discussion.

seanyoung (approved):

solang-parser/src/helpers/fmt.rs
LucasSte: @DaniPopes Inlining functions is not guaranteed to improve performance. It might also make the code slower:
https://isocpp.org/wiki/faq/inline-functions#inline-and-perf

LucasSte (approved):

seanyoung:
@DaniPopes thank you very much, great stuff! :pray:

PR #1238, opened 2023-03-27
Check for duplicate contract names on cli command line

seanyoung:
If we are compiling two files, and they define the same contract name from multiple locations, error out.

### DIFF
--- a/src/bin/solang.rs
+++ b/src/bin/solang.rs
@@ -600,7 +600,7 @@ fn contract_results(
     }
 
     if ns.top_file_no() != resolved_contract.loc.file_no() {
-        // contracts which were imported should not be considered. For example, if we have a file
+        // contracts that were imported should not be considered. For example, if we have a file
         // a.sol which imports b.sol, and b.sol defines contract B, then:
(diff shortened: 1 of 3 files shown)


src/bin/solang.rs
```diff
@@ -589,6 +599,26 @@ fn contract_results(
         return;
     }
 
+    if ns.top_file_no() != resolved_contract.loc.file_no() {
+        // contracts which were imported should not be considered. For example, if we have a file
```
LucasSte, line 603:
```suggestion
        // contracts that were imported should not be considered. For example, if we have a file
```

LucasSte: If you use `which`, you have a non-restrictive clause. `contracts, which were imported, should not be considered` means that all contracts were imported and all of them will not be considered.
When you use `that`, you have a restrictive clause. `contracts that were imported should not be considered` means that you are only not considering imported contract .

https://academicguides.waldenu.edu/writingcenter/grammar/clauses

xermicus (approved):
Nice, thanks

Issue #1234, opened 2023-03-22
Event not found after importing interface


andywer:
**Describe the bug**
Assume a contract C implementing an interface I that's been imported from another file. I declares an event, C emits it. Try to compile for Substrate.

solc: Works
solang: `event 'MyEvent' not found`

**To Reproduce**
See <https://github.com/0xamberhq/solang-issues/tree/main/01-emit-iface-events>.

**Expected behavior**
Should compile without errors.

**Hyperledger Solang version**
Solang v0.2.2

**Include the complete solidity source code**
See <https://github.com/0xamberhq/solang-issues/tree/main/01-emit-iface-events>.

**Additional context**
Seems to fail only when naming the symbols to import (`import {X} from "./X.sol";` vs `import "./X.sol";`).

xermicus:
No worries :rocket:

PR #1241, opened 2023-03-28
When resolving events in base contracts, use correct file number

seanyoung:
The file number should be of where the base contract is defined, not the contract where we're trying to do the lookup.

Fixes: https://github.com/hyperledger/solang/issues/1234


tests/imports.rs
```diff
@@ -148,8 +148,10 @@ fn contract_name_defined_twice() {
     let output = not_ok.get_output();
     let err = String::from_utf8_lossy(&output.stderr);
 
+    println!("{}", err);
```
LucasSte, line 151: Is this supposed to be here?

LucasSte (approved):

tests/imports.rs
seanyoung: Yes, it just helps with debugging.

PR #1243, opened 2023-03-30
Bugfix: Struct with fields of BytesN types can not be encoded linear

xermicus:
Fixes an encoding bug: When a struct has any field with `BytesN` type, this field requires an endianess swap and can't be encoded or decoded directly (e.g. with a `MemCopy`).

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -2067,7 +2067,8 @@ impl Namespace {
     pub fn calculate_struct_non_padded_size(&self, struct_type: &StructType) -> Option<BigInt> {
         let mut size = BigInt::from(0u8);
         for field in &struct_type.definition(self).fields {
-            if !field.ty.is_primitive() {
+            let ty = field.ty.clone().unwrap_user_type(&self);
+            if !ty.is_primitive() || matches!(ty, Type::Bytes(_)) {
                 // If a struct contains a non-primitive type, we cannot calculate its
                 // size during compile time
                 if let Type::Struct(struct_ty) = &field.ty {
@@ -2078,7 +2079,7 @@ impl Namespace {
                 }
                 return None;
             } else {
-                size.add_assign(field.ty.memory_size_of(self));
+                size.add_assign(ty.memory_size_of(self));
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/codegen/cfg.rs
```diff
@@ -2067,7 +2067,8 @@ impl Namespace {
     pub fn calculate_struct_non_padded_size(&self, struct_type: &StructType) -> Option<BigInt> {
```
LucasSte, line 2067: This is not the right solution. `calculate_struct_non_padded_size` is used to calculate the size of a struct, before we start the encoding or decoding. If we have a `Type::Bytes(N)` we can still calculate the size of the struct during compile time.

You should be modifying `allow_direct_copy`

xermicus: I don't get it. Changing `allow_direct_copy` is pointless. It is only used during the encoding and decoding in arrays. How would that help for structs?

Can we please get this fix in, I uncovered it during the dispatcher refactoring.

xermicus: I refactored `allow_direct_copy` so it works for structs too.

> we can move these methods to impl Type

I refrained from that. Right now the encoder is used for SCALE and Borsh, so anything in it is specific to those. Other encoding schemes might behave completely different. Implementing this for the type would suggest that `allow_memcpy` is universal for any encoding scheme, which is probably false.

src/codegen/encoding/mod.rs
```diff
@@ -1718,37 +1702,28 @@ fn array_outer_length(
...
-    } else {
-        elem_ty.is_primitive()
-    };
-
-    if array_ty.is_dynamic(ns) {
-        // If this is a dynamic array, we can only MemCpy if its elements are of
-        // any primitive type and we don't need to index it.
-        dims.len() == 1 && type_direct_copy
-    } else {
-        // If the array is not dynamic, we can MemCpy elements if their are primitive.
```
LucasSte, line 1750: My only suggestion is to keep these comments.

LucasSte (approved):

seanyoung (approved):

LucasSte (approved):

PR #1244, opened 2023-03-31
Bugfix: Unwrap user types while processing WriteBuffer instructions

xermicus:
User types can wrap `bytesN` which need an endianess swap during encoding.

### DIFF
--- a/tests/substrate_tests/builtins.rs
+++ b/tests/substrate_tests/builtins.rs
@@ -725,6 +725,11 @@ fn hash() {
             function get() public view returns (Hash) {
                 return Hash.wrap(current2);
             }
+
+            function test_encoding() public view {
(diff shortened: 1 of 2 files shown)


seanyoung (approved):

LucasSte (approved):

tests/substrate_tests/builtins.rs
```diff
@@ -725,6 +725,11 @@ fn hash() {
             function get() public view returns (Hash) {
                 return Hash.wrap(current2);
             }
+
+            function test_encoding() public view {
+                Hash h = Hash.wrap(current2);
+                assert(abi.encode(current2) == abi.encode(h));
```
LucasSte, line 731: Be careful with tests like this. They might mask a case when both the encoding and decoding are wrong.

PR #1245, opened 2023-03-31
Remove --math-overflow flag

### LINKED ISSUE #1216: Always enable math overflow unless in unchecked {} block

seanyoung:
Fixes #1216

### DIFF
--- a/src/emit/expression.rs
+++ b/src/emit/expression.rs
@@ -121,8 +121,8 @@ pub(super) fn expression<'a, T: TargetRuntime<'a> + ?Sized>(
             let left = expression(target, bin, l, vartab, function, ns).into_int_value();
             let right = expression(target, bin, r, vartab, function, ns).into_int_value();
 
-            if bin.options.math_overflow_check && !*unchecked {
-                let signed = l.ty().is_signed_int();
+            if !unchecked {
+                let signed = l.ty().is_signed_int(ns);
                 build_binary_op_with_overflow_check(
                     target,
                     bin,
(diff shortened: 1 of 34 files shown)


src/emit/expression.rs
```diff
@@ -121,8 +121,8 @@ pub(super) fn expression<'a, T: TargetRuntime<'a> + ?Sized>(
             let left = expression(target, bin, l, vartab, function, ns).into_int_value();
             let right = expression(target, bin, r, vartab, function, ns).into_int_value();
 
-            if bin.options.math_overflow_check && !*unchecked {
-                let signed = l.ty().is_signed_int();
+            if !unchecked {
```
LucasSte, line 124: Why aren't we dereferencing anymore?

src/emit/expression.rs
```diff
@@ -143,8 +143,8 @@ pub(super) fn expression<'a, T: TargetRuntime<'a> + ?Sized>(
...
```
LucasSte, line 146: Why aren't we dereferencing anymore?

LucasSte (approved):

xermicus (approved):
LGTM

src/emit/expression.rs
seanyoung: The deref is implicit

src/emit/expression.rs
seanyoung: The deref is implicit

Issue #1251, opened 2023-04-04
Improving Solana account management


LucasSte:
- [] `@payer` must not support literals.
- [] The account declared in `@payer` must be accessible via `tx.accounts`
- [] If the parameter of an annotation is a constructor argument, the annotation should appear before the argument declaration so we do not confuse this with an account declaration, which has nothing to do with arguments.
- [] All accounts declared must be validated (check if signer and/or mutable bits are properly set and ensure we have all the accounts needed).
- [] Provide `@signer`, `@reader`, `@mutable` and `@mutableSigner` so that developers can specify the accounts they need. These accounts must be accessible via `tx.accounts` and validated.
- [] Account declaration can happen in functions that have a body.

PR #1254, opened 2023-04-05
Unify `external_call` and `create_contract` on Solana

LucasSte:
`create_contract` and `external_call` have basically the same meaning on Solana, so I united the functions we have in `solana.c`.

This is part of my task to improve the account management on Solana. Please, read my task list #1251 To understand what I am trying to achieve.


seanyoung (approved):
Nice catch.

Please remove all the changed `.bc` files which do not matter. For example this is Solana specific so no wasm file needs to change.

xermicus (approved):
LGTM

LucasSte:
> Please remove all the changed `.bc` files which do not matter. For example this is Solana specific so no wasm file needs to change.

Ok, I'll remove them. Just out of curiosity, why do we maintain these compiled filed in the repo if we build them whenever we run `cargo build`?

seanyoung:
> Ok, I'll remove them. Just out of curiosity, why do we maintain these compiled files in the repo if we build them whenever we run `cargo build`?

They don't get built by `cargo build`. It might be possible to build them in `build.rs`. That would also remove the need for a `Makefile`.

I think it's a quick hack I did once upon a time, because I was not sure how to do this, and it has been fixed.

PR #1257, opened 2023-04-07
Ensure that only the last instruction is an edge

seanyoung:
This adds a check on the CFG ensuring that only the last instruction does control flow.

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -147,8 +147,6 @@ pub enum Instr {
         address: Expression,
         value: Expression,
     },
-    /// Insert unreachable instruction after e.g. self-destruct
-    Unreachable,
     /// Self destruct
     SelfDestruct { recipient: Expression },
     /// Emit event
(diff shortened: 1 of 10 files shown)


seanyoung:
@LucasSte the yul code has `is_reachable()` and `is_next_reachable()`. Is there a distinction?

xermicus (approved):
LGTM

LucasSte:
> @LucasSte the yul code has `is_reachable()` and `is_next_reachable()`. Is there a distinction?

Yes, there is. `is_reachable` tells if the block is reachable (i.e. there exists a code path from entry to this block). `is_next_reachable` determines if this block leads to another one. If the last instruction of a block is, for instance, an `assert-failure` or a `return`, `is_reachable` can be true, but `is_next_reachable` will be false. This way, I know that I should not add any new instructions to the block (e.g. a `Instr::Branch`). Check this example:

```solidity
for(uint32 i=0; i<40; i++) {
    if(condition) {
         return i; // There is no need to add a `Instr::Branch` here, because we are returning, so `is_next_reachable` is false.
    }
}
```

seanyoung:
Replaced by https://github.com/hyperledger/solang/pull/1260

PR #1259, opened 2023-04-08
 Parse error definitions and revert with custom errors

seanyoung:
These are not supported on Solana or Substrate, however this will help
with testing sema with evm solidity contracts.

This changes `revert()` from an builtin expression to a statement type.


tests/contract_testcases/evm/error.sol
```diff
@@ -0,0 +1,38 @@
...
+		if (1 != 0) {
+			revert E1({bar: 1, foo: true});
+		} else if (1 != 0) {
+			revert E2({bar: 1});
+		} else if (1 != 0) {
+			revert E3({bar: 1, foo: true, baz: "hey"});
+		} else if (1 != 0) {
+			revert e.E3({bar: 1, foo: true, foo: false});
+		} else {
+			revert E4({bar: 2, baz: true});
```
LucasSte, line 22: If I do `revert E3({bar: 2, bar: 1.5, foo: true})`, will there be any spurious error?

tests/contract_testcases/solana/error2.dot
```diff
@@ -0,0 +1,50 @@
...
+	diagnostic_17 [label="X2 is already defined as an error\nlevel Error\ntests/contract_testcases/solana/error2.sol:7:6-8"]
+	note_18 [label="location of previous definition\ntests/contract_testcases/solana/error2.sol:6:7-9"]
+	diagnostic_19 [label="found contract 'C'\nlevel Debug\ntests/contract_testcases/solana/error2.sol:10:1-18:2"]
+	diagnostic_20 [label="X2 is already defined as an error\nlevel Warning\ntests/contract_testcases/solana/error2.sol:11:8-10"]
+	note_21 [label="location of previous definition\ntests/contract_testcases/solana/error2.sol:6:7-9"]
+	diagnostic_22 [label="error 'X3' has duplicate field name 'X'\nlevel Error\ntests/contract_testcases/solana/error2.sol:12:23-24"]
+	note_23 [label="location of previous declaration of 'X'\ntests/contract_testcases/solana/error2.sol:12:11-16"]
+	diagnostic_24 [label="mapping type is not permitted as error field\nlevel Error\ntests/contract_testcases/solana/error2.sol:14:11-32"]
+	diagnostic_25 [label="type 'X' not found\nlevel Error\ntests/contract_testcases/solana/error2.sol:15:11-12"]
+	diagnostic_26 [label="expecting 'error', got 'meh'\nlevel Error\ntests/contract_testcases/solana/error2.sol:16:2-5"]
```
LucasSte, line 24: Were we expecting an error? A function definition in line 16 could be valid.

tests/contract_testcases/solana/error2.dot
```diff
@@ -0,0 +1,50 @@
...
+	X2_4 [label="name:X2\ntests/contract_testcases/solana/error2.sol:6:7-9\nfield name:bar ty:int256"]
+	X2_5 [label="name:X2\ncontract: 0\ntests/contract_testcases/solana/error2.sol:11:8-10\nfield name:bar ty:int256"]
+	X3 [label="name:X3\ncontract: 0\ntests/contract_testcases/solana/error2.sol:12:8-10\nfield name:X ty:int256"]
+	X4 [label="name:X4\ncontract: 0\ntests/contract_testcases/solana/error2.sol:13:8-10\nfield name:X ty:int256\nfield name: ty:bool"]
+	X5 [label="name:X5\ncontract: 0\ntests/contract_testcases/solana/error2.sol:14:8-10\nfield name: ty:unresolved"]
+	X6 [label="name:X6\ncontract: 0\ntests/contract_testcases/solana/error2.sol:15:8-10\nfield name: ty:unresolved"]
+	contract [label="contract C\ntests/contract_testcases/solana/error2.sol:10:1-18:2"]
+	diagnostic [label="X1 is already defined as an enum\nlevel Error\ntests/contract_testcases/solana/error2.sol:3:7-9"]
+	note [label="location of previous definition\ntests/contract_testcases/solana/error2.sol:2:6-8"]
+	diagnostic_15 [label="expecting 'error', got 'foo'\nlevel Error\ntests/contract_testcases/solana/error2.sol:4:1-4"]
```
LucasSte, line 13: Do we always expect and error after an error definition? What about structs and enums?

tests/contract_testcases/solana/error2.dot
```diff
@@ -0,0 +1,50 @@
...
+	note_18 [label="location of previous definition\ntests/contract_testcases/solana/error2.sol:6:7-9"]
+	diagnostic_19 [label="found contract 'C'\nlevel Debug\ntests/contract_testcases/solana/error2.sol:10:1-18:2"]
+	diagnostic_20 [label="X2 is already defined as an error\nlevel Warning\ntests/contract_testcases/solana/error2.sol:11:8-10"]
+	note_21 [label="location of previous definition\ntests/contract_testcases/solana/error2.sol:6:7-9"]
+	diagnostic_22 [label="error 'X3' has duplicate field name 'X'\nlevel Error\ntests/contract_testcases/solana/error2.sol:12:23-24"]
+	note_23 [label="location of previous declaration of 'X'\ntests/contract_testcases/solana/error2.sol:12:11-16"]
+	diagnostic_24 [label="mapping type is not permitted as error field\nlevel Error\ntests/contract_testcases/solana/error2.sol:14:11-32"]
+	diagnostic_25 [label="type 'X' not found\nlevel Error\ntests/contract_testcases/solana/error2.sol:15:11-12"]
+	diagnostic_26 [label="expecting 'error', got 'meh'\nlevel Error\ntests/contract_testcases/solana/error2.sol:16:2-5"]
+	diagnostic_27 [label="expression found where 'error' was expected\nlevel Error\ntests/contract_testcases/solana/error2.sol:17:2-6"]
```
LucasSte, line 25: I think this message is misleading. Can't we have anything other than an error after an error definition?

LucasSte (changes requested):

tests/contract_testcases/evm/error.sol
seanyoung: You'll get an error saying that `bar` is duplicate.

src/sema/statements.rs
```diff
@@ -878,45 +877,289 @@ fn statement(
...
+                    ));
+
+                    let _ = expression(
+                        &arg.expr,
+                        context,
+                        ns,
+                        symtable,
+                        diagnostics,
+                        ResolveTo::Unknown,
+                    );
```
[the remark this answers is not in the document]
seanyoung: For example:

```solidity
error Feh();
revert Feh({bar: X});
```
You'll get two diagnostics: field `bar` does not exists  and `X` is not found.

tests/contract_testcases/solana/error2.dot
seanyoung: I don't know what you mean.

`error` is not a keyword. The parser generates a parse tree like `<expression> <name>(<fields>..)`. The first `expression` should be the identifier `error`.

src/sema/statements.rs
```diff
@@ -878,45 +877,289 @@ fn statement(
...
+                    context,
+                    ns,
+                    symtable,
+                    diagnostics,
+                    ResolveTo::Unknown,
+                );
+            }
+        }
+
+        if ns.target != Target::EVM {
```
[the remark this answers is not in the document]
seanyoung: If we moved the check to the top of the function, the result would be exactly the same. There is no early exit.

src/sema/statements.rs
```diff
@@ -878,45 +877,289 @@ fn statement(
...
+                    error.fields.len(),
+                    args.len()
+                ),
+                error.loc,
+                format!("definition of '{}'", error.name),
+            ));
+        }
 
-            res.push(Statement::Expression(*loc, reachable, expr));
+        if ns.target != Target::EVM {
```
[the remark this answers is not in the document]
seanyoung: If we moved the check to the top of the function, the result would be exactly the same. There is no early exit.

src/sema/namespace.rs
```diff
@@ -490,6 +507,106 @@ impl Namespace {
...
 
+    pub(super) fn resolve_error(
+        &mut self,
+        file_no: usize,
+        contract_no: Option<usize>,
+        path: &pt::IdentifierPath,
+        diagnostics: &mut Diagnostics,
+    ) -> Result<usize, ()> {
+        // if we are resolving an event name without namespace (so no explicit contract name
+        // or import symbol), then we should look both in the current contract and global scope
```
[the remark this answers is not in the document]
seanyoung: I don't get what's wrong with the original text or what you are trying to improve.

The only problem I see is `event name` rather than `error name`.

tests/contract_testcases/solana/error2.dot
seanyoung: I am not sure what the developers intent was when we get in this situation. They may have misspelt `error`, or they may have been trying to do somthing completely different. I am not sure.

src/sema/namespace.rs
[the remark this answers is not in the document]
LucasSte: `look` here is ambiguous. I may mean `search` or `see`. If it means `see`, the preposition should be `at` not `in`.

src/sema/statements.rs
```diff
@@ -878,45 +877,289 @@ fn statement(
...
+                args.push(arg);
+            } else if let Some(id) = &error.fields[pos].id {
+                ns.diagnostics.push(Diagnostic::error_with_note(
+                    path.loc,
+                    format!("missing field '{}'", id.name),
+                    error.loc,
+                    format!("definition of '{}'", error.name),
+                ));
+            }
+        }
```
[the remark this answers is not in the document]
LucasSte: I could not*

tests/contract_testcases/evm/error.sol
LucasSte: Won't I be warned that `1.5` is not an integer?

tests/contract_testcases/solana/error2.dot
LucasSte:
```solidity
	error X6(X);
	meh X7();
```
In this example, `meh` is not unexpected construct, but is does not mean we are expecting another error declaration. I could have been an event. I think that saying `expected 'error'` in the message is misleading in this sense.

tests/contract_testcases/evm/error.sol
seanyoung: No, duplicates get resolved with `ResolveTo::Unknown`

src/sema/namespace.rs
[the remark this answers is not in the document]
seanyoung: We mean `search`, not `see`

tests/contract_testcases/solana/error2.dot
seanyoung: What should the diagnostic be then?

src/sema/namespace.rs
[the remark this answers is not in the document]
LucasSte: Please, use `search` then, because it is preciser.

src/sema/namespace.rs
```diff
@@ -490,6 +507,106 @@ impl Namespace {
...
+                        file_no,
+                        Some(contract_no),
+                        id.name.to_owned(),
+                    )) {
+                        None => (),
+                        Some(Symbol::Error(_, error_no)) => {
+                            return Ok(*error_no);
+                        }
+                        sym => {
+                            let error = Namespace::wrong_symbol(sym, id);
```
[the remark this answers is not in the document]
LucasSte: Intersting problem.

tests/contract_testcases/solana/error2.dot
LucasSte: How about `declaration expected, but found 'meh' instead`?

src/sema/namespace.rs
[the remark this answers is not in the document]
seanyoung: I've changed to search, although I think the improvement is marginal.

tests/contract_testcases/solana/error2.dot
seanyoung: There is another case:
```solidity
contract c {
       int foo(int bar);
}
```
I am not sure if there is much we can say other than `expecting 'function', 'error', or 'event'`

seanyoung:
I've added a check that gives a warning if an error definition is not used.

tests/contract_testcases/solana/error2.dot
seanyoung: > This looks like a good option.

I don't think it makes much sense, because it assumes that the developer's intention was to declare a variable and that seems unlikely - why would it have  parameters then.

seanyoung: What's your suggestion?

LucasSte: > What's your suggestion?

I can only think of generic phrases right now, like `incorrect declaration`.

src/sema/namespace.rs
```diff
@@ -490,6 +507,106 @@ impl Namespace {
         }
     }
 
+    pub(super) fn resolve_error(
```
xermicus, line 510: Would be nice to have a comment for this function describing its return value, e.g. what is the usize about and in what case it will return  `Err()`

src/sema/namespace.rs
```diff
@@ -490,6 +507,106 @@ impl Namespace {
...
+
+        let (id, namespace) = path
+            .identifiers
+            .split_last()
+            .map(|(id, namespace)| (id, namespace.iter().collect()))
+            .unwrap();
+
+        let s = self.resolve_namespace(namespace, file_no, contract_no, id, diagnostics)?;
+
+        if let Some(Symbol::Error(_, error_no)) = s {
```
xermicus, line 599: On a quick glance it looks like [ok_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) could be used here

src/sema/statements.rs
```diff
@@ -878,45 +877,293 @@ fn statement(
...
+            {
+                res.push(stmt);
+            }
+            Ok(false)
+        }
+        pt::Statement::Error(_) => unimplemented!(),
+    }
+}
+
+fn revert_pos_arg(
```
xermicus, line 898: Here too, a small comment here would be nice

src/sema/statements.rs
```diff
@@ -878,45 +877,293 @@ fn statement(
...
+
+        Ok(Statement::Revert {
+            loc: *loc,
+            error_no: None,
+            args: arguments,
+        })
+    }
+}
+
+fn revert_named_arg(
```
xermicus, line 1021: Same

src/sema/types.rs
```diff
@@ -779,6 +894,88 @@ fn event_decl(
     (doc, fields)
 }
 
+fn error_decl(
```
xermicus, line 897: Same, deserves a comment

src/sema/unused_variable.rs
```diff
@@ -487,3 +487,26 @@ pub fn check_unused_events(ns: &mut Namespace) {
         }
     }
 }
+
+/// Check for unused errors
```
xermicus, line 491: Sorry about that :rofl: BUT while there are missing comments on other functions, this comment seems overly superficial. I'd either remove or add more context than what's already obvious from the function name (given this is a public function, giving more context would be nice).

xermicus (approved):
LGTM

tests/contract_testcases/evm/error.sol
```diff
@@ -0,0 +1,38 @@
...
+	error E4(int bar, bool foo);
+
+	function named1() public {
+		if (1 != 0) {
+			revert E1({bar: 1, foo: true});
+		} else if (1 != 0) {
+			revert E2({bar: 1});
+		} else if (1 != 0) {
+			revert E3({bar: 1, foo: true, baz: "hey"});
+		} else if (1 != 0) {
```
LucasSte, line 19: Rustc would complain here that all the branches check the same condition. If I am not wrong, clang would do the same!

LucasSte: Perhaps this is something we can do too (in the future).

LucasSte (approved):

src/sema/namespace.rs
seanyoung: `Option.ok_or_else()` transforms `Some(x)` to `Ok(x)` or else calls a closure. In the `Ok()` cases we still want to extract `error_no` from `Symbol::Error(_, error_no)` and only match `Symbol::Error`. I am not sure how `ok_or_else()` can be used here.

Did not know about this function, looks very useful!

src/sema/namespace.rs
seanyoung: Added comment

src/sema/statements.rs
seanyoung: I've added a comment

src/sema/statements.rs
seanyoung: Added comment

src/sema/types.rs
seanyoung: Added comment

src/sema/unused_variable.rs
seanyoung: Absolutely right, I've added a much better comment

PR #1260, opened 2023-04-10
Ensure that the last instruction of a block is reachable

LucasSte:
The PR supersedes #1257.

### DIFF
--- a/src/sema/yul/ast.rs
+++ b/src/sema/yul/ast.rs
@@ -22,12 +22,13 @@ pub struct YulBlock {
     pub loc: pt::Loc,
     pub reachable: bool,
     pub next_reachable: bool,
-    pub body: Vec<YulStatement>,
+    pub statements: Vec<YulStatement>,
 }
 
 impl YulBlock {
+    /// Returns if whatever follows the YulBlock is reachable
     pub fn is_next_reachable(&self) -> bool {
-        self.body.is_empty() || (!self.body.is_empty() && self.next_reachable)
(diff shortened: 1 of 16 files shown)


src/sema/yul/ast.rs
```diff
@@ -124,7 +125,7 @@ pub struct YulFunction {
     pub name: String,
     pub params: Arc<Vec<Parameter>>,
     pub returns: Arc<Vec<Parameter>>,
-    pub body: Vec<YulStatement>,
+    pub body_block: YulBlock,
```
seanyoung, line 128: I'm not sure why this was renamed.

I'm guessing here: did you want to avoid `.body.body`? Maybe it's better to rename `YulBlock` member `body` to something else like `statements`

LucasSte: Yes, I wanted to avoid `.body.body`. I am not keen on `statements`, because the function's body is indeed a YulBlock. Does anything else come to your mind?

seanyoung: Just to be clear, this is what I had in mind (hope this is not redundant).

```rust
pub struct YulBlock {
    pub loc: pt::Loc,
    pub reachable: bool,
    pub next_reachable: bool,
    pub statements: Vec<YulStatement>,
}
```

LucasSte: Ah, I misunderstood. I thought you were referring to `struct YulFunction`.

xermicus (approved):

seanyoung (approved):

PR #1270, opened 2023-04-14
Use rationals rather than float for calculating flow

seanyoung:
Flow calculation involves division followed by addition, e.g. 10/3 + 10/3 + 10/3 is not exactly 10 when using floats. So, use rationals instead.

I *think* `Rational<usize>` will suffice but I chicked out and used BigRational instead.


seanyoung:
There is a divide by zero in the original code and I don't get why it isn't triggered - I needed to add an `if` statement to prevent it from panicing the rational code.

```

seanyoung:
> Good catch! The result of the zero division was never used, because the node has no children to pass the flow to, so this did not impact the correctness of the algorithm.

I'm surprised that it did not cause a division by zero exception (I did not investigate). In release mode the compiler may have optimized this case away, but what about debug mode?

LucasSte (approved):

xermicus (approved):

PR #1271, opened 2023-04-14
Enable the accounts call argument for constructors

LucasSte:
When calling the constructor of a contract, we should be able to specify the account metas for the transaction. We need this feature, so we can automatically pass the accounts for the contract construction.

This PR is the second task in https://github.com/hyperledger/solang/issues/1251


src/emit/solana/mod.rs
```diff
@@ -1127,4 +1128,252 @@ impl SolanaTarget {
...
+                    signer_seeds_len.into(),
+                ],
+                "",
+            )
+            .try_as_basic_value()
+            .left()
+            .unwrap()
+            .into_int_value();
+
+        self.return_error(function, binary, ret, error_message, ns, loc);
```
seanyoung, line 1346: I think `sol_invoke_signed_c` cannot fail on Solana. It either returns success or the  transaction is aborted. So, this code is never executed (the error path).

src/emit/solana/mod.rs
```diff
@@ -1127,4 +1128,252 @@ impl SolanaTarget {
...
+                    parameters.into(),
+                ],
+                "",
+            )
+            .try_as_basic_value()
+            .left()
+            .unwrap()
+            .into_int_value();
+
+        self.return_error(function, binary, ret, error_message, ns, loc);
```
seanyoung, line 1196: Maybe `external_call()` should be re-written to also abort on error (i.e. call `sol_panic()`).

seanyoung (approved):
LGTM

xermicus (approved):

src/emit/solana/mod.rs
LucasSte: > I think

Are you not sure? We should be sure that is does not fail in order to remove this code path.

src/emit/solana/mod.rs
LucasSte: We should panic if `sol_invoke_signed_c` also panics, right?

src/emit/solana/mod.rs
seanyoung: Please look at the solana source code.

src/emit/solana/mod.rs
seanyoung: if `sol_invoke_signed_c` fails, the execution has already been aborted

src/emit/solana/mod.rs
LucasSte: If I am checking this correctly, the function can return something other than zero if the I pass invalid data to it. In this case, we can have a return without aborting.

https://github.com/solana-labs/solana/blob/a27ff1c65efe7858164255362a938cced6d4f055/programs/bpf_loader/src/syscalls/cpi.rs#L925

LucasSte: I'll create a test to eliminate all my doubts.

seanyoung: > If I am checking this correctly, the function can return something other than zero if the I pass invalid data to it. In this case, we can have a return without aborting.
> 
> https://github.com/solana-labs/solana/blob/a27ff1c65efe7858164255362a938cced6d4f055/programs/bpf_loader/src/syscalls/cpi.rs#L925

Where exactly do you see this path?

LucasSte: `cpi_common` is registered as the sys call for SolInvokeSignedC: https://github.com/solana-labs/solana/blob/a27ff1c65efe7858164255362a938cced6d4f055/programs/bpf_loader/src/syscalls/cpi.rs#L492-L514

seanyoung: So where exactly do you see this error path where non-zero is returned?

LucasSte:
@xermicus You merged the main into my branch. Now, there are many things that do not compile. Please, do not do that again (at least on my PRs).

LucasSte:
> FYI our CI was broken, the `ink 4.1.0` umbrella crate pulled in newer versions which no longer compiled on our runners

This has nothing to do with my PR.

seanyoung:
> This has nothing to do with my PR.

Yes, it does. Your PR needs to be up to date with main for it be merged.

LucasSte:
> > This has nothing to do with my PR.
> 
> Yes, it does. Your PR needs to be up to date with main for it be merged.

The fact that something is broken in our runners has nothing to do with my PR.

xermicus:
There was nothing broken in our runners (see the drive by fix in #1284).

So.. I merged main because this PR had 2 approvals already and there were no conflicts anyways. What do you mean with `Now, there are many things that do not compile.`? After merging main[ it was all green](https://github.com/hyperledger/solang/actions/runs/4751864090). I mean eventually you'll have to merge main unless you don't want the PR merged anyways. But sure I can restrain from doing this in the future.

src/emit/solana/mod.rs
LucasSte: I was sure I have seen the error message `failed to invoke sol_invoke_signed_c` I output there in case the function does not return zero while I was getting the PR ready, but I could not recreate any case that fails today. I'll remove the branch.

src/emit/solana/mod.rs
LucasSte: > Maybe `external_call()` should be re-written to also abort on error (i.e. call `sol_panic()`).

Done. A few tests had to be adjusted to accommodate the change.

LucasSte:
> There was nothing broken in our runners (see the drive by fix in #1284).
> 
> So.. I merged main because this PR had 2 approvals already and there were no conflicts anyways. What do you mean with `Now, there are many things that do not compile.`? After merging main[ it was all green](https://github.com/hyperledger/solang/actions/runs/4751864090). I mean eventually you'll have to merge main and fix all problems emerging from that anyways.

I did have some compilation errors and clippy warnings after the merge, but later I found my comment unnecessary, so I deleted it.

>  But sure I can restrain from doing this in the future.

I would appreciate that.

PR #1272, opened 2023-04-14
Encode addreses in base58 for print

LucasSte:
Addresses on Solana are often shown in base58 encoding, so we should print them like so.


tests/substrate_tests/base58_encoding.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    runtime.constructor(0, Vec::new());
+
+    let addr = Address([
+        0x7d, 0x58, 0x39, 0xe2, 0x4a, 0xca, 0xda, 0x33, 0x8c, 0x25, 0x76, 0x43, 0xa7, 0xd2, 0xe0,
+        0x25, 0x45, 0x3f, 0x77, 0xd0, 0x58, 0xb8, 0x33, 0x5c, 0x1c, 0x37, 0x91, 0xbc, 0x67, 0x42,
+        0xb3, 0x20,
+    ]);
+    runtime.function("print_this", addr.encode());
+    assert!(runtime.printbuf.contains(addr.0.to_base58().as_str()));
```
seanyoung, line 28: This is not how address are base58 encoded on Substrate. There is a prefix of some sort, and it uses the blake2 hash: https://github.com/hyperledger/solang/blob/main/src/sema/expression/literals.rs#L211

xermicus: Yes, Substrate uses their so called [SS58 address format](https://docs.substrate.io/reference/address-formats/).

LucasSte: Ok, I return the formatting in Substrate to hex.

stdlib/format.c
```diff
@@ -239,3 +239,33 @@ uint256dec(char *output, uint256_t *val256)
...
 }
+
+static const char b58digits[] = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+
+// https://github.com/bitcoin/libbase58/blob/b1dd03fa8d1be4be076bb6152325c6b5cf64f678/base58.c inspired this code.
+void base58_encode(uint8_t * data, uint32_t data_len, uint8_t * output, uint32_t output_len)
+{
+    uint32_t j, carry, zero_count = 0;
+
+    while(zero_count < data_len && !data[zero_count])
```
seanyoung, line 250: The formatting is different than the rest of the file, e.g. no space after `while` or `for`.

tests/solana_tests/base58_encoding.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    vm.constructor(&[]);
+
+    for _ in 0..10 {
+        let account = account_new();
+        let _ = vm.function("print_this", &[BorshToken::Address(account)]);
+        let mut base_58 = account.to_base58();
+        while base_58.len() < 44 {
+            base_58.insert(0, '1');
+        }
```
seanyoung, line 27: Why is the output different?

seanyoung (changes requested):

stdlib/format.c
LucasSte: Can we adopt `clang-format`?

tests/solana_tests/base58_encoding.rs
LucasSte: Rust's `to_base58` ignores leading zeros in the byte array, so it won't transform them into ones. On the other hand, Solana addresses can start with leading ones: `11128aXFh5abEooZ2ouNDjPjk2TqDaHjG6JkX74vK4q` is a valid address.

src/emit/strings.rs
```diff
@@ -550,3 +570,7 @@ pub(super) fn string_location<'a, T: TargetRuntime<'a> + ?Sized>(
         }
     }
 }
+
+fn address_base58_size(ns: &Namespace) -> usize {
```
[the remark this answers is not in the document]
LucasSte: Fixed

stdlib/format.c
```diff
@@ -239,3 +239,33 @@ uint256dec(char *output, uint256_t *val256)
...
+            carry += 256 * output[j];
+            output[j] = carry % 58;
+            carry /= 58;
+            if (!j)
+            {
+                break;
+            }
+        }
+    }
+
```
xermicus, line 266: Why does the bitcoin implementation memset the leading null bytes to 1 while we don't?

stdlib/format.c
```diff
@@ -239,3 +239,33 @@ uint256dec(char *output, uint256_t *val256)
 
     return output;
 }
+
+static const char b58digits[] = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+
+// https://github.com/bitcoin/libbase58/blob/b1dd03fa8d1be4be076bb6152325c6b5cf64f678/base58.c inspired this code.
+void base58_encode(uint8_t * data, uint32_t data_len, uint8_t * output, uint32_t output_len)
```
xermicus, line 246: Would be great if this function could be tested directly. But I guess we lack the testing infrastructure for our stdlib

tests/solana_tests/base58_encoding.rs
seanyoung: That deserves a comment

stdlib/format.c
seanyoung: There is another PR for this

stdlib/format.c
seanyoung: That's the way solana addresses are encoded. For example the systems account is all zeros (in hex/binary) and represented as `11111111111111111111111111111111` and not as a empty string

src/emit/strings.rs
```diff
@@ -183,27 +188,42 @@ pub(super) fn format_string<'a, T: TargetRuntime<'a> + ?Sized>(
                     };
                 }
                 Type::Address(_) | Type::Contract(_) => {
-                    // for Solana/Substrate, we should encode in base58
+                    // For substrate we should encode in the SS58 format
```
seanyoung, line 191: Should be a FIXME

seanyoung: On Solana, it would be good if `"address_in_hex:{:x}".format(addr)` would still work

seanyoung (changes requested):

src/emit/strings.rs
LucasSte: Fixed

stdlib/format.c
LucasSte: Fixed

xermicus (approved):

PR #1273, opened 2023-04-15
modifiers are allowed in libraries

seanyoung:
Solang did not permit this, but solc does.


xermicus (approved):
LGTM

src/sema/functions.rs
```diff
@@ -355,19 +355,19 @@ pub fn contract_function(
...
-                "function in a library cannot override".to_string(),
+                format!("{} in a library cannot override", func.ty),
             ));
             success = false;
         } else if let Some(pt::Mutability::Payable(_)) = mutability {
             ns.diagnostics.push(Diagnostic::error(
                 func.loc,
-                "function in a library cannot be payable".to_string(),
+                format!("{} in a library cannot be payable", func.ty),
             ));
```
LucasSte, line 371: A sema test would make more sense in this case, so we can not only verify if we are allowing such a case, but also check that the error message is properly displayed.

seanyoung: We have contract_tests for this, which ensure that the diagnostics and its location are correct. I've added some more tests so they're all covered.

tests/evm.rs
```diff
@@ -173,25 +173,21 @@ contract testing  {
...
+        WalkDir::new(
+            Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("testdata/solidity/test/libsolidity/syntaxTests"),
+        )
+        .into_iter(),
+    );
 
     let errors: usize = entries
-        .into_par_iter()
+        .par_bridge()
```
xermicus, line 190: Nice, this is very slick now

xermicus (approved):

Issue #1261, opened 2023-04-11
Duplicate code (part 1)


LucasSte:
My IDE points 320 places with repeated code. I am selecting the glaring ones and opening issues. Here are 12 of them (more to come in other issues).

1. https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/expression/assign.rs#L126-L137

https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/expression/assign.rs#L371-L381

2. https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/expression/assign.rs#L205-L225

https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/expression/arithmetic.rs#L927-L947

3. https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/expression/assign.rs#L350-L368

https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/expression/assign.rs#L384-L402

4.
https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/ast.rs#L190-L197

https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/ast.rs#L212-L217

https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/ast.rs#L538-L543

5.
https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/builtin.rs#L1490-L1514

https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/sema/builtin.rs#L1553-L1577

6.
https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/codegen/cfg.rs#L1686-L1708

https://github.com/hyperledger/solang/blob/77d78369fafc2948f7ce26abcfff86edc18de144/src/codegen/cfg.rs#L1876-L1898

7.
[...]

PR #1274, opened 2023-04-16
[Fix] Removing the duplicate code.

anmode:
fix: #1261

Just wanna know If I am going in the right direction or not. Please guide me so I'll then fix other parts as I am trying to understand the codebase too.

### DIFF
--- a/src/sema/expression/assign.rs
+++ b/src/sema/expression/assign.rs
@@ -13,6 +13,26 @@ use solang_parser::diagnostics::Diagnostic;
 use solang_parser::pt;
 use solang_parser::pt::CodeLocation;
 
+macro_rules! check_immutable {
+    ($context:expr, $diagnostics:expr, $immutable:expr) => {
+        if $immutable {
+            if let Some(function_no) = $context.function_no {
+                if !$context.ns.functions[function_no].is_constructor() {
+                    $diagnostics.push(Diagnostic::error(
+                        $context
+                            .current_source_unit
+                            .source_line($context.current_source_unit_offset)
+                            .unwrap()
+                            .0,
+                        "cannot assign to immutable outside of constructor".to_string(),
+                    ));
+                    return Err(());
+                }
+            }
+        }
+    };
+}
+
 /// Resolve an assignment
 pub(super) fn assign_single(
     loc: &pt::Loc,
@@ -123,18 +143,7 @@ pub(super) fn assign_single(
                 right: Box::new(val.cast(&right.loc(), r_ty, true, ns, diagnostics)?),
             }),
             Type::StorageRef(immutable, r_ty) => {
-                if *immutable {
-                    if let Some(function_no) = context.function_no {
-                        if !ns.functions[function_no].is_constructor() {
-                            diagnostics.push(Diagnostic::error(
-                                *loc,
-                                "cannot assign to immutable outside of constructor".to_string(),
-                            ));
-                            return Err(());
-                        }
-                    }
-                }
-
+                check_immutable!(context, diagnostics, immutable);
                 Ok(Expression::Assign {
                     loc: *loc,
                     ty: *r_ty.clone(),
@@ -368,17 +377,7 @@ pub(super) fn assign_expr(
                 }
             },
             Type::StorageRef(immutable, r_ty) => {
-                if *immutable {
-                    if let Some(function_no) = context.function_no {
-                        if !ns.functions[function_no].is_constructor() {
-                            diagnostics.push(Diagnostic::error(
-                                *loc,
-                                "cannot assign to immutable outside of constructor".to_string(),
-                            ));
-                            return Err(());
-                        }
-                    }
-                }
+                check_immutable!(context, diagnostics, immutable);
 
                 match r_ty.as_ref() {
                     Type::Bytes(_) | Type::Int(_) | Type::Uint(_) => Ok(Expression::Assign {
--- a/src/sema/expression/assign.rs
+++ b/src/sema/expression/assign.rs
@@ -14,16 +14,12 @@ use solang_parser::pt;
 use solang_parser::pt::CodeLocation;
 
 macro_rules! check_immutable {
-    ($context:expr, $diagnostics:expr, $immutable:expr) => {
-        if $immutable {
+    ($immutable:expr, $context:expr, $ns:expr, $diagnostics:expr, $loc:expr) => {
+        if *$immutable {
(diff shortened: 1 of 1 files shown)


xermicus:
Hi @anmode, thanks for the pull request, appreciate your request for early feedback :+1: 

I really don't like such a macro just to spare us very few lines of codes. The only thing this does is adding up to the maintenance burden. Also, please try to use functions, preferably implementations for enums / structs, instead of macros if  possible. 

If you still want to go for the first case, I have some suggestions (they might apply for other cases too).

The first match arm could be rewritten as follows:
```rust
            Type::StorageRef(immutable, r_ty) => match context.function_no {
                Some(function_no) if *immutable && !ns.functions[function_no].is_constructor() => {
                        *loc,
                        "cannot assign to immutable outside of constructor".to_string(),
                    ));
                    Err(())
                }
                _ => Ok(Expression::Assign {
                    loc: *loc,
                    ty: *r_ty.clone(),
                    left: Box::new(var),
                    right: Box::new(val.cast(&right.loc(), r_ty, true, ns, diagnostics)?),
                }),
            },
```

A further improvement could be to avoid constructing the error message twice. To achieve that you can offload it into a `OnceCell`.

Additionally, in this file there are a lot of places where a `diagnostic.push` is followed by `Err()` or `return Err(())`. This could  be offloaded to a helper function instead.

LucasSte:
> Hi @anmode, thanks for the pull request, appreciate your request for early feedback 👍
> 
> I really don't like such a macro just to spare us very few lines of codes. The only thing this does is adding up maintenance burden. Also, please try to use functions, preferably implementations for enums / structs, instead of macros if possible.
> 
> If you still want to go for the first case in that issue, I have some suggestions (they might apply for other cases too).
> 
> The first match arm could be rewritten as follows, which basically removes the duplicated lines. This would also improve readability and gets rid of the triple nested `if` statement:
> 
> ```rust
>             Type::StorageRef(immutable, r_ty) => match context.function_no {
>                 Some(function_no) if *immutable && !ns.functions[function_no].is_constructor() => {
>                     diagnostics.push(Diagnostic::error(
>                         *loc,
>                         "cannot assign to immutable outside of constructor".to_string(),
>                     ));
>                     Err(())
>                 }
>                 _ => Ok(Expression::Assign {
>                     loc: *loc,
>                     ty: *r_ty.clone(),
>                     left: Box::new(var),
>                     right: Box::new(val.cast(&right.loc(), r_ty, true, ns, diagnostics)?),
>                 }),
>             },
> ```
> 
> A further improvement could be to avoid constructing the same error message twice. To achieve that you can offload it into a `OnceCell`.
> 
> Furthermore, in this file there are a lot of places where a `diagnostic.push` is followed by `Err()` or `return Err(())`. This could be offloaded to a helper function instead.

The idea of removing repeated code is to decrease maintenance. If we have to fix the logic in this section, we better fix it in a single place, rather than having to find all occurrences of it in the code. The code you suggested will still repeat the same logic and the same error message in two different places.

xermicus:
On a very quick glance it looks like you can implement a method on `Type` called something like `get_assign_type`. It would return the type (`r_ty` in the current code) if it is allowed, otherwise `Err(()))` out and fill in the diagnostics. Then you'll be able to remove the two outer match statements concerned by this completely (which would remove quite more duplicated code than the IDE detected first hand).

anmode:
> I agree with @LucasSte, there is room for improvement here. If you want you are welcome to proceed. On a very quick glance it looks like you can implement a method on `Type` called something like `get_assign_type`. It would return the type (`r_ty` in the current code) if it is allowed, otherwise `Err(())` out and fill in the diagnostics. Then you'll be able to remove the two outer match statements concerned by this completely (which would remove quite more duplicated code than the IDE detected first hand).

oh okay got that!! Thanks... One more thing. My IDE does not detect duplicate code lol. how to make my IDE of capable detecting duplicate code? I'm using VS code.

xermicus:
> oh okay got that!! Thanks... One more thing. My IDE does not detect duplicate code lol. how to make my IDE of capable detecting duplicate code? I'm using VS code.

Ask ChatGPT :shrug: Jokes aside, it's probably the JetBrains stuff that detects this

anmode:
> > oh okay got that!! Thanks... One more thing. My IDE does not detect duplicate code lol. how to make my IDE of capable detecting duplicate code? I'm using VS code.
> 
> Ask ChatGPT 🤷 Jokes aside, it's probably the JetBrains stuff that detects this

ohh okayy 😂

anmode:
Just to clarify! so i can make type like this?? and this would help me to simplify `match &var_ty `
```
impl Type {
    fn get_rh_assign_type(&self, ns: &Namespace, diagnostics: &mut Vec<Diagnostic>) -> Result<Type, ()> {
        match self {
            Type::Bytes(_) | Type::Int(_) | Type::Uint(_) => Ok(self.clone()),
            Type::Ref(r_ty) => match r_ty.as_ref() {
                Type::Bytes(_) | Type::Int(_) | Type::Uint(_) => Ok(self.clone()),
                _ => {
                        loc,
                        format!("assigning to incorrect type {}", r_ty.to_string(ns)),
                    ));
                    Err(())
                }
            },
            Type::StorageRef(immutable, r_ty) => {
                if *immutable {
                    if let Some(function_no) = context.function_no {
                        if !ns.functions[function_no].is_constructor() {
                                *loc,
                                "cannot assign to immutable outside of constructor".to_string(),
                            ));
                            return Err(());
                        }
                    }
                }

                match r_ty.as_ref() {
                    Type::Bytes(_) | Type::Int(_) | Type::Uint(_) => Ok(self.clone()),
                    _ => {
                            loc,
                            format!("assigning to incorrect type {}", r_ty.to_string(ns)),
                        ));
                        Err(())
                    }
                }
            },
            _ => {
                    loc,
                    "expression is not assignable".to_string(),
                ));
                Err(())
            }
        }
    }
}

xermicus:
Yes this looks like its going into the right direction.

xermicus:
@anmode looks like your PR got stale, hence I'm closing it. Just reply here if you planned to drive this forward within a reasonable time.

anmode:
Hii @xermicus ..so sorry actually my end semester arrives in the middle of..
Those will be completed by 25th of may. I'm interested in solving this issue. I'll continue this.
Thanks

xermicus:
Fine by me

xermicus:
This PR has been stale for a while now; if you still want to finish it please just ping me here

PR #1275, opened 2023-04-17
Allow this.function.selector as bytesN and in pure mutability

seanyoung:
Improve compatibility with solc.

### DIFF
--- a/src/sema/mutability.rs
+++ b/src/sema/mutability.rs
@@ -253,6 +253,18 @@ fn read_expression(expr: &Expression, state: &mut StateCheck) -> bool {
             state.read(loc)
         }
         Expression::Builtin {
+            kind: Builtin::FunctionSelector,
+            args,
+            ..
+        } => {
+            if let Expression::ExternalFunction { .. } = &args[0] {
+                // in the case of `this.func.selector`, the address of this is not used and
+                // therefore does not read state. Do not recurse down the `address` field of
+                // Expression::ExternalFunction
+                return false;
+            }
+        }
+        Expression::Builtin {
             loc,
             kind:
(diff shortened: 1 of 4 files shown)


src/sema/mutability.rs
```diff
@@ -252,6 +252,18 @@ fn read_expression(expr: &Expression, state: &mut StateCheck) -> bool {
         Expression::Subscript { loc, array_ty, .. } if array_ty.is_contract_storage() => {
             state.read(loc)
         }
+        Expression::Builtin {
+            kind: Builtin::FunctionSelector,
```
xermicus, line 256: Is this binding necessary?

xermicus (approved):

src/sema/mutability.rs
seanyoung: I am not following.  Here we are filtering on `kind == Builtin::FunctionSelector`

tests/contract_testcases/solana/selector.sol
```diff
@@ -0,0 +1,7 @@
+contract c {
+    function g() pure public returns (bytes8) {
+        return this.f.selector ^ this.x.selector;
+    }
```
LucasSte, line 4: This deserves a runtime test.

LucasSte (commented):
Why is this change necessary?

src/sema/mutability.rs
xermicus: Of course, I was not paying attention

tests/contract_testcases/solana/selector.sol
seanyoung: Runtime test added

PR #1277, opened 2023-04-17
`memory_size_of()` needs guarded recursion

xermicus:
This fixes the crash in ` ./solang-parser/testdata/solidity/test/libsolidity/syntaxTests/structs/recursion/return_recursive_structs2.sol`:

```
❯ cargo run -- compile --target substrate  ./solang-parser/testdata/solidity/test/libsolidity/syntaxTests/structs/recursion/return_recursive_structs2.sol

error: struct 'S' has infinite size
  ┌─ /home/cyrill/mess/solang/solang-parser/testdata/solidity/test/libsolidity/syntaxTests/structs/recursion/return_recursive_structs2.sol:4:12
  │
4 │     struct S { uint a; S[2][] sub; }
  │            ^           ---------- recursive field 'sub'

error: missing return statement
  ┌─ /home/cyrill/mess/solang/solang-parser/testdata/solidity/test/libsolidity/syntaxTests/structs/recursion/return_recursive_structs2.sol:6:6
  │
6 │     }
  │      ^
```

Ideally, instead of adding a dedicated test for this bug, we might want to run against all parser tests on CI too?

### DIFF
--- a/src/sema/types.rs
+++ b/src/sema/types.rs
@@ -1441,7 +1441,15 @@ impl Type {
 
     /// Returns the size a type occupies in memory
     pub fn memory_size_of(&self, ns: &Namespace) -> BigInt {
-        match self {
+        self.memory_size_of_internal(ns, &mut HashSet::new())
+    }
+
+    pub fn memory_size_of_internal(
+        &self,
+        ns: &Namespace,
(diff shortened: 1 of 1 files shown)


seanyoung (approved):
I am creating a PR which will test this  - no need for a test here.

PR #1279, opened 2023-04-18
Run Ethereum Solidity semantic tests through sema

seanyoung:
There are many failures. First there are some tests which cause panics in sema. These are explicitly excluded. There are a further 1062 failures. Note the line:

	assert_eq!(errors, 1062);

To see the failures, see:

	cargo test --test evm ethereum_solidity_tests -- --nocapture

### DIFF
--- a/tests/evm.rs
+++ b/tests/evm.rs
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use solang::{file_resolver::FileResolver, parse_and_resolve, sema::ast, Target};
-use std::ffi::OsStr;
+use std::{ffi::OsStr, fs, path::Path};
+use walkdir::WalkDir;
 
 fn test_solidity(src: &str) -> ast::Namespace {
     let mut cache = FileResolver::new();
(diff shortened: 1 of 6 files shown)


tests/evm.rs
```diff
@@ -166,3 +167,134 @@ contract testing  {
...
+    let errors: usize = semantic_tests
+        .into_iter()
+        .chain(syntax_tests.into_iter())
+        .map(|entry| {
+            let file_name = entry.file_name().to_string_lossy();
+            let path = entry.path().parent().unwrap();
+
+            // FIXME: max_depth_reached_4.sol causes a stack overflow in resolve_expression.rs
+            // FIXNE: others listed explicitly cause panics and need fixing
+            if !file_name.ends_with("max_depth_reached_4.sol")
```
xermicus, line 200: This could go into a `filter_map`

tests/evm.rs
```diff
@@ -166,3 +167,134 @@ contract testing  {
...
+#[test]
+fn ethereum_solidity_tests() {
+    let error_matcher = regex::Regex::new(r"// ----\r?\n// \w+Error( \d+)?:").unwrap();
+
+    let semantic_tests = WalkDir::new(
+        Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("testdata/solidity/test/libsolidity/semanticTests"),
+    )
+    .into_iter()
+    .collect::<Result<Vec<_>, _>>()
```
xermicus, line 180: Is this `collect` necessary? The next thing we do is `into_iter`

tests/evm.rs
```diff
@@ -166,3 +167,134 @@ contract testing  {
...
+        Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("testdata/solidity/test/libsolidity/syntaxTests"),
+    )
+    .into_iter()
+    .collect::<Result<Vec<_>, _>>()
+    .unwrap();
+
+    let errors: usize = semantic_tests
+        .into_iter()
+        .chain(syntax_tests.into_iter())
```
xermicus, line 193: Could be worthwhile a parallel iter with rayon?

xermicus (approved):
Looks good overall, I have left some nits

tests/evm.rs
xermicus: Nice.

Has been a while since I last used rayon, but sometimes you need to use [the parallel bridge](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelBridge.html), might be such a case. But whatever, done is better than perfect :+1:

seanyoung: Right, I did not know about that - it works! I'll add this in another PR

PR #1281, opened 2023-04-19
Do not panic if the value of a constant cannot be evaluated

seanyoung:
If a constant value makes no sense, then using that constant should not result in any error (previously this paniced).


xermicus (approved):
LGTM

Issue #1283, opened 2023-04-20
Inconsistent parsing behavior in `for` loops


DaniPopes:
Code:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.19;

contract C {
    function f() external {
        for (uint256 i = 0; i < 100; uint256 x = 0) {}
    }
}
```

`solang` compiles this just fine, while `solc` throws an error:

```error
Error: Expected ')' but got identifier
 --> a.sol:6:46:
  |
6 |         for (uint256 i = 0; i < 100; uint256 x = 0) {}
  |                                              ^
```

This is because the last part (`update`) of the for loop is parsed as a statement in `solidity.lalrpop`:

https://github.com/hyperledger/solang/blob/5d395e41f031eb087ff8033b615c5195c7899a0b/solang-parser/src/solidity.lalrpop#L771-L773

While Solidity's specification defines this as an expression ([link](https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.forStatement)).

I'm opening this issue because I don't know if this is intended/part of the Solang dialect, or if it's a bug.

DaniPopes:
Thanks @seanyoung!

PR #1285, opened 2023-04-22
The next part of a for loop is an expression, not a statement

seanyoung:
This should not parse:

	contract C {
	    function f() external {
		for (uint256 i = 0; i < 100; uint256 x = 0) {}
	    }
	}

Fixes https://github.com/hyperledger/solang/issues/1283.


src/codegen/statements.rs
```diff
@@ -421,22 +421,10 @@ pub(crate) fn statement(
...
-                        );
-                        body_reachable = stmt.reachable();
-                    }
+                if let Some(next) = next {
+                    expression(next, cfg, contract_no, Some(func), ns, vartab, opt);
+
+                    body_reachable = next.ty() != Type::Unreachable;
                 }
 
                 if body_reachable {
```
xermicus, line 430: Looks like this could be entangled a bit:

```rust
                if let Some(next) = next {
                    expression(next, cfg, contract_no, Some(func), ns, vartab, opt);

                   if next.ty() != Type::Unreachable {
                       cfg.add(vartab, Instr::Branch { block: body_block });
                   }
                }
```

src/codegen/statements.rs
```diff
@@ -530,21 +518,10 @@ pub(crate) fn statement(
...
-                    loops,
-                    placeholder,
-                    return_override,
-                    opt,
-                );
+            if let Some(next) = next {
+                expression(next, cfg, contract_no, Some(func), ns, vartab, opt);
 
-                next_reachable = stmt.reachable();
+                next_reachable = next.ty() != Type::Unreachable;
```
xermicus, line 524: Same here, the below if statement could go directly in here? And the same on line 511

xermicus (approved):

src/codegen/statements.rs
seanyoung: We still want the branch even if there is no `next`

src/codegen/statements.rs
seanyoung: Again we want the branch also if there is no next

src/codegen/statements.rs
xermicus: Ah yes you a right

Issue #1264, opened 2023-04-12
Duplicate code (part 4)


LucasSte:
2.
https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/sema/statements.rs#L1043-L1063

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/sema/expression/function_call.rs#L1715-L1735

======================================
3.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/storage.rs#L152-L158

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/solana/target.rs#L776-L783

=======================================
4.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/storage.rs#L582-L590

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/storage.rs#L760-L769

=======================================
5.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/storage.rs#L693-L703

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/storage.rs#L739-L749

========================================
6.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/codegen/storage.rs#L114-L135

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/codegen/storage.rs#L288-L309

========================================
7.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/strings.rs#L236-L246

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/strings.rs#L260-L270

[...]

PR #1286, opened 2023-04-23
[Fix] duplicate code (part 4)

tareknaser360:
Related token: #1264
This PR addresses the issue of duplicate code (part 4). I have added a new function `process_event` that helps avoid code duplication.
If this commit is good, I will continue working on resolving the remaining issues in separate commits.


src/sema/statements.rs
```diff
@@ -1174,6 +1174,35 @@ fn revert_named_arg(
     }
 }
 
+fn process_event(
```
xermicus, line 1177: The name of the function should describe what it does or at least deserves a comment.

src/sema/statements.rs
```diff
@@ -1174,6 +1174,35 @@ fn revert_named_arg(
...
+    ns: &mut Namespace,
+    function_no: &usize,
+    event_no: &usize,
+    event_loc: &pt::Loc,
+    loc: &pt::Loc,
+    cast_args: &[Expression],
+    event_nos: &Vec<usize>,
+    diagnostics: &mut Diagnostics,
+    temp_diagnostics: &Diagnostics,
+) -> Result<Option<Statement>, ()> {
```
xermicus, line 1188: I think this function could be a closure.


```
let to_stmt =
        |ns: &mut Namespace, event_no: usize, event_loc: pt::Loc, cast_args: Vec<Expression>| {
            if !ns.functions[function_no].emits_events.contains(&event_no) {
                ns.functions[function_no].emits_events.push(event_no);
            }

            Statement::Emit {
                loc: *loc,
                event_no,
                event_loc,
                args: cast_args,
            }
        };
```

Then you can do:
```
               if matches {
                    return Ok(to_stmt(ns, *event_no, event_loc.clone(), cast_args));
                }
                if event_nos.len() > 1 && diagnostics.extend_non_casting(&errors) {
                    return Err(());
                }
```

xermicus (changes requested):
Just moving some code into a couple helper function should be the last resort. Did you try to refactor the logic of the function `emit_event`? This function has many loops and deep nesting, also it seems to have more duplicate code in it. For example there are two for loops `for event_no in &event_nos {}`

seanyoung:
Your commit also needs a `Signed-off-by: Real Name <email@address>`

src/sema/statements.rs
tareknaser360: Thanks for the review. I called this closure to avoid duplication.

src/sema/statements.rs
tareknaser360: Does the new closure `to_stmt` need a comment?

tareknaser360:
> Your commit also needs a `Signed-off-by: Real Name <email@address>`

Thanks for the feedback.
I signed the new commit.

seanyoung (approved):

LucasSte (approved):

xermicus (approved):

PR #1289, opened 2023-04-25
Solidity keywords were not correctly handled in yul

seanyoung:
Many Solidity keywords like `abstract` were incorrectly handled as keywords in yul. 

Cc: @DaniPopes


xermicus (approved):

src/codegen/yul/builtin.rs
```diff
@@ -122,23 +125,43 @@ pub(crate) fn process_builtin(
...
+        YulBuiltInFunction::Return => {
+            if ns.target != Target::EVM {
+                let function_ty = builtin_ty.get_prototype_info();
+                unreachable!("{} yul builtin not implemented", function_ty.name);
+            }
+
+            let addr = expression(&args[0], contract_no, ns, vartab, cfg, opt).cast(&Type::Ref(Type::Uint(256).into()), ns);
+            let length = expression(&args[0], contract_no, ns, vartab, cfg, opt).cast(&Type::Uint(256), ns);
+
+            cfg.add(vartab, Instr::Return { value: vec![addr, length] });
```
LucasSte, line 150: I am not sure this is the correct return values. According to [Solidity documentation](https://docs.soliditylang.org/en/latest/yul.html#evm-dialect), the returned value is not the address and the length but `mem[p…(p+s))`. You are not loading the data for returning. 

`return(p, s) => end execution, return data mem[p…(p+s)]`

src/codegen/yul/builtin.rs
```diff
@@ -122,23 +125,43 @@ pub(crate) fn process_builtin(
...
+            Expression::Poison
+        }
+        YulBuiltInFunction::Stop // Stop is the same as return(0, 0)
+        | YulBuiltInFunction::Revert => {
```
LucasSte, line 154: Revert should revert the changes in this call. If we only return, we are not reverting anything.

solang-parser/src/solidity.lalrpop
```diff
@@ -879,9 +879,57 @@ YulIdentifier: Identifier = {
...
+    <l:@L> "throw" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "throw".to_string()},
+    <l:@L> "try" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "try".to_string()},
+    <l:@L> "using" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "using".to_string()},
+    <l:@L> "view" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "view".to_string()},
+    <l:@L> "virtual" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "virtual".to_string()},
+    <l:@L> "while" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "while".to_string()},
+    <l:@L> <i:Int> <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: format!("int{}", i)},
+    <l:@L> <u:Uint> <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: format!("uint{}", u)},
+    <l:@L> "string" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "string".to_string()},
+    <l:@L> "unchecked" <r:@L> => Identifier{loc: Loc::File(file_no, l, r), name: "unchecked".to_string()},
```
LucasSte, line 932: How is this tested?

LucasSte (changes requested):

src/codegen/yul/builtin.rs
LucasSte: In addition, these changes are not tested in any runtime test.

src/codegen/yul/builtin.rs
LucasSte: These new implementations must be tested as well.

src/codegen/yul/builtin.rs
seanyoung: We don't have EVM code gen so these cannot be tested. These were simply added to make the `fn ethereum_solidity_tests()` work.

src/codegen/yul/builtin.rs
seanyoung: We don't have EVM code gen so these cannot be tested. These were simply added to make the `fn ethereum_solidity_tests()` work.

solang-parser/src/solidity.lalrpop
seanyoung: `fn ethereum_solidity_tests()`

src/codegen/yul/builtin.rs
LucasSte: Not only the implementation is not following the documentation, but also we cannot test if this indeed works. I don't like the idea of adding code that we cannot ensure it is working. Why don't we raise a warning in sema when these buildings are used? This way, we don't go through this code path.

Also, why do we still have codegen for EVM? Maybe it is time to disable this.

seanyoung: Adding EVM support is hugely beneficial to solang. 
* We can run the eth solidity semantic tests through sema and codegen which is yielding an enormous amount of issues in our code. This PR would not exist if it weren't for those tests.
* You can see that [solang-parser](https://crates.io/crates/solang-parser) is quite poplar because it can parse eth solidity. If sema could also work with eth solidity, then we could publish `solang-sema` and that crate would get users, contributors, testers, and bug reports
* Our language server would be much more popular if we have better evm support
* It would be great if Solang had EVM support including codegen and emit, and could create contracts which can execute on EVM chains. This would require a lot of work and is not something we would directly spend time on, but we can do mentorships and such things.

Yes, this particular bit of code path is not tested, but we need to run evm contracts through codegen to get all the warnings/errors. The problems with evm codegen will hopefully be fixed later. In the mean time, this is sufficient for our purposes (getting all diagnostics and ast for evm solidity contracts).

LucasSte: I suggest you to at least implement the builtin correctly then.
Adding a large comment explaining that the implementation is not correct and should be fixed would also help.

LucasSte: > I suggest you to at least implement the builtin correctly then.

This is indeed important and would allow you to test for the other targets.

seanyoung: This is totally unnecessary work. We're trying to fix a problem in the parser and make the test pass.

LucasSte: As we have a time constraint, please, write a comment there explaining the situation (i.e the builtin is not correctly implemented because we wanted the parser tests to work). I would accept your PR this way.

I have been thinking too if a `Instr::EVMUnimplemented` would make sense in this case, but I worry it would consume a little time to implement.

seanyoung: I've given this some thought. In order to implement these yul builtins, we really need pointers in CFG. We could re-implement `Instr::Unreachable { ... }` as having a pointer and a length; I think that would make sense. However, for `Instr::Return { .. }` this is not so clear. What does mean if we do a yul return from a private function?

So I've gone down the route of `Instr::Unimplemented` for now. Maybe implementing an evm target will also help us understand better what to do with evm builtins on non-evm targets.

src/codegen/yul/builtin.rs
```diff
@@ -135,8 +138,14 @@ pub(crate) fn process_builtin(
...
         | YulBuiltInFunction::Origin
         => {
-            let function_ty = builtin_ty.get_prototype_info();
-            unreachable!("{} yul builtin not implemented", function_ty.name);
+            if ns.target != Target::EVM {
+                let function_ty = builtin_ty.get_prototype_info();
+                unreachable!("{} yul builtin not implemented", function_ty.name);
+            }
+
+            // Sema will only allow this for EVM, this is a placeholder until correct codegen is in place
```
LucasSte, line 146:
```suggestion
            // Sema will only allow this for EVM. This is a placeholder until correct codegen is in place
```

LucasSte (approved):

src/codegen/cfg.rs
```diff
@@ -181,6 +181,12 @@ pub enum Instr {
     },
     /// Return a code at the end of a function
     ReturnCode { code: ReturnCode },
+    /// For unimplemented code, e.g. unsupported yul builtins. This instruction should
+    /// only occur for the evm target, for which no emit is implemented yet. Once evm emit
+    /// is implemented and all yul builtins are supported, this instruction should
+    /// be removed. We only have this so we can pass evm code through sema/codegen, which is used
+    /// by the language server and the ethereum solidity tests.
+    Unimplemented { reachable: bool },
```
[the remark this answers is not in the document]
seanyoung: `fn edges()` checks this, see https://github.com/hyperledger/solang/pull/1289/files#diff-0a93faee00c132da152755d6caf824558ac734b5c8d010f5b9b62ec696a27d48R448-R450

seanyoung:
@xermicus the PR has changed a lot since your reviewed it

src/codegen/cfg.rs
```diff
@@ -437,7 +444,8 @@ impl BasicBlock {
                 | Instr::SelfDestruct { .. }
                 | Instr::ReturnCode { .. }
                 | Instr::ReturnData { .. }
-                | Instr::Return { .. } => {
+                | Instr::Return { .. }
+                | Instr::Unimplemented { reachable: false } => {
```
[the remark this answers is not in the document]
LucasSte: Ah, I just found out. Please, disregard.

src/codegen/cfg.rs
```diff
@@ -181,6 +181,12 @@ pub enum Instr {
     },
     /// Return a code at the end of a function
     ReturnCode { code: ReturnCode },
+    /// For unimplemented code, e.g. unsupported yul builtins. This instruction should
```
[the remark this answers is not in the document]
seanyoung: So, we can't error out on unimplemented yul builtins in sema.

PR #1290, opened 2023-04-25
storage references return values must be set

seanyoung:
solc permits returns values to not have a value assigned, unless its a storage reference. This means that

	function foo() public returns (bool) {}

compiles without warnings, however

	function bar() public returns (string storage) {}

fails with an error diagnostic.

This PR also changes the `contract_testcases`. The expected diagnostics are listed in the source file, in the same way as this is done for solc tests. This makes the `contract_testcases` much more useful. 

I am wondering if there is much justification for keeping the .dot files in the repo any more.


tests/contract.rs
```diff
@@ -153,3 +158,49 @@ fn add_file(cache: &mut FileResolver, path: &Path, target: Target) -> io::Result
...
+
+        if found.is_empty() && !line.starts_with("// ----") {
+            continue;
+        }
+
+        found.push_str(&line);
+        found.push('\n');
+    }
+
+    if !found.is_empty() {
```
xermicus, line 203: I wonder if this should be an `assert!(!found.is_empty())` instead? It would force the developer of the test to think about it

xermicus (commented):
This is great, thanks!

I don't see any reason to keep the DOT files any longer. Working with this seems so much more convenient.

tests/contract.rs
seanyoung: Totally agree, I've changed this as suggested

src/sema/statements.rs
```diff
@@ -210,8 +210,8 @@ pub fn resolve_function_body(
         ns.functions[function_no].modifiers = modifiers;
     }
 
-    // a function with no return values does not need a return statement
-    let mut return_required = !def.returns.is_empty();
+    // only returns with storage references must be a given a value
```
LucasSte, line 213: This sentence is unclear.

src/sema/statements.rs
```diff
@@ -234,6 +232,10 @@ pub fn resolve_function_body(
                 symtable.returns.push(pos);
             }
         } else {
+            if ret.ty.is_contract_storage() {
+                return_required = true;
+            }
```
LucasSte, line 237:
```suggestion
                return_required |= ret.ty.is_contract_storage();
```

tests/contract_testcases/evm/call/call.sol
```diff
@@ -4,4 +4,6 @@
                 (bool s, bytes memory bs) = a.delegatecall{value: 2}("");
             }
         }
-        
\ No newline at end of file
+        
+// ----
+// error (117-145): 'delegatecall' cannot have value specifed
```
LucasSte, line 9: What if we write the line numbers inside the parenthesis? This way, the test is even more human readable.

tests/contract.rs
```diff
@@ -153,3 +136,49 @@ fn add_file(cache: &mut FileResolver, path: &Path, target: Target) -> io::Result
 
     Ok(filename.to_string())
 }
+
+fn check_diagnostics(path: &Path, ns: &Namespace) {
+    let mut expected = "// ----\n".to_owned();
+
+    for diag in ns.diagnostics.iter() {
+        if diag.level == Level::Warning || diag.level == Level::Error {
+            expected = format!(
```
LucasSte, line 145: Shouldn't this be `expected.push`? This looks like you are reassigning the variable.

tests/evm.rs
```diff
@@ -255,7 +255,7 @@ fn ethereum_solidity_tests() {
         })
         .sum();
 
-    assert_eq!(errors, 1155);
+    assert_eq!(errors, 1094);
```
LucasSte, line 258: Why has this changed?

tests/contract_testcases/evm/error.sol
```diff
@@ -36,3 +36,23 @@ contract c is e {
 
 	}
 }
+
+// ----
```
LucasSte, line 40: What do you think of the entry header being this?
`// ---- EXPECTED DIAGNOSITCS ----`

src/sema/statements.rs
seanyoung: Why change this? What's the difference?

tests/evm.rs
seanyoung: Because this change fixes a whole bunch of failures in the eth-solidity-tests-through-sema tests

src/sema/statements.rs
seanyoung: Good point - I've re-written it.

tests/contract_testcases/evm/call/call.sol
seanyoung: I agree and updated the tests accordingly

tests/contract_testcases/evm/error.sol
seanyoung: I've changed it to something similar without the upper case

src/sema/statements.rs
LucasSte: It is just for shortness. Feel free to leave your code as it is.

src/sema/statements.rs
```diff
@@ -210,8 +210,12 @@ pub fn resolve_function_body(
         ns.functions[function_no].modifiers = modifiers;
     }
 
-    // a function with no return values does not need a return statement
-    let mut return_required = !def.returns.is_empty();
+    // If there is no return statement, then any unnamed return types will
```
LucasSte, line 213:
```suggestion
    // If there is no return statement, any unnamed return types will
```

src/sema/statements.rs
```diff
@@ -210,8 +210,12 @@ pub fn resolve_function_body(
...
+    // implicitly be 0. If there is a return type which is a storage
+    // reference (e.g. int[] storage), then a value must be given via
```
LucasSte, line 215:
```suggestion
    // reference (e.g. int[] storage), a value must be given via
```

tests/contract_testcases/evm/call/call.sol
```diff
@@ -4,4 +4,6 @@
                 (bool s, bytes memory bs) = a.delegatecall{value: 2}("");
             }
         }
-        
\ No newline at end of file
+        
+// ---- Expect: diagnostics ----
+// error: 4:45-73: 'delegatecall' cannot have value specifed
```
LucasSte, line 9:
```suggestion
// error: 4:45-73: 'delegatecall' cannot have value specified
```

tests/contract_testcases/solana/import_free_function.sol
```diff
@@ -9,3 +9,6 @@ function bar() {
 	x = X.foo();
 	x = X.foo({});
 }
+
+// ---- Expect: diagnostics ----
+// warning: 1:1-29: function can be declared 'pure'
```
LucasSte, line 14: Does your formatting deal with diagnostics that span over multiple lines?

LucasSte (approved):

src/sema/dotgraphviz.rs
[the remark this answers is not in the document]
seanyoung: We still have `solang --emit ast-dot`, that is useful right?

tests/contract_testcases/solana/import_free_function.sol
seanyoung: Yes, we use the standard conversion routine so this works fine

tests/contract_testcases/evm/call/call.sol
seanyoung: Fixed

src/sema/dotgraphviz.rs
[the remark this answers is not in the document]
LucasSte: > We still have `solang --emit ast-dot`, that is useful right?

I have never used it, so I can't tell.

xermicus (approved):

PR #1294, opened 2023-05-02
Update ink! caller integration test

xermicus:
Updates the caller contract to use the latest ink! release. I also re-structured the solidity code to try and make what's going on in there obvious.


integration/substrate/inkee.sol
```diff
@@ -1,6 +1,24 @@
...
+function ink_result(uint32 _value) returns (InkResult) {
+    return InkResult({ok: 0, value: _value});
+}
+
 contract Inkee {
+    // We use this selector just for convenience
     @selector([1, 2, 3, 4])
-    function echo(uint32 v) public pure returns (bool, uint32) {
-        return (false, v);
+    function echo(uint32 v) public returns (InkResult) {
```
seanyoung, line 21: why can't this function be pure any more? Is this a limitation of free standing functions?

seanyoung: I just couldn't help myself and tested it - if you declare `function ink_result()` as `pure` then `function echo()` can be pure too, if that matters at all.

integration/substrate/inkee.sol
```diff
@@ -1,6 +1,24 @@
+// ink! contracts are expected to return a value of type `Result<T, LangError>` (rusts "Result" type).
```
LucasSte, line 1:
```suggestion
// ink! contracts are expected to return a value of type `Result<T, LangError>` (Rust's "Result" type).
```

integration/substrate/inkee.sol
```diff
@@ -1,6 +1,24 @@
...
+// The ink! call builder only supports decoding the result into this `Result` type.
+// This works fine for ink! because ink! contracts are supposed to always return this type.
+// 
+// In solidity, nor does this type exist, neither does the concept of a "LangError".
```
LucasSte, line 5:
```suggestion
// In Solidity, not only does this type not exist, but also the concept of a "LangError" is inexistent.
```

LucasSte (approved):

integration/substrate/inkee.sol
xermicus: Good idea

PR #1295, opened 2023-05-03
Compare the full 64 bits address on Solana

seanyoung:
vector_new() takes an initializer arguments, which has as special value
for "no initializer". We should check the full 64 bits, rather than
casting the lower 32 bits to int.

This change exposes a bug in the old BPF target in Solana's LLVM fork. Switching 
to the newer SBF target fixes the issue.


LucasSte:
> This change exposes a bug in the old BPF target in Solana's LLVM fork. Switching to the newer SBF target fixes the issue.

The LLVM compilation instructions in our docs must be updated as well.

seanyoung:
> The LLVM compilation instructions in our docs must be updated as well.

I don't think there is anything that needs updating there, as the LLVM target are not listed. I have replaced occurrences of BPF with SBF.

.github/workflows/test.yml
```diff
@@ -41,14 +41,14 @@ jobs:
...
-    container: ubuntu:20.04
+    container: ubuntu:22.04
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
       - name: Install Python and git
         run: |
           apt-get update
-          apt-get install -y python3-pip git
+          apt-get install -y python3-pip git pkg-config libcairo-dev
```
LucasSte, line 51: These changes are also in #1298. I think it is best the leave them only in one place.

LucasSte (approved):

.github/workflows/test.yml
seanyoung: That was done to make the tests pass. Once one PR is merged, the rebase will remove the duplicate code

LucasSte: > That was done to make the tests pass. Once one PR is merged, the rebase will remove the duplicate code

This is why these changes better suit in a separate PR.

xermicus (approved):

docs/language/expressions.rst
```diff
@@ -143,7 +143,7 @@ an bitwise XOR of all function selectors in the interface. This makes it possibl
 an interface at runtime, which can be used to write a `supportsInterface()` function as described
 in the EIP.
 
-The contract code for a contract, i.e. the binary WebAssembly or BPF, can be retrieved using the
+The contract code for a contract, i.e. the binary WebAssembly or SBF, can be retrieved using the
```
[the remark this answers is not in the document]
seanyoung: I've replaced them all with `Solana SBF`. Thanks for the suggestion, this is much clearer

PR #1297, opened 2023-05-03
Solve YUL panics for EVM semantic tests

LucasSte:
I fixed a problem in which we were trying to fetch a function header that was not added to the set of processed headers, because there was an error in processing it.


seanyoung:
it's evm semantic tests (not parsing)

seanyoung (approved):

xermicus (approved):

PR #1299, opened 2023-05-05
Enable tests for stdlib

LucasSte:
This PR introduces the following changes:
1. Stdlib is now build directly with `cargo build`, so all the wasm and bpf files have been removed from the repository.
2. Enable the tests for the stdlib. If nothing fails within two minutes, the test passes.
3. Enable the linter `clang-format` for the stdlib. All C files have been reformatted.


LucasSte:
Solana integration tests are failing, but I believe they are not related to this PR. @seanyoung reported he got the same error in his PR.

build.rs
```diff
@@ -5,6 +5,16 @@ use std::process::Command;
...
+        Command::new("make")
+            .args(["-C", "stdlib", "solana_stdlib"])
+            .output()
+            .expect("Could not build Solana stdlib");
+
+        Command::new("make")
+            .args(["-C", "stdlib", "substrate_stdlib"])
+            .output()
+            .expect("Could not build Substrate stdlib");
+
```
seanyoung, line 17: Why is this separated out into two commands?

stdlib/Makefile
```diff
@@ -2,18 +2,23 @@ CC=clang
 BIT_INT_FLAGS=-Xclang -fexperimental-max-bitint-width=512
 CFLAGS=$(TARGET_FLAGS) -emit-llvm -O3 -ffreestanding -fno-builtin -Wall -Wno-unused-function $(BIT_INT_FLAGS)
 
-bpf/%.bc wasm/%.bc: %.c
+bpf/%.bc wasm/%.bc: %.c | outputs_dirs
```
seanyoung, line 5: Does this work on pattern recipe? I would expect this
```
$(SOLANA) $(WASM): | outputs_dirs
```
If it works, then fine of course :)

seanyoung (changes requested):

stdlib/Makefile
```diff
@@ -2,18 +2,23 @@ CC=clang
...
+
 
 $(SOLANA): TARGET_FLAGS=--target=sbf
 $(WASM): TARGET_FLAGS=--target=wasm32
 
-bpf/solana.bc: solana.c solana_sdk.h
+bpf/solana.bc: solana.c solana_sdk.h | outputs_dirs
+
+outputs_dirs:
+	$(info $(shell mkdir -p bpf wasm))
```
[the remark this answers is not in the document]
LucasSte: I am not entirely proficient in Makefile. For this, I have followed the fourth comment in [this SO answer.](https://stackoverflow.com/a/45048948):

This is better: `$(info $(shell mkdir -p $(DIRS)))` Without the `$(info ...)`, the output of the mkdir command will be pasted into the Makefile, leading to syntax errors at best. The `$(info ...)` call ensures that a) the errors (if any) are visible to the user, and b) that the function call expands to nothing.

stdlib/Makefile
```diff
@@ -2,18 +2,23 @@ CC=clang
...
+bpf/%.bc wasm/%.bc: %.c | outputs_dirs
 	$(CC) -c $(CFLAGS) $< -o $@
 
 SOLANA=$(addprefix bpf/,solana.bc bigint.bc format.bc stdlib.bc ripemd160.bc heap.bc)
 WASM=$(addprefix wasm/,ripemd160.bc stdlib.bc bigint.bc format.bc heap.bc)
 
-all: $(SOLANA) $(WASM)
+solana_stdlib: $(SOLANA)
+substrate_stdlib: $(WASM)
+
```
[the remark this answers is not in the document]
LucasSte: I totally agree that a single target is the best option, but I couldn't make it work. This is what I have tried:

1. Using `all: $(SOLANA) $(WASM)`, `make all` and `make -B` only build Solana.
2. Using `all: $(WASM) $(SOLANA)`, `make all` and `-B` only build WASM and `bpf/solana.bc`.
3. Using `all: solana_stdlib substrate_stdlib`, `make all` only builds Solana.
4.  Using `all: solana_stdlib | substrate_stdlib`, `make all` only builds Solana.
5.  Using `all: substrate_stdlib solana_stdlib`, `make all` only builds Solana.

By having two separate targets, I ensure both will be built.

If you know how to solve this, or can point me to the correct solution, I would adopt a single target.

build.rs
LucasSte: As I explained in your previous comment, I couldn't get `make all` to build everything.

stdlib/Makefile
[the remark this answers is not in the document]
LucasSte: Ah, I might have overread stack overflow. Your solution actually works.

.github/workflows/test.yml
```diff
@@ -360,3 +360,21 @@ jobs:
       working-directory: ./vscode
     - run: vsce package
       working-directory: ./vscode
+
+  std-lib-tests:
+    name: STDLIB
```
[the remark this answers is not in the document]
LucasSte: Changed it to `Stdlib`

stdlib/stdlib.c
```diff
@@ -86,128 +86,131 @@ int __memcmp_ord(uint8_t *a, uint8_t *b, uint32_t len)
...
+
+#ifndef TEST
+
+#ifdef __wasm__
+#define VECTOR_EMPTY ((uint8_t *)~0l)
+#else
+#define VECTOR_EMPTY ((uint8_t *)0l)
+#endif
+
+// Create a new vector. If initial is -1 then clear the data. This is done since a null pointer valid in wasm
```
xermicus, line 164:
```suggestion
// Create a new vector. If initial is -1 then clear the data. This is done since a null pointer is valid in Wasm
```

.github/workflows/test.yml
```diff
@@ -360,3 +360,21 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Lints for stdlib
+        run: |
+          clang-format *.c *.h --style=file --dry-run -Werror
+        working-directory: ./stdlib
+      - name: Stdlib tests
+        run: |
+          clang -DTEST -DSOL_TEST -O3 -Wall solana.c stdlib.c -o test
```
xermicus, line 378: What about a `make test` target?

.github/workflows/test.yml
```diff
@@ -360,3 +360,21 @@ jobs:
...
+    name: Stdlib
+    runs-on: solang-ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Lints for stdlib
+        run: |
+          clang-format *.c *.h --style=file --dry-run -Werror
```
xermicus, line 374: Could add a `make lints` target?

build.rs
```diff
@@ -5,6 +5,11 @@ use std::process::Command;
 fn main() {
     #[cfg(feature = "llvm")]
     {
+        Command::new("make")
```
xermicus, line 8: We now require `make` for building solang, we could mention this in the docs.

While on any developer setup it should be already be installed anyways. But it would be nice to mention for example if someone trying to package it or use it in a container.

xermicus (approved):
LGTM

build.rs
seanyoung: I have been wondering whether this could be replaced with all the commands for building the stdlib, and doing away with the `Makefile`, and along with that the `make` dependency. I was kind of undecided on that.

.github/workflows/test.yml
```diff
@@ -360,3 +360,21 @@ jobs:
...
       working-directory: ./vscode
+
+  std-lib-tests:
+    name: Stdlib
+    runs-on: solang-ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
```
seanyoung, line 371: I don't think it needs submodules. Removing these two lines will make it faster as solc does not have to be cloned.

seanyoung (approved):

PR #1304, opened 2023-05-10
Unbreak readthedocs build

seanyoung:
Longer term I would like to generate the docs in github actions rather then from readthedocs. If there is a failure we would like to know about it.

So, this PR does a bunch of work to fix the docs build. Now, the correct version number is rendered.

The pdf is not generated yet. Not sure if we care about the pdf though.


LucasSte (approved):

xermicus (approved):
Thanks!

PR #1305, opened 2023-05-10
Implement chain extensions builtin

xermicus:
Provide a builtin to call chain extensions. This is the building block for implementing more high level interfaces directly in Solidity.


src/emit/substrate/target.rs
```diff
@@ -1580,14 +1580,73 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
+        .into_int_value();
+
+        log_return_code(binary, "call_chain_extension", ret);
+
+        let ok = binary.context.append_basic_block(function, "ok_block");
+        let done = binary.context.append_basic_block(function, "done_block");
+        let success = binary
+            .builder
+            .build_int_compare(IntPredicate::EQ, ret, i32_zero!(), "ret");
+        binary.builder.build_conditional_branch(success, ok, done);
```
seanyoung, line 1629: What is supposed to happen if the return value from the syscall is not 0?

src/emit/substrate/target.rs
```diff
@@ -1580,14 +1580,73 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
+        .try_as_basic_value()
+        .left()
+        .unwrap();
+        binary
+            .builder
+            .build_store(args[2].into_pointer_value(), buf.into_pointer_value());
+        binary.builder.build_unconditional_branch(done);
+
+        binary.builder.position_at_end(done);
+        ret.into()
```
seanyoung, line 1649: This looks wrong. Why is the syscall return value returned and not the vector?

src/emit/substrate/target.rs
xermicus: [It'll return it to the caller](https://github.com/hyperledger/solang/blob/f85e43ba643e67826dffca0910202b6134db29fd/src/emit/instructions.rs#LL548C50-L548C50)

However now that I think of it, it might be better to just trap. One could come up with the idea to `try-catch` it, but according to the solidity docs this is only supposed to be used for external calls

src/emit/substrate/target.rs
xermicus: Are you sure? [An int value is expected](https://github.com/hyperledger/solang/blob/main/src/emit/instructions.rs#LL532C23-L532C23), a pointer won't work. I guessed that this int value is supposed to be the return code from the builtin.

The return value goes [into the last argument to the builtin](https://github.com/hyperledger/solang/blob/main/src/emit/instructions.rs#L525). I'm quite confident that I read this code correctly and implemented it accordingly. The unit test passes, and out of curiosity I tried it out on a real substrate node with the `fetch-rand` chain extension example. It works as expected.

src/emit/substrate/target.rs
xermicus: Actually I don't know about that logic in `instruction.rs`. Is it guaranteed that we always stay in `call` or `deploy` and this return there will return from the contract? I'm not sure about this. I think we should just trap.

xermicus: I changed the behavior entirely. We should not trap or do anything on a non-zero return value. Instead, we should return this value back to the caller.

In the case of `call_chain_extension`, the return value being non-zero has no apparent meaning. It depends on the chain extensions concrete implementation. Hence we should just return it to the caller, as he should know how to handle this.

tests/substrate.rs
```diff
@@ -711,6 +711,25 @@ impl Runtime {
...
+        id: u32,
+        input_ptr: u32,
+        input_len: u32,
+        output_ptr: u32,
+        output_len_ptr: u32,
+    ) -> Result<u32, Trap> {
+        assert_eq!(id, 123, "unkown chain extension");
+        assert_eq!(read_len(mem, output_len_ptr), 16384);
+
+        let data = read_buf(mem, input_ptr, input_len);
```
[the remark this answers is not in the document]
xermicus: Why not, I changed it so it reverses the input data and writes that back.

src/sema/builtin.rs
```diff
@@ -1658,5 +1658,64 @@ impl Namespace {
...
+            "chain_extension".to_string(),
+            None,
+            Vec::new(),
+            pt::FunctionTy::Function,
+            None,
+            pt::Visibility::Public(Some(pt::Loc::Builtin)),
+            vec![
+                Parameter {
+                    loc: pt::Loc::Builtin,
+                    id: None,
```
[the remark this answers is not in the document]
xermicus: I gave them a name

tests/substrate.rs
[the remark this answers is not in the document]
xermicus: But just to mention, because we memset 0 the output buffer before calling the chain extension, this shouldn't really matter, because the input and output buffer will only be identical if the chain extension actually did copy it.

docs/language/builtins.rst
```diff
@@ -311,6 +311,35 @@ Its underlying type is ``bytes32``, but it will be reported correctly as the ``H
 .. include:: ../examples/substrate/hash_type.sol
   :code: solidity
 
+chain_extension(uint32 ID, bytes input) returns (uint32, bytes)
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+Only available on Substrate. Call the chain extension with the given ``ID`` and ``input`` data.
+Returns the return value from the chain extension and the output data.
+
+This function is a low level interface; the input and output data buffers are just ``bytes``.
```
seanyoung, line 320: Total nitpick but I would replace "just" with "simply" or totally remove it. I am having a hard time explaining why it's better so I'll just leave this here as a suggestion.

docs/language/builtins.rst
```diff
@@ -311,6 +311,35 @@ Its underlying type is ``bytes32``, but it will be reported correctly as the ``H
...
+    This function calls the runtime API `call_chain_extension <https://docs.rs/pallet-contracts/latest/pallet_contracts/api_doc/trait.Version0.html#tymethod.call_chain_extension>`.
+	It assumes that the implementation of the chain extension does
+	- read the input from the ``input_ptr`` parameter, used as a buffer pointer
+	- write potential output into the buffer found at the ``output_ptr`` pointer
+	- respect the output buffer length in ``output_len_ptr`` to prevent OOB writes. The output buffer is 16KB in size.
+	- write the amount of bytes written to ``output_ptr`` into the buffer at ``output_len_ptr``
+	
+	Unlike with other runtime API calls, the contracts pallet can not guarantee this behaviour.
+	Instead, it's specific to each chain extension. Hence, before using this builtin,
+	you must make sure that the chain extension being called is compatible.
```
seanyoung, line 334: How about replacing
`Hence, before using this builtin, you must make sure that the chain extension being called is compatible. `
with
`Hence, when using this builtin, ensure that the chain extension is used in the correct way.`

src/emit/substrate/target.rs
```diff
@@ -1580,14 +1580,65 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
+
+        assert_eq!(builtin_func.name, "chain_extension", "unimplemented");
+
+        let input_ptr = binary.vector_bytes(args[1].into_pointer_value().into());
+        let input_len = binary.vector_len(args[1].into_pointer_value().into());
+        let (output_ptr, output_len_ptr) = scratch_buf!();
+        let len = 16384;
+        binary.builder.build_store(output_len_ptr, i32_const!(len));
+        call!(
+            "__memset8",
```
seanyoung, line 1600: we have a `__bzero8` which should be slightly faster/less gas

src/emit/substrate/target.rs
```diff
@@ -1580,14 +1580,65 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
-        unimplemented!()
+    ) -> Option<BasicValueEnum<'a>> {
+        emit_context!(binary);
+
+        assert_eq!(builtin_func.name, "chain_extension", "unimplemented");
+
+        let input_ptr = binary.vector_bytes(args[1].into_pointer_value().into());
+        let input_len = binary.vector_len(args[1].into_pointer_value().into());
+        let (output_ptr, output_len_ptr) = scratch_buf!();
+        let len = 16384;
```
seanyoung, line 1597: Needs a comment

tests/substrate_tests/builtins.rs
```diff
@@ -755,3 +755,23 @@ fn hash() {
 
     runtime.function("test_encoding", vec![]);
 }
+
+#[test]
+fn call_chain_extension() {
+    let mut runtime = build_solidity(
+        r##"
+        import "substrate";
```
seanyoung, line 763: Suggestion:

You could do `import {chain_extension as ChainExtension} from "substrate";` and test if that that works.

seanyoung (approved):
Very nice!!

docs/language/builtins.rst
xermicus: This suggests that the chain extension function / ID of interest can be used in various ways, and the caller can do something about that, doesn't it? But this is not the case. Either the implementation of the chain extension adheres to this, or it doesn't. There is nothing you can do about it (unless your are also the author of the parachain runtime, but this is a bit of a stretch).

My main goal of this warning is to make contract authors to reconsider whether they know what they are doing. Calling into a chain extension that you're not 100% certain about what it does is a really, really bad idea anyways.

docs/language/builtins.rst
xermicus: You are right. I removed the entire sentence. It's already obvious that the input and output are `bytes` by looking at the function signature. No need to repeat in words what's already obvious from the code.

docs/language/builtins.rst
```diff
@@ -311,6 +311,35 @@ Its underlying type is ``bytes32``, but it will be reported correctly as the ``H
...
+The caller is responsible for encoding the input and decoding the output correctly.
+We expect parachain authors to write their own higher level libraries on top.
+
+.. warning::
+    This function calls the runtime API `call_chain_extension <https://docs.rs/pallet-contracts/latest/pallet_contracts/api_doc/trait.Version0.html#tymethod.call_chain_extension>`.
+	It assumes that the implementation of the chain extension does
+	- read the input from the ``input_ptr`` parameter, used as a buffer pointer
+	- write potential output into the buffer found at the ``output_ptr`` pointer
+	- respect the output buffer length in ``output_len_ptr`` to prevent OOB writes. The output buffer is 16KB in size.
+	- write the amount of bytes written to ``output_ptr`` into the buffer at ``output_len_ptr``
```
LucasSte, line 330:
```suggestion
	It assumes that the implementation of the chain extension
	- reads the input from the ``input_ptr`` parameter, used as a buffer pointer
	- writes potential output into the buffer found at the ``output_ptr`` pointer
	- respects the output buffer length in ``output_len_ptr`` to prevent OOB writes. The output buffer is 16KB in size.
	- writes the amount of bytes written to ``output_ptr`` into the buffer at ``output_len_ptr``
```

src/sema/diagnostics.rs
```diff
@@ -167,8 +167,6 @@ fn convert_diagnostic(
                 diagnostic::Label::secondary(file_id[&file_no], start..end)
                     .with_message(note.message.to_owned()),
             );
-        } else {
-            unreachable!("note without file position");
```
LucasSte, line 171: Why did you remove this?

xermicus: We can't panic just because a note doesn't point to a file location. Builtins don't have a file location, so we just ignore the note

seanyoung: If a note has a builtin file location the note should be rendered differently. I don't think we generate them.

Silently dropping them does not seem like a good idea.

xermicus: The `diagnostics::Label` structs needs a file location. Feels like this would require some refactoring best done in a dedicated PR?

I don't think this is a huge problem, as it only occurs for our builtins and it only drops the note (the primary diagnostic is still there).

I can revert this change if you want. It's not specific to this chain extension builtin. I discovered it by accident and thought, instead of crashing the compiler, it'd be nicer to just not display the secondary note.

seanyoung: I don't agree that silently dropping something is better than a panic. If there is a problem, we would like to know about it. The earlier we catch a problem the better.

xermicus: It makes for a bad user experience. Given this code:

```solidity
import "substrate";
contract Foo {
   function chain_extension() public pure {}
}
```

Would you rather have 

```
thread 'main' panicked at 'internal error: entered unreachable code: note without file position', src/sema/diagnostics.rs:171:13
```

Instead of this?

```
warning: chain_extension is already defined as a function
  ┌─ /home/glow/code/solang/test.sol:3:13
  │
3 │    function chain_extension() public pure {}
  │             ^^^^^^^^^^^^^^^
```

I mean we are already aware of the problem. I'll open an issue so we don't loose track.

LucasSte: > The `diagnostics::Label` structs needs a file location. Feels like this would require some refactoring best done in a dedicated PR?

A dedicated PR would be the best option, so it is better now to revert this change if it does not affect this PR in any way.

xermicus: See #1308

xermicus: I don't see how it is better to panic. This is note about the diagnostic itself. We are not silently swallowing an error. We are just omitting a note containing additional information.

xermicus: Removing the panic is valid fix.

A panic is a last resort; you have an error that you can't do anything about it. Hence you decide to crash your program.

This is just wrong here. If we can't print the note because it makes no sense to print it. Then we just should not print it.

But anyways, it's not important for this PR. I reverted this change.

LucasSte (approved):

PR #1306, opened 2023-05-10
Add padding in account deserialization

LucasSte:
According to Solana's C SDK, there is a padding between accounts: https://github.com/solana-labs/solana/blob/427ad7b5bde27fe71cff13a1ed0a6a3f26302ed4/sdk/bpf/c/inc/sol/deserialize.h#L50-L68


LucasSte:
> Needs a testcase

I've added two test cases

stdlib/solana_sdk.h
```diff
@@ -272,6 +274,10 @@ static uint64_t sol_deserialize(
         input = (uint8_t *)(((uint64_t)input + 8 - 1) & ~(8 - 1)); // padding
         input += sizeof(uint64_t);
       }
+      else
+      {
+        input += 7; // padding
```
xermicus, line 279: Could mention why there is the padding (and/or why it is 7bytes) in the comment too

xermicus (approved):

stdlib/solana_sdk.h
```diff
@@ -246,6 +246,8 @@ static uint64_t sol_deserialize(
   {
     return ERROR_INVALID_ARGUMENT;
   }
+
+  uint64_t array_size = SOL_ARRAY_SIZE(params->ka);
```
seanyoung, line 250: This is not a good name. How about `max_accounts` or something similar

seanyoung (approved):

PR #1307, opened 2023-05-11
Read compiler configurations from toml file.

salaheldinsoliman:
1- Refactor `solang.rs`
2- Create a Struct `Configurations`, that can be created from the command line or from a `.toml` file.
3- Create a new subcommand `solang new`, which creates a project directory that includes a solidity example `flipper.sol` and `Solang.toml`.


src/bin/configurations/mod.rs
```diff
@@ -0,0 +1,483 @@
...
+    file_resolver::FileResolver,
+    Target,
+};
+
+use std::{collections::HashSet, ffi::OsString, fs, path::PathBuf, process::exit};
+use toml::Table;
+mod unit_tests;
+
+// A macro for error handling.
+macro_rules! match_err {
```
xermicus, line 15: Is this macro necessary? I think instead of invoking this macro in most places you could just do `.map_err(ToString::to_string)?`, or, in case the error does not implement `ToString` but `Display`, just directly call `.map_err(|e| format!("{e}"))?`.

src/bin/configurations/mod.rs
```diff
@@ -0,0 +1,483 @@
...
+
+// A type that defines the package to be compiled: file names, contract names, import maps and import paths.
+type Package = (
+    Vec<OsString>,
+    HashSet<String>,
+    Vec<PathBuf>,
+    Vec<(String, PathBuf)>,
+);
+
+type CompilerOutput = (
```
xermicus, line 34: Should this (and the type def below) be a struct instead?

Otherwise, when you see `package.2`, without reading this definition and the comment (the comment above should rather be a doc-comment) you'll have no idea what the second field of tuple is.

src/bin/configurations/mod.rs
```diff
@@ -0,0 +1,483 @@
...
+    pub imports: FileResolver,
+    pub emit: Option<String>,
+    pub std_json_output: bool,
+    pub output_directory: Option<OsString>,
+    pub output_meta: Option<OsString>,
+}
+
+impl Configurations {
+    /// Parse compile configurations from a toml file.
+    pub fn config_from_toml(path: &PathBuf) -> Result<Self, String> {
```
xermicus, line 58: Why do we want to manually parse the TOML file instead of using serde and [the TOML crate](https://crates.io/crates/toml)?

xermicus (changes requested):
Gave it a quick first review. My main gripe is parsing the config file manually.

PR #1310, opened 2023-05-12
Use solc v0.8.20 testdata

seanyoung:
Ethereum Solidity 0.8.20 has been released. Update our tests to use their latest test files.

### DIFF
--- a/src/sema/expression/resolve_expression.rs
+++ b/src/sema/expression/resolve_expression.rs
@@ -439,6 +439,13 @@ pub fn expression(
                 } else {
                     get_int_length(&expr_type, loc, false, ns, diagnostics)?;
 
+                    if !expr_type.is_signed_int(ns) {
+                        diagnostics.push(Diagnostic::error(
+                            *loc,
+                            "negate not allowed on unsigned".to_string(),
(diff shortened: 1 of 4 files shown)


xermicus (approved):

src/sema/expression/resolve_expression.rs
```diff
@@ -454,6 +462,11 @@ pub fn expression(
 
             get_int_length(&expr_type, loc, false, ns, diagnostics)?;
 
+            diagnostics.push(Diagnostic::error(
+                *loc,
+                "unary plus not permitted".to_string(),
+            ));
+
             Ok(expr)
```
LucasSte, line 470: Why are you still returning `Ok` if there has been an error?

seanyoung: There is no reason to stop resolving the expression/function. We want to resolve as much as possible.

LucasSte: > There is no reason to stop resolving the expression/function. We want to resolve as much as possible.

You returned an error in the previous case.

LucasSte (approved):

PR #1311, opened 2023-05-15
Remove seal prefixes

xermicus:
Since a while ago, the seal runtime API function imports do no longer need the `seal_` prefix. Since in Wasm the import name is found in the contract, it saves a few bytes in contract size. Unprefixed `return`  does not exist, and i left `seal_call` prefixed because it'll overwrite the `call` export.

### DIFF
--- a/src/emit/substrate/mod.rs
+++ b/src/emit/substrate/mod.rs
@@ -174,7 +174,7 @@ impl SubstrateTarget {
             "seal_gas_price",
             "seal_gas_left",
             "seal_caller",
-            "seal_terminate",
+            "terminate",
             "seal_deposit_event",
             "seal_transfer",
         ]);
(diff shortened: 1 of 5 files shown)


seanyoung:
> Since a while ago, the seal runtime API function imports do no longer need the `seal_` prefix. Since in Wasm the import name is found in the contract, it saves a few bytes in contract size. Unprefixed `return` does not exist, and i left `seal_call` prefixed because it'll overwrite the `call` export.

The problem with `call` is that the function is defined in the same llvm module as the contract llvm modules. It should be defined in a different module and then linked in via external functions.

Ideally this is done correctly in emit or it needs a comment explaining why `seal_call` still has a prefix.

xermicus:
Why not, I'll give it a look

xermicus:
So this would require our `Binary` struct to carry at least one other `Module`, where the externals are defined. But this seems to be quite a change for such a tiny optimization? I'd rather do that in another PR.

seanyoung:
> So this would require our `Binary` struct to carry at least one other `Module`, where the externals are defined. But this seems to be quite a change for such a tiny optimization? I'd rather do that in another PR.

Sure, I have never attempted to make the change so I don't know how much code it would be. I would say a comment explaining why `seal_call` is still used and not `call`.

xermicus:
Yeah I tried to just declare the externals on another module and link that in right from the start, but that's not how it works. I added a comment, thanks!

src/emit/substrate/mod.rs
```diff
@@ -300,6 +279,8 @@ impl SubstrateTarget {
             u8_ptr,
             u32_val
         );
+        // We still use prefixed seal_call because it would collide with the expored call function.
```
seanyoung, line 282: exported

PR #1312, opened 2023-05-15
Restrict external function callable scope

LucasSte:
On Solc, external functions can exclusively be called outside a contract. In order for use to accomplish #1251, we need to restrict their scope to can declare and organize the accounts the function needs.

### DIFF
--- a/docs/language/functions.rst
+++ b/docs/language/functions.rst
@@ -19,13 +19,24 @@ which are provided in the return statement, the values of the return variables a
 of the function is returned. It is still possible to explicitly return some values
 with a return statement.
 
-Functions which are declared ``public`` will be present in the ABI and are callable
-externally. If a function is declared ``private`` then it is not callable externally,
-but it can be called from within the contract. If a function is defined outside a
-contract, then it cannot have a visibility specifier (e.g. ``public``).
-
 Any DocComment before a function will be include in the ABI. Currently only Substrate
 supports documentation in the ABI.
+
+Function visibility
+___________________
+
+Solidity functions have a visibility specifier that restricts the scope in which they can be called.
+Functions can be declared public, private, internal or external with the following definitions:
+
+    - ``public`` functions can be invoked inside and outside a contract (e.g. by an RPC). They are
+      present in the contract's ABI or IDL.
+    - ``private`` functions can only be called inside the contract they are declared.
+    - ``internal`` functions can only be called internally within the contract or by any contract
+      that inherits that one where the function is declared.
+    - ``external`` function can exclusively be invoked by other contracts or directly by an RPC. They
+      are also present in the contract's ABI or IDL.
+
+If a function is defined outside a contract,  it cannot have a visibility specifier (e.g. ``public``).
 
 Arguments passing and return values
 ___________________________________
--- a/docs/language/functions.rst
+++ b/docs/language/functions.rst
@@ -32,11 +32,17 @@ Functions can be declared public, private, internal or external with the followi
       present in the contract's ABI or IDL.
     - ``private`` functions can only be called inside the contract they are declared.
     - ``internal`` functions can only be called internally within the contract or by any contract
-      that inherits that one where the function is declared.
-    - ``external`` function can exclusively be invoked by other contracts or directly by an RPC. They
+      inherited contract.
+    - ``external`` functions can exclusively be invoked by other contracts or directly by an RPC. They
       are also present in the contract's ABI or IDL.
 
-If a function is defined outside a contract,  it cannot have a visibility specifier (e.g. ``public``).
+Both internal and external functions can be called using the syntax ``this.func()``. In this case, the
+arguments are ABI encoded for the call, as it is treated like an external call. This is the only way to
+call an external function from inside the same contract it is defined. This method, however, should be avoided
+for public functions, as it will be more costly to call them than simply using ``func()``.
+
+If a function is defined outside a contract, it cannot have a visibility specifier (e.g. ``public``).
+
 
 Arguments passing and return values
 ___________________________________
--- a/docs/language/functions.rst
+++ b/docs/language/functions.rst
@@ -28,15 +28,15 @@ ___________________
 Solidity functions have a visibility specifier that restricts the scope in which they can be called.
 Functions can be declared public, private, internal or external with the following definitions:
 
-    - ``public`` functions can be invoked inside and outside a contract (e.g. by an RPC). They are
+    - ``public`` functions can be called inside and outside a contract (e.g. by an RPC). They are
       present in the contract's ABI or IDL.
     - ``private`` functions can only be called inside the contract they are declared.
     - ``internal`` functions can only be called internally within the contract or by any contract
       inherited contract.
-    - ``external`` functions can exclusively be invoked by other contracts or directly by an RPC. They
+    - ``external`` functions can exclusively be called by other contracts or directly by an RPC. They
       are also present in the contract's ABI or IDL.
 
-Both internal and external functions can be called using the syntax ``this.func()``. In this case, the
+Both public and external functions can be called using the syntax ``this.func()``. In this case, the
 arguments are ABI encoded for the call, as it is treated like an external call. This is the only way to
 call an external function from inside the same contract it is defined. This method, however, should be avoided
 for public functions, as it will be more costly to call them than simply using ``func()``.
--- a/src/sema/expression/function_call.rs
+++ b/src/sema/expression/function_call.rs
@@ -2591,9 +2591,10 @@ fn resolve_internal_call(
         if is_base(base_no, derived_no, ns) && matches!(func.visibility, Visibility::External(_)) {
             errors.push(Diagnostic::error_with_note(
                 *loc,
-                "external functions can only be invoked outside the contract".to_string(),
+                "functions declared external cannot be called via an internal function call"
+                    .to_string(),
                 func.loc,
-                "function defined here".to_string(),
+                format!("declaration of {} '{}'", func.ty, func.name),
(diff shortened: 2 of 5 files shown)


seanyoung (changes requested):
Your change made fewer solc semantic tests pass.

seanyoung:
Sorry, just noticed this PR was still marked draft. Please ignore me.

LucasSte:
> Your change made fewer solc semantic tests pass.

There was a mistake in the error condition. After that I've fixed it, the number of failing tests decreased :)

docs/language/functions.rst
```diff
@@ -19,14 +19,25 @@ which are provided in the return statement, the values of the return variables a
...
+    - ``public`` functions can be invoked inside and outside a contract (e.g. by an RPC). They are
+      present in the contract's ABI or IDL.
+    - ``private`` functions can only be called inside the contract they are declared.
+    - ``internal`` functions can only be called internally within the contract or by any contract
+      that inherits that one where the function is declared.
+    - ``external`` function can exclusively be invoked by other contracts or directly by an RPC. They
+      are also present in the contract's ABI or IDL.
+
+If a function is defined outside a contract,  it cannot have a visibility specifier (e.g. ``public``).
+
```
xermicus, line 40: If we "internally" call a public function, it'll cause a round trip of ABI encoding and decoding the call data. For economic reasons, calling into public functions should be avoided. Could be worth mentioning that here.

src/sema/expression/function_call.rs
```diff
@@ -2620,3 +2546,81 @@ pub(crate) fn function_type(func: &Function, external: bool, resolve_to: Resolve
...
+
+    match arg.cast(&arg.loc(), arg_ty, true, ns, errors) {
+        Ok(expr) => cast_args.push(expr),
+        Err(_) => return false,
+    }
+
+    true
+}
+
+/// This function finishes resolving internal function calls
```
xermicus, line 2576: Could mention here what a `None` return value means

docs/language/functions.rst
```diff
@@ -19,14 +19,25 @@ which are provided in the return statement, the values of the return variables a
...
+___________________
+
+Solidity functions have a visibility specifier that restricts the scope in which they can be called.
+Functions can be declared public, private, internal or external with the following definitions:
+
+    - ``public`` functions can be invoked inside and outside a contract (e.g. by an RPC). They are
+      present in the contract's ABI or IDL.
+    - ``private`` functions can only be called inside the contract they are declared.
+    - ``internal`` functions can only be called internally within the contract or by any contract
+      that inherits that one where the function is declared.
```
seanyoung, line 35: ` by any contract that inherits that one where the function is declared. ` is a bit awkward. How about `by any inherited contract`

src/sema/expression/function_call.rs
```diff
@@ -2620,3 +2546,81 @@ pub(crate) fn function_type(func: &Function, external: bool, resolve_to: Resolve
...
+        ));
+
+        return None;
+    } else if func.contract_no == context.contract_no
+        && matches!(func.visibility, Visibility::External(_))
+    {
+        errors.push(Diagnostic::error(
+            *loc,
+            "external functions can only be invoked outside the contract".to_string(),
+        ));
```
seanyoung, line 2604: needs a note with reference to function definition

src/sema/expression/function_call.rs
```diff
@@ -2620,3 +2546,81 @@ pub(crate) fn function_type(func: &Function, external: bool, resolve_to: Resolve
...
+    if func.contract_no != context.contract_no && func.is_private() {
+        errors.push(Diagnostic::error_with_note(
+            *loc,
+            format!("cannot call private {}", func.ty),
+            func.loc,
+            format!("declaration of {} '{}'", func.ty, func.name),
+        ));
+
+        return None;
+    } else if func.contract_no == context.contract_no
```
seanyoung, line 2598: inherited `external` functions are callable?

seanyoung: Needs test

seanyoung (changes requested):

src/sema/expression/function_call.rs
```diff
@@ -2620,3 +2546,81 @@ pub(crate) fn function_type(func: &Function, external: bool, resolve_to: Resolve
...
+            return false;
+        }
+    };
+
+    match arg.cast(&arg.loc(), arg_ty, true, ns, errors) {
+        Ok(expr) => cast_args.push(expr),
+        Err(_) => return false,
+    }
+
+    true
```
xermicus, line 2573:
```suggestion
    expression(arg, context, ns, symtable, errors, ResolveTo::Type(arg_ty))
        .and_then(|arg| arg.cast(&arg.loc(), arg_ty, true, ns, errors))
        .map(|expr| cast_args.push(expr))
        .is_ok()
```

docs/language/functions.rst
seanyoung: > If we "internally" call a public function, it'll cause a round trip of ABI encoding and decoding the call data. For economic reasons, calling into public functions should be avoided. Could be worth mentioning that here.

That only happens if it is called via `this.func(1, 2, 3)`. 

It should also be noted that `external` functions can be called this way.

src/sema/expression/function_call.rs
LucasSte: > inherited `external` functions are callable?

Oh, I missed this case.

docs/language/functions.rst
```diff
@@ -19,14 +19,31 @@ which are provided in the return statement, the values of the return variables a
...
+
+    - ``public`` functions can be invoked inside and outside a contract (e.g. by an RPC). They are
+      present in the contract's ABI or IDL.
+    - ``private`` functions can only be called inside the contract they are declared.
+    - ``internal`` functions can only be called internally within the contract or by any contract
+      inherited contract.
+    - ``external`` functions can exclusively be invoked by other contracts or directly by an RPC. They
+      are also present in the contract's ABI or IDL.
+
+Both internal and external functions can be called using the syntax ``this.func()``. In this case, the
```
seanyoung, line 39: `internal` functions cannot be called via `this.func()`, only `public` and `external`

docs/language/functions.rst
```diff
@@ -19,14 +19,31 @@ which are provided in the return statement, the values of the return variables a
...
 Any DocComment before a function will be include in the ABI. Currently only Substrate
 supports documentation in the ABI.
 
+Function visibility
+___________________
+
+Solidity functions have a visibility specifier that restricts the scope in which they can be called.
+Functions can be declared public, private, internal or external with the following definitions:
+
+    - ``public`` functions can be invoked inside and outside a contract (e.g. by an RPC). They are
```
seanyoung, line 31: In the previous paragraph you use the term "call" but now you use "invoke". I think nomenclature should be consistent, so it is better to use "call" throughout.

src/sema/expression/function_call.rs
```diff
@@ -2620,3 +2546,75 @@ pub(crate) fn function_type(func: &Function, external: bool, resolve_to: Resolve
...
+        ));
+
+        return None;
+    } else if let (Some(base_no), Some(derived_no)) = (func.contract_no, context.contract_no) {
+        if is_base(base_no, derived_no, ns) && matches!(func.visibility, Visibility::External(_)) {
+            errors.push(Diagnostic::error_with_note(
+                *loc,
+                "external functions can only be invoked outside the contract".to_string(),
+                func.loc,
+                "function defined here".to_string(),
```
seanyoung, line 2596: "here" is not great, also it might not be a definition (e.g. interface). How about reusing the note from above:

`format!("declaration of {} '{}'", func.ty, func.name),`

seanyoung (changes requested):

src/sema/expression/function_call.rs
```diff
@@ -2620,3 +2546,75 @@ pub(crate) fn function_type(func: &Function, external: bool, resolve_to: Resolve
...
+            func.loc,
+            format!("declaration of {} '{}'", func.ty, func.name),
+        ));
+
+        return None;
+    } else if let (Some(base_no), Some(derived_no)) = (func.contract_no, context.contract_no) {
+        if is_base(base_no, derived_no, ns) && matches!(func.visibility, Visibility::External(_)) {
+            errors.push(Diagnostic::error_with_note(
+                *loc,
+                "external functions can only be invoked outside the contract".to_string(),
```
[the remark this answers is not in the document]
LucasSte: Invoke and call are synonyms. I don't understand why you always complain about this word.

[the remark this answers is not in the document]
seanyoung: The solang documentation should be as clear and simple as possible. Solidity aims to be a simple language which javascript developers can understand. I know that invoke and call are synonymous, but a reader may not know this.

docs/language/functions.rst
```diff
@@ -19,14 +19,31 @@ which are provided in the return statement, the values of the return variables a
...
-but it can be called from within the contract. If a function is defined outside a
-contract, then it cannot have a visibility specifier (e.g. ``public``).
-
 Any DocComment before a function will be include in the ABI. Currently only Substrate
 supports documentation in the ABI.
 
+Function visibility
+___________________
+
+Solidity functions have a visibility specifier that restricts the scope in which they can be called.
```
xermicus, line 28: Very often we duplicate things that are already documented in the official Solidity docs. I'm aware that this is a whole other discussion, but I think we could at least link the official docs (https://docs.soliditylang.org/en/v0.8.20/contracts.html#function-visibility)

xermicus (approved):

seanyoung (approved):

PR #1313, opened 2023-05-16
fix: Remove Duplicate Code

mayank-pathakk:
This PR aims to remove the duplicate codes are mentioned in the issue #1261


mayank-pathakk:
@LucasSte @xermicus Hey! I wanted to work towards solving the several duplicate code issues.
Do you think this is the right way to proceed ?

PR #1314, opened 2023-05-16
fix: Remove Duplicate Code

mayank-pathakk:
This PR aims to remove the duplicate codes are mentioned in the issue #1261

### DIFF
--- a/src/sema/expression/assign.rs
+++ b/src/sema/expression/assign.rs
@@ -13,6 +13,20 @@ use solang_parser::diagnostics::Diagnostic;
 use solang_parser::pt;
 use solang_parser::pt::CodeLocation;
 
+/// Ensure immutable are only mutated in the constructor
+fn check_immutable(context: &ExprContext, ns: &mut Namespace, loc: &pt::Loc, immutable: &bool, diagnostics: &mut Diagnostics) -> () {
+    if *immutable {
+        if let Some(function_no) = context.function_no {
+            if !ns.functions[function_no].is_constructor() {
+                diagnostics.push(Diagnostic::error(
+                    *loc,
+                    "cannot assign to immutable outside of constructor".to_string(),
+                ));
+            }
+        }
+    }
+}
+
 /// Resolve an assignment
 pub(super) fn assign_single(
     loc: &pt::Loc,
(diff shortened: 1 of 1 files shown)


src/sema/expression/assign.rs
```diff
@@ -13,6 +13,22 @@ use solang_parser::diagnostics::Diagnostic;
...
+                diagnostics.push(Diagnostic::error(
+                    *loc,
+                    "cannot assign to immutable outside of constructor".to_string(),
+                ));
+                return Err(());
+            }
+        }
+    }
+    Ok(())
+}
```
seanyoung, line 30: Actually adding this diagnostic should not be fatal and cause resolving to fail. Please make this function return nothing.

src/sema/expression/assign.rs
```diff
@@ -13,6 +13,22 @@ use solang_parser::diagnostics::Diagnostic;
 use solang_parser::pt;
 use solang_parser::pt::CodeLocation;
 
+///Helper functions
```
seanyoung, line 16: That's not a useful comment. How about `// ensure immutable are only mutated in the constructor`

src/sema/expression/assign.rs
```diff
@@ -123,17 +139,7 @@ pub(super) fn assign_single(
...
-                        if !ns.functions[function_no].is_constructor() {
-                            diagnostics.push(Diagnostic::error(
-                                *loc,
-                                "cannot assign to immutable outside of constructor".to_string(),
-                            ));
-                            return Err(());
-                        }
-                    }
-                }
+                check_immutable(&context, ns, loc, &immutable, diagnostics)?;
```
seanyoung, line 142: .. and then remove the `?` from the call

seanyoung (changes requested):
Looks good, please make this change while refactoring the code

Thanks!

src/sema/expression/assign.rs
mayank-pathakk: Hi @seanyoung ! Thanks for the review :)
I have fixed the code, according to your suggestions.
If the rest of it is okay, I shall start working on the other refactorings.

xermicus:
This PR is front-running  #1274. However it looks like  #1274 got stale

xermicus:
The author of #1274 actually wants to complete his work. First comes first serves.

PR #1315, opened 2023-05-18
Virtual functions are available for calls

xermicus:
Virtual functions may not have a body, but they should still be callable.

Without this fix, the added testcase fails with this error which is wrong:

```
error: unknown function or type '_virtual'
  ┌─ /home/glow/code/solang/tests/contract_testcases/substrate/functions/virtual.sol:3:9
  │
3 │         _virtual();
  │         ^^^^^^^^
```


tests/contract_testcases/substrate/function_types/variable_or_func_type.sol
```diff
@@ -16,6 +16,4 @@ contract bar {
         }
         
 // ---- Expect: diagnostics ----
-// warning: 10:13-33: function can be declared 'view'
-// warning: 12:21-29: local variable 'variable' has been assigned, but never read
-// warning: 14:51-60: local variable 'func_type' has been assigned, but never read
+// error: 14:63-68: function 'total' is overloaded
```
seanyoung, line 19: This test is no longer working. The test is supposed to check that `total` can be resolved as a variable name, or as a function name (the accessor function).

Note that solc can't do this, this is a solang feature.

seanyoung (changes requested):
This change is wrong. You need to make sure you are calling a function with an body when doing an internal call. Of course you can call a virtual function without a body in abstract contract, but the final contract needs a body.

For example:
```solidity
interface I {
	function total() external returns (int);
}

contract A {
	function total() external returns (int) {
		return 4;
	}
}

contract B is I, A {
	function x() public {
		int total = total();
	}
}
```
Compile this:
```
$ cargo run compile --target substrate x.sol 
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/solang compile --target substrate x.sol`
warning: function can be declared 'pure'
  ┌─ /home/sean/git/solang/x.sol:6:2
  │
6 │     function total() external returns (int) {
  │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: local variable 'total' has been assigned, but never read
   ┌─ /home/sean/git/solang/x.sol:13:7
   │
13 │         int total = total();
   │             ^^^^^

thread 'main' panicked at 'no entry found for key', src/emit/instructions.rs:462:29
```
This is because the virtual definition in the interface is being called, not the function with a body.

xermicus:
Ok, thanks for pointing this out. I was unaware that we lack a `sema` pass that checks whether all called functions have an implementation. I'll add such a pass and add your example to the test cases.

tests/contract_testcases/substrate/function_types/variable_or_func_type.sol
xermicus: I was under the impression that the test is wrong. I was confused as to why solang does not detect the same error as `solc` (solc points out the invalid type, which I think is much better). 

> this is a solang feature

How is it used? Feels to me like we should rewrite this test

seanyoung:
> Ok, thanks for pointing this out. I was unaware that we lack a `sema` pass that checks whether all called functions have an implementation. I'll add such a pass and add your example to the test cases.

The old `has_body` check told it whether the function has an implementation. I *think* you can change codegen to find the correct function definition when you come across a `Expression::InternalFunctionCall { .. }` which calls a virtual function without a body (just a virtual function lookup).

xermicus:
@seanyoung I think the main issue was just that I didn't think about ` you can call a virtual function without a body in abstract contract` carefully enough, with emphasis on the _abstract contract_ part. I changed the code so it accounts for this and added your test case too. Now the change also fixes 4 evm parser tests :tada:

tests/contract_testcases/substrate/functions/virtual_function_call_01.sol
```diff
@@ -0,0 +1,9 @@
+abstract contract A {
+    constructor() {
+        _virtual();
+    }
+
+    function _virtual() internal virtual;
+}
+
```
seanyoung, line 8: This test should ensure that a concrete contract with A as a base actually works, i.e. adding:

```solidity
contract B is A {
	function _virtual() internal override {}

	function m() public {}
}
```
(It does, I've just tested it)

seanyoung (approved):

tests/contract_testcases/substrate/functions/virtual_function_call_01.sol
xermicus: Good idea,  thanks

tests/contract_testcases/substrate/functions/virtual_function_call_02.sol
```diff
@@ -0,0 +1,19 @@
...
+
+contract B is I, A {
+    function x() public {
+        int256 total = total();
+    }
+}
+
+// ---- Expect: diagnostics ----
+// error: 13:24-31: functions declared external cannot be called via an internal function call
+// 	note 6:5-47: declaration of function 'total'
```
LucasSte, line 19: What were you trying to test in this contract?
These errors do not look like they are related to your changes.

xermicus: It ensures functions without a body are[ only allowed on abstract contracts](https://github.com/hyperledger/solang/pull/1315#pullrequestreview-1433941958)

LucasSte: That makes sense.
Do you need the `external` keyword in those functions? It might be masking other errors in this file, but your goal is for it to have no errors.

xermicus: Interface functions must be declared `external`

seanyoung (approved):

PR #1316, opened 2023-05-19
Refactor Solang's CLI

salaheldinsoliman:
Refactor Solang's CLI to prepare to parse CLI arguments easily from a `toml` file.


xermicus (approved):
Love it!

src/bin/cli/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+                exit(1);
+            }
+        }
+    }
+
+    resolver
+}
+
+pub fn options_arg(debug: &DebugFeatures, optimizations: &Optimizations) -> Options {
+    let opt_level = match optimizations.opt_level.as_str() {
```
xermicus, line 245: Nit: Could probable implement `From` for `OptimizationLevel`

LucasSte:
How is this PR tested?

salaheldinsoliman:
> How is this PR tested?

Nice point @LucasSte , I will add further coverage of CLI's functionality to `tests/cli.rs`

xermicus:
> How is this PR tested?

Well it is tested the same way as it was before. By adding new tests, we don't actually check whether this PR does not break anything. What additional tests do you have in mind?

salaheldinsoliman:
> > How is this PR tested?
> 
> Well it is tested the same way as it was before. By adding new tests that weren't in `main` already, we don't actually check whether this PR does not break anything; what additional tests do you have in mind?

@xermicus One way to do that is comparing the `clap::Command` with the previous generated `Command`, but this would look ugly I guess.  I used this:  https://docs.rs/clap/latest/clap/struct.Command.html#method.debug_assert

xermicus:
The `debug_assert` is a good idea :+1: 

> comparing the clap::Command with the previous generated Command

I don't think this is time well spent. There are so many possible combinations. We use the binary in the codegen tests and in the integration tests, that should give as enough confidence about the most important parts of the CLI in my opinion.

salaheldinsoliman:
> The `debug_assert` is a good idea 👍
> 
> > comparing the clap::Command with the previous generated Command
> 
> I don't think this is time well spent. There are so many possible combinations. We use the binary in the codegen tests and in the integration tests, that should give as enough confidence about the most important parts of the CLI in my opinion.

I agree

LucasSte:
> > How is this PR tested?
> 
> Well it is tested the same way as it was before. By adding new tests that weren't in `main` already, we don't actually check whether this PR does not break anything; what additional tests do you have in mind?

We do not have proper CLI tests today in our repository. For the context of this PR, it might suffice to add a unit tests to check if a certain input command generates the correct `struct Cli` and its substructures.

xermicus:
> tests to check if a certain input command generates the correct struct Cli and its substructures.

We should trust `clap` to do that correctly. And at least the `compile` command (which is by far the most important one) is implicitly tested heavily in the codegen and integration tests. Per se I'm not against increasing test coverage, but without concrete suggestions for additional test cases, we can not move forward here.

LucasSte:
> > tests to check if a certain input command generates the correct struct Cli and its substructures.
> 
> We should trust `clap` to do that correctly. And at least the `compile` command (which is by far the most important one) is implicitly tested heavily in the codegen and integration tests. Per se I'm not against increasing test coverage, but without concrete suggestions for additional test cases, we can not move forward here.

I am sure clap works correctly. The reason behind the unit test is ensuring the correct behavior in our code (all commands are correctly mapped) and making sure mischanges do not go unnoticed. The codegen and integration test coverage for the CLI command is limited. They do not test combined options.

This is a suggestion for a complete PR. If you think there is no need to implement these tests, we can move forward.

seanyoung:
> I am sure clap works correctly. The reason behind the unit test is ensuring the correct behavior in our code (all commands are correctly mapped) and making sure mischanges do not go unnoticed. The codegen and integration test coverage for the CLI command is limited. They do not test combined options.

Do you have specific cases in mind? We have coverage for the `idl` and `language-server` subcommand already, and as you point out the integration/codegen tests already a lot of compile. There are import tests for the import cli options.

I am all for increasing coverage but we need to be specific here.

LucasSte:
> Do you have specific cases in mind? We have coverage for the `idl` and `language-server` subcommand already, and as you point out the integration/codegen tests already a lot of compile. There are import tests for the import cli options.
> 
> I am all for increasing coverage but we need to be specific here.

The untested options that I could find: enabling/disabling the optimizations passes (codegen tests do not cover all), using the `-O` option, setting a custom address length and a value length.

I might have missed something, so please ignore the cases if we already have tests for them.

salaheldinsoliman:
@LucasSte I added some test cases covering the poorly covered parts. If you found other parts that could use more coverage please list them. Thanks!

seanyoung (approved):

LucasSte (approved):

seanyoung:
@salaheldinsoliman thank you - very nice, this is much better!

PR #1317, opened 2023-05-19
`this` is not a keyword

seanyoung:
`this` should not be a keyword, and variables can be named `this` although this will hide the usual `this` which gives the current contract.


src/sema/expression/variable.rs
```diff
@@ -151,6 +151,21 @@ pub(super) fn variable(
...
+            Some(contract_no) => Ok(Expression::Builtin {
+                loc: id.loc,
+                tys: vec![Type::Contract(contract_no)],
+                kind: Builtin::GetAddress,
+                args: Vec::new(),
+            }),
+            None => {
+                diagnostics.push(Diagnostic::error(
+                    id.loc,
+                    "this not allowed outside contract".to_owned(),
```
LucasSte, line 164: This code path is not tested.

seanyoung: Good catch, I've added at test

LucasSte (approved):

PR #1323, opened 2023-05-23
Update to `ink!` v4.2 and bump MSRV to 1.68

xermicus:
Spared this for after the release. Upgrading to the latest `ink!` version requires a MSRV of `1.68.0`. In my opinion this is fine because `1.69.0` is already out anyways.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ jobs:
       with:
         submodules: recursive
     - name: Rust stable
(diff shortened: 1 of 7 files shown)


seanyoung:
Ideally we'd update the ci image as well, because now the new rust version has to be downloaded for each run.

seanyoung (approved):

xermicus:
Good point

LucasSte (approved):

PR #1325, opened 2023-05-24
Load readonly members in codegen

LucasSte:
Whenever I retrieve a `StructMember` of the `AccountInfo` struct, Solang loads the pointer in emit. As a consequence, if I need to use the pointer again in codegen, I must save values on the stack and get their stack address.

For the new Solana account management, we are generating the `AccountMeta` array in codegen after retrieving the account addresses from `AccountInfo`. In this case, it is more efficient to use the pointer directly.


xermicus (commented):
This PR touches changes a bunch of LOC, might be worthwhile to add a codegen test?

LucasSte:
> This PR touches changes a bunch of LOC, might be worthwhile to add a codegen test?

Solana integration and runtime tests already cover the majority of changes in this PR. I've added a codegen test for completeness.

seanyoung (approved):

xermicus (approved):
LGTM

PR #1328, opened 2023-05-25
Make GetAddress return a pointer to an address

LucasSte:
In the same lines of #1325, loading addresses in codegen instead of emit allows us to leverage pointers to create `AccountMeta` arrays.


src/emit/solana/target.rs
```diff
@@ -1707,11 +1708,7 @@ impl<'a> TargetRuntime<'a> for SolanaTarget {
                     "key_pointer",
                 );
 
-                binary.builder.build_load(
-                    binary.address_type(ns),
-                    key_pointer.into_pointer_value(),
-                    "key",
-                )
+                key_pointer.as_basic_value_enum()
```
seanyoung, line 1711: This is concerning. The scratch buffer is a global buffer which gets overwritten by lots of syscalls/code. We should have a comment explaining that this needs to be taken into account for codegen.

LucasSte: I'll add the comment. This is what already happens with `Expression::ReturnData`, isn't it?

LucasSte: @seanyoung The comment should go in the substrate file, right? You commented in the solana target.

seanyoung: yes, you're right. Also please comment the returndata one if it isn't already

xermicus (approved):

src/emit/solana/target.rs
LucasSte: Fixed!

seanyoung (approved):

PR #1329, opened 2023-05-25
Bump lalrpop and other dependencies

seanyoung:

PR #1335, opened 2023-05-27
Bump `lalrpop` to 0.20

DaniPopes:
<https://github.com/lalrpop/lalrpop/blob/master/RELEASES.md#0200-2023-05-02>

Test changes are due to "expected tokens in failed parses are more accurate".


seanyoung:
I had already made this change in an earlier pull request, see https://github.com/hyperledger/solang/pull/1329

PR #1336, opened 2023-05-29
Destructure can assign to storage variables

seanyoung:
Fix building of the following:

        contract c {
            address factory;
            int decimals;

            constructor() {
                (factory, decimals) = foo();
            }

            function foo() internal pure returns (address, int) {
                return (address(2), 5);
            }
        }

### DIFF
--- a/src/sema/unused_variable.rs
+++ b/src/sema/unused_variable.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::sema::ast::{Builtin, CallArgs, Diagnostic, EventDecl, Expression, Namespace};
+use crate::sema::ast::{
+    Builtin, CallArgs, Diagnostic, EventDecl, Expression, Namespace, RetrieveType,
+};
 use crate::sema::symtable::{Symtable, VariableUsage};
 use crate::sema::{ast, symtable};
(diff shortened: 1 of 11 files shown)


src/sema/unused_variable.rs
```diff
@@ -43,6 +49,49 @@ pub fn assigned_variable(ns: &mut Namespace, exp: &Expression, symtable: &mut Sy
...
+//  contract c {
+//      int[2] case1;
+//
+//      function f(int[2] storage case2) {
+//          case1[0] = 1;
+//          case2[0] = 1;
+//      }
+//  }
+//  The subscript for case1 is an assignment
+//  The subscript for case2 is a read
```
LucasSte, line 62: Isn't `case2` an assignment too?

seanyoung: An assignment to `case2` would look like:
```solidity
  case2 = ...;
```
Another way to think about this:

```solidity
contract test {
     function bar(int[] storage foo) internal view {
                foo[0] = 102;
  }
}
```
If you compile this with current `main`, you get a warning saying parameter `foo` is unused. It's clearly not

xermicus (approved):

LucasSte (approved):

PR #1337, opened 2023-05-29
@inheritdoc should work with bases of bases

seanyoung:
We were only checking the immediate base contracts of the current contract, not their bases.

### DIFF
--- a/src/sema/tags.rs
+++ b/src/sema/tags.rs
@@ -11,7 +11,7 @@ pub fn resolve_tags(
     tags: &[DocComment],
     params: Option<&[Parameter]>,
     returns: Option<&[Parameter]>,
-    bases: Option<&[String]>,
+    bases: Option<Vec<usize>>,
     ns: &mut Namespace,
 ) -> Vec<Tag> {
     let mut res: Vec<Tag> = Vec::new();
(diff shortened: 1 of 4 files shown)


src/sema/tags.rs
```diff
@@ -11,7 +11,7 @@ pub fn resolve_tags(
     tags: &[DocComment],
     params: Option<&[Parameter]>,
     returns: Option<&[Parameter]>,
-    bases: Option<&[String]>,
+    bases: Option<Vec<usize>>,
```
xermicus, line 14: Nit: This type argument can probably be flattened into just a `Vec` (I can't see a case where passing `Some` empty Vec makes sense here, and in fact the code does nowhere handle this case explicitly, making it superfluous)

xermicus (approved):

src/sema/tags.rs
seanyoung: `@inheritdoc` only makes sense for some items. For others, e.g. structs outside defined out of contracts, you want the diagnostic "@inheritdoc not allowed".

tests/contract_testcases/substrate/tags/inheritdoc.sol
```diff
@@ -0,0 +1,18 @@
...
+    /// @inheritdoc a
+    int v1;
+
+    /// @inheritdoc x
+    int v2;
+
+}
+
+// ---- Expect: diagnostics ----
+// error: 12:21-22: base contract 'x' not found in tag '@inheritdoc'
```
LucasSte, line 18: This read negative symbol in the file means we are not adding a new line at the end of it. GitHub complains about this all the time. Should we adopt the standard of adding a new line at the end of all files?

LucasSte (approved):

tests/contract_testcases/substrate/tags/inheritdoc.sol
xermicus: It's not really a warning. It's an indicator that there is no newline, and it's displayed because[ the diff would look the same](https://stackoverflow.com/a/5813359) regardless whether the file has a newline at the end or not

Issue #997, opened 2022-08-31
Solang should not ask functions that contain 'address.call' to be declared as view


LucasSte:
Functions that modify the state, like the SPL-token `mint_to`, are not `view` functions, but the compiler warns that they can be declared as view.

Example: 
```solidity
	function mint_to(address mint, address account, address authority, uint64 amount) internal {
		bytes instr = new bytes(9);

		instr[0] = uint8(TokenInstruction.MintTo);
		instr.writeUint64LE(amount, 1);

		AccountMeta[3] metas = [
			AccountMeta({pubkey: mint, is_writable: true, is_signer: false}),
			AccountMeta({pubkey: account, is_writable: true, is_signer: false}),
			AccountMeta({pubkey: authority, is_writable: true, is_signer: true})
		];

		tokenProgramId.call{accounts: metas}(instr);
	}
```

Warning:
```
warning: function can be declared 'view'
   ┌─ /solang/solana-library/spl_token.sol:46:2
   │
46 │     function mint_to(address mint, address account, address authority, uint64 amount) internal {
   │
```

PR #1338, opened 2023-05-29
functions that use .call() should be not allowed in view functions

seanyoung:
Functions that modify the state, like the SPL-token `mint_to`, are not `view` functions, but the compiler warns that they can be declared as view.

Example: 
```solidity
	function mint_to(address mint, address account, address authority, uint64 amount) internal {
		bytes instr = new bytes(9);

		instr[0] = uint8(TokenInstruction.MintTo);
		instr.writeUint64LE(amount, 1);

		AccountMeta[3] metas = [
			AccountMeta({pubkey: mint, is_writable: true, is_signer: false}),
			AccountMeta({pubkey: account, is_writable: true, is_signer: false}),
			AccountMeta({pubkey: authority, is_writable: true, is_signer: true})
		];

		tokenProgramId.call{accounts: metas}(instr);
	}
```

Warning:
```
warning: function can be declared 'view'
   ┌─ /solang/solana-library/spl_token.sol:46:2
   │
46 │     function mint_to(address mint, address account, address authority, uint64 amount) internal {
   │
```

Fixes https://github.com/hyperledger/solang/issues/997

Another PR will implement staticcall for Solana, which is allowed in view functions.

### DIFF
--- a/src/sema/mutability.rs
+++ b/src/sema/mutability.rs
@@ -1,19 +1,36 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use super::ast::{
-    Builtin, DestructureField, Diagnostic, Expression, Function, Mutability, Namespace, Statement,
-    Type,
+use super::{
+    ast::{
+        Builtin, CallTy, DestructureField, Diagnostic, Expression, Function, Mutability, Namespace,
+        RetrieveType, Statement, Type,
+    },
+    yul::ast::{YulExpression, YulStatement},
+    Recurse,
 };
-use crate::sema::ast::RetrieveType;
-use crate::sema::yul::ast::{YulExpression, YulStatement};
-use crate::sema::Recurse;
-use solang_parser::pt;
+use solang_parser::{helpers::CodeLocation, pt};
+
+#[derive(PartialEq, PartialOrd)]
+enum Access {
+    None,
+    Read,
+    Write,
+    Value,
+}
+
+impl Access {
+    fn increase_to(&mut self, other: Access) {
+        if *self < other {
+            *self = other;
+        }
+    }
+}
 
 /// check state mutability
 pub fn mutability(file_no: usize, ns: &mut Namespace) {
     if !ns.diagnostics.any_errors() {
         for func in &ns.functions {
-            if func.loc.try_file_no() != Some(file_no) {
+            if func.loc.try_file_no() != Some(file_no) || func.ty == pt::FunctionTy::Modifier {
                 continue;
             }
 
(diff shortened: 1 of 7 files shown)


seanyoung:
@xermicus I don't know why there was a special case for Substrate, it may have been wrong  or maybe there is a good reason. Do you know if `seal_call` is permitted in `pure` or `view` functions, i.e. non-mutable messages?

xermicus:
No I don't know why there was a special case. I'm not aware on why it should not be permitted in `pure` or `view` functions, as long as it is guaranteed that the callee is also `pure` or `view` respectively.

src/sema/mutability.rs
```diff
@@ -27,9 +27,15 @@ pub fn mutability(file_no: usize, ns: &mut Namespace) {
...
     diagnostics: Vec<Diagnostic>,
+    /// Does the function have an expression or statement that reads from state?
     does_read_state: bool,
+    /// Does the function have an expression or statement that writes to state?
     does_write_state: bool,
+    /// Does the function have an expression or statement that may write to state?
+    may_write_state: bool,
+    /// Is the function declared view, payer or default mutability
     can_read_state: bool,
+    /// Is the function declared with default or payer mutability
```
LucasSte, line 38: You used the question mark on the other questions, but not here. Please, adopt a single notation.

src/sema/mutability.rs
```diff
@@ -315,11 +326,10 @@ fn read_expression(expr: &Expression, state: &mut StateCheck) -> bool {
...
         },
         Expression::ExternalFunctionCallRaw { loc, .. } => {
-            if state.ns.target.is_substrate() {
-                state.write(loc)
-            } else {
-                state.read(loc)
-            }
+            // We don't know if the external call will write to state, so don't give diagnostic about
+            // function may be declared view
+            state.may_write_state = true;
```
LucasSte, line 331: You've only used `may_write_to_state` in a single place. 
Why didn't you just reuse the `does_write_to_state`? You could have renamed it to `may_write_to_state`, as we are doing our best effort in this analysis.

LucasSte (commented):
Does this PR solve #997 ?

Edit: I've just read the PR summary. Can you add the example from #997?

src/sema/mutability.rs
seanyoung: I've added question marks but I think it questionable whether this improves things.

src/sema/mutability.rs
xermicus: If we don't know, why are we not assume for "the worst" and flag it as a state writing function?

src/sema/mutability.rs
```diff
@@ -27,9 +27,15 @@ pub fn mutability(file_no: usize, ns: &mut Namespace) {
 /// While we recurse through the AST, maintain some state
 struct StateCheck<'a> {
     diagnostics: Vec<Diagnostic>,
+    /// Does the function have an expression or statement that reads from state?
```
xermicus, line 30: Sorry for this nitpick, but I think using an `enum Mutability { Write, Read, None }` instead a bunch of bools would make for more idiomatic code overall. The struct could then just have two fields of this enum, like `desired_mutability` and `actual_mutability`.

I mean its unrelated to this bugfix; I'll just leave it as suggestion if you feel like it

src/sema/mutability.rs
seanyoung: The problem with that is that you cannot mark functions as `view` which means you need pay for a transaction in order to call it. You don't want that in many cases. In fact, non-`view` functions cannot have return values Solana/Anchor so it would not work at all.

src/sema/mutability.rs
seanyoung: Great idea, I've re-written it with an enum.

LucasSte (approved):

xermicus (approved):
Looks a lot cleanier now, thanks!

Issue #1330, opened 2023-05-25
cargo build fails with solang dependency


BenTheKush:
**Describe the bug**
I have a fresh project with a `solang = "0.3.0"` as its only dependency:

```toml
[package]
name = "playground"
version = "0.1.0"
edition = "2021"

[dependencies]
solang = "0.3.0"
```

Running `cargo build` fails with error message:

```
error: No suitable version of LLVM was found system-wide or pointed
              to by LLVM_SYS_150_PREFIX.
       
              Consider using `llvmenv` to compile an appropriate copy of LLVM, and
              refer to the llvm-sys documentation for more information.
       
              llvm-sys: https://crates.io/crates/llvm-sys
              llvmenv: https://crates.io/crates/llvmenv
   --> /Users/benku/.cargo/registry/src/github.com-1ecc6299db9ec823/llvm-sys-150.1.0/src/lib.rs:489:1
    |
489 | / std::compile_error!(concat!(
490 | |     "No suitable version of LLVM was found system-wide or pointed
491 | |        to by LLVM_SYS_",
492 | |     env!("CARGO_PKG_VERSION_MAJOR"),
...   |
499 | |        llvmenv: https://crates.io/crates/llvmenv"
500 | | ));
    | |__^
```

It seems like I might need a special version of llvm on my path? I'm running into issues running the instructions on the [solang readthedocs](https://solang.readthedocs.io/en/latest/installing.html#llvm-from-source); in particular, running the supplied `cmake` command is failing, complaining `Cmake was unable to find a build program corresponding to "Ninja"`.

Any hints on how to proceed? It would be helpful to have clearer instructions on how to use the `solang` crate as a dependency.

BenTheKush:
An example of unclear documentation, [ReadTheDocs](https://solang.readthedocs.io/en/latest/installing.html#install-llvm) says:
> Step 1: Install the LLVM Libraries
> Solang needs a build of [LLVM with some extra patches](https://github.com/solana-labs/llvm-project/). These patches make it possible to generate code for Solana, and fixes concurrency issues in the lld linker.
>
> You can either download the pre-built libraries from [github](https://github.com/hyperledger/solang/releases/tag/v0.3.0) or [build your own from source](https://solang.readthedocs.io/en/latest/installing.html#llvm-from-source).

I think the release binary link should be changed to `https://github.com/hyperledger/solang-llvm/releases`?

xermicus:
[...]

PR #1342, opened 2023-05-30
Fix build without llvm feature

seanyoung:
Also ensure it is tested in CI.

Fixes https://github.com/hyperledger/solang/issues/1330.


.github/workflows/test.yml
```diff
@@ -37,6 +37,11 @@ jobs:
       - name: Run cargo doc
         if: always()
         run: cargo doc --workspace --bins
+      - name: Test build without llvm feature
+        if: always()
```
xermicus, line 41: Can this wait until the "normal" build passes? I feel like this is very rarely of immediate concern for PR authors, it'll most likely fail too if the build fails and it wouldn't help anyways if this succeeds but the "normal" build fails. 

```suggestion
        if: always()
        needs: linux-x86-64
```

seanyoung: I don't think you can make a step depend on another job, I think we need a whole new job that depends on linux-x86-64. wdyt?

xermicus: Right, the `needs` should go to the job but I couldn't suggest it there. But now that I think about it, shouldn't this step be inside `linux-x86-64` anyways?

seanyoung: I've changed it again. Now I use `cargo clippy --no-default-features --lib -- -D warnings`. This gives build errors, warnings, and clippies for building without the llvm feature. Also, hopefully you'll agree this fits into the lints job better.

xermicus: This is even better

xermicus (approved):

src/codegen/expression.rs
```diff
@@ -3586,6 +3586,7 @@ pub(super) fn assert_failure(
 }
 
 /// Generate the binary code for a contract
+#[cfg(feature = "llvm")]
 fn code(loc: &Loc, contract_no: usize, ns: &Namespace, opt: &Options) -> Expression {
```
LucasSte, line 3590: Is it possible to merge this function and the one below like this?
```rust
#[cfg(feature = "llvm")]
let code = contract.emit(ns, opt);

#[cfg(not(feature = "llvm"))]
let code = b"code placeholder".to_vec();

// Rest of the function
```

seanyoung: You'll get clippy warnings that `ns`, `opt`, `contract` are unused variables.

PR #1343, opened 2023-05-30
Substrate: Write assumed environment configuration into metadata

xermicus:
In `ink!`, certain types are configurable. The latest version of the metadata crate stores this information in the ABI file.

### DIFF
--- a/src/abi/substrate.rs
+++ b/src/abi/substrate.rs
@@ -454,7 +454,7 @@ pub fn gen_project(contract_no: usize, ns: &ast::Namespace) -> InkProject {
         .chain_extension(Default::default()) // Does not exist in Solidity
         .max_event_topics(4)
         .account_id(TypeSpec::new(
-            resolve_ast(&ast::Type::Address(false).into(), ns, &mut registry).into(),
+            resolve_ast(&ast::Type::Address(false), ns, &mut registry).into(),
             path!("AccountId"),
         ))
         .balance(TypeSpec::new(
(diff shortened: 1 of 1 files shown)


xermicus:
What would this test be good for? I trust the EnvironmentSpec builder to do its job correctly

seanyoung (approved):

LucasSte:
> What would this test be good for? I trust the EnvironmentSpec builder to do its job correctly

If one refactors the code and moves everything around, one might mistakenly remove this portion of your code. As refactoring PRs are usually large, we might miss this during review. We trust our tests to detect regressions. Because there is no test for your modification, its removal will go unnoticed.

The rule of thumb is that if you remove your additional lines of code and no test fail, you need a test!

seanyoung:
It's not a bad idea to have some code which does some asserts on the output metadata, both to ensure that the metadata looks like we expect now and that there are no regressions in future. `EnvironmentSpec` surely works but we're interested in that the metadata looks like we want, have we made all the right calls to `EnvironmentSpec` etc.

seanyoung:
Having said that @xermicus  is the maintainer for Substrate, so it is up entirely up to him

xermicus:
@LucasSte in general I agree with your sentiment. However, for this case here it would be incredibly hard for someone to "move everything aroung", additionally forgetting to write the Environment spec, and then us reviewing the code not catch this. 

This environment data is just for frontends. The way this should be tested is to have integration that ensures we output compatible metadata. But right now I'm not aware on any frontend validating this.

LucasSte (approved):

xermicus:
Thanks.

Yes we still miss coverage in some places. For example we don't have any metadata unit tests at all. So before splitting hairs on this rather unimportant (at least at the moment) detail I'd rather spend my time implementing tests for the more important parts of the metadata first. Or other more important parts of the compiler.

PR #1345, opened 2023-05-31
`@payer` annotation should declare an account

LucasSte:
We aim to declare accounts for a function using only annotations. This PR is the first step towards this. The `@payer` annotation in the constructor now declares an account that is going to appear in the IDL. When a constructor is called from an external function, we automatically generate the `AccountMeta` array for the call so developers do not have to initialize such an array themselves. In addition, as we always need the account key in the transaction, `@payer` does not support addresses literals anymore.

This PR lays ground for the `@reader`, `@mutable` and `@mutableReader` annotations we wish to introduce later #1251 .

This PR already touches 45 files, so these changes are going to be in another PR:
1. When an annotation refers to a constructor argument, it should precede the argument declaration (just like Java). Right now, `@payer(account)` declares something while `@seed(arg1)` points to an argument. Such a syntax will lead to confusion.
2. We can't yet access the payer `AccountInfo` with the syntax `tx.accounts.payer`. This new command needs further changes in sema, codegen and emit.
3. The C function `external_call` will become useless as soon as I implement the account management for external calls, so it will be removed in the future.


src/codegen/mod.rs
```diff
@@ -159,7 +160,14 @@ pub fn codegen(ns: &mut Namespace, opt: &Options) {
     if ns.target == Target::Solana {
         for contract_no in 0..ns.contracts.len() {
             if ns.contracts[contract_no].instantiable {
-                collect_accounts_from_contract(contract_no, ns);
+                let diag = collect_accounts_from_contract(contract_no, ns);
+                ns.diagnostics.extend(diag);
+            }
+        }
+
+        for contract_no in 0..ns.contracts.len() {
```
xermicus, line 168: Nitpick using iterators:
```rust
for (n,_) in ns.contracts.iter().enumerate().filter(|(_, c)| c.instantiable) {}
```

src/sema/solana_accounts.rs
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use phf::phf_set;
+
+/// These are the accounts that we can collect from a contract and that Anchor will populate
+/// automatically if their names match the source code description:
+/// https://github.com/coral-xyz/anchor/blob/06c42327d4241e5f79c35bc5588ec0a6ad2fedeb/ts/packages/anchor/src/program/accounts-resolver.ts#L54-L60
+pub static CLOCK_ACCOUNT: &str = "clock";
```
xermicus, line 8: Why are those needed if they all are in the HashSet below?

src/codegen/solana_deploy.rs
```diff
@@ -362,7 +345,7 @@ pub(super) fn solana_deploy(
                 Expression::NumberLiteral {
                     loc: Loc::Codegen,
                     ty: Type::Uint(64),
-                    value: space_runtime_constant.into(),
+                    value: (space_runtime_constant).into(),
```
xermicus, line 348: Are these needed?

```suggestion
                    value: space_runtime_constant.into(),
```

src/codegen/solana_accounts/account_management.rs
```diff
@@ -0,0 +1,186 @@
...
+    } = instr
+    {
+        if accounts.is_some() || constructor_no.is_none() {
+            return;
+        }
+
+        let mut account_metas: Vec<Expression> = Vec::new();
+        let constructor_func = &functions[constructor_no.unwrap()];
+        for (name, account) in constructor_func.solana_accounts.borrow().iter() {
+            if name == DATA_ACCOUNT {
```
xermicus, line 73: These accounts might as well be an enum, and then you could implement this logic here (as well as to and from string implementations) on that enum

src/codegen/solana_accounts/account_management.rs
```diff
@@ -0,0 +1,186 @@
...
+use num_bigint::BigInt;
+use num_traits::Zero;
+use solang_parser::pt::Loc;
+use std::collections::{HashSet, VecDeque};
+
+/// This function walks over the CFG and automates the account management, so developers do not need
+/// to do so. For instance, when calling 'new construct{address: addr}()', we construct the correct
+/// AccountMeta array with all the accounts the constructor needs.
+pub(crate) fn manage_contract_accounts(contract_no: usize, ns: &mut Namespace) {
+    let contract_functions = ns.contracts[contract_no].functions.clone();
```
xermicus, line 16: Why is this clone needed? Can't this be an iterator over `ns.contracts[contract_no].function`?

xermicus (commented):
I'll be wrapping up my workday, here's what I found so far

src/sema/solana_accounts.rs
LucasSte: I think this section could have a better design, but I could think of any simple way to do this. I need these strings in many parts of the code to add the proper accounts to the IDL. Rewriting the same string repeatedly is prone to errors.

At the same time, these are reserved keywords for account names, so developers cannot declare them as account names.

I needed a way to get the string and check if a string is one reserved word.

src/codegen/solana_accounts/account_management.rs
LucasSte: I mutably borrow `ns.functions[func_no].cfg` in the loop. I cannot borrow `ns` twice.

docs/targets/solana.rst
```diff
@@ -446,3 +445,26 @@ The usage of system instructions needs the correct setting of writable and signe
...
+Solana Account Management
+_________________________
+
+In a contract constructor, one can optionally write the ``@payer`` annotation, which receives a character sequence as
+an argument. This annotation defines a Solana account that is going to pay for the initialization of the contract's data
+account. The syntax ``@payer(my_account)`` declares an account named ``my_account``, which will be
+required for every call to the constructor.
+
+In any Solana cross program invocation, including constructor calls, all the accounts a transaction needs must be
+informed. Whenever possible, the compiler will try to automatically generate the ``AccountMeta`` array that satisfy
```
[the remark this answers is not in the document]
LucasSte: The verb is incorrectly conjugated in my text. It should have been `satisfies`.

docs/examples/solana/payer_annotation.sol
```diff
@@ -0,0 +1,43 @@
+import 'solana';
+
+@program_id("SoLDxXQ9GMoa15i4NavZc61XGkas2aom4aNiWT6KUER")
```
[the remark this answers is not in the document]
LucasSte: This is an example for the docs. I have explained it in `docs/targets/solana.rst`. Should I write more words here too?

src/codegen/mod.rs
LucasSte: This is not possible, because the iterator borrows `ns`, while the function call in line 170 borrows `ns` as mutable.

src/sema/expression/function_call.rs
```diff
@@ -2179,6 +2179,21 @@ pub(super) fn parse_call_args(
...
+                ns.functions[context.function_no.unwrap()].visibility,
+                Visibility::External(_)
+            )
+            && !ns.functions[context.function_no.unwrap()].is_constructor()
+        {
+            std::println!("{}", ns.functions[context.function_no.unwrap()].visibility);
+            diagnostics.push(Diagnostic::error(
+                *loc,
+                "A constructor call in a non-external function needs the AccountMeta array. \
+                The compiler can only automatically handle accounts in functions declared as external"
```
seanyoung, line 2193: I don't think this is clear. It is written from the perspective of the compiler writer, not the Solidity developer

Accounts are required for calling a contract constructor. You can either provide the accounts with the {accounts: [ ...]} call arguments or change the visibility of this function to external.

seanyoung (approved):
Looks great apart from some minor issues.

docs/examples/solana/payer_annotation.sol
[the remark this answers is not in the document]
xermicus: I don't have a hard opinion on this. Usually it's just easier to just add some comments directly above the relevant LOC instead of trying to explain everything in detail in the doc paragraph. Additionally if someone is browsing our example he'll see the comments and better understand what's going on.

src/codegen/mod.rs
xermicus: So `iter_mut` should work?

src/sema/solana_accounts.rs
xermicus: Sounds like these should be enums variants then? You can implement (or even better derive)  from and to static methods

src/codegen/solana_accounts/account_management.rs
xermicus: Ok I see. No need to bikeshed this

src/codegen/mod.rs
LucasSte: It doesn't work. I cannot borrow `ns` both in the for loop header and in its body.

tests/solana_tests/runtime_errors.rs
```diff
@@ -200,7 +188,7 @@ contract calle_contract {
 
     assert_eq!(
         vm.logs,
-        "runtime_error: read integer out of bounds in test.sol:86:18-30,\n"
```
xermicus, line 203: Maybe it's just better to remove the line numbers and exact positions (I think we just want to ensure this error is present, there's no way it stems from a different place indicating a bug in the compiler)

tests/solana.rs
```diff
@@ -999,14 +1000,15 @@ impl Pubkey {
     }
 }
 
+// TODO: This struct is wrong
```
xermicus, line 1003: The comment could state what's wrong about it too

xermicus (approved):

tests/solana.rs
LucasSte: The struct is correct now. The comment was just to remind me to fix it later. I should have removed it.

PR #1347, opened 2023-06-02
Read compiler configurations from toml

salaheldinsoliman:
1- Implement taking solang's configurations from toml file. `solang compile --configuration-file`.
2- Implement solang new. a command that creates a new directory with a flipper example as well as the corresponding toml file. `solang new --target substrate`.

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -149,7 +149,7 @@ Options:
   1. Command line
   2. Configuration file
   3. Default values.
-  The default name for the toml file is "solang.toml". If two configuration files exist in the same directory, priority will be givin to the one passed explicitly to this argument.
+  The default name for the toml file is "solang.toml". If two configuration files exist in the same directory, priority will be given to the one passed explicitly to this argument.
   
 \-\-wasm-opt
    wasm-opt passes for Wasm targets (0, 1, 2, 3, 4, s or z; see the wasm-opt help for more details).
@@ -169,7 +169,7 @@ ______________________________
 
   solang new \-\-target solana my_project
 
-A solang project is a directory in which there is one or more solidity files, next to a ``solang.toml`` file where 
+A solang project is a directory in which there are one or more solidity files and a ``solang.toml`` file where 
 the compilation options are defined. Given these two components, a user can run ``solang compile`` in a similar fashion as ``cargo build``.
 
 The ``solang new`` command creates a new solang project with an example `flipper <https://github.com/hyperledger/solang/blob/main/examples/solana/flipper.sol>`_ contract,


xermicus (commented):
Nice! Gave it a first review round. Looks quite good already

src/bin/solang.rs
```diff
@@ -36,23 +36,86 @@ fn main() {
 
     match cli.command {
         Commands::Doc(doc_args) => doc(doc_args),
-        Commands::Compile(compile_args) => compile(&compile_args),
+        Commands::Compile(compile_args) => {
+            let config = if let Some(conf_file) = &compile_args.configuration_file {
+                eprintln!("info: reading config from toml file");
+                read_toml_config(conf_file)
+            } else {
```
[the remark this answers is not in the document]
salaheldinsoliman: This is the crate you meant I think: https://docs.rs/clap-serde/latest/clap_serde/
So this takes a toml specifying how the command should look like, not the arguments at runtime. (I was wrong in stating that this is what we need).

On the other hand though, there must be a way to check which args were explicitly present at run time. If so this should be fairly easy to do

src/bin/cli/mod.rs
```diff
@@ -274,3 +340,46 @@ fn parse_import_map(map: &str) -> Result<(String, PathBuf), String> {
...
+                _ => {Err(serde::de::Error::custom("Invalid option for `emit`. Valid options are: `ast-dot`, `cfg`, `llvm-ir`, `llvm-bc`, `object`, `asm`"))}
+            }
+        }
+        None => Ok(None),
+    }
+}
+
+fn default_true() -> bool {
+    true
+}
```
LucasSte, line 385: As per this PR https://github.com/vityafx/serde-aux/pull/23, you don't need this function. You could just write `default_true`.

src/bin/cli/test.rs
```diff
@@ -40,4 +45,103 @@ mod tests {
...
+            [
+                ("map1".to_owned(), PathBuf::from("path")),
+                ("map2".to_owned(), PathBuf::from("path2"))
+            ]
+        );
+
+        package_toml = r#"
+        contracts = ["flipper"] # Contracts to include from the compiled files
+        import_path = ["path1", "path2"]   
+        import_map = ["map1=path", "map2=path2"]    # Maps to import. Define as : import_paths = ["map=path/to/map", "map2=path/to/map2", ..]"#;
```
LucasSte, line 76: Other commands like target, optimizations and emit could also be tested.

LucasSte (changes requested):

src/bin/cli/mod.rs
```diff
@@ -35,6 +36,18 @@ pub enum Commands {
 
     #[command(about = "Generate Solidity interface files from Anchor IDL files")]
     Idl(IdlCommand),
+
+    #[command(about = "Create a new project with a ready example")]
+    New(New),
+}
+
+#[derive(Args)]
+pub struct New {
```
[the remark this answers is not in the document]
xermicus: I disagree, I don't see how this could not be very obvious that it's the struct for the `new` subcommand. Additionally it's confusing if the name diverges from the subcommand name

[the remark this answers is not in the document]
LucasSte: It might make sense here, but `fn new(args: New)` in line 70 from `src/bin/solang.rs` deserves either another name or a doc comment. I suggested a new name here because of this awkward construct.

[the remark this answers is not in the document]
xermicus: Hmm yeah that's a good point, the `new` function should definitively be renamed.

src/bin/cli/mod.rs
salaheldinsoliman: for some reason, `bool_true` doesn't work

src/bin/solang_new_examples/solana/flipper.sol
```diff
@@ -0,0 +1,22 @@
+@program_id("F1ipperKF9EfD821ZbbYjS319LXYiBmjhzkkf5a26rC")
+contract flipper {
+	bool private value;
+
+	/// Constructor that initializes the `bool` value to the given `init_value`.
```
xermicus, line 5: In Solidity, triple slashes are used with [NatSpec](https://docs.soliditylang.org/en/latest/style-guide.html#natspec). If we use them here we should add a couple of [NatSpec Tags](https://docs.soliditylang.org/en/latest/natspec-format.html#tags) too (alternatively just use normal comments)

src/bin/solang_new_examples/substrate/flipper.sol
```diff
@@ -0,0 +1,20 @@
+contract flipper {
+	bool private value;
+
+	/// Constructor that initializes the `bool` value to the given `init_value`.
```
xermicus, line 4: Same as in the Solana example

xermicus (commented):
This is great work, thanks! :heart:

Can you please add a short section in the docs about this?

What readers might be curios about is what the default values look like. E.g. if you have an empty TOML or empty sections in the TOML.

Also my gut feeling is that many users want to depend on some openzepplin contracts, would be nice if the docs would show how to use importmaps in the TOML file.

src/bin/cli/mod.rs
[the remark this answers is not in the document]
seanyoung: Should this be called `init`, just like `git init`?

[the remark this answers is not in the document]
seanyoung: cargo allows `cargo init` and `cargo new`. `git new` does not exist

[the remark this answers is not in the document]
salaheldinsoliman: But there is `cargo contract new`

src/bin/cli/mod.rs
```diff
@@ -127,10 +261,22 @@ pub struct TargetArg {
     pub value_length: Option<u64>,
 }
 
+#[derive(Args, Deserialize, Debug, PartialEq)]
+pub struct CompileTargetArg {
+    #[arg(name = "TARGET", required_unless_present = "CONFFILE" ,long = "target", value_parser = ["solana", "substrate", "evm"], help = "Target to build for [possible values: solana, substrate]", num_args = 1, hide_possible_values = true)]
```
xermicus, line 266: Is there a reason we need `required_unless_present = "CONFFILE"` here?

src/bin/cli/mod.rs
```diff
@@ -142,63 +288,147 @@ pub struct Package {
     pub import_map: Option<Vec<(String, PathBuf)>>,
 }
 
-#[derive(Args)]
+#[derive(Args, Deserialize, Debug, PartialEq)]
+pub struct CompilePackage {
+    #[arg(name = "INPUT", help = "Solidity input files",value_parser = ValueParser::path_buf(), num_args = 1.., required_unless_present = "CONFFILE")]
```
xermicus, line 293: Same as my comment above. Then, given a `solang.toml` file is present, we can just do `solang compile` (akin to how you can just `cargo build` in Rust projects). Which I'd find really neat

xermicus (changes requested):

src/bin/cli/mod.rs
salaheldinsoliman: Good pointer, Thanks

docs/running.rst
```diff
@@ -142,6 +160,18 @@ Options:
...
 
+
+Starting a new project
+______________________________
+
+
+  solang new --target solana my_project
+
+Start a new Solang project with an example `flipper.sol <https://github.com/hyperledger/solang/blob/main/examples/solana/flipper.sol>`_ contract,
+next to an example configuration file.
```
xermicus, line 171:
```suggestion
containing a default configuration file.
```

docs/running.rst
```diff
@@ -142,6 +160,18 @@ Options:
...
+
+
+
```
xermicus, line 174: This Paragraph needs to explain the user in more detail what exactly a "Solang project" is (that it creates a new directory with a default config file is just the product of this command, but we also need to explain the purpose of these Projects)

examples/substrate/substrate_config.toml
```diff
@@ -0,0 +1,32 @@
...
+generate-debug-info = false  # Add debug info to the generated llvm IR.
+
+[optimizations]
+dead-storage = true
+constant-folding = true
+strength-reduce = true
+vector-to-slice = true
+common-subexpression-elimination = true
+llvm-IR-optimization-level = "default"  # Set llvm optimizer level. Valid options are "none", "less", "default", "aggressive"
+
```
xermicus, line 26: FIY I just merged #1365, could you include that here too?

src/bin/solang.rs
```diff
@@ -32,27 +33,100 @@ mod idl;
...
+
+    let dir_path = args.project_name.unwrap_or(default_path);
+
+    match create_dir(&dir_path) {
+        Ok(_) => (),
+        Err(error) => {
+            eprintln!("couldn't create project directory, reason: {error}");
+            exit(1)
+        }
+    };
```
xermicus, line 95:
```suggestion
    if let Err(error) = create_dir(&dir_path) {
        eprintln!("couldn't create project directory, reason: {error}");
        exit(1);
    }
```

src/bin/cli/mod.rs
```diff
@@ -218,10 +449,50 @@ pub(crate) fn target_arg(target_arg: &TargetArg) -> Target {
...
-pub fn imports_arg(package: &Package) -> FileResolver {
+pub trait PackageTrait {
+    fn get_input(&self) -> &Vec<PathBuf>;
+    fn get_import_path(&self) -> &Option<Vec<PathBuf>>;
+    fn get_import_map(&self) -> &Option<Vec<(String, PathBuf)>>;
+}
+
+impl PackageTrait for CompilePackage {
+    fn get_input(&self) -> &Vec<PathBuf> {
+        match &self.input {
```
xermicus, line 460: Could use `map_err` instead of the match here

src/bin/cli/mod.rs
```diff
@@ -218,10 +449,50 @@ pub(crate) fn target_arg(target_arg: &TargetArg) -> Target {
     target
 }
 
-pub fn imports_arg(package: &Package) -> FileResolver {
+pub trait PackageTrait {
```
xermicus, line 452: This needs a comment explaining the trait and what it's used for

src/bin/solang.rs
```diff
@@ -32,27 +33,100 @@ mod idl;
...
+
+    let flipper = match target {
+        "solana" => include_str!("./solang_new_examples/solana/flipper.sol"),
+        "substrate" => include_str!("./solang_new_examples/substrate/flipper.sol"),
+        _ => unreachable!(),
```
xermicus, line 100: The cli arg parser allows `--target evm` resulting in a panic here

docs/running.rst
```diff
@@ -131,6 +131,24 @@ Options:
...
+  .. code-block:: toml
+
+    [package]
+    input_files = ["flipper.sol"]  # Solidity files to compile
+
+    [target]
+    name = "solana"  # Target name
+
+
+  Fields that are not explicitly present in the .toml file are defaulted. If any other argument is provided in the command line, for example, ``solang compile --config-file --target substrate``, the argument will be overridden.
```
LucasSte, line 146: Do I need to specify the `.toml` file I am referring to when I run `solang compile --config-file`? What happens when I have more than one `.toml` file in my repository?

If you are creating a `.toml` with a reserved name, that should be noted in the documentation.

src/bin/solang_new_examples/solana/flipper.sol
```diff
@@ -0,0 +1,22 @@
...
+	@payer(payer)
+	constructor(address payer, bool initvalue) {
```
LucasSte, line 7:
```suggestion
	constructor(bool initvalue) {
```
There is not need to pass the payer as an argument.

src/bin/cli/mod.rs
salaheldinsoliman: I think `.expect()` would look better

xermicus: Panicking is not the same as `std::process::exit(1)`. While technically the result is almost the same (I'm not sure whether we care about the different exit code). An unwrap or expect implies that this is a situation we do not anticipate for, and if it should ever happen regardless there's nothing to do about it. Here we can terminate in a more graceful way.

docs/running.rst
salaheldinsoliman: The default one is `solang.toml` , but if a special name is used, it will be taken instead.

src/bin/cli/mod.rs
salaheldinsoliman: but a map_err would imply converting the Option<> to a Result<> beforehand. I think an `if let Some()` would suffice, what do you think?

xermicus: Right, I didn't pay attention close enough, sorry. Yes you could return inside the `if let` and exit afterwards. Or leave it like it is

examples/substrate/substrate_config.toml
```diff
@@ -0,0 +1,32 @@
...
+log-runtime-errors = true   # Log runtime errors to the environment.
+generate-debug-info = false  # Add debug info to the generated llvm IR.
+
+[optimizations]
+dead-storage = true
+constant-folding = true
+strength-reduce = true
+vector-to-slice = true
+common-subexpression-elimination = true
+llvm-IR-optimization-level = "default"  # Set llvm optimizer level. Valid options are "none", "less", "default", "aggressive"
```
seanyoung, line 25: This does not apply to llvm-IR only, also to target specific machine code. This should be called `llvm-optimization-level`

src/bin/solang_new_examples/solana/flipper.sol
seanyoung: If not NatSpec tags are used, it defaults to `@notice`, which works fine for this.

seanyoung: BTW @salaheldinsoliman just copied the examples which I think are fine to begin with. If we want to add natspec, then it should be done everywhere in the repo (I don't think it is necessary)

seanyoung: FWIW I think NatSpec is pointless garbage. Rust does not have it and it is better for it.

src/bin/solang_new_examples/substrate/flipper.sol
seanyoung: Same as above

seanyoung (approved):

xermicus (approved):
LGTM

LucasSte (approved):

PR #1350, opened 2023-06-02
Do not allow push and pop in fixed length arrays

LucasSte:
We presently allow pushed and pops from fixed length arrays, leading to a panic in emit. This PR disallows such operations.

### DIFF
--- a/tests/contract_testcases/solana/functions/push_pop_fixed_length_array.sol
+++ b/tests/contract_testcases/solana/functions/push_pop_fixed_length_array.sol
@@ -16,5 +16,5 @@ contract MyTest {
 
 
 // ---- Expect: diagnostics ----
-// error: 5:11-15: method 'push' does not exist
-// error: 12:11-14: method 'pop' does not exist
+// error: 5:11-15: method push() is not available for fixed length arrays
+// error: 12:11-14: method pop() is not available for fixed length arrays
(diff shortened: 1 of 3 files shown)


tests/contract_testcases/solana/functions/push_pop_fixed_length_array.sol
```diff
@@ -0,0 +1,20 @@
...
+        uint64[10] a; 
+        a[9]  =  0x41; 
+        a.pop(); 
+        return  (a[9]);
+    }
+}
+
+
+// ---- Expect: diagnostics ----
+// error: 5:11-15: method 'push' does not exist
```
seanyoung, line 19: a nicer diagnostic saying that push and pop are not available for fixed sizes array would be preferable

seanyoung (approved):

xermicus (approved):

tests/contract_testcases/solana/functions/push_pop_fixed_length_array.sol
LucasSte: Let me check how to fix this.

LucasSte: The function call matching algorithm needs improvements. In the way it is now, it is not possible to say that push and pops are not available for external calls, because adding an error to diagnostics prevents the algorithm from finding a match for the library function. Likewise, trying the library case first makes other test cases fail.

xermicus: I agree, briefly looked at it and came to the same conclusion. Just matching on the function name is not gonna cut it.

LucasSte: Now fixed.

PR #1356, opened 2023-06-05
Improve unused variable elimination

LucasSte:


src/codegen/unused_variable.rs
```diff
@@ -50,6 +51,25 @@ pub fn should_remove_assignment(
...
+            ..
+        } => should_remove_assignment(ns, &args[0], func, opt),
+
+        Expression::Builtin {
+            kind: Builtin::ArrayPop | Builtin::ArrayPush,
+            args,
+            ..
+        } => {
+            // If the argument is a storage reference, the operation shall not be removed
+            if matches!(args[0].ty(), Type::StorageRef(..)) {
```
xermicus, line 66: Does this work on nested types?

xermicus (approved):
LGTM

src/codegen/unused_variable.rs
LucasSte: Do you mean user defined types?

LucasSte: `Type::UserType` only supports elementary types (not vectors or structs).

`Type::Ref` is a pointer to a memory address (non-storage). `Type::StorageRef` is a storage reference. I don't think it is possible to nest one within the other.

seanyoung (approved):

src/codegen/unused_variable.rs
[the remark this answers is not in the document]
LucasSte: > We have an `args[0].ty().is_storage_reference()` shorthand for this

That should be `is_contract_storage()`

Issue #1265, opened 2023-04-12
Duplicate code (part 5)


LucasSte:
1.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/target.rs#L364-L371

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/target.rs#L452-L459

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/target.rs#L528-L535

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/emit/substrate/target.rs#L638-L645

========================================
2.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/sema/types.rs#L588-L594

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/sema/types.rs#L682-L689

========================================
3.

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/codegen/vector_to_slice.rs#L60-L65

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/codegen/vector_to_slice.rs#L79-L83

https://github.com/hyperledger/solang/blob/1b46d73b32449a09ba1c1e54c8073ecb4d7cf4a0/src/codegen/vector_to_slice.rs#L91-L95

LucasSte:
Hi @Arun9650 ,
Yes, the idea is just to remove the duplicate code using either functions or Rust macros.

seanyoung:
> hi i like to take this issue , so we need to just remove the duplicate the code

Great, go for it! Please let us know if you have any questions.

Sanket-0510:
hello, I would like to take this issue, I have started working to fix repetitive code by adding function to generate the pointer in target.rs, I would like to show you code to check if am going in right direction and also the naming convention I am using for it is correct or not.

LucasSte:
[...]

PR #1359, opened 2023-06-07
fix(refactor): Fix duplicate code for issue #1265

mayank-pathakk:
This PR aims to solve the duplicate code issue in #1265 .

### DIFF
--- a/src/emit/substrate/target.rs
+++ b/src/emit/substrate/target.rs
@@ -19,6 +19,19 @@ use inkwell::{AddressSpace, IntPredicate};
 use solang_parser::pt::Loc;
 use std::collections::HashMap;
 
+macro_rules! create_gep_instruction {
+    ($binary:expr, $base_pointer:expr, $index:expr) => {
+        unsafe {
+            $binary.builder.build_gep(
+                $binary.context.i8_type().array_type(SCRATCH_SIZE),
+                $base_pointer.as_pointer_value(),
+                &[i32_zero!(), $index],
+                "data_offset",
+            )
+        }
+    };
+}
+
 impl<'a> TargetRuntime<'a> for SubstrateTarget {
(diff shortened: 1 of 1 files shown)


src/emit/substrate/target.rs
```diff
@@ -526,15 +532,8 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
-        let offset = unsafe {
-            binary.builder.build_gep(
-                binary.context.i8_type().array_type(SCRATCH_SIZE),
-                binary.scratch.unwrap().as_pointer_value(),
-                &[i32_zero!(), length],
-                "data_offset",
-            )
-        };
-
+        let offset = create_gep_instruction!(binary, binary.scratch.unwrap(), length);
```
LucasSte, line 535: The name should be specific to what this function/macro is doing. `create_gep_instruction` is too generic. What you are doing here is fetching a pointer to an offset of the scratch buffer.

mayank-pathakk: Okay. I'll amend this and keep it in mind for the remaining of the refactoring

src/emit/substrate/target.rs
```diff
@@ -19,6 +19,19 @@ use inkwell::{AddressSpace, IntPredicate};
 use solang_parser::pt::Loc;
 use std::collections::HashMap;
 
+macro_rules! create_gep_instruction {
```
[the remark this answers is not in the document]
xermicus: That's not what macros are for. If you'd like to learn more about macros, I can recommend the [macro section in the rust book](https://doc.rust-lang.org/book/ch19-06-macros.html#the-difference-between-macros-and-functions).

Nirhar:
I'd like to work on this issue, if its still open!

PR #1364, opened 2023-06-13
Fix codegen tests

xermicus:
Make the failing case less brittle


LucasSte (approved):

seanyoung (commented):
I don't understand this change. Why does it make it less "brittle"?

xermicus:
Well, exact matches on strings like that are brittle. By removing a part of which we don't care about in that case makes it less brittle because it decreases the chance to fail because of random totally unrelated changes.

seanyoung (approved):

PR #1365, opened 2023-06-13
`wasm-opt` for Substrate 

xermicus:
`wasm-opt` brings some great optimizations and we should use it (`cargo contract` uses it for ink contracts). For example, the Wasm blob of our flipper example (compile `--release`) goes down from 2.2kb to 1.2kb.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,8 +60,8 @@ scale-info = "2.4"
 petgraph = "0.6.3"
 wasmparser = "0.106.0"
 wasm-encoder = "0.28"
-wasm-opt = "0.112.0"
-contract-build = "3.0.1"
+wasm-opt = { version = "0.112.0", optional = true }
+contract-build = { version = "3.0.1", optional = true }
 
 
 [dev-dependencies]
@@ -94,8 +94,9 @@ no-default-features = true
 lto = true
 
 [features]
-default = ["llvm"]
+default = ["llvm", "wasm_opt"]
 llvm = ["inkwell", "libc"]
+wasm_opt = ["wasm-opt", "contract-build"]
 
 [workspace]
 members = ["solang-parser", "tests/wasm_host_attr"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,8 @@ scale-info = "2.4"
 petgraph = "0.6.3"
 wasmparser = "0.106.0"
 wasm-encoder = "0.28"
+wasm-opt = "0.112.0"
+contract-build = "3.0.1"
 
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,7 +96,7 @@ lto = true
 [features]
 default = ["llvm", "wasm_opt"]
 llvm = ["inkwell", "libc"]
-wasm_opt = ["wasm-opt", "contract-build"]
+wasm_opt = ["llvm", "wasm-opt", "contract-build"]
 
 [workspace]
 members = ["solang-parser", "tests/wasm_host_attr"]
--- a/docs/code_gen_options.rst
+++ b/docs/code_gen_options.rst
@@ -145,6 +145,21 @@ This also means that, whenever the length of an array is accessed using '.length
 Note that this optimization does not cover every case. When an array is passed
 as a function argument, for instance, the length is unknown.
 
+``wasm-opt`` optimization passes
+--------------------------------
+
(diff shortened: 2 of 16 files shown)


docs/installing.rst
```diff
@@ -87,11 +87,16 @@ Then you can build the image using:
 Option 5: Build Solang from source
 ----------------------------------
 
-In order to build Solang from source, you will need rust 1.68.0 or higher,
-and a build of LLVM based on the Solana LLVM tree. There are a few LLVM patches required that are not upstream yet.
+In order to build Solang from source, you will need:
+
+* Rust version 1.68.0 or higher
+* A C++ compiler with support for C++17
+* A build of LLVM based on the Solana LLVM tree. There are a few LLVM patches required that are not upstream yet.
```
LucasSte, line 94: Our LLVM binary already comes with clang, which can build C++ 17.

seanyoung: > Our LLVM binary already comes with clang, which can build C++ 17.

Only the wasm and sbf targets are enabled, not the native target (e.g. intel or arm). So, this compiler won't work. We may need to adjust the build for that. I am guessing the native clang/gcc is being used in CI and @xermicus' local machine

Cargo.toml
```diff
@@ -60,6 +60,8 @@ scale-info = "2.4"
 petgraph = "0.6.3"
 wasmparser = "0.106.0"
 wasm-encoder = "0.28"
+wasm-opt = "0.112.0"
+contract-build = "3.0.1"
```
seanyoung, line 64: These should be optional crates enabled by the llvm feature

docs/code_gen_options.rst
```diff
@@ -145,6 +145,21 @@ This also means that, whenever the length of an array is accessed using '.length
 Note that this optimization does not cover every case. When an array is passed
 as a function argument, for instance, the length is unknown.
 
+``wasm-opt`` optimization passes
+--------------------------------
+
+For the Substrate target, optimization passes from the `binaryen <https://github.com/WebAssembly/binaryen>`_ ``wasm-opt`` 
+tool can be applied. This may vastly shrink the Wasm code size and make it more efficient.
```
seanyoung, line 152: - wasm should be either lower case or upper case
- "may vastly" is a bit odd. Sometimes it only shrinks it a little bit. How about simply "may shrink"?
- "makes it more effcient" (make => makes)

seanyoung (changes requested):

docs/code_gen_options.rst
LucasSte: > * wasm should be either lower case or upper case

Ideally it should conform to how we've been writing `wasm` in our docs.

xermicus: The abbreviation is called [`Wasm`](https://webassembly.org/) as I wrote it

docs/installing.rst
xermicus: That's a very good point. I need to investigate how the `build.rs` of wasm-opt selects the compiler, as we want to make sure that this works consistently

seanyoung (approved):
Enabling the wasm_opt feature without the llvm feature makes no sense, but I don't know what we can do about that (and if it matters).

docs/installing.rst
xermicus: @seanyoung `wasm-opt` uses the `cc` crate under the hood, which expects [`cc` on unix](https://github.com/rust-lang/cc-rs#compile-time-requirements) for unix like systems. So on Linux and MacOS it'll always work, unless someone overwrites their `cc` with the solana labs LLVM project. And because that toolchain only supports wasm and SBF, this would represent a severe misconfiguration of their system inducing a lot of problems elsewhere.

For windows it seems fine too as [`cl.exe`](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options?view=msvc-170) seems to be the equivalent of unix `cc`.

xermicus:
Agreed, however we can't tie it together with the `llvm` flag, because the other way round, e.g. enabling `llvm` but not `wasm_opt`, it does make sense.

But what we can do is to automatically enable `llvm` for the `wasm_opt` feature, WDYT?

seanyoung:
> But what we can do is to automatically enable `llvm` for the `wasm_opt` feature, WDYT?

The more I think about it, the more I think it doesn't matter

docs/code_gen_options.rst
LucasSte: > * "may vastly" is a bit odd. Sometimes it only shrinks it a little bit. How about simply "may shrink"?
> * "makes it more effcient" (make => makes)

@xermicus This has not been solved yet.

LucasSte (approved):

PR #1366, opened 2023-06-14
Refactor constructor annotations

LucasSte:
Presently, we can declare an account on a Solana constructor using the `@payer(my_account)` annotation, while `@seed(my_seed)` refers to a constructor parameter. Such a construction is confusing for the same syntax have two different meanings.

This PR limits the scope of annotation above a constructor to accept only literals as parameters (e.g. `@seed("pine_tree")`). Annotations that refer to function parameters must appear before them: `constructor (@seed bytes arg1)`. This is another in #1251.


docs/language/contracts.rst
```diff
@@ -119,6 +114,9 @@ For the creation of a contract, the data account must the **first** element in s
 .. include:: ../examples/solana/create_contract_with_metas.sol
   :code: solidity
 
+The ordering of the account in the ``AccountMeta`` array matters and must follow the
+:ref:`IDL disposal <account_management>`.
```
seanyoung, line 118: IDL disposal? Am I missing something?

seanyoung (commented):
Looks good. I've checked and I think you have all the cases covered

seanyoung:
There are no testcases with whitespace between `@` and the identifier. I am not sure how we can disallow whitespace there (and if we want to)

xermicus (approved):
LGTM

LucasSte:
> There are no testcases with whitespace between `@` and the identifier. I am not sure how we can disallow whitespace there (and if we want to)

The entire annotation is now a token. There is no way of declaring an annotation with a whitespace between the `@` and the identifier. I've added a test for this case as well.

docs/language/contracts.rst
LucasSte: I've changed to `IDL ordering`. What do you think?

solang-parser/src/lexer.rs
```diff
@@ -178,6 +178,8 @@ pub enum Token<'input> {
     Case,
     Default,
     YulArrow,
+
+    Annotation(&'input str),
     At,
```
seanyoung, line 183: `Token::At` is now unused, right? I don't think there is any reason to keep it.

LucasSte: Fixed!

docs/targets/solana.rst
```diff
@@ -227,16 +227,20 @@ is passed to the transaction that runs the constructor code.
 
 Alternatively, the data account can be created by the constructor, on chain. When
 this method is used, some parameters must be specified for the account
-using annotations. Those are placed before the constructor. If there is no
-constructor present, then an empty constructor can be added. The constructor
-arguments can be used in the annotations.
+using annotations. Annotations placed on top of a constructor can only contain literals or
```
seanyoung, line 230: How about replacing "placed on top of" with above or before

"on top of" sounds like it needs to be on the same line somehow

docs/targets/solana.rst
```diff
@@ -253,9 +257,9 @@ If the data account is going to be a
 `program derived address <https://docs.solana.com/developing/programming-model/calling-between-programs#program-derived-addresses>`_,
 then the seeds and bump have to be provided. There can be multiple seeds, and an optional
 single bump. If the bump is not provided, then the seeds must not create an
-account that falls on the curve. The ``@seed`` can be a string literal,
-or a hex string with the format ``hex"4142"``, or a constructor argument of type
-``bytes``. The ``@bump`` must a single byte of type ``bytes1``.
+account that falls on the curve. When placed on top of the constructor, the ``@seed`` can be a string literal,
```
seanyoung, line 260: "placed on top of the" => above or before

docs/targets/solana.rst
```diff
@@ -253,9 +257,9 @@ If the data account is going to be a
...
+or a hex string with the format ``hex"4142"``. If next to an argument, the seed annotation must refer to an argument
```
seanyoung, line 261: "If next to an" => before?

solang-parser/src/pt.rs
```diff
@@ -1258,6 +1273,8 @@ pub struct Parameter {
     pub storage: Option<StorageLocation>,
     /// The optional identifier.
     pub name: Option<Identifier>,
+    /// An optional annotation '@annotation'.
+    pub annotation: Option<Annotation>,
```
seanyoung, line 1277: The other fields are ordered by their appearance in the source code (exception: loc is always first). Maybe annotation can go first after loc?

seanyoung (approved):

PR #1369, opened 2023-06-16
Allow returns values to be ignored in try catch

seanyoung:
A try catch statement does not need to decode the return values if it is not interested.

```
interface I {
      function func()   externalpublic returns (int, bool);
}

contract C {
     function test(I i) public {
           try I.func() {
           }
           catch (bytes memory bs) { 
           }
     } 
}
```

Here `func()` returns two values but the try statement does not use them at all. This means they do not have abi decoded either.

### DIFF
--- a/tests/substrate_tests/calls.rs
+++ b/tests/substrate_tests/calls.rs
@@ -334,6 +334,65 @@ fn try_catch_external_calls() {
 }
 
 #[test]
+fn try_catch_external_calls_dont_decode_returns() {
+    // try not using the return values of test() - revert case
+    // note the absense of "try o.test() returns (int32 y, bool) {"
+    let mut runtime = build_solidity(
+        r##"
+        contract c {
+            function test() public returns (int32 x) {
+                other o = new other();
+                try o.test() {
+                    x = 1;
+                } catch (bytes c) {
+                    x = 2;
+                }
+            }
+        }
+
+        contract other {
+            function test() public returns (int32, bool) {
+                revert("foo");
+            }
+        }
+        "##,
+    );
+
+    runtime.constructor(0, Vec::new());
+    runtime.function("test", Vec::new());
+
+    assert_eq!(runtime.output(), 2i32.encode());
+
+    // try not using the return values of test() - normal case
+    // note the absense of "try o.test() returns (int32 y, bool) {"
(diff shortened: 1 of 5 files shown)


tests/substrate_tests/calls.rs
```diff
@@ -216,6 +216,32 @@ fn try_catch_external_calls() {
...
+            function test() public {
+                other o = new other();
+                int32 x = 0;
+                try o.test() {
+                    x = 1;
+                } catch (bytes c) {
+                    assert(c == hex"a079c3080c666f6f");
+                    x = 2;
+                }
+                assert(x == 2);
```
xermicus, line 231: Without further context, I couldn't tell why this test case exists or what exactly is being tested here.

I think it'd be better if this `test()` function would return `x` instead, so that it can be explicitly checked later on. Also it could go into a dedicated `#[test]`.

tests/substrate_tests/calls.rs
```diff
@@ -216,6 +216,32 @@ fn try_catch_external_calls() {
...
+        contract other {
+            function test() public returns (int32, bool) {
+                revert("foo");
+            }
+        }
+        "##,
+    );
+
+    runtime.constructor(0, Vec::new());
+    runtime.function_expect_failure("test", Vec::new());
```
LucasSte, line 244: I think it is worth adding a case in which we enter the `catch` branch and the contract does not fail.

seanyoung: You're right, that was a mistake. I've fixed it.

tests/substrate_tests/calls.rs
seanyoung: I've moved the test into it's own `#[test]` and added some comments, and expanded the description on this PR.

xermicus: Much clearer now, thanks!

xermicus (approved):
I think this change is correct. A codegen test, ultimately proofing that the decoding is no longer happening, would be the cherry on top.

LucasSte (approved):

seanyoung:
> I think this change is correct. A codegen test, ultimately proofing that the decoding is no longer happening, would be the cherry on top.

I've added a test

PR #1370, opened 2023-06-16
Remove unnecessary code from lexer

LucasSte:
The function `Lexer::next` has a `Result` as its returned type, but we never return `Err()`. This PR removes this unnecessary usage of `Result`.

I found this while creating a new token for `@annotation`.


seanyoung (approved):
Nice catch

xermicus (approved):

PR #1374, opened 2023-06-19
Implicit accessor function should return struct members, not struct

seanyoung:
If the acessor function returns a single struct, then the members should be returned. If any of those members are structs, then those members will remain structs.

	contract C {
		struct S { int a; bool b; }

		S public s;

		function foo() public {
			(int a, bool b) = this.s();
		}
	}

Also, the accesor function should be declared `external` and therefore not accessible as an internal function call.

### DIFF
--- a/src/abi/tests.rs
+++ b/src/abi/tests.rs
@@ -84,30 +84,10 @@ fn constants_and_types() {
 
     assert_eq!(idl.types.len(), 2);
 
-    assert_eq!(idl.types[0].name, "MyStruct");
+    assert_eq!(idl.types[0].name, "Week");
     assert!(idl.types[0].docs.is_none());
     assert_eq!(
         idl.types[0].ty,
-        IdlTypeDefinitionTy::Struct {
-            fields: vec![
-                IdlField {
-                    name: "g".to_string(),
-                    docs: None,
-                    ty: IdlType::U8,
-                },
-                IdlField {
-                    name: "d".to_string(),
-                    docs: None,
-                    ty: IdlType::Array(IdlType::U8.into(), 2)
-                }
-            ]
-        }
-    );
-
-    assert_eq!(idl.types[1].name, "Week");
-    assert!(idl.types[1].docs.is_none());
-    assert_eq!(
-        idl.types[1].ty,
         IdlTypeDefinitionTy::Enum {
             variants: vec![
                 IdlEnumVariant {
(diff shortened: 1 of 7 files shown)


src/abi/tests.rs
```diff
@@ -104,10 +109,10 @@ fn constants_and_types() {
         }
     );
 
-    assert_eq!(idl.types[1].name, "Week");
-    assert!(idl.types[1].docs.is_none());
+    assert_eq!(idl.types[0].name, "Week");
+    assert!(idl.types[0].docs.is_none());
     assert_eq!(
-        idl.types[1].ty,
+        idl.types[0].ty,
```
LucasSte, line 115: Would you mind placing the check for item `0` before that for item `1`? I believe it improves readability.

src/sema/variables.rs
```diff
@@ -460,6 +461,143 @@ pub fn variable_decl<'a>(
...
+                    body = vec![
+                        Statement::VariableDecl(pt::Loc::Implicit, var_no, param, Some(expr)),
+                        Statement::Return(
+                            pt::Loc::Implicit,
+                            Some(Expression::List {
+                                loc: pt::Loc::Implicit,
+                                list,
+                            }),
+                        ),
+                    ];
```
LucasSte, line 524: What do you think of building the function body during codegen?

src/sema/variables.rs
```diff
@@ -460,6 +461,143 @@ pub fn variable_decl<'a>(
                 ));
             }
 
+            let body;
+            let returns;
```
LucasSte, line 465: You don't have to declare the variable like this. In Rust you can do `let (body, returns) = if let Type::Struct(str) = ..`

src/sema/variables.rs
```diff
@@ -460,6 +461,143 @@ pub fn variable_decl<'a>(
...
+                            pt::Loc::Implicit,
+                            Some(Expression::List {
+                                loc: pt::Loc::Implicit,
+                                list,
+                            }),
+                        ),
+                    ];
+
+                    returns = def.fields.clone();
+                }
```
LucasSte, line 583: The two branches of `if constant { .. } else {...}` have nearly identical code. You can simplify most of it.

tests/contract_testcases/solana/accessor/accessor_is_external.sol
```diff
@@ -0,0 +1,10 @@
+contract C {
+	int constant public c = 102;
+
+	function f() public {
+		int x = c();
+	}
+}
+// ---- Expect: diagnostics ----
+// error: 5:11-14: functions declared external cannot be called via an internal function call
+// 	note 2:22-23: declaration of function 'c'
```
LucasSte, line 10: The note points to a public variable, while the error says `functions declared external`. Don't we need a clearer error message for this case?

src/sema/variables.rs
seanyoung: I can see the argument for keeping the ast more "pure", but on the other hand I don't think it is that ugly. However, I am not sure what this would look like, code wise.

We need to build the function prototype in sema, that includes the parameters and return values. Once we've calculated all of that, we've already done almost all the work to build up the body. We would need to do the same work again in codegen to build the body, or somehow share this code.

Maybe something for another day.

xermicus: In principle I agree with Lucas, it seems odd to sort of generate code not in _codegen_ but in sema. However there might be more such instances. I also thought a bit about it and didn't come to an immediate conclusion or solution to that question.

xermicus (approved):
LGTM

tests/contract_testcases/solana/accessor/accessor_is_external.sol
seanyoung: Good idea, I've improved this.

src/sema/variables.rs
LucasSte: Your new version looks much cleaner. Thanks!

LucasSte (approved):

PR #1376, opened 2023-06-20
Substrate: Implement call flags

xermicus:
We'll need them for more ergonomic `delegatecall` support.

### DIFF
--- a/integration/substrate/call_flags.sol
+++ b/integration/substrate/call_flags.sol
@@ -1,0 +1,58 @@
+contract CallFlags {
+    uint8 roundtrips;
+
+    // See https://github.com/paritytech/substrate/blob/5ea6d95309aaccfa399c5f72e5a14a4b7c6c4ca1/frame/contracts/src/wasm/runtime.rs#L373
+    enum CallFlag { FORWARD_INPUT, CLONE_INPUT, TAIL_CALL, ALLOW_REENTRY }
+    function bitflags(CallFlag[] _flags) internal pure returns (uint32 flags) {
+        for (uint n = 0; n < _flags.length; n++) {
+            flags |= (2 ** uint32(_flags[n]));
+        }
+    }
+
+    // Reentrancy is required for reaching the `foo` function for itself.
+    //
+    // Cloneing and forward should have the effect of calling this function again, regardless of what _address was passed.
+    // Furthermore:
+    // Cloneing the clone should work together with reentrancy.
+    // Forwarding the input should fail caused by reading the input more than once in the loop
+    // Tail call should work with any combination of input forward.
+    function echo(
+        address _address,
+        bytes4 _selector,
+        uint32 _x,
+        CallFlag[] _flags
+    ) public payable returns(uint32 ret) {
+        for (uint n = 0; n < 2; n++) {
+            if (roundtrips > 1) {
+                return _x;
+            }
+            roundtrips += 1;
+
+            bytes input = abi.encode(_selector, _x);
+            (bool ok, bytes raw) =  _address.call{flags: bitflags(_flags)}(input);
+            require(ok);
+            ret = abi.decode(raw, (uint32));
+
+            roundtrips -= 1;
+        }
+    }
+
+    @selector([0,0,0,0])
+    function foo(uint32 x) public pure returns(uint32) {
+        return x;
+    }
+
+    // Yields different result for tail calls
+    function tail_call_it(
+        address _address,
+        bytes4 _selector,
+        uint32 _x,
+        CallFlag[] _flags
+    ) public returns(uint32 ret) {
+        bytes input = abi.encode(_selector, _x);
+        (bool ok, bytes raw) =  _address.call{flags: bitflags(_flags)}(input);
+        require(ok);
+        ret = abi.decode(raw, (uint32));
+        ret += 1;
+    }
+}
--- a/integration/substrate/call_flags.spec.ts
+++ b/integration/substrate/call_flags.spec.ts
@@ -1,0 +1,80 @@
+import expect from 'expect';
+import { createConnection, deploy, aliceKeypair, query, debug_buffer, } from './index';
+import { ContractPromise } from '@polkadot/api-contract';
+import { ApiPromise } from '@polkadot/api';
+import { KeyringPair } from '@polkadot/keyring/types';
+
+enum CallFlags {
+    FORWARD_INPUT, CLONE_INPUT, TAIL_CALL, ALLOW_REENTRY
+}
+
+describe('Deploy the CallFlags contract and tests for various call flag combinations', () => {
+    let conn: ApiPromise;
+    let contract: ContractPromise;
+    let alice: KeyringPair;
+    const voyager = 987654321;
+    const foo = [0, 0, 0, 0];
+
+    before(async function () {
+        alice = aliceKeypair();
+        conn = await createConnection();
+        const deployment = await deploy(conn, alice, 'CallFlags.contract', 0n);
+        contract = new ContractPromise(conn, deployment.abi, deployment.address);
+    });
+
+    after(async function () {
+        await conn.disconnect();
+    });
+
+    it('works with the reentry flag', async function () {
+        const flags = [CallFlags.ALLOW_REENTRY];
+        const answer = await query(conn, alice, contract, "echo", [contract.address, foo, voyager, flags]);
+        expect(answer.output?.toJSON()).toStrictEqual(voyager);
+    });
+
+    it('works with the reentry and tail call flags', async function () {
+        const flags = [CallFlags.ALLOW_REENTRY, CallFlags.TAIL_CALL];
+        const answer = await query(conn, alice, contract, "echo", [contract.address, foo, voyager, flags]);
+        expect(answer.output?.toJSON()).toStrictEqual(voyager);
+    });
+
+    it('works with the reentry and clone input flags', async function () {
+        const flags = [CallFlags.ALLOW_REENTRY, CallFlags.CLONE_INPUT];
+        const answer = await query(conn, alice, contract, "echo", [contract.address, foo, voyager, flags]);
+        expect(answer.output?.toJSON()).toStrictEqual(voyager);
+    });
+
+    it('works with the reentry, tail call and clone input flags', async function () {
+        const flags = [CallFlags.ALLOW_REENTRY, CallFlags.TAIL_CALL, CallFlags.CLONE_INPUT];
+        const answer = await query(conn, alice, contract, "echo", [contract.address, foo, voyager, flags]);
+        expect(answer.output?.toJSON()).toStrictEqual(voyager);
+    });
+
+    it('fails without the reentry flag', async function () {
+        const flags = [CallFlags.TAIL_CALL];
+        const answer = await query(conn, alice, contract, "echo", [contract.address, foo, voyager, flags]);
+        const { index, error } = answer.result.asErr.asModule;
+        // Module 8 error 0x14 is ReentranceDenied in the contracts pallet
+        expect(index.toJSON()).toStrictEqual(8);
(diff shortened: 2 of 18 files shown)


src/emit/substrate/mod.rs
```diff
@@ -294,6 +295,16 @@ impl SubstrateTarget {
...
+        external!(
+            "delegate_call",
+            i32_type,
+            u32_val,
+            u8_ptr,
+            u8_ptr,
+            u32_val,
+            u8_ptr,
+            u32_val
+        );
```
seanyoung, line 307: This is unused and therefore untested

tests/substrate.rs
```diff
@@ -22,6 +23,20 @@ mod substrate_tests;
...
+    TailCall,
+    AllowReentry,
+}
+
+impl CallFlags {
+    fn set(&self, flags: u32) -> bool {
+        flags & 2u32.pow(*self as u32) != 0
+    }
+}
+
```
seanyoung, line 39: How about
```rust
enum CallFlags {
    ForwardInput = 1,
    CloneInput = 2,
    TailCall = 4,
    AllowReentry = 8,
}
```

src/codegen/cfg.rs
```diff
@@ -1201,7 +1203,8 @@ impl ControlFlowGraph {
                         format!("({contract_no}, {function_no})")
                     } else {
                         "_".to_string()
-                    }
+                    },
+                    flags.as_ref().map(|e| self.expr_to_string(contract, ns, e)).unwrap_or("0".to_string())
```
seanyoung, line 1207: I don't think it should be 0 for unspecified, it should be empty (i.e. `.unwrap_or("".to_string())` (maybe `.unwrap_or_default()`.)

integration/substrate/call_flags.sol
```diff
@@ -0,0 +1,58 @@
+contract CallFlags {
+    uint8 roundtrips;
+
+    // See https://github.com/paritytech/substrate/blob/5ea6d95309aaccfa399c5f72e5a14a4b7c6c4ca1/frame/contracts/src/wasm/runtime.rs#L373
+    enum CallFlag { FORWARD_INPUT, CLONE_INPUT, TAIL_CALL, ALLOW_REENTRY }
```
seanyoung, line 5: How about defining these flags in `import "substrate";`?

seanyoung (commented):
Needs documentation

integration/substrate/call_flags.sol
xermicus: I thought about it but decided against it. This doesn't need to be a compiler builtin; there's nothing that can't be written in plain Solidity. I'd rather provide such helper things directly in Solidity (sort of like a stdlib).

tests/substrate.rs
xermicus: Absolutely

docs/examples/substrate/call_flags.sol
```diff
@@ -0,0 +1,23 @@
+contract CallFlags {
+
+    enum CallFlag { FORWARD_INPUT, CLONE_INPUT, TAIL_CALL, ALLOW_REENTRY }
+
+    // Set all flags found in _flags to true.
+    function bitflags(CallFlag[] _flags) internal pure returns (uint32 flags) {
+        for (uint n = 0; n < _flags.length; n++) {
+            flags |= (2 ** uint32(_flags[n]));
+        }
+    }
```
seanyoung, line 10: This is not very readable, and it's part of the documentation. How about:

```solidity
uint32 constant FORWARD_INPUT = 1;
uint32 constant CLONE_INPUT = 2;
uint32 constant TAIL_CALL = 4;
uint32 constant ALLOW_REENTRY = 8;
```
You could even put it in a library:

```solidity
library CallFlags {
uint32 constant FORWARD_INPUT = 1;
uint32 constant CLONE_INPUT = 2;
uint32 constant TAIL_CALL = 4;
uint32 constant ALLOW_REENTRY = 8;
}

contract c {
   function call_with_flags(
        address _address,
        bytes4 _selector,
    ) public returns (bytes ret) {
        uint32 call_flags = bitflags(_flags);
        (bool ok, ret) = _address.call{flags: CallFlags.ALLOW_REENTRANT}(_selector);
        require(ok);
    }
}
```

It could also go into `substrate-library/`.

docs/targets/substrate.rst
```diff
@@ -49,3 +49,19 @@ import object.
...
+
+Call Flags
+__________
+
+The Substrate contracts pallet knows several 
+`flags <https://github.com/paritytech/substrate/blob/6e0059a416a5768e58765a49b33c21920c0b0eb9/frame/contracts/src/wasm/runtime.rs#L392>`_ 
+that can be used when calling other contracts.
+
+Solang allows a named ``flags`` argument of type ``uint32`` to the ``address.call()`` function to set desired flags.
+By default (if this argument is unset), no flag will be set.
```
seanyoung, line 61: This also works on normal calls, rights? e.g. `foo.bar{flags: CallFlags.ALLOW_REENTRANT}(102);`

docs/examples/substrate/call_flags.sol
xermicus: I like that!

docs/examples/substrate/call_flags.sol
```diff
@@ -0,0 +1,23 @@
...
+
+    // Call the contract at _address with the given _selector.
+    // Specify any flag used for the contract call in _flags.
+    function call_with_flags(
+        address _address,
```
LucasSte, line 15: Why do you prefer arguments that start with an underscore?
(I'm not asking you to change. It is just my curiosity.)

docs/targets/substrate.rst
```diff
@@ -49,3 +49,19 @@ import object.
...
     target.
+
+Call Flags
+__________
+
+The Substrate contracts pallet knows several 
+`flags <https://github.com/paritytech/substrate/blob/6e0059a416a5768e58765a49b33c21920c0b0eb9/frame/contracts/src/wasm/runtime.rs#L392>`_ 
+that can be used when calling other contracts.
+
+Solang allows a named ``flags`` argument of type ``uint32`` to the ``address.call()`` function to set desired flags.
```
LucasSte, line 60:
```suggestion
Solang allows a ``flags`` call argument of type ``uint32`` to the ``address.call()`` function to set desired flags.
```
We've been calling this syntax `call argument` in Solana.

integration/substrate/call_flags.sol
```diff
@@ -0,0 +1,63 @@
...
+    function bitflags(CallFlag[] _flags) internal pure returns (uint32 flags) {
+        for (uint n = 0; n < _flags.length; n++) {
+            flags |= (2 ** uint32(_flags[n]));
+        }
+    }
+
+    // Reentrancy is required for reaching the `foo` function for itself.
+    //
+    // Cloneing and forward should have the effect of calling this function again, regardless of what _address was passed.
```
LucasSte, line 14:
```suggestion
    // Cloning and forwarding should have the effect of calling this function again, regardless of what _address was passed.
```

integration/substrate/call_flags.sol
```diff
@@ -0,0 +1,63 @@
...
+    // Furthermore:
+    // Cloneing the clone should work together with reentrancy.
```
LucasSte, line 16:
```suggestion
    // Cloning the clone should work together with reentrancy.
```

integration/substrate/call_flags.sol
```diff
@@ -0,0 +1,63 @@
...
+    // Forwarding the input should fail caused by reading the input more than once in the loop
```
LucasSte, line 17:
```suggestion
    // Forwarding the input should fail due to reading the input more than once in the loop
```

docs/examples/substrate/call_flags.sol
```diff
@@ -0,0 +1,23 @@
+contract CallFlags {
+
+    enum CallFlag { FORWARD_INPUT, CLONE_INPUT, TAIL_CALL, ALLOW_REENTRY }
+
+    // Set all flags found in _flags to true.
+    function bitflags(CallFlag[] _flags) internal pure returns (uint32 flags) {
+        for (uint n = 0; n < _flags.length; n++) {
+            flags |= (2 ** uint32(_flags[n]));
```
LucasSte, line 8: Isn't is cheaper to do `2 << (uint32(_flags[n]) - 1)`?

docs/examples/substrate/call_flags.sol
xermicus: I changed the example to a much simpler and more to the point version.

Gonna introduce the `substrate-library/` in a later PR. The enum version, while a bit less readable is really nice for end-users. But the constants are nice to have as well. I think we want both. Additionally, I think the `Hash` type from the `substrate` builtin could also be shoved off into a Solidity version. But it's all out of scope for this PR.

docs/examples/substrate/call_flags.sol
xermicus: I thought it was in the Solidity style guide because I saw that a lot. But I checked, it's not. It seems to be something many Solidity devs just do.

However in this case for `_address` I left it in place because `address`  is already the name of the type...

integration/substrate/call_flags.sol
```diff
@@ -0,0 +1,63 @@
...
+    // Tail call should work with any combination of input forward.
```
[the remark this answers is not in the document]
xermicus: Yes, gonna fix it

docs/examples/substrate/call_flags.sol
xermicus: Might well be 🧠  gonna keep this in mind for a library version, thanks

LucasSte: Ah, there is an even better choice: `1 << uint32(_flags[n])`.
Now power and no subtraction!

xermicus: Ah yes of course, left shift is equal to 2 to the power of it in that case, why didn't I think of this :sweat_smile:

integration/substrate/call_flags.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    it('fails with the input forwarding flag', async function () {
+        const flags = [CallFlags.ALLOW_REENTRY, CallFlags.FORWARD_INPUT];
+        const answer = await query(conn, alice, contract, "echo", [contract.address, foo, voyager, flags]);
+        const { index, error } = answer.result.asErr.asModule;
+        // Module 8 error 0x0b is ContractTrapped in the contracts pallet
+        expect(index.toJSON()).toStrictEqual(8);
+        expect(error.toJSON()).toStrictEqual("0x0b000000");
+    });
+
+    it('test for the tail call flag to work correct', async function () {
```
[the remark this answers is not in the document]
LucasSte: @xermicus This has not been fixed.

docs/targets/substrate.rst
xermicus: @LucasSte weird I added them all to the suggestion batch. Maybe I misclicked in the so that's why only one of your suggestions went in

docs/targets/substrate.rst
```diff
@@ -49,3 +49,19 @@ import object.
...
     target.
+
+Call Flags
+__________
+
+The Substrate contracts pallet knows several 
+`flags <https://github.com/paritytech/substrate/blob/6e0059a416a5768e58765a49b33c21920c0b0eb9/frame/contracts/src/wasm/runtime.rs#L392>`_ 
+that can be used when calling other contracts.
+
+Solang allows a ``flags`` call argument of type ``uint32`` to the ``address.call()`` function to set desired flags.
```
LucasSte, line 60:
```suggestion
Solang allows a ``flags`` call argument of type ``uint32`` in the ``address.call()`` function to set desired flags.
```

LucasSte (approved):

seanyoung (approved):

Issue #1367, opened 2023-06-15
Getting LLVM ERROR: Branch target out of insn range


kupermind:
Hi guys! When compiling a code for solana via solang I am getting the following error if I add one more state variable in the contract:
```
info: contract ServiceRegistrySolana uses at least 4152 bytes account data
info: Generating LLVM IR for contract ServiceRegistrySolana with target solana
LLVM ERROR: Branch target out of insn range
zsh: abort      npm run build
```

The working code is located here:
https://github.com/valory-xyz/autonolas-registries/tree/solana

Steps to reproduce the working code:
```
git clone --recursive https://github.com/valory-xyz/autonolas-registries.git
git checkout 573ced9e1d1d3d070bc26a58da8b6b04e5291be6
cd autonolas-registries/integrations/solana
yarn
npm run build
npm run test
```

Now try to declare a single variable on line 66, for example, `address public var`, in the following file:
https://github.com/valory-xyz/autonolas-registries/tree/573ced9e1d1d3d070bc26a58da8b6b04e5291be6/integrations/solana/contracts/ServiceRegistrySolana.sol

Then try to recompile with `npm run build` or a custom solang build command, and the build will fail with the specified error.

The Solang version is v0.3.0, the platform is MacOs, although tried on the latest Linux as well and got the same result.

In order to make it work with that additional state variable, I need to comment the code out starting from the line 797, which might point to some contract size issues. Although if I were to add mapping instead of a primitive variable, then the original code would compile, just increasing the bytes account data. So the issue is really related to the primitive state variables.

Thanks a lot for your help!

LucasSte:
Hey @kupermind,
Can you provide the exact contract that fails for you?
I copied the file you liked and added a new variable `address public var` in line 66, but I could not replicate the error.

kupermind:
Hi @LucasSte,

Please checkout the following branch: `llvm_insn_range_solana`
The `ServiceRegistrySolana.sol` contract fails right from the start, you just need to do the described steps:
```
git clone --recursive https://github.com/valory-xyz/autonolas-registries.git
git checkout llvm_insn_range_solana
cd autonolas-registries/integrations/solana
yarn
npm run build
```

The build must fail with the specified error.

Thanks a lot!

LucasSte:
I am still investigating the problem, but I've found some options that make your contract work, so you can continue your development until we find a fix.

[...]

PR #1377, opened 2023-06-20
Avoid repeated global strings

LucasSte:
When debugging the issue #1367, I found out that we keep emitting repeated global strings. The LLVM IR for the contract pointed in such an issue had the string `math overflow,\0A` defined 1008 times.

This PR DOES NOT solve #1367, but it is something that needed a fix.

### DIFF
--- a/tests/codegen_testcases/solidity/global_strings.sol
+++ b/tests/codegen_testcases/solidity/global_strings.sol
@@ -1,0 +1,153 @@
+// RUN: --target solana --emit llvm-ir
+// READ: ServiceRegistrySolana.ll
+
+
+// BEGIN-CHECK: "math overflow,\0A"
+// NOT-CHECK: "math overflow,\0A"
+
+import "solana";
+contract ServiceRegistrySolana {
+    struct Service {
+        address serviceOwner;
+        uint64 securityDeposit;
+        address multisig;
+        bytes32 configHash;
+        uint32 threshold;
+        uint32 maxNumAgentInstances;
+        uint32 numAgentInstances;
+        uint8 state;
+        uint32[] agentIds;
+        uint32[] slots;
+        uint64[] bonds;
+        address[] operators;
+        address[] agentInstances;
+        uint32[] agentIdForAgentInstances;
+    }
+
+    address public owner;
+    address public escrow;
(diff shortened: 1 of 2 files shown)


xermicus:
Are these "global strings" guaranteed to be readonly?

seanyoung:
> Are these "global strings" guaranteed to be readonly?

only when `constant` is set

LucasSte:
> > Are these "global strings" guaranteed to be readonly?
> 
> only when `constant` is set

I missed that. I should reuse them when they are constant. Let me update the PR.

LucasSte:
> > Are these "global strings" guaranteed to be readonly?
> 
> only when `constant` is set

Fixed!
The changes worked because presently we do not generate any non-constant global string.

seanyoung (commented):
How are we testing this?

LucasSte:
> How are we testing this?

The test file I added `tests/codegen_testcases/solidity/global_strings.sol` generates the LLVM-IR file and reads it. It ensures that we only have a single `math overflow` string there, instead of 1008.

tests/codegen_testcases/solidity/global_strings.sol
```diff
@@ -0,0 +1,153 @@
...
+    string public constant VERSION = "1.0.0";
+    mapping(bytes32 => uint64) public mapOperatorAndServiceIdOperatorBalances;
+    mapping (address => address) public mapAgentInstanceOperators;
+    mapping (address => bool) public mapMultisigs;
+    Service public services;
+
+
+
+    constructor(address a, string memory _baseURI)
+    {
```
xermicus, line 43: Does the contract need all those empty functions for this case?

LucasSte: I can shorten the contract, but I left it as it is because it used to generate 1008 definitions of `math overflow`.
What is your opinion in shortening it?

xermicus: Well, it's not the end of the world. I'll leave that up to you. I just thought that if in the end only 1 single `math overflow` string instead of 1008 should be generated, the test code could be more to the point.

xermicus (approved):

PR #1378, opened 2023-06-21
Wasm linker: Do not strip off empty data sections

xermicus:
This PR gets rid of the wasm-opt complaining about `warning: data index out of bounds in name section, data subsection: .bss at index 0`.

This was caused by us just skipping these sections. By doing so, the resulting Wasm blob size will be smaller without (it's just uninitialized data):

```
 - segment[10] <.bss.selector> memory=0 size=4 - init i32=608
  - 0000260: 0000 0000                                ....
 - segment[11] <.bss.calldata_len> memory=0 size=4 - init i32=612
  - 0000264: 0000 0000                                ....
 - segment[12] <.bss.scratch_len> memory=0 size=4 - init i32=616
  - 0000268: 0000 0000                                ....
 - segment[13] <.bss.scratch> memory=0 size=32768 - init i32=624
  - 0000270: 0000 0000 0000 0000 0000 0000 0000 0000  ................
  - ...
```

However, `wasm-opt` will do the same, so we can just rely on `wasm-opt` to do it correctly. I additionally checked the Wasm blob size of various contracts (with `wasm-opt` enabled) and the code sizes stay exactly the same after this change.

### DIFF
--- a/src/linker/wasm.rs
+++ b/src/linker/wasm.rs
@@ -6,10 +6,10 @@ use std::io::Read;
 use std::io::Write;
 use tempfile::tempdir;
 use wasm_encoder::{
-    ConstExpr, DataSection, DataSegment, DataSegmentMode, EntityType, GlobalSection, GlobalType,
-    ImportSection, MemoryType, Module, RawSection, ValType,
+    ConstExpr, EntityType, GlobalSection, GlobalType, ImportSection, MemoryType, Module,
+    RawSection, ValType,
 };
-use wasmparser::{Data, DataKind, Global, Import, Parser, Payload::*, SectionLimited, TypeRef};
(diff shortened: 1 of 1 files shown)


seanyoung (approved):
The scratch buffer is 32Kb bss which gets included in every wasm file. I wrote this code to remove the bss data. 

I knew that wasm-opt could do the same (and even better) but I didn't want to introduce a dependency on the wasm-opt binary for Substrate.

Now that wasm-opt is used as a crate we have the best of both world.

xermicus:
Yeah that makes sense. Somehow, just skipping it (how I implemented after deprecating the `parity-wasm` crate) was only part of the solution. But I thought the same, why bother if `wasm-opt` can do it instead (especially because contract authors want `wasm-opt` anyways).

LucasSte (approved):
😃

PR #1380, opened 2023-06-22
Fix a bunch of subxt-tests rust warnings

seanyoung:
If you don't have a substrate node running you still get errors. I don't know what to do about that.

I've tried putting it in `#[cfg(test)]` which didn't help.

### DIFF
--- a/integration/subxt-tests/src/lib.rs
+++ b/integration/subxt-tests/src/lib.rs
@@ -1,4 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
+#![allow(unused_imports)]
+#![allow(dead_code)]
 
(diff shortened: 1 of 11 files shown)


integration/subxt-tests/src/lib.rs
```diff
@@ -1,4 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
+#![allow(unused_imports)]
+#![allow(dead_code)]
```
xermicus, line 3: Is this needed?

xermicus (approved):
Thanks for fixing this oversight. Ideally we should add a `cargo clippy --workspace --tests` for the workspace in `integration/subx-test`.

LucasSte (approved):

Issue #1354, opened 2023-06-04
Feature flags for targets


xermicus:
Since the custom LLVM fork should only be necessary for the Solana backend, we should decouple the substrate backend. Otherwise anyone wanting to use solang as a library is forced to use the solana labs LLVM fork, even though they might not want to build SBF contracts at all.

Ideally the code base and build scripts can be controlled with a `solana` or `SBF` feature flag.

LucasSte:
Solang is built on top of LLVM. Even if we decouple from Solana's LLVM fork, developers need LLVM installed locally.

xermicus:
(ACK, but what I meant is we should have a way to say "I want use Solang as a crate and I only care for the Substrate target" and it will work with vanilla LLVM. Otherwise, people will need the Solana LLVM labs as soon as they depend on Solang, even if they don't wanna build SBF contracts, which seems very odd)

seanyoung:
If you use upstream llvm, `cargo test` will on Substrate due to the linker not being re-entrant. Maybe that's been fixed in the latest llvm, I haven't looked at this for ages.

PR #1384, opened 2023-06-23
Access payer account using `tx.accounts.account_name`

LucasSte:
This PR creates the infrastructure for accessing accounts with `tx.accounts.account_name`. Presently, we can only access accounts declared with the `@payer` annotation, but our goal is to make this possible for all accounts declared with annotations (this is not implemented yet). 

The documentation I wrote is very terse, because my plan is to add more details when I have the `@reader`, `@mutable`, `@signer` and `@mutableSigner` implemented.

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -189,6 +189,14 @@ pub enum Instr {
     /// be removed. We only have this so we can pass evm code through sema/codegen, which is used
     /// by the language server and the ethereum solidity tests.
     Unimplemented { reachable: bool },
+    /// This instruction serves to track account accesses through 'tx.accounts.my_account'
+    /// on Solana, and has no emit implementation. It is exchanged by the proper
+    /// Expression::Subscript at solana_accounts/account_management.rs
+    AccountAccess {
+        loc: pt::Loc,
+        var_no: usize,
+        name: String,
+    },
 }
 
 /// This struct defined the return codes that we send to the execution environment when we return
@@ -348,6 +356,7 @@ impl Instr {
             | Instr::Nop
             | Instr::ReturnCode { .. }
             | Instr::Branch { .. }
+            | Instr::AccountAccess { .. }
             | Instr::PopMemory { .. }
             | Instr::Unimplemented { .. } => {}
         }
@@ -1351,6 +1360,10 @@ impl ControlFlowGraph {
 
             Instr::Unimplemented { .. } => {
                 "unimplemented".into()
+            }
+
+            Instr::AccountAccess { .. } => {
+                unreachable!("Instr::AccountAccess shall never be in the final CFG")
             }
         }
     }
--- a/src/codegen/solana_accounts/account_management.rs
+++ b/src/codegen/solana_accounts/account_management.rs
@@ -14,7 +14,11 @@ use std::collections::{HashSet, VecDeque};
 /// AccountMeta array with all the accounts the constructor needs.
 pub(crate) fn manage_contract_accounts(contract_no: usize, ns: &mut Namespace) {
     let contract_functions = ns.contracts[contract_no].functions.clone();
+    let mut constructor_no = None;
     for function_no in &contract_functions {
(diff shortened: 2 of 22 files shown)


src/sema/ast.rs
```diff
@@ -1072,6 +1072,12 @@ pub enum Expression {
         array: Box<Expression>,
         index: Box<Expression>,
     },
+    NamedSubscript {
```
seanyoung, line 1075: "Subscript" makes me think of the subscript operator `[]`. I don't think this is a good name.

What do you think of calling it either `Member` or `NamedMember`?

src/sema/expression/member_access.rs
```diff
@@ -242,6 +244,27 @@ pub(super) fn member_access(
...
+                        ty: Type::Ref(Box::new(Type::Struct(StructType::AccountInfo))),
+                        array: Box::new(expr),
+                        name: id.name.clone(),
+                    })
+                } else {
+                    diagnostics.push(Diagnostic::error(
+                        id.loc,
+                        "unrecognized account".to_string(),
+                    ));
+                    Err(())
```
seanyoung, line 266: Returning `Err(())` means we stop resolving the body of this function. Can we return the `Ok(Expression..)` even though it is not valid? As an error diagnostic has been generated, codegen should never be entered for the generated ast.

seanyoung (approved):

src/codegen/cfg.rs
```diff
@@ -189,6 +189,14 @@ pub enum Instr {
     /// be removed. We only have this so we can pass evm code through sema/codegen, which is used
     /// by the language server and the ethereum solidity tests.
     Unimplemented { reachable: bool },
+    /// This instruction serves to track account accesses through 'tx.accounts.my_account'
+    /// on Solana, and has no emit implementation. It is exchanged by the proper
+    /// Expression::Subscript at solana_accounts/account_management.rs
+    AccountAccess {
```
xermicus, line 195: We really need  #1354

src/codegen/solana_accounts/account_management.rs
```diff
@@ -118,11 +131,54 @@ fn process_instruction(instr: &mut Instr, functions: &[Function], ast_no: usize)
 
         *address = None;
         *accounts = Some(metas_vector);
+    } else if let Instr::AccountAccess { loc, name, var_no } = instr {
```
xermicus, line 134: Why is this not just a `match` on `instr`?

xermicus (approved):

src/codegen/cfg.rs
LucasSte: How does this help here?
I made a hack in codegen, but you suggested features flags for LLVM. Would they encompass codegen too?

src/codegen/solana_accounts/account_management.rs
LucasSte: I though about that too, but as I only had two cases I considered a match unnecessary.

src/sema/expression/member_access.rs
LucasSte: Should we have an `Expression::Unresolved`?

LucasSte: I tried returning a fake expression, but now we stop resolving the file.

LucasSte: > I tried returning a fake expression, but now we stop resolving the file.

This is not what happens. I analyzed the issue and if I do not return an error, the diagnostics are not propagated to Namespace and they are not printed.

LucasSte: If I write `assert(tx.accounts.unrecognized_account == address(this))` and I don't return `Err()`, the diagnostic is ignored here:

https://github.com/hyperledger/solang/blob/a1913b695374e869ee2714bd0c3dd7886545bcf8/src/sema/builtin.rs#L962-L975

src/sema/ast.rs
LucasSte: Changed it to `NamedMember`

PR #1390, opened 2023-06-24
Substrate: Implement `delegatecall()`

xermicus:
Required by upgradable proxy patterns.


docs/language/interface_libraries.rst
```diff
@@ -37,9 +37,9 @@ When writing libraries there are restrictions compared to contracts:
 .. note::
 
     When using the Ethereum Foundation Solidity compiler, library are a special contract type and libraries are
-    called using `delegatecall`. Parity Substrate has no ``delegatecall`` functionality so Solang statically
```
xermicus, line 40: @seanyoung WDYT, now that we have that should we change it to align with with `solc`? I think that inlining them makes much more sense economically anyways. However, we should also try do be as close to EVM Solidity as possible; are there further implications in-lining them?

src/emit/substrate/target.rs
```diff
@@ -967,46 +967,126 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
-            ]
-        )
-        .try_as_basic_value()
-        .left()
-        .unwrap()
-        .into_int_value();
+        let ret = match call_type {
+            ast::CallTy::Regular => {
+                // FIXME: Using the namespace value type, this comment is no longer valid?
+                // balance is a u128
```
seanyoung, line 985: Yes, just get rid of it

src/emit/substrate/target.rs
```diff
@@ -967,46 +967,126 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
+                ret
+            }
+            ast::CallTy::Delegate => {
+                // delegate_call asks for a code hash instead of an address
+                let hash_len = i32_const!(32); // FIXME: This is configurable like the address length
+                let code_hash_out_ptr = binary.builder.build_array_alloca(
+                    binary.context.i8_type(),
+                    hash_len,
+                    "code_hash_out_ptr",
+                );
```
seanyoung, line 1019: Can we use the scratch buffer for this and avoid the alloca?

seanyoung: Having said that, alloca is cheap

src/emit/substrate/target.rs
```diff
@@ -967,46 +967,126 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
+                let done_block = binary.context.append_basic_block(function, "done_block");
+                binary.builder.build_conditional_branch(
+                    code_hash_found,
+                    call_block,
+                    not_found_block,
+                );
 
-        log_return_code(binary, "seal_call", ret);
+                binary.builder.position_at_end(not_found_block);
+                self.log_runtime_error(binary, "not a contract account".to_string(), Some(loc), ns);
```
seanyoung, line 1059: Maybe include delegatecall in the message?

seanyoung (approved):

integration/substrate/UpgradeableProxy.sol
```diff
@@ -0,0 +1,147 @@
...
+ * The success and return data of the delegated call will be returned back to the caller of the proxy.
+ */
+abstract contract Proxy {
+    uint32 constant FORWARD_INPUT = 1;
+    uint32 constant TAIL_CALL = 4;
+
+    /**
+     * @dev Delegates the current call to `implementation`.
+     *
+     * This function does not return to its internal call site, it will return directly to the external caller.
```
LucasSte, line 24:
```suggestion
     * This function does not return to its internal call site. It will return directly to the external caller.
```
Comma splice.

integration/substrate/UpgradeableProxy.sol
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * @dev Fallback function that delegates calls to the address returned by `_implementation()`. Will run if no other
+     * function in the contract matches the call data.
+     */
+    fallback() external virtual {
+        _fallback();
+    }
+
+    /**
+     * @dev Fallback function that delegates calls to the address returned by `_implementation()`. Will run if call data
```
LucasSte, line 55:
```suggestion
     * @dev Fallback function that delegates calls to the address returned by `_implementation()`. It will run if call data
```

integration/substrate/delegate_call.sol
```diff
@@ -0,0 +1,33 @@
+// https://solidity-by-example.org/delegatecall/
+
+// SPDX-License-Identifier: MIT
+// pragma solidity ^0.8.17;
+
+// NOTE: Deploy this contract first
+contract Delegatee {
+    // NOTE: storage layout must be the same as contract A
```
LucasSte, line 8: Where is contract `A`?

integration/substrate/delegate_call.sol
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+contract Delegator {
+    uint public num;
+    address public sender;
+    uint public value;
+
+    function setVars(address _contract, uint _num) public payable {
+        // A's storage is set, B is not modified.
```
LucasSte, line 26: Which one is A and which one is B?

integration/substrate/delegate_call.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        // Set delegatee storage to default values and alice address
+        const gasLimit = await weight(conn, delegatee, 'setVars', [0n]);
+        await transaction(delegatee.tx.setVars({ gasLimit }, [0n]), alice);
+    });
+
+    after(async function () {
+        await conn.disconnect();
+    });
+
+    it('Executes the delegatee in the contex of the delegator', async function () {
```
LucasSte, line 35:
```suggestion
    it('Executes the delegatee in the context of the delegator', async function () {
```

docs/language/functions.rst
```diff
@@ -308,6 +308,39 @@ calling.
...
+Calling an external function using ``delegatecall``
+___________________________________________________
+
+External functions can also be called using ``delegatecall``.
+The difference to a regular ``call`` is that  ``delegatecall`` executes the callee code in the context of the caller,
+which has the following implications:
+
+* The callee will read from and write to the `caller` storage
+* ``value`` can't be specified for ``delegatecall`` (``msg.value`` stays the same in the callee)
+* ``msg.sender`` stays the same in the callee
```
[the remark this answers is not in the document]
xermicus: Yes but they are different things worth of pointing them out individually

docs/language/interface_libraries.rst
```diff
@@ -37,9 +37,9 @@ When writing libraries there are restrictions compared to contracts:
 .. note::
 
     When using the Ethereum Foundation Solidity compiler, library are a special contract type and libraries are
```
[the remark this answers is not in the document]
xermicus: I just took the existing form over without changing anything

docs/language/interface_libraries.rst
xermicus: Ah yes thanks for pointing this out. I need to give this a more close look in the near future. Ideally we can have the best of both worlds. I think if not inlined, the libraries could be instantiated in the storage initializers (or some similar mechanism).

src/emit/substrate/target.rs
xermicus: I like `alloca` better, see #1399

docs/language/interface_libraries.rst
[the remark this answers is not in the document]
xermicus: ACK, I changed it

docs/language/functions.rst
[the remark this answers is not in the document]
LucasSte: > Yes but they are different things worth of pointing them out individually

I don't understand why they are different things. Would you mind writing a few words about that in the docs?
I am a completely oblivious to Substrate, so someone who is only starting with it might wonder the same thing as I do.

docs/language/interface_libraries.rst
```diff
@@ -36,10 +36,10 @@ When writing libraries there are restrictions compared to contracts:
 
 .. note::
 
-    When using the Ethereum Foundation Solidity compiler, library are a special contract type and libraries are
-    called using `delegatecall`. Parity Substrate has no ``delegatecall`` functionality so Solang statically
-    links the library calls into your contract code. This does make for larger contract code, however this
-    reduces the call overhead and make it possible to do compiler optimizations across library and contract code.
+    When using the Ethereum Foundation Solidity compiler, libraries are a special contract type are
```
LucasSte, line 39:
```suggestion
    When using the Ethereum Foundation Solidity compiler, libraries are a special contract type and are
```

LucasSte (approved):

docs/language/functions.rst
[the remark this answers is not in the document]
xermicus: (The first sentence talks about `msg.value`, not `msg.sender`).

PR #1392, opened 2023-06-26
add contract author and version flags

salaheldinsoliman:
This PR implements a way to specify the author and version of all contracts via the CLI or via "solang.toml"

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -154,6 +154,17 @@ Options:
 \-\-wasm-opt
    wasm-opt passes for Wasm targets (0, 1, 2, 3, 4, s or z; see the wasm-opt help for more details).
 
+\-\-contract-authors
+   Specify authors for all contracts. If a `@author` tag is present, it will override this argument for the targeted contract.
+   For specifying multiple authors, use this format: `--contract-authors author1,author2,..`
+
+  .. note::
+
+      This will only affect the metadata in case of substrate target.
+
+\-\-version
+   Specify contracts version. According to `semver <https://semver.org/>`_ A normal version number must take the form X.Y.Z where X, Y, and Z are non-negative integers, and must not contain leading zeroes.
+
 .. warning::
 
     If multiple Solidity source files define the same contract name, you will get a single
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -163,7 +163,7 @@ Options:
       This will only affect the metadata in case of substrate target.
 
 \-\-version
-   Specify contracts version. According to `semver <https://semver.org/>`_ A normal version number must take the form X.Y.Z where X, Y, and Z are non-negative integers, and must not contain leading zeroes.
+   Specify contracts version. According to `semver <https://semver.org/>`_, a normal version number must take the form X.Y.Z where X, Y, and Z are non-negative integers, and must not contain leading zeroes.
 
 .. warning::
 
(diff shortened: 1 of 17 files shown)


docs/running.rst
```diff
@@ -154,6 +154,16 @@ Options:
...
 
+\-\-contract-authors
+   Specify authors for all contracts. If a `@author` tag is present, it will override this argument for the targeted contract.
+
+  .. note::
+
+      This will only affect the metadata in case of substrate target.
+
+\-\-version
+   Specify contracts version.
```
LucasSte, line 165:
```suggestion
   Specify contracts' version.
```

LucasSte (approved):

docs/running.rst
```diff
@@ -154,6 +154,16 @@ Options:
 \-\-wasm-opt
    wasm-opt passes for Wasm targets (0, 1, 2, 3, 4, s or z; see the wasm-opt help for more details).
 
+\-\-contract-authors
+   Specify authors for all contracts. If a `@author` tag is present, it will override this argument for the targeted contract.
+
+  .. note::
+
+      This will only affect the metadata in case of substrate target.
```
seanyoung, line 162: Then the command line option should not be allowed

seanyoung (changes requested):
This is broken.

```
$ cargo run compile --target substrate --contract-authors foo,bar examples/substrate/flipper.sol 
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/solang compile --target substrate --contract-authors foo,bar examples/substrate/flipper.sol`
No input files specified, please specifiy them in solang.toml or in command line

File: src/bin/solang.rs [hunk 3]
```diff
@@ -241,6 +242,21 @@ fn compile(compile_args: &Compile) {
...
+        let authors = if let Some(authors) = &compile_args.package.authors {
+            if target == Target::Solana {
+                eprintln!("warning: the `authors` flag will be ignored for solana target")
+            }
+            authors.clone()
+        } else {
+            vec!["unknown".to_string()]
+        };
+
+        let version = if let Some(version) = &compile_args.package.version {
```
xermicus, line 254: Nit, can use `unwrap_or`?

docs/running.rst
```diff
@@ -154,6 +154,16 @@ Options:
...
+
+\-\-version
```
xermicus, line 164: This should mention that a specific format is expected (also which one).

src/lib.rs
```diff
@@ -158,7 +158,14 @@ pub fn compile(
         if contract.instantiable {
             let code = contract.emit(&ns, &opts);
 
-            let (abistr, _) = abi::generate_abi(contract_no, &ns, &code, false);
+            let (abistr, _) = abi::generate_abi(
+                contract_no,
+                &ns,
+                &code,
+                false,
+                &vec!["unknown".to_string()],
```
xermicus, line 166: Could use the default values here. Also it'd be nice if this would work in the library version too.

xermicus (commented):
LGTM

src/bin/solang.rs
salaheldinsoliman: It looks better, but it wants to take ownership of `compile_args.package.version`. Then inside `unwrap_or`, it needs to return a reference, so I have to use a let statement beforehand. E.G the final product would look like this: 
`let default_version = "0.0.1".to_string();
 let version = compile_args.package.version.as_ref().unwrap_or(&default_version);`
 So I think `if let some` works better here

xermicus: Not terribly important, the `if let` is fine too

seanyoung (approved):

xermicus (approved):
Please include what versioning schema is expected in the docs

PR #1393, opened 2023-06-26
add option to set trace level

chioni16:
This allows the user to configure the verbosity of trace of messages exchanged between client and the server.
The user can set this in `User Settings` (press ctrl/cmd+,) and then searching for solidity related configuration.


seanyoung (approved):
Is there any way we can have a test for this?

xermicus (approved):

LucasSte (approved):

seanyoung:
Thank you @chioni16

PR #1394, opened 2023-06-26
Pick up unsaved changes in a file

chioni16:
Currently the language server reads the file, rather than using the source code which is sent to it over language server requests. In order for the language server to update with the latest text of the source file, one needs to save the file. 
These changes enable the language server to update as one types.


src/bin/languageserver/mod.rs
```diff
@@ -70,9 +72,12 @@ pub async fn start_server(language_args: &LanguageServerCommand) -> ! {
 impl SolangServer {
     /// Parse file
     async fn parse_file(&self, uri: Url) {
+        // if let Some(contents) = self.text_buffers.lock().await.get(&path) {
```
seanyoung, line 75: Please remove commented out lines of code

src/bin/languageserver/mod.rs
```diff
@@ -70,9 +72,12 @@ pub async fn start_server(language_args: &LanguageServerCommand) -> ! {
...
+        let mut resolver = FileResolver::new();
+        for (path, contents) in self.text_buffers.lock().await.iter() {
+            resolver.set_file_contents(path.to_str().unwrap(), contents.clone());
+        }
```
seanyoung, line 79: I do wonder if it makes more sense to keep a `FileResolver` around and keep that up to date. `FileResolver` could have methods for updating the file (or cache).

Just an suggestion.

src/bin/languageserver/mod.rs
```diff
@@ -1158,18 +1163,78 @@ impl LanguageServer for SolangServer {
...
+                        }
+                        new
+                    } else {
+                        // when no range is provided, entire file is sent in the request
+                        content_change.text
+                    };
+                }
+            }
+        }
+
```
seanyoung, line 1223: This code could really do with some tests.

seanyoung (changes requested):
This already looks very good!

src/bin/languageserver/mod.rs
```diff
@@ -1158,18 +1163,78 @@ impl LanguageServer for SolangServer {
...
+        if let Ok(path) = uri.to_file_path() {
+            if let Some(text_buf) = self.text_buffers.lock().await.get_mut(&path) {
+                for content_change in params.content_changes {
+                    *text_buf = if let Some(range) = content_change.range {
+                        let start_line = range.start.line as usize;
+                        let start_col = range.start.character as usize;
+                        let end_line = range.end.line as usize;
+                        let end_col = range.end.character as usize;
+                        
+                        // directly add the changes to the buffer when changes are present at the end of the file
```
LucasSte, line 1185:
```suggestion
                        // Directly add the changes to the buffer when changes are present at the end of the file.
```

src/bin/languageserver/mod.rs
```diff
@@ -1158,18 +1163,78 @@ impl LanguageServer for SolangServer {
...
+                            }
+
+                            if i == end_line {
+                                new.push_str(&line[end_col..]);
+                                new.push('\n');
+                            }
+                        }
+                        new
+                    } else {
+                        // when no range is provided, entire file is sent in the request
```
LucasSte, line 1217:
```suggestion
                        // When no range is provided, entire file is sent in the request.
```

LucasSte (commented):
These changes need a test. In addition, please, run `cargo clippy --workspace --tests --bins -- -D warnings` and fix the issues pointed by it.

src/bin/languageserver/mod.rs
```diff
@@ -30,6 +30,7 @@ pub struct SolangServer {
     importpaths: Vec<PathBuf>,
     importmaps: Vec<(String, PathBuf)>,
     files: Mutex<HashMap<PathBuf, Hovers>>,
+    text_buffers: Mutex<HashMap<PathBuf, String>>,
```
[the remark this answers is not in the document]
chioni16: Is something like this alright? 
```rust 
    struct Files {
        hovers: HashMap<PathBuf, Hovers>,
        text_buffers: HashMap<PathBuf, String>,
    }
    pub struct SolangServer {
        client: Client,
        target: Target,
        importpaths: Vec<PathBuf>,
        files: Mutex<Files>,
    }
```

Having both `hovers` and `text_buffers` stored in the same `HashMap` may require use of `Option` as they are not created simultaneously.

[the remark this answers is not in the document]
seanyoung: That looks great, thanks!

src/bin/languageserver/mod.rs
chioni16: I have added the tests in the same file `src/bin/languageserver/mod.rs`. 
Do I need to move them to a different location? In that case I will have to make the function that I am testing `public`. Not sure if that's okay.

src/bin/languageserver/mod.rs
```diff
@@ -1158,26 +1168,52 @@ impl LanguageServer for SolangServer {
...
 
     async fn did_change(&self, params: DidChangeTextDocumentParams) {
         let uri = params.text_document.uri;
 
+        if let Ok(path) = uri.to_file_path() {
+            if let Some(text_buf) = self.files.lock().await.text_buffers.get_mut(&path) {
+                *text_buf = params
+                    .content_changes
+                    .into_iter()
+                    .fold(text_buf.clone(), update_file_contents);
```
seanyoung, line 1190: Usually the `.clone()` here is unnecessary. `fn update_file_contents()` could take `&str` rather than `String`, and this would save having to clone the text every time.

src/bin/languageserver/mod.rs
```diff
@@ -1230,3 +1266,157 @@ fn loc_to_range(loc: &pt::Loc, file: &ast::File) -> Range {
 
     Range::new(start, end)
 }
+
+fn update_file_contents(
+    mut prev_content: String,
```
seanyoung, line 1271: How about `prev_content: &str` here

src/bin/languageserver/mod.rs
```diff
@@ -1230,3 +1266,157 @@ fn loc_to_range(loc: &pt::Loc, file: &ast::File) -> Range {
...
+    if let Some(range) = content_change.range {
+        let start_line = range.start.line as usize;
+        let start_col = range.start.character as usize;
+        let end_line = range.end.line as usize;
+        let end_col = range.end.character as usize;
+
+        // Directly add the changes to the buffer when changes are present at the end of the file.
+        if start_line == prev_content.lines().count() {
+            prev_content.push_str(&content_change.text);
+            // return format!("{}{}", prev_content, content_change.text);
```
seanyoung, line 1283: Please don't leave commented out code in your PR. However, in this case, this is the line you want uncommenting, I think.

src/bin/languageserver/mod.rs
seanyoung: That's fine. If the file gets too big it might be time to split it up.

src/bin/languageserver/mod.rs
chioni16: `fold` requires the function that is passed (in this case, `update_file_contents`) to have the same same input and output type. So, I had to use `clone` on `text_buf` to make both input and output to `update_file_contents` a `String`. https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold
I decided to use `fold` as I believe it makes the code more readable.

src/bin/languageserver/mod.rs
chioni16: `prev_content` needs to be of type `String` due to the restrictions placed by the `fold` method. `fold` requires the input and output of the function passed to it, to have both input and output of the same type (`String` here). https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold

src/bin/languageserver/mod.rs
chioni16: `FileResolver` seems to be designed for the use case where the files are just added once. 
For example, certain methods return an error when the path being added already exists. Also, `FileResolver` stores the files as a `Vec` and not `HashMap` or `HashSet` which can be used to deduplicate entries. 
So, I believe some changes need to be made to `FileResolver` before this idea can be incorporated.

seanyoung (approved):

src/bin/languageserver/mod.rs
```diff
@@ -1158,26 +1168,52 @@ impl LanguageServer for SolangServer {
     async fn did_open(&self, params: DidOpenTextDocumentParams) {
         let uri = params.text_document.uri;
 
+        if let Ok(path) = uri.to_file_path() {
```
xermicus, line 1171: Is there nothing to do in the error case?

src/bin/languageserver/mod.rs
```diff
@@ -1158,26 +1168,52 @@ impl LanguageServer for SolangServer {
...
+                .insert(path, params.text_document.text);
+        }
+
         self.parse_file(uri).await;
     }
 
     async fn did_change(&self, params: DidChangeTextDocumentParams) {
         let uri = params.text_document.uri;
 
+        if let Ok(path) = uri.to_file_path() {
```
xermicus, line 1185: Same, I don't know this code so might be fine. But this is very unusual

src/bin/languageserver/mod.rs
chioni16: the `did_open` method from the trait `tower_lsp::LanguageServer` trait doesn't return anything back to the client. 
So, if there is something wrong with the `uri` that the editor (VSCode in our case) sent us, we can't report it back. 

Moreover, because the `uri` is not built by us, but by the `tower_lsp` library itself, we should hardly encounter an error  here. Looking at the implementation of the function `to_file_path`, errors are returned when the `uri` isn't a valid URI, which is unlikely to occur given that they are always constructed by the same library, from where the `to_file_path` function originates.

src/bin/languageserver/mod.rs
chioni16: same as the previous one. Also, please feel free to let me know if you had something specific in mind that we can do when `Err` is returned in this scenario. 
Because we can't notify the client/editor about this (lack of return value) and because we don't store any state between the calls, I couldn't think of anything we could do here, other than silently ignore the request.

src/bin/languageserver/mod.rs
seanyoung: From what I understand, the `uri` is parsed by the `URL` crate. Just because it is a valid `URL` or `URI` does not mean it's a valid file path. It could be any kind of `URI`. For example, if we start using the the language server in web based editor like remix, it seems unlikely that file paths are used.

Should for some reason `.to_file_path()` fail, the current code silently swallows the error. The options we have are:
 - Silently ignore the error
 - panic
 - log a message

I think it is best if we log a message in this case.

src/bin/languageserver/mod.rs
```diff
@@ -24,12 +24,17 @@ struct Hovers {
 
 type HoverEntry = Interval<usize, String>;
 
+struct Files {
```
LucasSte, line 27: This struct deserves a doc comment.

src/bin/languageserver/mod.rs
```diff
@@ -1230,3 +1266,156 @@ fn loc_to_range(loc: &pt::Loc, file: &ast::File) -> Range {
...
+        let end_line = range.end.line as usize;
+        let end_col = range.end.character as usize;
+
+        // Directly add the changes to the buffer when changes are present at the end of the file.
+        if start_line == prev_content.lines().count() {
+            prev_content.push_str(&content_change.text);
+            return prev_content;
+        }
+
+        let mut new = String::new();
```
LucasSte, line 1286: This variable deserves a better name.

LucasSte (approved):

src/bin/languageserver/mod.rs
xermicus: Ok, probably we want to log a warning then as Sean suggested above.

src/bin/languageserver/mod.rs
chioni16: I decided to rename it to `new_content` in accordance with its counterpart `prev_content`

xermicus (approved):

seanyoung:
Good stuff @chioni16, thanks!

PR #1395, opened 2023-06-26
Drop diagnostics on closing the file

chioni16:
When closing a file in VS code, all the warnings and errors remain in the list of problems. This removes all diagnostics pertaining to the close file.

### DIFF
--- a/src/bin/languageserver/mod.rs
+++ b/src/bin/languageserver/mod.rs
@@ -1179,7 +1179,7 @@ impl LanguageServer for SolangServer {
         if let Ok(path) = uri.to_file_path() {
             self.files.lock().await.remove(&path);
         }
-        
+
         self.client.publish_diagnostics(uri, vec![], None).await;
     }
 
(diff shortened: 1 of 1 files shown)


src/bin/languageserver/mod.rs
```diff
@@ -1179,6 +1179,8 @@ impl LanguageServer for SolangServer {
         if let Ok(path) = uri.to_file_path() {
             self.files.lock().await.remove(&path);
         }
+        
+        self.client.publish_diagnostics(uri, vec![], None).await;
```
seanyoung, line 1183: Wonderful!

Again, this could do with a test

xermicus (approved):

LucasSte:
Please, run `cargo fmt --all` to format your file.

src/bin/languageserver/mod.rs
chioni16: I tried writing a test using the `vscode` typescript API to open and close files. 
I am using `vscode.commands.executeCommand('workbench.action.closeActiveEditor')` to close the open tab. But this doesn't seem to emit the `didClose` event which is what the language server listens for. 
I tried registering a listener for this event in javascript as well. But, this listener did not run either. 
But manually closing the files works every time and the old diagnostics disappear when this happens.

seanyoung (approved):

LucasSte (approved):

PR #1396, opened 2023-06-26
Add markdown formatting to hover messages

chioni16:


seanyoung:
Rather than caching the rendered text in the hovers structure, is it possible to only render it on demand, i.e. the hovers contains enough information to render the text, but not the rendered text? I guess this would mean storing some sort of `enum` in the hovers structure.

chioni16:
> Rather than caching the rendered text in the hovers structure, is it possible to only render it on demand, i.e. the hovers contains enough information to render the text, but not the rendered text? I guess this would mean storing some sort of `enum` in the hovers structure.

@seanyoung  Sure, that should be possible, but with a caveat. 
Suppose we need to display the type of a variable in the hover message, here we can't store the `Type` directly in the enum.
mainly because in order to get the string representation of the type, we need the `Namespace` struct which may not be available when we want to render the whole enum as a string at the end, right before we send it back to the client. 
So, we will. have to store the string version of `Type` in the enum. Also storing the whole `Type` or `Parameter` struct can be memory intensive. 
So, all we are doing at the end is to put these individual string components together to form the final hover message that we want to send to the client.

src/bin/languageserver/mod.rs
```diff
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
+#![allow(unstable_name_collisions)]
```
xermicus, line 3: I'd prefer to use the fully qualified syntax `itertools::Itertools::intersperse(...)` instead of disabling the lint (as clippy suggests)

src/bin/languageserver/mod.rs
```diff
@@ -1208,9 +1153,14 @@ impl LanguageServer for SolangServer {
                     let range = loc_to_range(&loc, &hovers.file);
 
                     return Ok(Some(Hover {
-                        contents: HoverContents::Scalar(MarkedString::String(
-                            hover.val.to_string(),
-                        )),
+                        // contents: HoverContents::Scalar(MarkedString::String(
```
xermicus, line 1156: Commented out code is almost never kept in the source (if there is a good exception, you can explain in a comment why it should be kept instead).

LucasSte:
> Builtins have non-markdown hovers that a bit ugly.

I have descriptions for the builtins in `builtins.rs`. Is our intention to use then in the hovers description box?

chioni16:
1. Struct type definition members do not use markdown:
    I have added the markdown for this. 
2. Struct member expression has no hover:
    Hovering over struct member expression now displays its type, which is in accordance with the behaviour observed with hovers over variables. 
3. Missing hovers for for loops
    Variable names of length one sometimes don't show hovers. I believe this has to do with the position sent (0 based indexing vs 1 based indexing, perhaps?)
  
  Screenshot 2023-07-11 at 6 55 49 AM

4. Builtins have non-markdown hovers that a bit ugly.
    I believe this happens only when the `solang::sema::builtin::get_prototype` function can't find a suitable entry in `solang::sema::builtin::BUILTIN_VARIABLE:`. Maybe it's better to not display a hover message in such cases? Or display a generic message without the params, return values etc? 
5. Hovers should really say where a variable is a constant (and what value it has, if it is a constant) or contract variable.
    I have added code to display if the variable is a constant / contract variable.

  Screenshot 2023-07-11 at 8 40 11 AM

  Screenshot 2023-07-11 at 8 40 47 AM

    But I had some difficulties displaying the constant value. I decided to make use of `Namespace::var_constants`, which according to the comments

```
/// For a variable reference at a location, give the constant value
/// This for use by the language server to show the value of a variable at a location
```

But this doesn't seem to be working as expected.

chioni16:
> > Builtins have non-markdown hovers that a bit ugly.
> 
> We have descriptions for the builtins in `builtins.rs`. Is our intention to use them in the hovers description box?

We are already making use of `solang::sema::builtin::BUILTIN_VARIABLE` in `builtin.rs`. I believe this is what you were referring to.

chioni16:
> Variable names of length one sometimes don't show hovers.
 
 This may be due to the fact that the range `(start, stop)` isn't inclusive on `stop`.  This causes an issue when `start` equals `stop` and the `find` method seems to skip such intervals. 
from `rust-lapper` docs (https://docs.rs/rust-lapper/1.1.0/rust_lapper/):  
> So [start, stop) is not inclusive of the stop position for neither the queries, nor the Intervals.
 
I was able to get hovers displayed for variables of length 1 by incrementing the intervals' `stop` value by 1. 

  Screenshot 2023-07-11 at 1 07 54 PM


  But adding 1 everywhere looks a bit weird. 

TL;DR - `rust-lapper` expects half open ranges (start is included , end is excluded) but we have been passing inclusive ranges (both start and end are included). This discrepancy doesn't cause any problem unless the range is of length 1 (which happens in case of variable names of length 1).

seanyoung:
@chioni16 this is already looking much better. I do have some more observations:

image
The Solidity syntax is `uint32 x, uint32 y`. `x: uint32, y: uint32` is rust syntax (it's nicer, I agree!). There are also unnecessary spaces in the formatting of the function, and it does not include the contract name. 
```solidity
function test::bar(uint32 x, uint32 y) public returns (uint32)
```
There is also a needless newline between the doc comment and the prototype.

image
The comment/description now follows the definition, rather than before. Also this is not a function, yet we have `(  )`.

image
Wrong order of comment/description which is also getting incorrect Solidity syntax highlighting. Also unnecessary spacing.

image
What is `: c`? It's the contract then it would be better to have `address c::token`. Would be possible to include the constant value?

seanyoung:
Single letter variables don't seem to work:
image

Enums give non-markdown hovers
image

Please can you can walk through all Solidity files in docs/ and ensure everything works.

(I do realize there is stuff missing because it's not in the ast, like `import` statements, but let's make sure that what's in the ast works.)

seanyoung (approved):

src/bin/languageserver/mod.rs
```diff
@@ -390,29 +371,29 @@ impl<'a> Builder<'a> {
...
+                    stop: loc.end() + 1,
+                    val: make_code_block("bytes"),
                 });
             }
             ast::Expression::NumberLiteral { loc, ty, .. } => {
                 self.hovers.push(HoverEntry {
                     start: loc.start(),
-                    stop: loc.end(),
-                    val: ty.to_string(self.ns),
+                    stop: loc.end() + 1,
```
LucasSte, line 395: Why do all stops need a `+1`?
If all of them need, there might be a problem in the loc.

seanyoung: That's a good point. The hover are working correctly, but I don't understand why we need the `+ 1` here. @chioni16 is it possible to not have these ` + 1` every where, and if they are needed, it needs a comment explaining why they are needed.

xermicus (approved):

src/bin/languageserver/mod.rs
chioni16: Hello! This is needed because the `rust-lapper` expects a half-open range and we pass inclusive ranges. So the `+1` is needed to bridge this gap. 
More information in this comment - 
https://github.com/hyperledger/solang/pull/1396#issuecomment-1630314919

LucasSte: Thanks for explanation.
Using `+1` is everywhere is error prone and we may forget it when performing general maintenance in your code in the future. 

I suggest you to create a function `loc.to_lapper_end()`, for instance, in which you format the end location according to your needs. In that function, please, add your comment explaining why the addition is necessary.

chioni16: Thanks for the suggestion, @LucasSte ! 
I believe the `rust-lapper` dependency is used uniquely in the language server. As such, it may be better to create a function in the language server module instead of adding a method to `pt::Loc`. 
I was thinking of something like this inside the language server module. 
```rust
fn to_lapper_end(end: usize) -> usize {
    end + 1
}
```
And I can call this function wherever required. For instance, 
```rust
self.hovers.push(HoverEntry {
    start: param.loc.start(),
    stop: to_lapper_end(param.loc),
    val: make_code_block(val),
});
```
Also, any suggestions for the name of the function? I'm all ears.

chioni16: Or else, we can create a new method on the `pt::Loc` struct called `exclusive_end` (naming things is hard!)
This basically returns the "open" end (not included in the range). This way, it is not specific to `rust-lapper` and it will be generic enough that its addition to `pt::Loc` will be justified. What do you think?

LucasSte: `exclusive_end` is a good name. I was thinking of doing something under `impl Loc`, perhaps (if Rust lets you) you can even create the `impl` in the language server file.

I might be missing something here. If you are excluding the end, why are you adding not subtracting? The way I understand if your interval is `[1, 2]` and you add one you have `[1, 3]`, this means you included an end.

chioni16: Hello! 
The `end` method on `pt::Loc` returns a closed end. So, the end is part of the range. 
For instance, 
```
[ a  b  c  d  e  f  g  h  i  j  k ]
[ 0  1  2  3  4  5  6  7  8  9 ]
```
if the `Loc` represents the part of the array from `b` to `f`, the `end` method would return `5`. Inclusive range `[1, 5]`

But when I need to pass this range to the `rust-lapper` crate, it falsely assumes that the range is a half-open one and wrongly interprets the range to represent the part of the array from `b` to `e` and not `f`. 
To correct this, we would have to increment the end value by 1, so that the `rust-lapper` receives the half open range, which is what it was expecting in the first place. So, `[1, 6)`. 
Please let me know if you have any questions. 

Also, I am not sure about the `#[track_caller]` and `#[inline]` attributes on this newly added method. I copied the attributes from the `end` method. Please let me know if they are needed or can be removed.

LucasSte: Everything looks good to me.
What do you think @seanyoung?

seanyoung: I think it looks good

LucasSte (approved):

seanyoung:
Thanks @chioni16

PR #1397, opened 2023-06-26
Substrate: Implement builtin for `set_code_hash` API

xermicus:
Exposes a low-level builtin function for calling `seal_set_code_hash`. The code hash arguments is of type `bytes` intentionally (and not of type `Hash`. `bytes32` will internally be represented as an `i256`. I came to the conclusion that, given this conversion can be written in Solidity, we do and should not want to deal with that in the compiler, be in emit or in codegen or elsewhere. Instead, a more high level function, that asks for a `Hash` argument and converts that to `bytes`, should rather be implement in the substrate Solidity library (once we have it).

### DIFF
--- a/docs/language/builtins.rst
+++ b/docs/language/builtins.rst
@@ -346,6 +346,28 @@ is_contract(address AccountId) returns (bool)
 
 Only available on Substrate. Checks whether the given address is a contract address. 
 
+set_code_hash(bytes hash) returns (uint32)
+++++++++++++++++++++++++++++++++++++++++++
+
+Only available on Substrate. Replace the contracts code with the code corresponding to ``hash``.
+Implies that the new code was already uploaded, otherwise the operation fails.
+A return value of 0 indicates success; a return value of 7 indicates that there was no corresponding code found.
+
+.. note::
+
+    This is a low level function. We strongly advise consulting the underlying 
+    `API documentation <https://docs.rs/pallet-contracts/latest/pallet_contracts/api_doc/trait.Version0.html#tymethod.set_code_hash>`_ 
+    to obtain a full understanding of its implications.
+
+This functionality is intended to be used for implementing upgradeable contracts. 
+Pitfalls generally applying to writing
+`upgradeable contracts <https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable>`_ 
+must be considered whenever using this builtin function, most notably:
+
+* The contract must safeguard access to this functionality, so that it is only callable by priviledged users.
+* The code you are upgrading to must be storage compatible with the existing code.
+* Constructors and any other initializers, including initial storage value definitions, won't be executed.
+
 Cryptography
 ____________
 
--- a/tests/substrate.rs
+++ b/tests/substrate.rs
@@ -70,12 +70,12 @@ impl HostError for HostReturn {}
 
 /// Represents a contract code artifact.
 #[derive(Clone)]
-struct WasmCode {
+pub struct WasmCode {
     /// A mapping from function names to selectors.
     messages: HashMap<String, Vec<u8>>,
     /// A list of the selectors of the constructors.
     constructors: Vec<Vec<u8>>,
-    hash: [u8; 32],
+    hash: Hash,
     blob: Vec<u8>,
 }
 
@@ -235,7 +235,7 @@ impl Runtime {
         Self {
             accounts: blobs
                 .iter()
-                .map(|blob| Account::with_contract(&blob.hash, blob))
+                .map(|blob| Account::with_contract(blob.hash.as_ref(), blob))
                 .collect(),
             blobs,
             ..Default::default()
@@ -309,7 +309,7 @@ impl Runtime {
     /// Returns `None` if there is no contract corresponding to the given `code_hash`.
     fn deploy(
         &mut self,
-        code_hash: [u8; 32],
+        code_hash: Hash,
         value: u128,
         salt: &[u8],
         input: Vec<u8>,
@@ -347,6 +347,10 @@ fn read_value(mem: &[u8], ptr: u32) -> u128 {
 
 fn read_account(mem: &[u8], ptr: u32) -> Address {
     Address::try_from(&mem[ptr as usize..(ptr + 32) as usize]).unwrap()
+}
+
+fn read_hash(mem: &[u8], ptr: u32) -> Hash {
+    Hash::try_from(&mem[ptr as usize..(ptr + 32) as usize]).unwrap()
 }
 
 /// Host functions mock the original implementation, refer to the [pallet docs][1] for more information.
@@ -572,7 +576,7 @@ impl Runtime {
         salt_ptr: u32,
         salt_len: u32,
     ) -> Result<u32, Trap> {
-        let code_hash = read_account(mem, code_hash_ptr);
+        let code_hash = read_hash(mem, code_hash_ptr);
         let salt = read_buf(mem, salt_ptr, salt_len);
         let input = read_buf(mem, input_data_ptr, input_data_len);
         let value = read_value(mem, value_ptr);
@@ -782,6 +786,16 @@ impl Runtime {
             .any(|account| account.contract.is_some() && account.address == address)
             .into())
     }
+
+    #[seal(0)]
+    fn set_code_hash(code_hash_ptr: u32) -> Result<u32, Trap> {
+        let hash = read_hash(mem, code_hash_ptr);
+        if let Some(code) = vm.blobs.iter().find(|code| code.hash == hash) {
+            vm.accounts[vm.account].contract.as_mut().unwrap().code = code.clone();
+            return Ok(0);
+        }
+        Ok(7) // ReturnCode::CodeNoteFound
+    }
 }
 
 /// Provides a mock implementation of substrates [contracts pallet][1]
@@ -871,6 +885,11 @@ impl MockSubstrate {
         let mut input = self.0.data().blobs[self.0.data().account].constructors[index].clone();
         input.append(&mut args);
         self.raw_constructor(input);
+    }
+
+    /// Get a list of all uploaded cotracts
+    pub fn blobs(&self) -> Vec<WasmCode> {
+        self.0.data().blobs.clone()
     }
 
     /// Call the "deploy" function with the given `input`.
--- a/integration/substrate/set_code_hash.sol
+++ b/integration/substrate/set_code_hash.sol
@@ -1,0 +1,28 @@
+import "substrate";
+
+abstract contract Upgradeable {
+    function set_code(Hash code) external {
+        bytes _code = Hash.unwrap(code);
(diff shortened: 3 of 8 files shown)


docs/language/builtins.rst
```diff
@@ -346,6 +346,28 @@ is_contract(address AccountId) returns (bool)
...
+    `API documentation <https://docs.rs/pallet-contracts/latest/pallet_contracts/api_doc/trait.Version0.html#tymethod.set_code_hash>`_ 
+    to obtain a full understanding of its implications.
+
+This functionality is intended to be used for implementing upgradeable contracts. 
+Pitfalls generally applying to writing
+`upgradeable contracts <https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable>`_ 
+must be considered whenever using this builtin function, most notably:
+
+* The contract must safeguard access to this functionality, so that it is only callable by priviledged users.
+* The code you are upgrading to must be storage compatible with the existing code.
```
LucasSte, line 368: Isn't it clear to say that the contract's storage layout should be compatible with the new code?
Alternatively, a few words on `storage compatible` here would help understanding.

tests/substrate.rs
```diff
@@ -782,6 +786,16 @@ impl Runtime {
...
     }
+
+    #[seal(0)]
+    fn set_code_hash(code_hash_ptr: u32) -> Result<u32, Trap> {
+        let hash = read_hash(mem, code_hash_ptr);
+        if let Some(code) = vm.blobs.iter().find(|code| code.hash == hash) {
+            vm.accounts[vm.account].contract.as_mut().unwrap().code = code.clone();
+            return Ok(0);
+        }
+        Ok(7) // ReturnCode::CodeNoteFound
```
LucasSte, line 797: The code could come from the enum itself.

xermicus: I agree. That enum is defined in the contracts pallet. We could either pull that dependency in or maintain our own version. However there are many more places where it's done like that, and I'm heading off to decoded, so I'd rather clean this up another time.

docs/language/builtins.rst
xermicus: I think saying that the new _code_ must be storage compatible is technically more correct than just saying _contract_. The _contract_ will stay at the same address. `set_code_hash` doesn't change the contract, but merely its _code_.

However I agree that we could state what this exactly means.I added a link to the OpenZepplin docs.

seanyoung:
> Exposes a low-level builtin function for calling `seal_set_code_hash`. The code hash arguments is of type `bytes` intentionally (and not of type `Hash`, so that it matches the the pointer expected by the API function. `Hash`, which is just an alias to `bytes32`, will internally be represented as an `i256`. I came to the conclusion that, given this type conversion can easily be written in Solidity, we do and should not want to deal with that in the compiler, be it in emit or in codegen or elsewhere. Instead, a more high level function, that asks for a `Hash` argument and converts that to `bytes`, should rather be implement in the substrate Solidity library (once we have it). Additionally, if some parachain decides to use a hash longer than 32bytes, using just a `bytes` pointer will still work.

We have a `Hash` type specifically for representing a `Hash` value on Substrate, but you don't want to use it because it needs conversion functions in emit. Either we consistently use `Hash` or we just remove `Hash` if it is no good, I think.

I think we need to fix `bytes32` in emit to be an fixed array of `i8` in llvm. Also, you could define `Hash` to be  `uint8[32]` rather than `bytes32`. That will allow you to easily define it larger than 32 bytes.

xermicus:
> you don't want to use it because it needs conversion functions in emit. 

Yes that is one reason I don't want to use it. The other one is that I don't see why we shouldn't just accept `bytes` if this is exactly what the runtime API function expects anyways? Why not offloading this into Solidity land to keep the compiler lean?

> Either we consistently use Hash or we just remove Hash if it is no good, I think.

Yes I want to get rid of this `Hash` builtin type after the June release. I want to implement it in Solidity.

xermicus:
`I think we need to fix bytes32 in emit to be an fixed array of i8 in llvm. Also, you could define Hash to be uint8[32] rather than bytes32. That will allow you to easily define it larger than 32 bytes.`

These are very good poins, however it would blow up the scope of this PR. Is this the case only for `bytes32`?

seanyoung:
> > you don't want to use it because it needs conversion functions in emit.
> 
> Yes that is one reason I don't want to use it. The other one is that I don't see why we shouldn't just accept `bytes` if this is exactly what the runtime API function expects anyways? Why not offloading this into Solidity land to keep the compiler lean?

The runtime API wants a pointer, it does not need a dynamic array. By "offloading this into Solidity" now the generated llvm does a malloc of a vector, which means it has to do more work, larger code size and more gas used.

> > Either we consistently use Hash or we just remove Hash if it is no good, I think.
> 
> Yes I want to get rid of this `Hash` builtin type after the June release. I want to implement it in Solidity.

My problem with this is that you're introducing a new builtin `set_code_hash()` for this release which will change with the next release. Not great for language stability.

xermicus:
The length of this hash _is_ dynamic. By making this builtin accepting `bytes` we already anticipate for different hash lengths in different runtime configurations. Why do you think we should change the builtin the next release? But it the other way round; if this was a fixed size array, how would a contract dev targeting a runtime with a hash length != 32 bytes still be able to use this builtin?

Regarding the overhead from the allocation. Next to the problem that an allocation is inevitable when dealing with dynamic size types. I think we can be pragmatic here. Upgrading the contracts code hash is something done rarely. I don't see how this matters here.

seanyoung:
> The length of this hash _is_ dynamic. 

It's no more dynamic than the length of `address` or value, which are both defined to fixed-length types right now.

> By making this builtin accepting `bytes` we already anticipate for different hash lengths in different runtime configurations. Why do you think we should change the builtin the next release? Put the other way round; if this was a fixed size array of size 32, how would a contract dev targeting a runtime with a hash length != 32 bytes still be able to use this builtin?

By ensuring that `Hash` is always defined to the correct length, just like `address`.

xermicus:
Ah yes of course. You are very right, thanks! We can make that appear static from the contracts point of view. And later on introduce `--hash-length`.  I changed the implementation, WDYT?

So I see 2 follow ups from this:
* In emit, `bytesN` should always be treated as an actual bytes array and not an integer
* Implement substrate `--hash-length`

Both can wait for now, especially the `--hash-length` because I'm not aware of any popular runtime with a hash len != 32 yet. Moving the `Hash` builtin into a Solidity library is something I wanted to do anyways.

xermicus:
> The Hash builtin is still not compatible with set_code_hash(), or do we have implicit version from bytes32 to uint8[32]?

No, but does it matter?

> You can't move Hash into a Solidity library (as in source code); if you do how are you going to define it depending on the --hash-length argument?

An option would be to generate the correct type when using `solang new`.

seanyoung:
> > The Hash builtin is still not compatible with set_code_hash(), or do we have implicit version from bytes32 to uint8[32]?
> 
> No, but does it matter?

It's just poor language design. The `Hash` type should be used consistently everywhere.

I get the plan is to implement a `--hash-length` in the next release, so then any source code with June release should still work with the next release.
 
> > You can't move Hash into a Solidity library (as in source code); if you do how are you going to define it depending on the --hash-length argument?
> 
> An option would be to generate the correct type when using `solang new`.

So  `address` length and value length are gathered from the commandline or `solang.toml`, but the hash length is from an file generated by `solang new`? What is the advantage of that other than providing a footgun?

xermicus:
Well, I agree with your points. But we have a dilemma. `Hash` was chosen to be of `bytes32` because `uint8[32]` is not an elementary type. So it's not valid to declare `type Hash is uint8[32];`. It'll lead to problems in other places in the compiler. It won't be possible to wrap and unwrap neither. 

The main reason `Hash` was introduced is to be able to use that type in the ABI. However I agree that it would also make for good consistency to use it everywhere. I need to give this more thought.

seanyoung:
> Well, I agree with your points. But we have a dilemma. `Hash` was chosen to be of `bytes32` because `uint8[32]` is not an elementary type. So it's not valid to declare `type Hash is uint8[32];`. It'll lead to problems in other places in the compiler. It won't be possible to wrap and unwrap neither.

solang and solc don't allow this. We could also think about lifting this restriction. I am not sure why this solc limitation exists or what problems you mention.

xermicus:
`uint8[N]` isn't a primitive type, right? It requires allocations. If we currently just assume that user types are primitives types everywhere. And now this restriction is lifted. This will lead to problems. There might be other problems, we should think really careful about this.

In fact, I just tried this out (switching the `Hash` builtin to `uint8[32]` and there are indeed problems:

```solidity
function foo(Hash h) public pure {}
```

```
thread 'substrate_tests::builtins::hash' panicked at 'Found ArrayValue(ArrayValue { name: "", address: 0x7fb61010e3a0, is_const: false, is_const_array: false, is_const_data_array: false, is_null: false, llvm_value: "[32 x i8] %0", llvm_type: ArrayType { array_type: Type { address: 0x7fb6100b4aa0, llvm_type: "[32 x i8]" } } }) but expected PointerValue variant', /home/cyrill/.cargo/registry/src/index.crates.io-6f17d22bba15001f/inkwell-0.2.0/src/values/enums.rs:302:13
```

But regarding this PR, how should we proceed? We can change the signature of this builtin function in the future to the user type. It'll be a breaking change but not the end of the world. If the underlying type will stay the same, it'll just appear differently in the metadata. Alternatively this PR just needs to sit it out until we found a Solution to this problem.

seanyoung:
> But regarding this PR, how should we proceed? We can change the signature of this builtin function in the future to the user type. It'll be a breaking change but not the end of the world. If the underlying type will stay the same, it'll just appear differently in the metadata. Alternatively this PR just needs to sit it out until we found a Solution to this problem.

I've given it some thought. If we agree, let's proceed with the PR as-is. Clearly the `Hash` builtin needs some thought and it's the not fault of this PR.

seanyoung (approved):

xermicus:
Great, in that case we have reached consensus :-)

docs/language/builtins.rst
```diff
@@ -346,6 +346,30 @@ is_contract(address AccountId) returns (bool)
 
 Only available on Substrate. Checks whether the given address is a contract address. 
 
+set_code_hash(bytes hash) returns (uint32)
+++++++++++++++++++++++++++++++++++++++++++
+
+Only available on Substrate. Replace the contracts code with the code corresponding to ``hash``.
```
LucasSte, line 352:
```suggestion
Only available on Substrate. Replace the contract's code with the code corresponding to ``hash``.
```

docs/language/builtins.rst
```diff
@@ -346,6 +346,30 @@ is_contract(address AccountId) returns (bool)
...
+Implies that the new code was already uploaded, otherwise the operation fails.
```
LucasSte, line 353:
```suggestion
Assumes that the new code was already uploaded, otherwise the operation fails.
```

LucasSte (approved):

PR #1400, opened 2023-06-26
Runtime error strings must be LLVM constant strings instead of dynamically allocated

LucasSte:
The contract linked at #1367 calls `vector_new` too many times, because the runtime error strings are dynamically allocated during execution. This PR makes the runtime error strings constant attributes in LLVM, so they are not allocated.

In addition, this PR has these changes:
1. The test case caused a stack overflow in constant folding, so I split `fn expression` into multiple small functions.
2. `log_runtime_error` contained repeated code for both targets and has been removed from the `trait TargetRuntime`.
3. `error_message_with_loc` now adds the new line character.

### DIFF
--- a/src/codegen/constant_folding.rs
+++ b/src/codegen/constant_folding.rs
@@ -9,6 +9,7 @@ use num_bigint::{BigInt, Sign};
 use num_traits::{ToPrimitive, Zero};
 use ripemd::Ripemd160;
 use sha2::{Digest, Sha256};
+use solang_parser::pt;
 use solang_parser::pt::Loc;
 use std::ops::{Add, BitAnd, BitOr, BitXor, Div, Mul, Rem, Shl, Shr, Sub};
 use tiny_keccak::{Hasher, Keccak};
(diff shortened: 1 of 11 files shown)


src/codegen/mod.rs
```diff
@@ -1832,16 +1832,12 @@ impl From<&ast::Builtin> for Builtin {
...
-            _ => error.to_string(),
+pub(super) fn error_msg_with_loc(ns: &Namespace, error: String, loc: Option<Loc>) -> String {
+    match &loc {
+        Some(loc @ Loc::File(..)) => {
+            let loc_from_file = ns.loc_to_string(PathDisplay::Filename, loc);
+            format!("runtime_error: {error} in {loc_from_file},\n")
         }
-    } else {
-        error.to_string()
+        _ => error + ",\n",
```
seanyoung, line 1841: I know you did not introduce this, but why the trailing `,`?

src/codegen/constant_folding.rs
```diff
@@ -386,1192 +387,192 @@ fn expression(
...
-                        ty: ty.clone(),
-                        overflowing: *overflowing,
-                        left: Box::new(left.0),
-                        right: Box::new(right.0),
-                    },
-                    left.1 && right.1,
-                )
-            }
-        }
+        } => add(loc, ty, *overflowing, left, right, vars, cfg, ns),
```
seanyoung, line 390: Did you make any changes apart from factoring code into functions?

src/codegen/mod.rs
```diff
@@ -1832,16 +1832,12 @@ impl From<&ast::Builtin> for Builtin {
...
 
-pub(super) fn error_msg_with_loc(ns: &Namespace, error: &str, loc: Option<Loc>) -> String {
-    if let Some(loc) = loc {
-        match loc {
-            Loc::File(..) => {
-                let loc_from_file = ns.loc_to_string(PathDisplay::Filename, &loc);
-                format!("runtime_error: {error} in {loc_from_file}")
-            }
-            _ => error.to_string(),
+pub(super) fn error_msg_with_loc(ns: &Namespace, error: String, loc: Option<Loc>) -> String {
```
seanyoung, line 1835: Is it really worth it to change this from `&str` to `String`?

src/codegen/mod.rs
LucasSte: I guess that @salaheldinsoliman's reasoning was that there can be multiple errors from a file, so he would split them using a comma.

src/codegen/constant_folding.rs
LucasSte: I united the calculation of hashes in a single function. The match statement had different branches for each of them and I found that to be repeated code. Please, see `fn builtin_hash` on line 1279.

src/codegen/mod.rs
LucasSte: `fn log_runtime_error` is the main function that calls `error_msg_with_loc`. It was receiving a `String` as the argument, that was converted to `&str` for `error_msg_with_loc`, which would convert again `&str` to `String`. This scheme meant two allocations for the same string.

src/codegen/mod.rs
salaheldinsoliman: As far as I remember the contracts node would print (entered unreachable code) or sth like that after the error is printed, so that makes the error more readable

seanyoung: On Substrate, the strings get concatenated to one big string, while on Solana you get a vector of strings. Maybe the `,` makes more sense on Substrate that Solana. We can leave it for now

seanyoung (approved):

src/codegen/constant_folding.rs
```diff
@@ -1672,3 +673,1251 @@ fn constants_equal(left: &Expression, right: &Expression) -> bool {
         _ => false,
     }
 }
+
+fn add(
```
xermicus, line 677: These functions could use a comment explaining what they do and what their return values mean.

xermicus (approved):

PR #1402, opened 2023-06-27
Ensure that a constructor cannot be run twice on the same data account

seanyoung:
The constructor code path was not checking that the data account magic number is not populated.

The change also exposed another issue: when a pure function is called, there should be no load of data magic number, this may result in a access violation.

### DIFF
--- a/integration/solana/simple.spec.ts
+++ b/integration/solana/simple.spec.ts
@@ -9,6 +9,11 @@ describe('Simple solang tests', function () {
     it('flipper', async function () {
         let { program, storage } = await loadContract('flipper', [true]);
 
+        // make sure we can't run the constructor twice
+        await expect(program.methods.new(false)
+            .accounts({ dataAccount: storage.publicKey })
+            .rpc()).rejects.toThrow();
+
         let res = await program.methods.get().accounts({ dataAccount: storage.publicKey }).view();
 
         expect(res).toStrictEqual(true);
--- a/src/codegen/dispatch/solana.rs
+++ b/src/codegen/dispatch/solana.rs
@@ -6,10 +6,7 @@ use crate::codegen::{
     vartable::Vartable,
     Builtin, Expression, Options,
 };
-use crate::{
-    sema::ast::{Namespace, StructType, Type},
-    Target,
-};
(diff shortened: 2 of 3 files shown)


LucasSte:
> The change also exposed another issue: when a pure function is called, there should be no load of data magic number, this may result in a access violation.

This is not solved here, is it?

seanyoung:
> > The change also exposed another issue: when a pure function is called, there should be no load of data magic number, this may result in a access violation.
> 
> This is not solved here, is it?

Yes, it is. When I added the magic check for the constructor code path, suddenly I was getting access violations. I've moved the magic load to where it is needed

xermicus (approved):
I'd expect this change to break some existing codegen tests; I also don't see this fix being verified in a codegen.

integration/solana/simple.spec.ts
```diff
@@ -9,6 +9,11 @@ describe('Simple solang tests', function () {
     it('flipper', async function () {
         let { program, storage } = await loadContract('flipper', [true]);
 
+        // make sure we can't run the constructor twice
+        await expect(program.methods.new(false)
+            .accounts({ dataAccount: storage.publicKey })
+            .rpc()).rejects.toThrow();
```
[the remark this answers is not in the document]
seanyoung: Yes, this is formatted by vscode.

seanyoung:
> I'd expect this change to break some existing codegen tests; I also don't see this fix being verified in a codegen.

I've added a MockVM test

src/codegen/dispatch/solana.rs
```diff
@@ -445,6 +444,57 @@ fn add_constructor_dispatch_case(
...
+    cfg.set_basic_block(magic_bad);
+
+    cfg.add(
+        vartab,
+        Instr::ReturnCode {
+            code: ReturnCode::InvalidDataError,
+        },
+    );
+
+    cfg.set_basic_block(magic_ok);
```
LucasSte, line 496: You have here 48 lines of repeated code. What do you think of creating a function?

seanyoung: Good point, fixed

LucasSte (approved):
The Mac Arm test has failed.

PR #1405, opened 2023-06-28
Substrate: Distinct dispatch CFG for `call` and `deploy`

xermicus:
Fix an oversight form #1249. The dispatch logic can not be the same for `call` and `deploy` as constructors must _not_ be allowed to be accessed from the `call` function.

### DIFF
--- a/integration/substrate/constructor_dispatch.spec.ts
+++ b/integration/substrate/constructor_dispatch.spec.ts
@@ -3,6 +3,7 @@ import { ContractPromise } from '@polkadot/api-contract';
 import { ApiPromise } from '@polkadot/api';
 import { KeyringPair } from '@polkadot/keyring/types';
 import expect from 'expect';
+import { error } from 'console';
 
 describe('Test that the constructor can not be reached from the call function', () => {
     let conn: ApiPromise;
@@ -18,7 +19,6 @@ describe('Test that the constructor can not be reached from the call function', 
         contract = new ContractPromise(conn, contract_deployment.abi, contract_deployment.address);
         const caller_deployment = await deploy(conn, alice, 'HappyCaller.contract', 0n);
         caller = new ContractPromise(conn, caller_deployment.abi, caller_deployment.address);
-
     });
 
     after(async function () {
@@ -26,10 +26,18 @@ describe('Test that the constructor can not be reached from the call function', 
     });
 
     it('Should fail to overwrite the admin account of the target contract', async function () {
-        // "Call" the constructor
-        const input = contract.abi.constructors[0].selector;
-        let gasLimit = await weight(conn, caller, 'call', [contract.address, input]);
-        await transaction(caller.tx.call({ gasLimit }, contract.address, input), alice);
+        // Expect the caller to succeed on normal function 
+        let input = contract.abi.messages[0].selector;
+        let attempt = await query(conn, alice, caller, "call", [contract.address, input]);
+        expect(attempt.result.isOk).toStrictEqual(true);
+
+        // "Calling" the constructor should fail
+        input = contract.abi.constructors[0].selector;
+        attempt = await query(conn, alice, caller, "call", [contract.address, input]);
+        expect(attempt.result.isErr).toStrictEqual(true);
+
+        const gasLimit = await weight(conn, caller, 'call', [contract.address, input]);
+        await expect(transaction(caller.tx.call({ gasLimit }, contract.address, input), alice)).rejects.toBeDefined();
 
         // Alice must still be admin
         let admin = await query(conn, alice, contract, "boss");
--- a/integration/substrate/constructor_dispatch.spec.ts
+++ b/integration/substrate/constructor_dispatch.spec.ts
@@ -1,0 +1,39 @@
+import { createConnection, deploy, aliceKeypair, query, debug_buffer, weight, transaction, daveKeypair, } from './index';
+import { ContractPromise } from '@polkadot/api-contract';
+import { ApiPromise } from '@polkadot/api';
+import { KeyringPair } from '@polkadot/keyring/types';
+import expect from 'expect';
+
+describe('Test that the constructor can not be reached from the call function', () => {
+    let conn: ApiPromise;
+    let contract: ContractPromise;
+    let caller: ContractPromise;
+    let alice: KeyringPair;
+
+    before(async function () {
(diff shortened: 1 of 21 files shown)


seanyoung (approved):
I can't see a test for ensuring that normal functions cannot be reached through the deploy function.

xermicus:
Added this as an additional mock VM test

LucasSte:
I might be missing some context in this change. Do call and deploy have different entry points for contracts on Susbtrate? Or is it something like `fn solana_deploy`?

seanyoung:
> I might be missing some context in this change. Do call and deploy have different entry points for contracts on Susbtrate? Or is it something like `fn solana_deploy`?

On Substrate, the wasm file has two entry points: `deploy` and `call`. The first is used for calling the constructor, and the second for a function call. They should have separate dispatch functions

tests/substrate_tests/calls.rs
```diff
@@ -1041,3 +1041,26 @@ contract Flagger {
...
+fn constructors_and_messages_distinct_in_dispatcher() {
+    let mut runtime = build_solidity(
+        r##"
+        contract c {
+            constructor() {}
+            function foo() public pure {}
+        }"##,
+    );
+
+    let constructor = 0xcdbf608du32.to_be_bytes().to_vec();
```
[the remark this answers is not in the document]
xermicus: Yeah why not

src/codegen/mod.rs
```diff
@@ -242,9 +242,10 @@ fn contract(contract_no: usize, ns: &mut Namespace, opt: &Options) {
             ns.contracts[contract_no].default_constructor = Some((func, cfg_no));
         }
 
-        let dispatch_cfg = function_dispatch(contract_no, &all_cfg, ns, opt);
-        ns.contracts[contract_no].dispatch_no = all_cfg.len();
-        all_cfg.push(dispatch_cfg);
+        for dispatch_cfg in function_dispatch(contract_no, &all_cfg, ns, opt) {
+            ns.contracts[contract_no].dispatch_no = all_cfg.len();
```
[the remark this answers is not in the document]
xermicus: I agree; removed it.

xermicus:
Yes Sean is correct about that. I slightly refactored to code and also documented that. @LucasSte

integration/substrate/constructor_dispatch.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    after(async function () {
+        await conn.disconnect();
+    });
+
+    it('Should fail to overwrite the admin account of the target contract', async function () {
+        // "Call" the constructor
+        const input = contract.abi.constructors[0].selector;
+        let gasLimit = await weight(conn, caller, 'call', [contract.address, input]);
+        await transaction(caller.tx.call({ gasLimit }, contract.address, input), alice);
```
seanyoung, line 32: Why does this not fail? I would expect an `await expect(..).rejects.toThrow();` here

integration/substrate/upgradeable_proxy.spec.ts
```diff
@@ -53,14 +59,15 @@ describe('Deploy the upgradable proxy and implementations; expect the upgrade me
...
+        const selector = implV2.abi.messages.find((m: IMsg) => m.identifier === "setVersion").selector;
+        await upgrade_version(implV2.address, selector);
         counter = new ContractPromise(conn, implV2.abi, proxy.address);
 
         // Test implementation V2
         count = await query(conn, alice, counter, "count");
         expect(BigInt(count.output?.toString() ?? "")).toStrictEqual(3n);
 
-        gasLimit = await weight(conn, counter, 'dec', []);
+        //gasLimit = await weight(conn, counter, 'dec', []);
```
seanyoung, line 70: why this?

src/codegen/solana_accounts/account_management.rs
```diff
@@ -32,12 +33,12 @@ pub(crate) fn manage_contract_accounts(contract_no: usize, ns: &mut Namespace) {
...
-            &mut ns.contracts[contract_no].cfg[dispatch],
-            &ns.functions,
-            constructor,
-        );
+        let dispatch = ns.contracts[contract_no]
+            .cfg
+            .iter_mut()
+            .find(|cfg| cfg.name == SOLANA_DISPATCH_CFG_NAME)
+            .expect("dispatch CFG is always generated");
+        traverse_cfg(dispatch, &ns.functions, constructor);
```
LucasSte, line 41: I'd rather store the dispatch number in a variable to avoid the loop.

tests/substrate_tests/calls.rs
```diff
@@ -1041,3 +1041,29 @@ contract Flagger {
...
+        contract c {
+            @selector([0, 1, 2, 3])
+            constructor() {}
+
+            @selector([4, 5, 6, 7])
+            function foo() public pure {}
+        }"##,
+    );
+
+    let constructor = 0x00010203u32.to_be_bytes().to_vec();
```
seanyoung, line 1058: nitpick: `vec![0,1,2,3]`?

LucasSte (approved):

tests/substrate_tests/calls.rs
```diff
@@ -1041,3 +1041,29 @@ contract Flagger {
...
+    // Given this constructor selector works as intended
+    runtime.raw_constructor(constructor.clone());
+    // Expect calling the constructor via "call" to trap the contract
+    runtime.raw_function_failure(constructor);
+
+    let function = 0x04050607u32.to_be_bytes().to_vec();
```
seanyoung, line 1064: nitpick: `vec![4,5,6,7]`?

seanyoung (approved):

integration/substrate/constructor_dispatch.spec.ts
xermicus: I extended the test

integration/substrate/upgradeable_proxy.spec.ts
xermicus: Slipped in, thanks

src/codegen/solana_accounts/account_management.rs
xermicus: Considered it, but ultimately decided against. This iter shouldn't have substantial impact. Also I think it's nice to have just a single instance of the CFG identifier, defining it explicitly.

PR #1409, opened 2023-06-28
Fix undefined variable detection

LucasSte:
The contract I added in the tests was working properly in Solang v0.3.0, but it caused a panic in [this line](https://github.com/hyperledger/solang/blob/a84b0ad3b67a17b524ef6b7437fd4c5376833807/src/codegen/strength_reduce/expression_values.rs#L79) on the main branch. 

I couldn't identify what has changed from the release till now, but I deemed the logic for reporting undefined variables and stop optimizations erroneous.

### DIFF
--- a/src/codegen/undefined_variable.rs
+++ b/src/codegen/undefined_variable.rs
@@ -43,9 +43,10 @@ pub fn find_undefined_variables(
     }
 
     let mut all_diagnostics: Vec<Diagnostic> = diagnostics.into_values().collect();
+    let has_diagnostic = !all_diagnostics.is_empty();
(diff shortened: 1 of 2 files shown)


xermicus:
What's the root cause of this? If this is a regression we should investigate what exactly broke

seanyoung (approved):

LucasSte:
> What's the root cause of this? If this is a regression we should investigate what exactly broke

I dug into the code, but I couldn't pinpoint exactly which change caused the problem.

xermicus (approved):

PR #1411, opened 2023-07-01
Implement Goto Definition functionality in the language server

chioni16:


src/bin/languageserver/mod.rs
```diff
@@ -1,32 +1,52 @@
...
-use std::{collections::HashMap, ffi::OsString, fmt::Write, path::PathBuf};
+use std::{collections::HashMap, ffi::OsString, path::PathBuf};
 use tokio::sync::Mutex;
 use tower_lsp::{jsonrpc::Result, lsp_types::*, Client, LanguageServer, LspService, Server};
 
 use crate::cli::{target_arg, LanguageServerCommand};
 
-struct Hovers {
+type HoverEntry = Interval<usize, String>;
+type ReferenceEntry = Interval<usize, DefinitionIndex>;
```
seanyoung, line 23: What do you think of `DefinitionEntry`?

src/bin/languageserver/mod.rs
```diff
@@ -1,32 +1,52 @@
...
+    Event(usize),
+    UserType(usize),
+}
+
+type Definitions = HashMap<DefinitionIndex, (PathBuf, Range)>;
+
+struct Cache {
     file: ast::File,
     hovers: Lapper<usize, String>,
+    references: Lapper<usize, DefinitionIndex>,
```
seanyoung, line 44: What do you think of `definitions`?

src/bin/languageserver/mod.rs
```diff
@@ -1,32 +1,52 @@
...
+    Enum(usize),
+    Variant(usize, usize),
+    Contract(usize),
+    Event(usize),
+    UserType(usize),
+}
+
+type Definitions = HashMap<DefinitionIndex, (PathBuf, Range)>;
+
+struct Cache {
```
seanyoung, line 41: `FileCache` ?

src/bin/languageserver/mod.rs
```diff
@@ -163,17 +185,24 @@ impl SolangServer {
 
             let res = self.client.publish_diagnostics(uri, diags, None);
 
-            let hovers = Builder::build(&ns);
-
-            self.files.lock().await.hovers.insert(path, hovers);
+            let (caches, definitions) = Builder::build(&ns);
 
+            let mut files = self.files.lock().await;
+            for (f, c) in ns.files.iter().zip(caches.into_iter()) {
```
seanyoung, line 191: You should skip over builtin files like `solana` and `polkadot`.

src/bin/languageserver/mod.rs
```diff
@@ -326,36 +376,45 @@ impl<'a> Builder<'a> {
...
+                let fields = itertools::Itertools::intersperse(
+                    event.fields.iter().map(|field| {
+                        format!(
+                            "\t{}{}{}",
+                            field.ty.to_string(self.ns),
+                            if field.indexed { " indexed " } else { " " },
+                            field.name_as_str()
+                        )
+                    }),
+                    ",\n".to_string(),
```
seanyoung, line 392: Surely that's equivalent to:
```rust
                let fields = event.fields.iter().map(|field| {
                        format!(
                            "\t{}{}{}",
                            field.ty.to_string(self.ns),
                            if field.indexed { " indexed " } else { " " },
                            field.name_as_str()
                        )
                    }).join(",\n");
```

src/bin/languageserver/mod.rs
```diff
@@ -1282,6 +1726,77 @@ fn loc_to_range(loc: &pt::Loc, file: &ast::File) -> Range {
     Range::new(start, end)
 }
 
+fn get_file_no_from_loc(loc: &pt::Loc) -> Option<usize> {
+    if let pt::Loc::File(fno, _, _) = loc {
+        Some(*fno)
+    } else {
+        None
+    }
+}
```
seanyoung, line 1735: This is already implemented with `loc.try_file_no()` and `loc.file_no()`

src/bin/languageserver/mod.rs
```diff
@@ -1282,6 +1726,77 @@ fn loc_to_range(loc: &pt::Loc, file: &ast::File) -> Range {
...
+
+fn remove_reference(ty: Type) -> Type {
+    match ty {
+        Type::Ref(ty) => *ty,
+        Type::StorageRef(_, ty) => *ty,
+        ty => ty,
+    }
+}
```
seanyoung, line 1756: `ty.deref_any()`

src/bin/languageserver/mod.rs
```diff
@@ -1237,29 +1642,32 @@ impl LanguageServer for SolangServer {
     }
 
     async fn hover(&self, hverparam: HoverParams) -> Result<Option<Hover>> {
+        self.client
+            .log_message(MessageType::INFO, "hover executed!")
+            .await;
```
seanyoung, line 1647: Is this useful?

chioni16: I left it by mistake, I have removed it now. Thanks!

src/bin/languageserver/mod.rs
chioni16: Thanks! Replaced it with `deref_any` function

src/bin/languageserver/mod.rs
```diff
@@ -36,6 +63,7 @@ pub struct SolangServer {
     importpaths: Vec<PathBuf>,
     importmaps: Vec<(String, PathBuf)>,
     files: Mutex<Files>,
+    definitions: Mutex<Definitions>,
```
seanyoung, line 66: Here you have a single, global vector of definitions. That's not correct, I think. Say you have two files:
```solidity
contract X {
    struct S { int f1; }
    constructor(S s) {} 
}
```
and 
```solidity
contract X {
    struct S { int64 f1; }
    constructor(S s) {} 
}
```
Now you have two files that have definitions `DefinitionIndex::Struct(0)`. I think there should be a list of definitions per file.

chioni16: Hello @seanyoung ! 
In this case, won't the two contracts have different ids (i.e, they will be located at different indices of `Namespace::contracts`)? 
Also, this is the definition of the type `Definitions`. 
```rust
type Definitions = HashMap<DefinitionIndex, (PathBuf, Range)>;
```
It's a mapping from the `DefinitionIndex` (which will be unique for each "code object" in the namespace) to a tuple of filename where the "code object" is defined and its location within the file. 

So, we are in a way keeping track of the file information where something is defined, just in a different way. 
I chose to do it this way as the most frequent operation on `Definitions` is to look them up using their `DefinitionIndex` to get the file and location within the file where they are defined. So, I decided that grouping them by files(which is what we do with `hovers`, `references`) makes less sense for `definitions`.

seanyoung: > In this case, won't the two contracts have different ids (i.e, they will be located at different indices of `Namespace::contracts`)?

Since they are different files, and do not import each other, they are separate `Namespace` structures.

chioni16: The `definitions` field is now part of `Files`. 
But this leads to another problem when the file is closed. This triggers the `did_close` handler function where we remove the entry corresponding to the closed file from `Files::cache`. 
But there can be live references in other files whose definitions are defined in the closed file. This is an issue. 
So, we either don't remove entries on file close, or we need to implement some kind of garbage collector that removes the entries only when we can confirm that there are no references pointing to the definitions in the closed file.

seanyoung: You are right, this is a problem.

In fact if you open a file, then all the files this file imports are added to cache, but only removed if the files are opened and then closed. 

One solution is to only add the file which was requested in `fn parse_file` to the cache, and keep all its definitions in the same `FileCache` struct.

chioni16: Hello @seanyoung ! 
I have been thinking about this problem .
1. The path of the file where something is defined should be part of `DefinitionIndex` (which is not the case, as of now).  This is because the role of `DefinitionIndex` is to uniquely identify the location where something is defined and as it stands now, `DefinitionIndex` fails to do so when files belonging to multiple namespaces are opened. 
Making the filepath part of `DefinitionIndex` will help solve this issue. 
File path not being part of `DefinitionIndex` requires us to store the filepath in a tuple and makes things awkward down the line. 
2. If we make the above change, then we don't need to make `definitions` field part of `Files`. This will enable a "flat" lookup as opposed to going through `FileCache` first. This can help with the above issue that occurs when a file is closed. (https://github.com/hyperledger/solang/pull/1411#discussion_r1278318147)
By making the structs not be part of `Files`, we are free to handle the removal of definitions independently from that of `Files`. 
3. The whole filepath will have to be stored in the `DefinitionIndex` as of now. Using numbers to represent files will not work as the same number can point to two or more different files when multiple namespaces are involved. 
But I believe we should be able to handle this in the future with a layer of indirection that properly maps numbers to filepaths, despite the presence of multiple namespaces.

seanyoung: Hi @chioni16
That makes a lot of sense, sounds good to me. It's fine to start the filepath in the `DefinitionIndex`. We can think of optimizing this later. I would expect there is a rust crate to help with this, but I have not been able to find one this morning.

chioni16:
During the course of implementation, I discovered a few issues that need changes to be made to the parser or new things to be added altogether. This is a non-exhaustive list of such things. New entries will be added as and when discovered. This also explains some of the decisions taken as to how the data is stored in the `SolangServer` struct. 

    The language server currently stores some of the data grouped by the file to which the data belongs (`Files` struct). Other data (`Definitions`) is not grouped by file due to problems faced during cleanup, but is stored as a "global" field which is common to all files. (https://github.com/hyperledger/solang/pull/1411#discussion_r1278318147).
    Files from multiple namespaces can be open at any time in VSCode. But compiler currently works on the granularity of a a namespace, i.e, all the analyses + code generated is for the whole namespace. 
    So, we will need some way to update data that is part of the language server between calls to the `parse_file` method that provides new information for a namespace. 
1. Propagate changes made to a file to all the files that depend on the file.
    This requires a data structure that shows import dependencies between different files in the namespace. 
2. Need a way to safely remove stored `Definitions` that are no longer used by any of the `References`

src/bin/languageserver/mod.rs
```diff
@@ -15,42 +17,92 @@ use solang_parser::pt;
...
     },
     Client, LanguageServer, LspService, Server,
 };
 
 use crate::cli::{target_arg, LanguageServerCommand};
 
-struct Hovers {
-    file: ast::File,
-    hovers: Lapper<usize, String>,
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
seanyoung, line 38: Could use a comment explaining what it is (and what the `usize` is)

src/bin/languageserver/mod.rs
```diff
@@ -650,17 +874,49 @@ impl<'a> Builder<'a> {
...
+                    loc.file_no(),
+                    HoverEntry {
+                        start: loc.start(),
+                        stop: loc.exclusive_end(),
+                        val: make_code_block(ty.to_string(self.ns)),
+                    },
+                ));
+
+                let t = expr.ty().deref_any().clone();
+                if let Type::Struct(StructType::UserDefined(_)) = t {
```
seanyoung, line 890: I don't think the clone is necessary.
```rust
                if let Type::Struct(StructType::UserDefined(_)) = expr.ty().deref_any() {
```

seanyoung (approved):
Looks great - please address the comments if they make sense

src/bin/languageserver/mod.rs
```diff
@@ -15,42 +17,92 @@ use solang_parser::pt;
...
-    hovers: HashMap<PathBuf, Hovers>,
+    caches: HashMap<PathBuf, FileCache>,
     text_buffers: HashMap<PathBuf, String>,
 }
 
+// The language server currently stores some of the data grouped by the file to which the data belongs (Files struct).
+// Other data (Definitions) is not grouped by file due to problems faced during cleanup,
+// but is stored as a "global" field which is common to all files.
+//
+// When the file is closed. This triggers the `did_close` handler function
```
LucasSte, line 81:
```suggestion
// Closing the file triggers the `did_close` handler function
```

src/bin/languageserver/mod.rs
```diff
@@ -15,42 +17,92 @@ use solang_parser::pt;
...
+// where we remove the entry corresponding to the closed file from Files::cache.
+// If the definitions are part of `FileCache`, they are also removed with the rest of `FileCache`
+// But there can be live references in other files whose definitions are defined in the closed file.
+//
+// Files from multiple namespaces can be open at any time in VSCode.
+// But compiler currently works on the granularity of a a namespace,
```
LucasSte, line 87:
```suggestion
// But compiler currently works on the granularity of a namespace,
```

src/bin/languageserver/mod.rs
```diff
@@ -15,42 +17,92 @@ use solang_parser::pt;
...
+// i.e, all the analyses + code generated is for the whole namespace.
```
LucasSte, line 88: You said that the compiler works on the granularity (i.e. individual pieces), but then you explain that the analysis and code generated is for the whole namespace. Did you mean entirety instead of granularity?

src/bin/languageserver/mod.rs
```diff
@@ -15,42 +17,92 @@ use solang_parser::pt;
...
+//
+// So, we will need some way to update data that is part of the language server between calls to
```
LucasSte, line 90:
```suggestion
// So, we will need some way to update data that part of the language server between calls to
```

src/bin/languageserver/mod.rs
```diff
@@ -15,42 +17,92 @@ use solang_parser::pt;
...
+// the parse_file method that provides new information for a namespace.
+//
+// 1. Propagate changes made to a file to all the files that depend on the file.
```
LucasSte, line 93:
```suggestion
// 1. Propagate changes made to a file to all the files that depend on that.
```

src/bin/languageserver/mod.rs
```diff
@@ -377,41 +498,83 @@ impl<'a> Builder<'a> {
         }
     }
 
-    // Constructs lookup table(messages) by traversing over the expressions and storing
-    // the respective expression type messages in the table.
+    // Constructs lookup table by traversing over the expressions and storing
```
LucasSte, line 501:
```suggestion
    // Constructs lookup table by traversing the expressions and storing
```

src/bin/languageserver/mod.rs
LucasSte: Another alternative would be:
```Rust

if matches!(Type::Struct(StructType::UserDefined(_)), expr.ty().deref_any()) {
```

src/bin/languageserver/mod.rs
```diff
@@ -685,95 +941,125 @@ impl<'a> Builder<'a> {
...
-                    // modifiers do not have mutability, bases or modifiers itself
-                    let fnc = &self.ns.functions[*function_no];
-                    let mut msg_tg = render(&fnc.tags[..]);
-                    if !msg_tg.is_empty() {
-                        msg_tg.push_str("\n\n");
-                    }
 
-                    let params = fnc.params.iter().map(|parm| format!("{} {}", parm.ty.to_string(self.ns), parm.name_as_str())).join(", ");
+            ast::Expression::ExternalFunction { loc, address, function_no, .. } => {
+                // modifiers do not have mutability, bases or modifiers itself
```
LucasSte, line 1002:
```suggestion
                // modifiers do not have mutability, bases or modifiers themselves
```

src/bin/languageserver/mod.rs
chioni16: We need `clone` here as we need to pass the type to the `DefinitionType::Field` in line 898. 

Trying to move `clone` to line 898 instead of its current location (line 889), as shown below, leads to another error. 
```rust
let t = expr.ty().deref_any();
if let Type::Struct(StructType::UserDefined(_)) = t {
    self.references.push((
        loc.file_no(),
        ReferenceEntry {
            start: loc.start(),
            stop: loc.exclusive_end(),
            val: DefinitionIndex {
                def_path: Default::default(),
                def_type: DefinitionType::Field(t.clone(), *field),
            },
        },
    ));
}
```

```
error[E0716]: temporary value dropped while borrowed
   --> src/bin/languageserver/mod.rs:889:25
    |
889 |                 let t = expr.ty().deref_any();
    |                         ^^^^^^^^^            - temporary value is freed at the end of this statement
    |                         |
    |                         creates a temporary value which is freed while still in use
890 |                 if let Type::Struct(StructType::UserDefined(_)) = t {
    |                                                                   - borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value
```
So, in order to prevent this error and to avoid calling the `expr.ty()` twice, I think `clone` is required.

seanyoung: Yes, you are right @chioni16. I could not get rid of the clone either (without making the code worse).

xermicus (approved):

src/bin/languageserver/mod.rs
chioni16: I was going for the following meaning of "granularity" here. 
> the scale or level of detail in a set of data.

LucasSte: That makes sense. Perhaps you need to rephrase what you wrote after the `i.e` then. It is a little confusing.

src/bin/languageserver/mod.rs
```diff
@@ -178,22 +243,31 @@ impl SolangServer {
...
             res.await;
         }
     }
 }
 
 struct Builder<'a> {
-    hovers: Vec<HoverEntry>,
+    hovers: Vec<(usize, HoverEntry)>,
+    definitions: Definitions,
+    references: Vec<(usize, ReferenceEntry)>,
```
LucasSte, line 265: Please, add a comment explaining what the `usize` means in hovers and references.

LucasSte (approved):

src/bin/languageserver/mod.rs
chioni16: Hello!
I decided to remove the line as it didn't add any more information and was redundant. Thanks!

seanyoung:
Thank you @chioni16 !

PR #1413, opened 2023-07-02
Ensure that doccomments still work when a function has an annotation

seanyoung:
Annotations stop doc tags from working on functions

```solidity
contract c {
      /// comment is free
      @selector([1,2,3,4])
      function foo() public {}
}
```

### DIFF
--- a/src/sema/mod.rs
+++ b/src/sema/mod.rs
@@ -442,7 +442,10 @@ fn collect_annotations_doccomments<'a>(
 
             for part in &contract.parts {
                 match part {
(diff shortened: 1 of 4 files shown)


xermicus (approved):

LucasSte (approved):
Should annotations for `@signer`, `@payer`  and so on accept doc comments?

seanyoung:
> Should annotations for `@signer`, `@payer` and so on accept doc comments?

Yes, it should.

```solidity
contract c {
      /// Comment me this 
      @signer(foo)
      function func() public {}
}
```

Issue #1398, opened 2023-06-26
Rename Substrate to Polkadot


xermicus:
I'm thinking about renaming the Substrate target to Polkadot.

* Polkadot is a much stronger brand than Substrate. I don't have evidence for this, but I'd bet that way more people (especially not very into the Polkadot ecosystem, which part of our target audience) know the term Polkadot than Substrate.
* The contracts pallet can be seen as "Polkadots native contracts"
* Substrate is a dying brand (the brand, not the technology). Chances are that it'll just be something like "[Polkadot SDK](https://forum.polkadot.network/t/psa-parity-is-currently-working-on-merging-the-polkadot-stack-repositories-into-one-single-repository/2883/13)" or something similar in the future anyways.
* Highly unlikely that there will be multiple popular contract runtimes, which we'd both support, in the foreseeable future.
* The Solana backend is also just called Solana and not sealevel or SPL or anchor or SBF

seanyoung:
Any rename should be complete, including file names/paths

xermicus:
Yes you are absolutely right; didn't write that explicitly but its what I meant. It's not just about renaming the target in the CLI, but consistently switching out these two terms everywhere.
* Source code, including tests
* File names, paths, etc
* Occurances in the CLI and config TOML
* Documentation
* [This repo](https://github.com/hyperledger/solang-substrate-ci) comes to my mind, but not sure if that's worth the struggle (after all it's a fork of the _substrate_\_contracts_node

However, we should be careful with a blind search and replace, as there might be specific occurances where "Substrate" is still the correct term to use.

seanyoung:
The builtin import `import "substrate";` will also need to be renamed. I think we should do this rename sooner rather than later, to reduce user pain.

Harsh06-afk:
Hey there,
I would like to work on this issue, can you provide me with some more references to start with.
I would like to join community can you provide relevant links.

xermicus:
> Hey there, I would like to work on this issue, can you provide me with some more references to start with. I would like to join community can you provide relevant links.

https://discord.gg/hyperledger

I think this needs some coordination. It involves changing a bunch of things out of the scope of PRs. Also it'll require a massive PR changing many lines, causing many conflicts if not merged timely.

PR #1414, opened 2023-07-02
Rename Substrate to Polkadot

xermicus:
As per #1398


seanyoung:
Should we support `--target substrate` so we're backwards compatible?

```bash
$ cargo run compile --target substrate examples/polkadot/flipper.sol 
   Compiling solang-parser v0.3.1 (/home/sean/git/solang/solang-parser)
   Compiling solang v0.3.1 (/home/sean/git/solang)
    Finished dev [unoptimized + debuginfo] target(s) in 23.80s
     Running `target/debug/solang compile --target substrate examples/polkadot/flipper.sol`
error: invalid value 'substrate' for '--target <TARGET>'
  [possible values: solana, polkadot, evm]
```

xermicus:
Backwards compatibility with what do you have in mind? I'm not aware of any tooling or something like that integrating solang via the binary yet.

tests/contract_testcases/polkadot/functions/external_call_args.sol
```diff
@@ -0,0 +1,71 @@
...
+        }({
+            amountOutMin: 0,
+            path: paths,
+            to: address(this),
+            deadline: block.timestamp + deadline
+        });
+    }
+}
+// ---- Expect: diagnostics ----
+// warning: 61:20-53: conversion truncates uint256 to uint128, as value is type uint128 on target polkadot
```
LucasSte, line 71: `polkadot` is written here with a lower case p, while it is written with an upper case P elsewhere. Please, use a single form.

tests/contract_testcases/polkadot/inheritance/inherit_types_05.sol
```diff
@@ -0,0 +1,22 @@
...
+        f2
+    }
+}
+
+contract a is b, c {
+    function test(foo x) public {}
+}
+
+// ---- Expect: diagnostics ----
+// error: 1:1-6:2: contracts without public storage or functions are not allowed on Polkadot. Consider declaring this contract abstract: 'abstract contract b'
```
LucasSte, line 20: Upper case P on polkadot, while it is writing with a lower case in other diagnostics.

.gitmodules
```diff
@@ -2,5 +2,5 @@
 	path = testdata/solidity
 	url = https://github.com/ethereum/solidity
 [submodule "integration/substrate/tornado/tornado-cli"]
```
LucasSte, line 4: This line still shows substrate.

src/codegen/dispatch/polkadot.rs
LucasSte, line 1: In the lines 33 and 34 from this file, you mention `substrate_deploy_dispatch` and `substrate_call_dispatch`. Should that be renamed to Polkadot?

tests/polkadot.rs
LucasSte, line 1: You have `MockSubstrate` defined in this file. Should this be `MockPolkadot`?

LucasSte (commented):
There is an inconsistent usage of `polkadot` and `Polkadot` in the diagnostics messages. Please, chose only one option.

xermicus:
> There is an inconsistent usage of polkadot and Polkadot in the diagnostics messages. Please, chose only one option.

This is not related to this PR. I just replaced the capitalized version if it was already there. This PR took me already almost a full day of going through our entire codebase, I'm not keen on scope creeping it.

.github/workflows/test.yml
```diff
@@ -309,18 +309,18 @@ jobs:
...
     steps:
     - name: Checkout sources
       uses: actions/checkout@v3
       with:
         submodules: recursive
-      # We can't run substrate as a github actions service, since it requires
+      # We can't run substrate node as a github actions service, since it requires
       # command line arguments. See https://github.com/actions/runner/pull/1152
-    - name: Start substrate
+    - name: Start substrate contracts node
```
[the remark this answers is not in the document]
xermicus: Our CI node which is a fork of the [substrate-contracts-node](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.24.0). Saying `Polkadot node` could be mistaken for an actual relay chain node on the polkadot network.

.github/workflows/test.yml
```diff
@@ -334,28 +334,28 @@ jobs:
...
-    name: Substrate Integration test with subxt
+  polkadot-subxt:
+    name: Polkadot Integration test with subxt
     runs-on: ubuntu-22.04
     needs: linux-x86-64
     steps:
     - name: Checkout sources
       uses: actions/checkout@v2
-      # We can't run substrate as a github actions service, since it requires
+      # We can't run substrate node as a github actions service, since it requires
```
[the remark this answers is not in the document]
xermicus: Same as above

.gitmodules
xermicus: Hmm, good catch. Apparently `git mv` didn't change this, trying to just change it manually now.

tests/polkadot.rs
xermicus: I kept it because it mocks the contracts pallet which is part of `FRAME` in Substrate. So, when talking about the MockVM we are talking about some technically and not about the our polkadot compile target, that's I like to keep it for now.

src/codegen/dispatch/polkadot.rs
xermicus: I thought it doesn't matter but why not change that too for consistency

tests/polkadot.rs
seanyoung: I could really do with glossary of Polkadot terms. frame, contracts pallet, seal - the list goes on!

xermicus: Yeah from an outside world it can get overwhelming quickly

tests/contract_testcases/polkadot/functions/external_call_args.sol
seanyoung: We often do `format!("you cannot do foo on target {}", ns.target);` The `fmt::Display` for `Target` prints lower case, and I don't know why. I think this can be changed to start with an upper case letter, it is Polkadot and Solana after all, not polkadot and solana.

LucasSte: >The `fmt::Display` for `Target` prints lower case, and I don't know why.

It is written with a lower case.
https://github.com/hyperledger/solang/blob/f9230fc5ab2d957e90153fdfaaa5701f241f6add/src/lib.rs#L39-L41

seanyoung: Yes, that's the code I am referring to.

xermicus: I think what Sean meant was that he doesn't know why it's implemented like that. I'm happy to change capitalization  in a follow up PR, but not in this one.

src/lib.rs
```diff
@@ -37,7 +37,7 @@ impl fmt::Display for Target {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Target::Solana => write!(f, "solana"),
-            Target::Substrate { .. } => write!(f, "substrate"),
+            Target::Polkadot { .. } => write!(f, "polkadot"),
```
LucasSte, line 40:
```suggestion
            Target::Polkadot { .. } => write!(f, "Polkadot"),
```

LucasSte (approved):

src/lib.rs
xermicus: I'll do that as a follow up for both targets as this breaks the `contract` tests

Issue #916, opened 2022-07-13
Support `require()` messages in Substrate debug output


xermicus:
Call `seal_debug_message` in `assert_failure`, to get messages from `require()` into the debug message buffer.

xermicus:
[This is solved](https://github.com/hyperledger/solang/blob/de11ed32aebc90b73e0822171f85afc785566f12/tests/polkadot_tests/errors.rs#L80-L99) since a while ago.

PR #1415, opened 2023-07-03
Polkadot: `AssertFailure` emits revert instead trap

### LINKED ISSUE #1292: Memory OOB access in try-catch
The following contract traps the execution with a `MemoryOutOfBounds` trap code:

```solidity
contract c {
    constructor() payable {}

    function test() public returns (int32) {
        int32 x = 0;
        try new other(true) {
            x = 1;
        } catch (bytes memory _c) {
            x = 2;
        }
        return x;
    }
}

contract other {
    constructor(bool rev) public {
        if (rev) {
            revert("foo");
        }
    }

    function _ext() public {}
}
```

A real substrate node will not tell about the OOB error, however the mock runtime will panic on this. See substrate test `calls::try_catch_constructor`. Uncommenting the last function call to the `test` function in this test will trigger the OOB access. 

 `cargo test --test substrate -- calls::try_catch_constructor`

xermicus:
Using the `seal_return` API with the revert flag instead of just trapping the contract execution allows to return data. This is required for further alignment with [solc](https://docs.soliditylang.org/en/latest/control-structures.html#panic-via-assert-and-error-via-require). 

Closes #1292
Unblocks #916


xermicus:
Seems like in the test, when building the same source multiple times, the contracts end up having different code. This did only happen on the CI hosts, neither on my PC nor my laptop. Also, I built the offending case on the Linux ARM host using this branch. It consistently produces the exact same binary as my PC. So this must be something going on in the MockVM, maybe because of the `FileResolver`.

tests/substrate_tests/expressions.rs
```diff
@@ -247,21 +247,19 @@ fn expressions() {
...
     let mut runtime = build_solidity(
         "
         contract test {
-            function do_test() public returns (uint){
-                uint256 val = 100;
-
-                return (val / 0);
+            uint256 divisor = 0;
+            function do_test(uint256 divisor) public returns (uint256 result){
+                result = 100 / divisor;
```
LucasSte, line 257: The argument is called `divisor` and you have a storage variable named `divisor`. This test seems a bit awkward because of the naming conflict.

xermicus: Very good catch, thanks!

integration/subxt-tests/src/cases/asserts.rs
```diff
@@ -45,10 +45,7 @@ async fn case() -> anyhow::Result<()> {
             &|t: &ContractMessageTranscoder| t.encode::<_, String>("test_assert_rpc", []).unwrap(),
         )
         .await;
-
-    if let Err(r) = res {
-        assert!(r.to_string().contains("ContractTrapped"));
-    }
+    assert!(res.is_err());
```
seanyoung, line 48: why not 
```rust
 assert!(res.unwrap_err().to_string().contains("ContractReverted"));
```
?

tests/polkadot_tests/calls.rs
```diff
@@ -35,11 +35,17 @@ fn revert() {
...
     runtime.function_expect_failure("a", Vec::new());
 
-    assert_eq!(runtime.output().len(), 0);
+    assert_eq!(
+        runtime.output().len(),
+        4 + "revert value has to be passed down the stack"
+            .to_string()
+            .encode()
+            .len()
+    );
```
seanyoung, line 49: Why not assert the contents of `runtime.output` rather than the just the length?

tests/polkadot_tests/contracts.rs
```diff
@@ -99,10 +99,7 @@ fn revert_constructor() {
     );
 
     runtime.constructor(0, Vec::new());
-
     runtime.function_expect_failure("test", Vec::new());
-
-    assert_eq!(runtime.output().len(), 0);
```
seanyoung, line 105: why not assert that the "Hello, World!" revert has been sent?

seanyoung (approved):

integration/subxt-tests/src/cases/asserts.rs
xermicus: I tried that but the error just contains `[]` somehow. However it works for the transaction just further down, where I do this assertion.

tests/polkadot_tests/calls.rs
xermicus: Good idea, added it. However I noticed that the endianness of this function selector is swapped, something to look into afterwards.

integration/subxt-tests/src/cases/asserts.rs
seanyoung: So there is a bug, right?

tests/polkadot_tests/contracts.rs
xermicus: The contract doesn't return the output back to the caller. Should we bubble it up?

src/emit/substrate/target.rs
```diff
@@ -779,25 +779,14 @@ impl<'a> TargetRuntime<'a> for SubstrateTarget {
...
+        emit_context!(binary);
 
-        binary
-            .builder
-            .build_indirect_call(asm_fn, asm, &[], "unreachable");
+        let flags = i32_const!(1).into(); // First bit set means revert
+        call!("seal_return", &[flags, data.into(), length.into()]);
 
+        // insert "unreachable" instruction; build_unreachable() tells the compiler
+        // that this code path is not reachable and may be discarded.
```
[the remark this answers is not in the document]
seanyoung: > The code path is not discarded. The instruction only signals there cannot be anything after it.

That's not true.

tests/polkadot_tests/calls.rs
seanyoung: Another bug.

xermicus: Yes. I guess we never found out until now because it still works in the try catch when the endianess issue is consistent enough.

integration/subxt-tests/src/cases/asserts.rs
xermicus: The way the execution is build for these tests, the error would contain the debug buffer output. However it's compiled with `--release` so there is no debug buffer output and that's why the error is empty. I'll change that.

tests/polkadot_tests/contracts.rs
xermicus: I think we should. Also, now we can return `Result<T, LangError>` where `T` is what the function or constructor returns, and `LangError` is `(bytes4, string)`. I can do that in this PR or in a followup

tests/polkadot_tests/calls.rs
```diff
@@ -944,102 +938,68 @@ contract Flagger {
...
-    struct Input {
-        address: [u8; 32],
-        selector: [u8; 4],
-        voyager: u32,
-        flags: Vec<CallFlags>,
-    }
+    let forward_input = 0b1u32;
+    let clone_input = 0b10u32;
+    let tail_call = 0b100u32;
+    let allow_reentry = 0b1000u32;
```
LucasSte, line 955: For clarity, you can write these as `1`, `1 << 1`, `1 << 2` and `1 << 3`.

LucasSte (approved):

src/emit/substrate/target.rs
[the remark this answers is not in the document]
LucasSte: > Well I just took over the comment as is. I rewrote the comment, @LucasSte WDYT?

That comment is great, Cyrill. Thanks!

tests/polkadot_tests/calls.rs
xermicus: Fixed in #1421

tests/polkadot_tests/calls.rs
seanyoung: I don't think `1 << 3` is necessarily clearer than `0b100`.

LucasSte: My reasoning was that I don't need to count the zeros to the right of `1` when I use shift, because `1 << 2` shows me exactly the set bit. 

Either way, it is up to Cyrill whether to change or not.

PR #1418, opened 2023-07-03
Unbreak build updating target to Solana in solana tests

seanyoung:
Merging https://github.com/hyperledger/solang/pull/1413 broke the build as `.substrate_default()` no longer exists.

### DIFF
--- a/tests/solana_tests/tags.rs
+++ b/tests/solana_tests/tags.rs
@@ -21,7 +21,7 @@ fn contract() {
             @selector([1,2,3,4,5,6,7,8])
             constructor() {}
         }"#,
-        Target::default_substrate(),
+        Target::Solana,
     );
 
     assert_eq!(ns.contracts[0].tags[0].tag, "notice");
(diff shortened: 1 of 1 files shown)


tests/solana_tests/tags.rs
```diff
@@ -184,10 +184,10 @@ fn functions() {
         }
 
         contract b {}"#,
-        Target::default_substrate(),
+        Target::Solana,
     );
 
-    assert_eq!(ns.diagnostics.len(), 3);
+    assert_eq!(ns.diagnostics.len(), 5);
```
LucasSte, line 190: Why has this changed?

xermicus (approved):

tests/solana_tests/tags.rs
seanyoung: Substrate does not like `contract b {}` without any functions and produces an error `contracts without public storage or functions are not allowed on Polkadot. Consider declaring this contract abstract: 'abstract contract b'`. The error prevents unused variable from running.

The test is looking for tags.

LucasSte (approved):

PR #1419, opened 2023-07-03
Track `import_no` in `File`

BenTheKush:
This PR:
1. Adds a `resolved_files: Vec<ResolvedFile>` field to `FileResolver`;
2. Populates it during the execution of `FileResolver.resolve_file()`;
3. Extends `FileResolver`'s API with `get_resolved_file(file_no: usize)` to allow access of this data.
 
Prior to this PR, the resolution root of each file is discarded after `parse_and_resolve` and has to be recomputed by a client (and this recomputation depends on implementation-specifics such as source root visit order during file resolution). This PR persists this data.


seanyoung:
> This PR:
> 
>     1. Adds a `resolved_files: Vec<ResolvedFile>` field to `FileResolver`;
> 
>     2. Populates it during the execution of `FileResolver.resolve_file()`;
> 
>     3. Extends `FileResolver`'s API with `get_resolved_file(file_no: usize)` to allow access of this data.
> 
> 
> Prior to this PR, the resolution root of each file is discarded after `parse_and_resolve` and has to be recomputed by a client (and this recomputation depends on implementation-specifics such as source root visit order during file resolution). This PR persists this data.

It's not clear to me with field of `ResolvedFile` you need, or how exactly you are going to use it. The only public field is `full_path`, how is that going to help you? You already have that in the `File` struct, right?

Would you mind giving some (pseudo) code that shows what you're planning to do?

BenTheKush:
@seanyoung good point, I hadn't noticed that everything was private.

My end goal is to determine the file name relative to the import root it was resolved from. So if I resolve `a/b/C.sol` inside of `/Users/ben/my/source/root/`, then I want `a/b/C.sol`. I was going to use `base` to do this but I think I misunderstood what `"base part relative to import"` means. 

It looks like I might actually want to use `import_no`, and I probably want to implement something like

```rust
impl FileResolver {
    fn get_import_path(&self, resolved: &ResolvedFile) -> &PathBuf {
        self.import_paths.get(resolved.import_no).expect("Resolved file should have a valid import number")
    }
}
```

seanyoung:
If you want the literal path from the `import "foo/bar.sol"` statement, then that is something that should be added to the `File` struct in the namespace `ns.files`:

One complication might be import maps:
```solidity
import "@openzeppelin/contracts/token/ERC721/ERC721Full.sol";
```
Do you want `@openzeppelin/contracts/token/ERC721/ERC721Full.sol` in this case?

BenTheKush:
> If you want the literal path from the `import "foo/bar.sol"` statement, then that is something that should be added to the `File` struct in the namespace `ns.files`:
> 
> One complication might be import maps:
> 
> ```solidity
> import "@openzeppelin/contracts/token/ERC721/ERC721Full.sol";
> ```
> 
> Do you want `@openzeppelin/contracts/token/ERC721/ERC721Full.sol` in this case?

In my case I won't be mutating external dependencies/projects (I'm not a solidity dev, are remappings ever used to point to a location within the same project?), so I don't have a strong preference. I'd say having the remapping information is sufficient, so yeah, having `@openzeppelin/...` would be ideal (if we were to mutate such a thing).

seanyoung:
> In my case I won't be mutating external dependencies/projects (I'm not a solidity dev, are remappings ever used to point to a location within the same project?),

mappings are always used with external dependencies from what I've seen. It is possible to make it point to a location within the same project, but I don't think this makes much sense.

> so I don't have a strong preference. I'd say having the remapping information is sufficient, so yeah, having @openzeppelin/... would be ideal (if we were to mutate such a thing).

It should be fairly simple to add another field to the `File` struct path, something like `source_path` or `solidity_path`?

BenTheKush:
I think this should work out fine, but just to sanity check...suppose I import `A.sol` twice: once with a relative import and once with an direct import:

From `RelativeImport.sol`:
```solidity
import "../A.sol";
...
```

From `DirectImport.sol`:

```solidity
import "contracts/A.sol";
...
```

My instinct is that `DirectImport.sol`'s import will resolve with respect to `FileResolver.import_paths`, and that `RelativeImport.sol`'s import will inheriut `RelativeImport.sol`'s import_path, and there should be no ambiguity (assuming import paths are not gross and nested/etc).

BenTheKush:
Okay, here's my proposal: we could add the solidity path directly to `File`, but this loses import path information which we may want later. By keeping track of  the `import_no` (from `ResolvedFile`) associated with each `File` we can infer the solidity path, as well as any remappings.

There are two ways to do this:

1. Adding a `Vec<ResolvedFile>` to `FileResolver` and adding a better public API (e.g., make `import_no` public, adding some logic to `FileResolver` to get the solidity name of a file `FileResolver::get_solidity_path(file: &File) -> Option<String>`). This is easier from an implementation side (I don't have to touch as many files). 

2. We could also add the `import_no` directly to `File`. We would still need to reference `FileResolver` to get the actual import path (and optional remapping name). One thing that I don't like about this conceptually is that not all files constructed seem to have an import path associated with them. Whenever a `FileResolver` resolves a file it infers an import path. However, we can construct a `File` object without an import path (e.g., in `sema/builtin.rs`, see listing below), and IMO this means that we should decouple importing from `File`s.

From [sema/builtin.rs:1463](https://github.com/hyperledger/solang/blob/9a53ecbb32d0a56162e2b894463076cedfafa7e3/src/sema/builtin.rs#L1463):
```rust
impl Namespace {
    pub fn add_solana_builtins(&mut self) {
        let file_no = self.files.len();
        // NOTE: It appears this file is never imported, so adding import information feels kludgy.
        self.files.push(File {
            path: PathBuf::from("solana"),
            line_starts: Vec::new(),
            cache_no: None,
        });
```

So yeah, I'm voting for 1: it's the easiest from an implementation standpoint and it's a bit cleaner from a conceptual standpoint (IMO), but I'm happy to go with 2 if you feel it's best.

BenTheKush:
@seanyoung Okay, pushed idea 1, but if you prefer idea 2 happy to revert and do that instead.

Letting CI run, tests should pass 🤞 🤞

xermicus (approved):
LGTM

src/file_resolver.rs
```diff
@@ -72,6 +80,18 @@ impl FileResolver {
...
+    /// Get the import path corresponding to a map
+    pub fn get_import_map(&self, map: &OsString) -> Option<&PathBuf> {
+        match self
+            .import_paths
+            .iter()
+            .find(|(m, _)| m.as_ref() == Some(map))
+        {
+            Some((_, pb)) => Some(pb),
+            None => None,
+        }
```
seanyoung, line 92: Surely this can be simplified with `.find_map()`

src/file_resolver.rs
```diff
@@ -86,6 +106,49 @@ impl FileResolver {
         self.files.get(file_no).cloned()
     }
 
+    /// Get the ResolvedFile associated with the `file_no`th `File`
+    pub fn get_resolved_file(&self, file_no: usize) -> Option<ResolvedFile> {
+        self.resolved_files.get(file_no).cloned()
+    }
```
seanyoung, line 112: Why are we cloning here and not returning a reference?

src/file_resolver.rs
```diff
@@ -86,6 +106,49 @@ impl FileResolver {
...
+                        "File at path {} should be prefixed by import path {}",
+                        full_path.display(),
+                        import_path.display()
+                    );
+                };
+                match map {
+                    Some(map) => Some(format!("{}/{}", map.to_str().unwrap(), rel_path.display())),
+                    None => Some(format!("{}", rel_path.display())),
+                }
+            }
```
seanyoung, line 147: Here you are reconstructing the solidity name but this does not work in many cases because the path is canonicalized.

- The filesystem used could be case insensitive,
- There could be symbol links or hard links
-  You're also using `Path.display()` and some characters may display differently
- A path could be relative to the working directory
- I notice a lack of test cases

I don't think this scheme is going to work TBH, It might be better to add an `import_no: Option<usize>` to `ast::File`.

src/file_resolver.rs
```diff
@@ -86,6 +106,49 @@ impl FileResolver {
...
+            None => None,
```
seanyoung, line 148: Why this? Looks broken.

seanyoung (changes requested):

src/file_resolver.rs
BenTheKush: ahh didn't know about this, thanks!

src/file_resolver.rs
BenTheKush: Alright, all good points, thank you! I'll go ahead and update.

Also Re missing test cases, it was Friday EOD and I wanted to get high-level feedback on approach before spending an extra couple hours on tests (it's kind of a big repo and I didn't see any test cases in obvious places so it was gonna take a bit of searching). Sorry that wasn't clear, should have mentioned that!

src/file_resolver.rs
BenTheKush: I made this prettier, didn't use `.find_map()` (wasn't clear how to simplify). Instead I did

```rust
    /// Get the import path corresponding to a map
    pub fn get_import_map(&self, map: &OsString) -> Option<&PathBuf> {
        self.import_paths
            .iter()
            .find(|(m, _)| m.as_ref() == Some(map))
            .map(|(_, pb)| pb)
    }
```

Let me know if you'd like something else

src/file_resolver.rs
BenTheKush: no longer applicable

src/file_resolver.rs
BenTheKush: I've updated to just track `import_no`.

BenTheKush:
Okay, went with the `Option<usize>` solution, lmk if this looks okay

src/file_resolver.rs
```diff
@@ -33,6 +33,18 @@ pub struct ResolvedFile {
     base: PathBuf,
 }
 
+impl ResolvedFile {
+    /// Get the import number associated with this `ResolvedFile`
+    pub fn get_import_no(&self) -> usize {
+        self.import_no
+    }
```
seanyoung, line 40: Import paths are not always used. For example, if you do `solang compile --target solana examples/solana/flipper.sol` then no import path is used at all. I think `ResolvedFile` `import_no` should be an `Option<usize>`, because it contains garbage in this case (it's set to 0).

seanyoung: As the `import_no` and `base` are only used in the solang crate, there is no need for the accessor functions.

seanyoung (changes requested):
Thanks for rewriting the PR @BenTheKush 

I think this code could really do with some test cases. We know it's broken in when no import path is used.

src/file_resolver.rs
BenTheKush: I added `get_import_no()` because I couldn't access `ResolvedFile.import_no` from `sema::mod::sema_file(...)` (according to my linter, didn't actually try to compile). I added `get_base()` since I was in the area, but am happy to remove it.

Re import paths: I'll update to use `Option<usize>`

BenTheKush:
> Thanks for rewriting the PR @BenTheKush

NP, I agree that this is definitely much nicer!

> I think this code could really do with some test cases. We know it's broken in when no import path is used.

Do you mean my changes specifically need tests, or this area of code in general? I'm definitely happy to write tests for both, but could we carve testing general functionality into a new PR? I'm blocked at work right now, so if we are confident I'm not introducing regressions I'd love to ship this and then add some tests in a separate PR.

As far as testing my changes, most of them should be correct-by-compilation (i.e., if it compiles it should be good).

The only piece of logic that is complicated enough to test would be `get_import_map`. I'll add some unit tests for that in-file.

Let me know if these should be somewhere else (I didn't see any tests in file), and if you want anything else tested.

BenTheKush:
Removed `get_base()`, I still need `get_import_no()`. Added tests for import logic/import number stuff (very basic, definitely not stress testing anything, e.g., nothing being imported, etc). Should be good to go, assuming you're okay carving the other stuff out into another PR. I'm going to branch off of here and start that refactor separately, can push if you want this to be in this same PR.

BenTheKush:
I'm trying to compare file paths, but this is failing on Windows:

```
 ---- sema::tests::get_import_map stdout ----
thread 'sema::tests::get_import_map' panicked at 'assertion failed: `(left == right)`
  left: `Some("C:\\a\\solang\\solang\\examples")`,
 right: `Some("\\\\?\\C:\\a\\solang\\solang\\examples")`', src\sema\tests\mod.rs:638:5

---- sema::tests::get_import_path stdout ----
thread 'sema::tests::get_import_path' panicked at 'assertion failed: `(left == right)`
  left: `Some((None, "C:\\a\\solang\\solang\\examples"))`,
 right: `Some((None, "\\\\?\\C:\\a\\solang\\solang\\examples"))`', src\sema\tests\mod.rs:658:9
```

Is there a good way to write this test case? Should I check which platform we are on and do a special case test?

seanyoung:
> I'm trying to compare file paths, but this is failing on Windows:
> 
> ```
>  ---- sema::tests::get_import_map stdout ----
> thread 'sema::tests::get_import_map' panicked at 'assertion failed: `(left == right)`
>   left: `Some("C:\\a\\solang\\solang\\examples")`,
>  right: `Some("\\\\?\\C:\\a\\solang\\solang\\examples")`', src\sema\tests\mod.rs:638:5
> 
> ---- sema::tests::get_import_path stdout ----
> thread 'sema::tests::get_import_path' panicked at 'assertion failed: `(left == right)`
>   left: `Some((None, "C:\\a\\solang\\solang\\examples"))`,
>  right: `Some((None, "\\\\?\\C:\\a\\solang\\solang\\examples"))`', src\sema\tests\mod.rs:658:9
> ```
> 
> Is there a good way to write this test case? Should I check which platform we are on and do a special case test?

That's a Windows UNC path. Run the path through `std::fs::canonicalize()` and then you'll get the UNC path. You can use it on all platforms.

Make sure you add a test for when a file is resolved via relative paths and not via import paths/maps.

BenTheKush:
> That's a Windows UNC path. Run the path through `std::fs::canonicalize()` and then you'll get the UNC path. You can use it on all platforms.

Great, done!

> Make sure you add a test for when a file is resolved via relative paths and not via import paths/maps.

This is testing functionality that I didn't add, can we put this in the other PR that I'm writing? I already have most of it written, I just really need to get access to import information.

Also, seems like you're coming online around 1am my time. I'm going to stay up so that I can respond to anything else that comes up and get this pushed through (time differences are a killer lol).

seanyoung (approved):

LucasSte (approved):

PR #1431, opened 2023-07-06
Remove unwanted files

LucasSte:
These files are temporary and automatically generated during tests. They have been included in the repository.


xermicus:
What about adding `tests/.tmp*` to `.gitignore`?

seanyoung (approved):

xermicus (approved):

LucasSte:
> What about adding `tests/.tmp*` to `.gitignore`?

Good point. I'll do that.

PR #1432, opened 2023-07-06
Identify externally callable functions in `sema`

xermicus:
The information whether a given function is supposed to be reachable via an external call to the contract is currently done directly in `codegen`. However, we need this information during the ABI generation too. This fixes a bug where inherited public functions would end up duplicated in the contract metadata.


src/sema/ast.rs
```diff
@@ -508,7 +508,11 @@ impl Function {
         matches!(self.mutability, Mutability::Pure(_))
     }
 
-    /// Is this function accessable externally
+    /// Is this function visible externally, based on it's visibilty modifiers.
+    ///
+    /// Due to inheritance, this alone does not determine whether a function is
+    /// externally callable in the final contract artifact; for that use
+    /// `Namesapce::function_externally_callable()` instead.
```
seanyoung, line 515: s/Namesapce/Namespace/

seanyoung (approved):
Looks great!

src/codegen/cfg.rs
```diff
@@ -2122,6 +2106,39 @@ impl Contract {
 }
 
 impl Namespace {
+    /// Determine whether a function should be included in the dispatacher and metadata,
```
LucasSte, line 2109:
```suggestion
    /// Determine whether a function should be included in the dispatcher and metadata,
```

src/codegen/cfg.rs
```diff
@@ -2122,6 +2106,39 @@ impl Contract {
...
+    /// taking inheritance into account.
+    ///
+    /// `function_no` is optional because default constructors require creating a CFG
+    /// without any corresponding function definition for default constructors.
```
LucasSte, line 2113:
```suggestion
    /// `function_no` is optional because default constructors require creating a CFG
    /// without any corresponding function definition.
```

src/sema/ast.rs
```diff
@@ -508,7 +508,11 @@ impl Function {
         matches!(self.mutability, Mutability::Pure(_))
     }
 
-    /// Is this function accessable externally
+    /// Is this function visible externally, based on it's visibilty modifiers.
+    ///
+    /// Due to inheritance, this alone does not determine whether a function is
+    /// externally callable in the final contract artifact; for that use
```
LucasSte, line 514:
```suggestion
    /// externally callable in the final contract artifact; for that, use
```

LucasSte (approved):

PR #1440, opened 2023-07-12
Bump should be hashed by last

LucasSte:
Presently, we hash include the bump in the `seeds` array passed to a Solana runtime call in the same order specified by developers when they are writing the contract's constructor. This is not the intended behavior, though. The bump must always be the last element in the array. This PR ensures that.


seanyoung (approved):

xermicus (approved):
LGTM

PR #1441, opened 2023-07-14
Fix compiler output section in solang.toml

salaheldinsoliman:
The compiler output section was not desterilized correctly from the `solang.toml` file created by `solang new`. this adds a fix for this.

### DIFF
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -108,7 +108,10 @@ pub struct Compile {
     pub package: CompilePackage,
 
     #[clap(flatten)]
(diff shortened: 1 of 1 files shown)


seanyoung (commented):
If there is a bug, there should a test

LucasSte:
> The compiler output section was not desterilized correctly

[Desterilized](https://www.collinsdictionary.com/dictionary/english/desterilize#)?

xermicus (approved):
As Sean pointed out, ideally comes with a small test

seanyoung (approved):
I am guessing a test is not forthcoming, since the milestone has been approved. I'd rather merge the fix without a test, than not merge it at all.

PR #1443, opened 2023-07-14
Make import_no Optional

BenTheKush:
This PR does 2 things:
1. Makes `ResolvedFile::import_no` an `Option<usize>`
2. Improves test suite for import logic

This is a first attempt, I imagine there are a few things that can be improved.

Possible issues include:
1. Some `import_no`s are set to `0` when no import is used [per this conversation](https://github.com/hyperledger/solang/pull/1419#discussion_r1259358226); in this case, the `import_no` should be set to `None`. However, I imagine others are set to 0 because it is the first import that is performed. I currently just changed all literal `import_no: 0` to `import_no: Some(0)` to get it to type check, but I'm guessing they may actually want to be `import_no: None`.
2. @seanyoung you wanted more tests, and in particular for:
    1. when no import path is used, and
    2. "when a file is resolved via relative paths and not via import paths/maps". 
  Could you expand on these more? For (i), how would I set up a case when no import path is used? For (ii), do you mean in an import statement (like `import ./blah.sol`)? I have a test that does this (imports `integrations/polkadots/createpair.sol` which has a relative import), but I'm not sure exactly what we want to test.

### DIFF
--- a/src/file_resolver.rs
+++ b/src/file_resolver.rs
@@ -206,7 +206,7 @@ impl FileResolver {
                 return Ok(ResolvedFile {
                     full_path,
                     base,
-                    import_no: Some(0),
+                    import_no: import_no.clone(),
                 });
             } else if let Ok(full_path) = base.join(&path).canonicalize() {
                 self.load_file(&full_path)?;
@@ -218,7 +218,7 @@ impl FileResolver {
                 return Ok(ResolvedFile {
                     full_path,
                     base,
-                    import_no: Some(0),
+                    import_no: import_no.clone(),
                 });
             }
 
--- a/src/file_resolver.rs
+++ b/src/file_resolver.rs
@@ -28,14 +28,14 @@ pub struct ResolvedFile {
     /// Full path on the filesystem
     pub full_path: PathBuf,
     /// Which import path was used, if any
-    import_no: usize,
+    import_no: Option<usize>,
     // Base part relative to import
     base: PathBuf,
 }
 
 impl ResolvedFile {
     /// Get the import number associated with this `ResolvedFile`
-    pub fn get_import_no(&self) -> usize {
+    pub fn get_import_no(&self) -> Option<usize> {
         self.import_no
     }
 }
@@ -179,7 +179,7 @@ impl FileResolver {
 
                             return Ok(ResolvedFile {
                                 full_path,
-                                import_no,
+                                import_no: Some(import_no),
                                 base,
                             });
                         }
@@ -206,7 +206,7 @@ impl FileResolver {
                 return Ok(ResolvedFile {
                     full_path,
                     base,
-                    import_no: 0,
+                    import_no: Some(0),
                 });
             } else if let Ok(full_path) = base.join(&path).canonicalize() {
                 self.load_file(&full_path)?;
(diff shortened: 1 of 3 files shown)


src/sema/tests/mod.rs
```diff
@@ -628,17 +628,27 @@ contract Child {
...
+
+    let mut cache = FileResolver::new();
+    let integration = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("integration");
+    let polkadot = integration.join("polkadot");
+    assert!(cache.add_import_path(polkadot.as_path()).is_ok());
+    let ns = parse_and_resolve(OsStr::new("createpair.sol"), &mut cache, Target::EVM);
+    for file in ns.files {
+        assert_eq!(
+            &polkadot,
+            &cache.get_import_path(file.import_no.unwrap()).unwrap().1
```
seanyoung, line 649: Nothing is imported via import path so these asserts should `assert!(file.import_no.is_none());`

src/file_resolver.rs
```diff
@@ -206,7 +206,7 @@ impl FileResolver {
                 return Ok(ResolvedFile {
                     full_path,
                     base,
-                    import_no: 0,
+                    import_no: Some(0),
```
seanyoung, line 209: None/copy of parent

src/file_resolver.rs
```diff
@@ -218,12 +218,12 @@ impl FileResolver {
...
```
seanyoung, line 221: copy of parent `import_no`

src/file_resolver.rs
```diff
@@ -236,7 +236,7 @@ impl FileResolver {
             return Ok(ResolvedFile {
                 full_path,
                 base,
-                import_no: 0,
+                import_no: Some(0),
```
seanyoung, line 239: copy of import_no in cache

seanyoung (changes requested):
>    i. when no import path is used, and

Yes, add a test which removes:
```rust
assert!(cache.add_import_path(polkadot.as_path()).is_ok());
```

>    ii. "when a file is resolved via relative paths and not via import paths/maps". 
>  Could you expand on these more? For (i), how would I set up a case when no import path is used? For (ii), do you mean in an import statement (like `import ./blah.sol`)? I have a test that does this (imports `integrations/polkadots/createpair.sol` which has a relative import), but I'm not sure exactly what we want to test.

```rust
let path = OsStr::from("examples/polkadot/flipper.sol");
let ns = parse_and_resolve(path, &mut cache, Target::EVM);
```

src/file_resolver.rs
BenTheKush: Where can I get this? I don't think this is accessible through FileResolver (it's in File which is stored in Namespace).

src/sema/tests/mod.rs
BenTheKush: I don't understand, "createpair.sol" is resolved by file resolver  by walking over import paths. If I don't add polkadot as an import path to the file resolver then tests fail.

bkushigian:
@seanyoung just wanted to follow up on this PR. I left a couple replies above.

seanyoung:
> @seanyoung just wanted to follow up on this PR. I left a couple replies above.

I've looked at this and wrote a new PR #1465 which solves a number of issues including `import_no` being optional. Would you mind having a look and see if that works for you.

seanyoung:
Fixed by https://github.com/hyperledger/solang/pull/1465

PR #1446, opened 2023-07-18
Add debug implementations required to print Namespace

chioni16:
Derive `Debug` implementations for structs and enums required to print `struct Namespace`. This is of use when one needs to print `Namespace` struct for debugging purposes.

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -32,7 +32,7 @@ use std::{fmt, fmt::Write};
 // IndexMap <ArrayVariable res , res of temp variable>
 pub type ArrayLengthVars = IndexMap<usize, usize>;
 
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 pub enum Instr {
     /// Set variable
@@ -392,7 +392,7 @@ impl fmt::Display for HashTy {
     }
 }
 
-#[derive(Clone, Default)]
+#[derive(Debug, Clone, Default)]
 pub struct BasicBlock {
     pub phis: Option<BTreeSet<usize>>,
     pub name: String,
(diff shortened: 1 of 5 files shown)


xermicus (approved):

seanyoung (approved):

LucasSte:
We don't actually need to print these structs during runtime, so adding the `#[derive(Debug)]` attribute creates overhead code that we don't use.
What do you think of enabling it only for debug builds?

https://users.rust-lang.org/t/will-derive-debug-be-compiled-in-release-build/30388

chioni16:
> We don't actually need to print these structs during runtime, so adding the `#[derive(Debug)]` attribute creates overhead code that we don't use. What do you think of enabling it only for debug builds?
> 
> https://users.rust-lang.org/t/will-derive-debug-be-compiled-in-release-build/30388

Yes, that's better. Thanks for the suggestion!

seanyoung:
@LucasSte what overhead are you concerned about? I does not increase code size (according to `cargo build --release` testing). You're suggesting extra code complexity, what does buy ys?

xermicus:
If this compiles to dead code, why would it end up in the final binary? I'd expect  dead code elimination to kick it out. I'd expect guarding these derives with `cfg_attr` to improve compile time of solang itself, however I'm not too worried about a few additional `Debug` derives

LucasSte:
> If this compiles to dead code, why would it end up in the final binary? I'd expect dead code elimination to kick it out. I'd rather expect guarding these derives with `cfg_attr` to improve compile time of solang itself, however I'm not too worried about a few additional `Debug` derives

I assumed it adds code to our binary from this question: https://users.rust-lang.org/t/will-derive-debug-be-compiled-in-release-build/30388, but I did not investigate further.

PR #1449, opened 2023-07-19
Polkadot: Reverts return encoded error data

xermicus:
This is a continuation of #1415. `require()`, `assert()` and `revert()` now return error data, according to the [Ethereum Solidity documentation](https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require). Additionally, many reverts inserted by the compiler now return the corresponding `Panic(uint256)`  error data, to align Solang closer with `solc`.

The error types known to the contract are added in the metadata `lang_error` field. At the moment there are only `Error` and `Panic` because we don't support custom errors yet.

Refactored revert-related code into a dedicated `codegen` module to. Refactored the `polkadot::errors` into distinct tests, made them less brittle and added assertions for the execution output.

I'm now working on a follow-up PR for bubbling up uncaught exceptions (this is why it's already included that in the documentation).


src/emit/binary.rs
```diff
@@ -1074,6 +1076,19 @@ impl<'a> Binary<'a> {
...
         );
     }
+
+    /// Emit encoded error data of "Panic(uint256)" as interned global string.
+    pub(super) fn error_data_const(
+        &self,
+        ns: &Namespace,
+        code: PanicCode,
+    ) -> (PointerValue<'a>, IntValue<'a>) {
+        let bytes = create_encoder(ns, false).const_error_panic(code);
```
LucasSte, line 1086: Using codegen in emit is not a good idea. You could have an `Expression:StringLiteral` for this case. We usually assume that, when we reach emit, everything from CFG has already been constructed.

xermicus:
#1452 will make the CI green

src/codegen/encoding/mod.rs
```diff
@@ -1732,10 +1735,57 @@ pub(super) trait AbiEncoding {
 
     /// Returns if the we are packed encoding
     fn is_packed(&self) -> bool;
+
+    /// Encode "Panic(uint256)" error data.
+    fn const_error_panic(&self, code: PanicCode) -> Vec<u8> {
```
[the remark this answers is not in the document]
xermicus: Wouldn't they be encoded the same in borsh?

src/emit/binary.rs
xermicus: I disagree. `create_encoder` doesn't take a `CFG` or anything, so it's impossible that it changes the CFG. By re-using it in emit, we use the _exact_ same code to create these constant values. I don't want to re-implement and test this again in emit.

src/codegen/encoding/mod.rs
```diff
@@ -1732,10 +1735,57 @@ pub(super) trait AbiEncoding {
...
+    /// Returns`None` if the error data is not a compile time constant value.
+    fn encode_error_data_const(&self, error: SolidityError) -> Option<Expression> {
+        let bytes = match error {
+            SolidityError::Empty => return None,
+            SolidityError::String(data) => match data {
+                Expression::AllocDynamicBytes {
+                    ty: Type::String,
+                    initializer: Some(data),
+                    ..
+                } => self.const_error_string(String::from_utf8(data).unwrap()),
```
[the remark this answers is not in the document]
xermicus: There is no `Expression::StringLiteral` on `codegen::Expression`. In `emit::expression.rs`:

```
Expression::AllocDynamicBytes {
            ty,
            size,
            initializer,
            ..
        } => {
            if matches!(ty, Type::Slice(_)) {
                let init = initializer.as_ref().unwrap();

                let data = bin.emit_global_string("const_string", init, true)
```

This is exactly what I want. This only works by having an `AllocDynamicBytes`. `emit` even checks if it's a slice and then emits an interned string, which is exactly what I want

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,35 @@ The following example shows how call flags can be used:
...
+``assert()``, ``require()``, or ``revert()`` will revert the contract execution, where the revert reason 
+is supplied as the contracts output (if any). Solidity contracts can also revert with a `Panic`.
+Uncaught exceptions from calling and instantiating contracts or transferring funds will be bubbled 
+up back to the caller.
+
+The metadata contains all error variants that the contract knows about in the ``lang_error`` field.
+The 4 bytes `selector` of the error data can be seen as the enum discriminator. However, because 
+SCALE encoding does not allow discriminators larger than 1 byte, the hex-encoded error selector 
+is provided as the enum variant name in the metadata (the selector could also be calculated by
+reconstructing and hashing the error signature based on the enum Variant types).
```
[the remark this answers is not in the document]
xermicus: What do you mean with built in? The error type is represented as an enum in the metadata, and it has variants.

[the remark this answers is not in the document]
LucasSte: You wrote `Variant` with a capital V, as if its a name of something.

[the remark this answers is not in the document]
LucasSte: That looks like `enum Variant`. If that is not what you mean, you should use a lower case letter.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: You can create an `Expression::StringLiteral` that does this. This is what I meant.

docs/targets/polkadot.rst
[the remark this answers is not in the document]
xermicus: Ah right, yeah that's true I'll fix it

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: It looks like so. My concern was that this feature is not implemented yet on Solana. Anchor does not decode anything if a function fails, so we return no values.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Ah I see. Yeah why not; I created #1453 so we don't forget doing it afterwards. For now I'd like to make progress with this PR, and creating a new Expression is out of scope here, especially if the same can be done by just using `AllocDynamicBytes`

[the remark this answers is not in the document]
LucasSte: The new expression is the easiest solution for this PR, IMO, as it would not require the hack with `Expression::AllocaDynamicBytes`.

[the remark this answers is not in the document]
xermicus: It would be the proper solution, but still out of scope for this PR. On a side note, I don't see how implementing and testing a whole new Expression, and then replace existing parts where this new expression can also be used, is an easier Solution than just  using `Expression::AllocaDynamicByte`. It works as I expected with minimal code.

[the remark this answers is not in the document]
LucasSte: `Expression::AllocDynamicBytes` gives you the constant string, but it is not the proper representation of it in the code. You are masking a constant string behind an allocation. We might implement passes in the future that discern between when you have a constant string and and you have slice (e.g. concatenating items during compile time) that simply won't work. The CFG should have a fairly trustworthy representation of the code we write.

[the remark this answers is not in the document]
LucasSte: I can't recall right now any other place where string literals (or a byte array literal) are need. The slice we have in `Expression::AllocDynamicBytes` is needed due to Solana runtime calls.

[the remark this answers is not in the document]
xermicus: > I can't recall right now any other place where string literals (or a byte array literal) are need.


Obviously  is already happening at other places, otherwise the code turning `AllocDynamicBytes` of type `Slice` into interned strings wouldn't be there. 

> The slice we have in Expression::AllocDynamicBytes is needed due to Solana runtime calls.

Why is it ok then to use `AllocDynamicBytes` of type Slice when the Solana runtime needs it but not for other things. I don't get it.

> We might implement passes in the future that discern between when you have a constant string and and you have slice

Then we just fix it before implementing these optimizations. Also I'm not at all worried about this optimization not happening for this case.

This discussion did convince me 0% so far to do this. Again: Out of scope and I don't have time for this. I'm trying to fix real problems so we can get the compiler audited and supporting OpenZepplin as soon as possible. I lack any resource to improve things that otherwise work but are not perfect.

I see your point. I agree with you that this could be done different. However given no security or other significant issues arise from using `AllocDynamicBytes` here I simply can neither afford to scope creep this PR nor to implement such cosmetics before merging this.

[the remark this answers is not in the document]
LucasSte: I re-reviewed your PR and re-analyzed the use-case. It makes sense to me now.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: What about moving these functions out of the trait? They do not use `self`.

src/emit/binary.rs
LucasSte: Perhaps the best here would be for `const_error_panic` to be a utility function that is not in the encoder trait. It transforms the panic code into bytes, while the encoder usually returns a Solidity array (an `Expression::Variable`).

This way, you can decouple the encoder from emit and still call that function in codegen.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: Then, `create_encoder` need not to be `pub(crate)` and we can decouple the encoder from `emit`.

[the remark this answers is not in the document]
xermicus: First, I did it exactly like that. I reconsidered.

I _want_ anyone to force using the encoder, even when encoding const data.

We must not allow _any_ encoding happening outside of `AbiEncoding` trait. Otherwise we create a perfect footgun.

By implementing this here, as a trait method, we force this. It's exactly what I want.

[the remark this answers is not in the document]
xermicus: I'm less concerned about creating and using a `AbiEncoding` instance in `emit`. Yes, when we pass in a CFG that gets modified in emit, this would be problematic. However, this is 1. hard to do anyways, 2. very easy to spot in a code reviews. Where as without having us to force using the encoder, it's 1. super easy to emit incorrected const ABI encoded values and 2. much harder to spot in code reviews.

src/emit/binary.rs
xermicus: (My [above comment ](https://github.com/hyperledger/solang/pull/1449#discussion_r1271074040) relates to this as well)

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Another benefit of having this in the `AbiEncoding` trait is that once we have an optimization to optimize encoding of compile time constant values, it can converge the arguments to constants and then just use this.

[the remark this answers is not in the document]
LucasSte: > Another benefit of having this in the `AbiEncoding` trait is that once we have an optimization to optimize encoding of compile time constant values, it can converge the arguments to constants and then just use this.

I need to quote yourself for this one:

> Then we just fix it before implementing these optimizations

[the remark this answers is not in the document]
xermicus: Yes that's true. We can't even test this for Solana. I'm going to remove the provided implementations and implement it in the SCALE version.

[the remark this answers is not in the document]
xermicus: > I need to quote yourself for this one:
>> Then we just fix it before implementing these optimizations 

Do you think when we want to implement this optimization the constant ABI encoding can't be a trait method? Why?

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,35 @@ The following example shows how call flags can be used:
 .. include:: ../examples/polkadot/call_flags.sol
   :code: solidity
 
+
+Reverts and error data decoding
+_______________________________
+
+When the contract decides to revert its execution, returned error data corresponds to what contracts on
```
seanyoung, line 68:
```suggestion
When a contract reverts, the returned error data is what
```
(Removing superfluous words)

docs/targets/polkadot.rst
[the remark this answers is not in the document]
seanyoung: You could simply list the two error selectors for `Panic` and `Error`. 

Custom errors are not implemented in your PR.

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
seanyoung: We could have generic `fn const_encode(args: &[Expression]) -> Vec<u8>` or something like that. That would be generic to useful for all targets and it can be used for both `Panic`, `Error()` and custom errors.

src/codegen/expression.rs
```diff
@@ -2011,7 +1901,8 @@ fn expr_builtin(
                 vartab,
                 ns,
             );
-            assert_failure(loc, None, ns, cfg, vartab);
+            let error = SolidityError::Panic(PanicCode::Generic);
```
seanyoung, line 1904: out of bounds

src/codegen/expression.rs
```diff
@@ -2071,7 +1962,8 @@ fn expr_builtin(
...
```
seanyoung, line 1965: out of bounds

src/codegen/expression.rs
```diff
@@ -2148,7 +2040,8 @@ fn expr_builtin(
...
```
seanyoung, line 2043: out of bounds

src/codegen/expression.rs
```diff
@@ -2412,7 +2305,8 @@ fn checking_trunc(
         vartab,
         ns,
     );
-    assert_failure(loc, None, ns, cfg, vartab);
+    let error = SolidityError::Panic(PanicCode::Generic);
```
seanyoung, line 2308: integer overflow

src/codegen/mod.rs
```diff
@@ -6,11 +6,14 @@ mod constant_folding;
 mod constructor;
 mod dead_storage;
 pub(crate) mod dispatch;
-mod encoding;
+pub(crate) mod encoding;
 mod events;
 mod expression;
 mod external_functions;
 mod reaching_definitions;
+/// Releated to code that ultimately compiles to the target
```
seanyoung, line 14: I don't know what that means. All code compiles to the target

src/codegen/mod.rs
```diff
@@ -6,11 +6,14 @@ mod constant_folding;
...
+/// equivalent instruction of EVM revert (0xfd).
```
seanyoung, line 15: comment belongs at top of module

src/codegen/revert.rs
```diff
@@ -0,0 +1,364 @@
...
+pub enum PanicCode {
+    Generic = 0x00,
+    Assertion = 0x01,
+    MathOverflow = 0x11,
+    DivisionByZero = 0x12,
+    EnumCastOob = 0x21,
+    StorageBytesEncodingIncorrect = 0x22,
+    EmptyArrayPop = 0x31,
+    ArrayIndexOob = 0x32,
+    OutOfMemory = 0x41,
```
seanyoung, line 95: Not wired up

src/codegen/revert.rs
```diff
@@ -0,0 +1,364 @@
...
+    InternalFunctionUninitialized = 0x51,
```
seanyoung, line 96: We don't have a check for this (we need it though!)

src/emit/binary.rs
seanyoung: If a const encoder doesn't take a cfg then why can't it be used in emit?

seanyoung (changes requested):

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Yes, that goes into the direction what I envision for #1451

src/codegen/revert.rs
xermicus: True, I added it

src/codegen/mod.rs
xermicus: Moved it

src/codegen/expression.rs
xermicus: Changed it

src/codegen/expression.rs
xermicus: Changed it

src/codegen/expression.rs
xermicus: Changed it

src/codegen/expression.rs
xermicus: changed

src/codegen/revert.rs
xermicus: True, I can wire it up on `Binary::vector_new()`. What is the max size on Solana

xermicus: Or maybe we should just point it out in the documentation and solve it later. I don't think it's terribly important

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
LucasSte: > We could have generic `fn const_encode(args: &[Expression]) -> Vec<u8>` or something like that. That would be generic to useful for all targets and it can be used for both `Panic`, `Error()` and custom errors.

That sounds like a good idea.

docs/targets/polkadot.rst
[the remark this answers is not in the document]
xermicus: Done that

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,42 @@ The following example shows how call flags can be used:
...
   :code: solidity
 
+
+Reverts and error data decoding
+_______________________________
+
+When a contract reverts, the returned error data is what
+`EVM would return <https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require>`_.
+``assert()``, ``require()``, or ``revert()`` will revert the contract execution, where the revert reason 
+is supplied as the contracts output (if any). Solidity contracts can also revert with a `Panic`.
```
seanyoung, line 71: Some words about where `Panic` occurs would be helpful

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,42 @@ The following example shows how call flags can be used:
...
+    Solang does not yet support custom errors. For now, only `Error` (selector of `0x08c379a0`) 
+    and `Panic` (selector of `4e487b71`) will be returned and occur in the metadata.
+
+The general process of decoding the output data for Solang Solidity contracts is as follows:
+
+1. If the revert flag is **not** set, the contract didn't revert and the output should be encoded as specified in the message spec.
+2. The compiler version must be solang > 0.3.1, or the error data can't be decoded (check the ``compiler`` field in the contract metadata).
+3. If the output length is smaller than 4 bytes, the error data can't be decoded (contracts may return empty error data, for example if ``revert()`` without arguments is used).
+4. If the first 4 bytes of the output do **not** match any of the selectors found in ``lang_error``, the error can't be decoded.
+5. **Skip** the selector (first 4 bytes) and decode the remaining data according to the corresponding variant type found in `lang_error`.
```
seanyoung, line 94: Some example code would be great

src/codegen/encoding/mod.rs
```diff
@@ -1732,10 +1732,17 @@ pub(super) trait AbiEncoding {
 
     /// Returns if the we are packed encoding
     fn is_packed(&self) -> bool;
+
+    /// Encode constant data at compile time.
+    ///
+    /// Returns `None` if the data can not be encoded at compile time.
+    fn const_encode(&self, _args: &[Expression]) -> Option<Vec<u8>> {
+        None
+    }
```
seanyoung, line 1741: How about returning `Vec<u8>` and a default function which just calls `unimplemented!();`

src/codegen/encoding/scale_encoding.rs
```diff
@@ -578,4 +580,40 @@ impl AbiEncoding for ScaleEncoding {
...
+                Expression::AllocDynamicBytes {
+                    initializer: Some(data),
+                    ty: Type::String,
+                    ..
+                } => result.extend_from_slice(&String::from_utf8(data.to_vec()).unwrap().encode()),
+                Expression::AllocDynamicBytes {
+                    initializer: Some(data),
+                    ty: Type::Slice(_),
+                    ..
+                } => result.extend_from_slice(data),
```
seanyoung, line 597: That's wrong, if `Type::Slice(ty)` has `ty == Type::Bytes(1)` then it should have the same implementation as `Type::String`.

src/codegen/encoding/scale_encoding.rs
```diff
@@ -578,4 +580,40 @@ impl AbiEncoding for ScaleEncoding {
...
         self.packed_encoder
     }
+
+    fn const_encode(&self, args: &[Expression]) -> Option<Vec<u8>> {
+        let mut result = vec![];
+        for arg in args {
+            match arg {
+                Expression::AllocDynamicBytes {
+                    initializer: Some(data),
+                    ty: Type::String,
```
seanyoung, line 590: `Type::String | Type::DynamicBytes`

src/emit/instructions.rs
```diff
@@ -292,14 +293,18 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
...
+                    bin.context
+                        .i8_type()
+                        .ptr_type(AddressSpace::default())
+                        .const_null(),
+                    bin.context.i32_type().const_zero(),
+                )
+            } else {
+                bin.panic_data_const(ns, PanicCode::EmptyArrayPop)
+            };
+            target.assert_failure(bin, revert_out, revert_out_len);
```
seanyoung, line 307: This sequence of code is repeated a few times, might be good to factor out into a function (maybe something on target?)

seanyoung (changes requested):
Great stuff just some minor things

src/codegen/revert.rs
seanyoung: Or maybe a comment next to the definition

src/codegen/encoding/mod.rs
xermicus: I like returning the Option way better. It allows to do `const_encode(..).unwrap_or(abi_encode())`

docs/targets/polkadot.rst
xermicus: What code example do you have in mind?

I don't see what a code example would add. Also, a Typescript or Rust snippet to do this will be quite long I don't see why this belongs into our docs, a solidity example is uselesse because try-catch should be used anyways.

src/codegen/encoding/scale_encoding.rs
xermicus: Good point

src/codegen/encoding/scale_encoding.rs
xermicus: Are `Type::DynamicBytes` always string? This will SCALE encode it using `String`, and not just yield the bytes as is

src/emit/instructions.rs
xermicus: Good point, I'll look into it

src/codegen/revert.rs
xermicus: Ah yes, definitively deserves a comment too

src/codegen/encoding/scale_encoding.rs
seanyoung: `Vec<u8>` and `String` are encoded the  same way in SCALE.

src/codegen/encoding/mod.rs
seanyoung: I had not thought of that way

src/codegen/encoding/scale_encoding.rs
xermicus: While that's true, parsing the vec into an UTF8 string might panic. I think I can add `Type::DynamicBytes` but will add another match arm for it :-)

docs/targets/polkadot.rst
seanyoung: How about a link to an integration test?

xermicus: Well, in the integration test we always just match the output directly against the expected output bytes (integration tests are not examples...)

docs/targets/polkadot.rst
xermicus: Gonna add a link to the Solidity docs

src/emit/instructions.rs
xermicus: I moved it into `panic_data_const`

src/codegen/encoding/scale_encoding.rs
```diff
@@ -578,4 +580,60 @@ impl AbiEncoding for ScaleEncoding {
...
+    /// TODO: This is used and tested for error data (Error and Panic) only.
+    fn const_encode(&self, args: &[Expression]) -> Option<Vec<u8>> {
+        let mut result = vec![];
+        for arg in args {
+            match arg {
+                Expression::AllocDynamicBytes {
+                    initializer: Some(data),
+                    ty: Type::String,
+                    ..
+                } => result.extend_from_slice(&String::from_utf8(data.to_vec()).unwrap().encode()),
```
seanyoung, line 593: This could be simply `result.extend_from_slice(&data.encode()),`

seanyoung (changes requested):
There are still comments open when you requested a review from me

src/codegen/encoding/mod.rs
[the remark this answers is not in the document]
xermicus: Done

src/codegen/encoding/scale_encoding.rs
xermicus: `data` is a vector, not a string. Just because it coincidentally works I still think it's better to encode it using the correct type

src/emit/binary.rs
```diff
@@ -1074,6 +1077,38 @@ impl<'a> Binary<'a> {
...
+        }
+
+        let expr = Expression::NumberLiteral {
+            loc: pt::Loc::Codegen,
+            ty: Type::Uint(256),
+            value: (code as u8).into(),
+        };
+        let bytes = create_encoder(ns, false)
+            .const_encode(&[SolidityError::Panic(code).selector_expression(), expr])
+            .unwrap();
```
LucasSte, line 1106: You can just do `code.abi_encode` here, can't you?
Perhaps this would be cleaner.

src/codegen/encoding/scale_encoding.rs
seanyoung: You're assuming that the contents is valid utf8, which it does not need to be.

src/codegen/encoding/scale_encoding.rs
seanyoung: `Type::String` and `Type::DynamicBytes` encode the same way

src/emit/binary.rs
xermicus: I see what you mean, but `SolidityError::AbiEncode` encodes any error data and not just const ones, which is why it takes a CFG. It shouldn't be used in emit, I made it `pub(super)`.

src/codegen/encoding/scale_encoding.rs
```diff
@@ -578,4 +580,60 @@ impl AbiEncoding for ScaleEncoding {
...
+    let data = vec![0x41, 0x41];
+    let encoder = ScaleEncoding::new(false);
+    let expr = Expression::AllocDynamicBytes {
+        loc: Codegen,
+        ty: Type::DynamicBytes,
+        size: Expression::Poison.into(),
+        initializer: data.clone().into(),
+    };
+    let encoded = encoder.const_encode(&[expr]).unwrap();
+    assert_eq!(encoded, data.encode());
```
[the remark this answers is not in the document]
xermicus: Done

src/codegen/encoding/scale_encoding.rs
xermicus: Hmm interesting, do we allow UTF16? Or why is that?

src/codegen/revert.rs
xermicus: Added a comment and a tracking issue #1477

src/codegen/encoding/scale_encoding.rs
xermicus: I changed it to use them same code for both types for now

seanyoung:
```solidity
revert("abc \xfe def");
```
is not valid unicode but it is a string

LucasSte (approved):

src/codegen/encoding/scale_encoding.rs
```diff
@@ -578,4 +580,88 @@ impl AbiEncoding for ScaleEncoding {
...
+                    ty: Type::Slice(inner),
+                    ..
+                } if matches!(**inner, Type::Bytes(1)) => result.extend_from_slice(data),
+                Expression::NumberLiteral {
+                    ty: Type::Bytes(4),
+                    value,
+                    ..
+                } => {
+                    let bytes = value.to_bytes_be().1;
+                    result.extend_from_slice(&bytes[..]);
```
seanyoung, line 605: That's  not right. Say the values is `0x0000fffff`, then bytes will be two bytes and you will be missing the two leading bytes. This case really needs a test

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,45 @@ The following example shows how call flags can be used:
...
+
+Reverts and error data decoding
+_______________________________
+
+When a contract reverts, the returned error data is what
+`EVM would return <https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require>`_.
+``assert()``, ``require()``, or ``revert()`` will revert the contract execution, where the revert reason 
+is supplied as the contracts output (if any). Solidity contracts can also revert with a `Panic` 
+(please refer to the 
+`Solidity language documentation <https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require>`_.
```
seanyoung, line 73: Please call this `Ethereum Solidity documentation. It's totally confusing to call this "Solidity language documentation".

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,45 @@ The following example shows how call flags can be used:
...
+SCALE encoding does not allow discriminators larger than 1 byte, the hex-encoded error selector 
+is provided as the enum variant name in the metadata (the selector could also be calculated by
+reconstructing and hashing the error signature based on the enum variant types).
+
+.. note::
+
+    Solidity knows about `Error`, `Panic` and 
+    `custom errors <https://docs.soliditylang.org/en/latest/abi-spec.html#errors>`_.
+    Solang does not yet support custom errors. For now, only `Error` (selector of `0x08c379a0`) 
+    and `Panic` (selector of `4e487b71`) will be returned and occur in the metadata.
```
seanyoung, line 89: Add `0x` prefix to 4e487b71 to make it consistent with other selector

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,45 @@ The following example shows how call flags can be used:
...
+`Solidity language documentation <https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require>`_.
+for more information about when `Panic` might be returned).
+Uncaught exceptions from calling and instantiating contracts or transferring funds will be bubbled 
+up back to the caller.
+
+The metadata contains all error variants that the contract knows about in the ``lang_error`` field.
+The 4 bytes `selector` of the error data can be seen as the enum discriminator. However, because 
+SCALE encoding does not allow discriminators larger than 1 byte, the hex-encoded error selector 
+is provided as the enum variant name in the metadata (the selector could also be calculated by
+reconstructing and hashing the error signature based on the enum variant types).
```
seanyoung, line 82: So where does the name go? E.g. `Panic`, `Error`, or `CustomError5`? 

I do think this this paragraph is inscrutable. I still don't have a good idea of what is going on here.

docs/targets/polkadot.rst
```diff
@@ -62,3 +61,45 @@ The following example shows how call flags can be used:
...
+
+.. note::
+
+    Solidity knows about `Error`, `Panic` and 
+    `custom errors <https://docs.soliditylang.org/en/latest/abi-spec.html#errors>`_.
```
seanyoung, line 87: Please don't refer to solc/Ethereum Solidity as "Solidity". That suggests we are not Solidity

seanyoung (changes requested):
Please

docs/targets/polkadot.rst
xermicus: Why is it confusing calling the Solidity Language Documentation the Solidity Language Documentation.

If anything, `Ethereum Solidity documentation` is confusing. What is `Ethereum Solidity`? What is not `Ethereum Solidity`?

I don't get why we can't call the Solidity Language by it's name.

You go to a website, the website is called "https://soliditylang.org/".

This implies that there is a thing called "Solidity Language".

On this website, there is a button, called "read the docs", leading to another website at "https://docs.soliditylang.org".

So, if there is the `Solidity Langauge`, and it has a `Documentation`. How is it "totally confusing" to calling it "Solidity language documentation"?

docs/targets/polkadot.rst
xermicus: > Please don't refer to solc/Ethereum Solidity as "Solidity"

I disagree. Solidity is what the language authors think it is. And that happens to be whats on soliditylang.org. We can't say we are a Solidity compiler when we in fact compile something that is not Solidity.

If you mean `solc/Ethereum` what is broadly understood as the _Solidity Language_, it doesn't follow to not call it _Solidity_. Saying  "Solidity knows about `X`" is correct.

src/codegen/encoding/scale_encoding.rs
xermicus: Hmm yeah good point, thanks. While it definitively works for this limited application on error codes. It's a bit a pain to get this right, because it's not natively in SCALE. And `num-bigint` doesn't document about trailing 0 bytes. I have to look into this again.

docs/targets/polkadot.rst
xermicus: Makes me think, do you think it'd be better if the Selector goes into the path of the Variant Type?

docs/targets/polkadot.rst
seanyoung: Also, I've called it "Ethereum Solidity" throughout the documentation. If you're going to change this, then change all the links that say "Ethereum Solidity".

LucasSte: In one hand, Solidity for Solana has its own syntax, so we can see it as a different flavor. Calling Ethereum Solidity to refer to the original Solidity implementation is reasonable, given that we are slightly different than the original.

seanyoung: > In one hand, Solidity for Solana has its own syntax, so we can see it as a different flavor. Calling Ethereum Solidity to refer to the original Solidity implementation is reasonable, given that we are slightly different than the original.

Exactly. On Polkadot there are also many small differences with Ethereum Solidity (or should be called Solidity on Ethereum/ Solidity on EVM)?

xermicus: > Your link suggests that https://docs.soliditylang.org/en/latest/abi-spec.html#errors is the authoritative answer.

Yes, intentionally so. I'd be largely surprised if this wasn't the authoritative answer. For a vast majority of anyone referring to the Solidity Language in any way.

> Where does that leave Solang?

This is concisely explained in literally the next sentence:

    Solang does not yet support custom errors.

> Also, I've called it "Ethereum Solidity" throughout the documentation. 

Not true. By just having a very quick look at the docs, I find _many_ instances we just talk about `Solidity`. No mention of `Ethereum Solidity`, `Solang Solidity` or whathever else. Just `Solidity`. And in this case I assume anyone reaading the docs will `Solidity` as to what's documented on docs.soliditylang.org and implemented by solc.

For example in the Solana section. The first sentence:

    As the underlying Solana environment is different than that of Ethereum, Solidity inner workings have been modified to function properly.

So, this should mention that the inner workings of `Ethereum Solidity` have been modified, not `Solidity`. There are _many_ more of these instances.

> If you're going to change this, then change all the links that say "Ethereum Solidity"

As pointed out, in our docs, this is not at all consistent. In fact, there are many instances we write just `Solidity`, without referring to either what you mean with "Ethereum Solidity", nor to "Solang Solidity" nor "Solidity on Solana" nor anything else. So why not do the only sane thing here which is referring the term `Solidity` as to what everyone else in the world refers it to.

xermicus: > > In one hand, Solidity for Solana has its own syntax, so we can see it as a different flavor. Calling Ethereum Solidity to refer to the original Solidity implementation is reasonable, given that we are slightly different than the original.
> 
> Exactly. On Polkadot there are also many small differences with Ethereum Solidity (or should be called Solidity on Ethereum/ Solidity on EVM)?

I'm fine with calling it `Solana Solidity` or `Solidity on Solana` or `Solang Solidity` or whatever. I'm just not convinced that we can not just use `Solidity` when referring to what the World understands as `Solidity` to be is what's documented on soliditylang.org and in `solc`.

And this is _very_ inconsistent in our docs.

But fine. Apparently it's the end of the world if someone talks about `Solidity` in our docs. I'm just gonna `s/Solidity/Ethereum Solidity` here to end this stupid discussion for good

xermicus: Ah, now I get what you mean.

But it doesn't work like that.

You can't invent random German sounding words and start using in them in a local community still call your Language German.

What you end up is a _dialect_.

So the correct thing to say would be that Solang is a Compiler for _Solidity Dialects_ running on other runtimes.

But that still leaves the solc/docs.soliditylang.org the authority. We can't just deny that fact.

seanyoung: I am fine with dialect. 

> But that still leaves the solc/docs.soliditylang.org the authority. We can't just deny that fact.

I am totally denying that fact. 

Just in the same way. there is no version German which is not a dialect. Each and every one is a dialect. Just in the same way, the King of England speaks a dialect.

xermicus: I summarize this discussion into the following points:

- Neither the `solc` compiler nor the `docs.soliditylang.org` resemble a language spec.
- Hence it's questionable whether this is authoritative because it's merely the predominant dialect or version of Solidity at the time of writing.
- Add to that, seeing the compiler or docs of a version of Solidity that can only run on a specific runtime is definitively not what we want.
- From that follows, saying just `Solidity` while linking to some specific documentation or implementation should be avoided. Because it implies that the documentation or implementation linked to to be an authoritative source while it is not.

xermicus: @seanyoung we can (we should!) continue discussion in #1479

as for this PR I can just change it to `Ethereum Solidity` because that's what seems to be used most often in other places.

src/emit/instructions.rs
```diff
@@ -601,6 +596,31 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
             let callable =
                 expression(target, bin, call_expr, &w.vars, function, ns).into_pointer_value();
 
+            let ptr_ok = bin.context.append_basic_block(function, "fn_ptr_ok");
+            let ptr_nil = bin.context.append_basic_block(function, "fn_ptr_nil");
+            let nil_ptr = bin
```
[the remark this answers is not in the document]
xermicus: I agree

seanyoung (approved):

Issue #1050, opened 2022-10-24
Substrate: Wrong or unclear behaviour in Try-Catch statements 


xermicus:
**Describe the bug**
In the [documentation](https://solang.readthedocs.io/en/latest/language/statements.html#try-catch-statement), it is stated that 

> There is an alternate syntax which avoids the abi decoding by leaving the catch Error(…) out. This might be useful when no error string is expected ...

How-ever, I found an interesting edge case which does not line up with what the documentation says.

As a side note this documentation page needs some overhaul; the examples are not (no longer?) working properly, I can take care of that.

**To Reproduce**
Consider the following contracts, adapted closely from the documentation mentioned above:

```solidity
contract aborting {
    constructor(bool abort) {
        if (abort) {
            revert("bar");
        }
    }

    function foo() public pure {}
}

contract runner {
    function test() public {
        try new aborting(true) returns (aborting a) {
            // new succeeded; a holds the a reference to the new contract
        } catch Error(string x) {
            if (x == "bar") {
                // "bar" revert or require was executed
            }
        } catch (bytes raw) {
            // if no error string could decoding, we end up here with the raw data
        }
    }
}
```

If the `runner` contract is instantiate without any funds, the underlying `seal_instantiate` call fails (with error code 5 / transfer failed). How-ever, based on what's written in the documentation I'd expect this to be catched inside the `catch (bytes raw)` block. Instead, the contract traps.

For this case, the `seal_instantiate` error _can_ actually be catched and handled, so that the runner contract does _not_ trap, like so:

```solidity
contract aborting {
    constructor(bool abort) {
        if (abort) {
            revert("bar");
        }
    }

    function foo() public pure {}
}

contract runner {
    function test() public {
        try new aborting(true) returns (aborting a) {
            // new succeeded; a holds the a reference to the new contract
        } catch {
            // NOTE: We end actually up here if seal_instantiate runs out of funds
        }
    }
}
```

But this is not documented that way.

**Expected behavior**
I think that either the documentation should be updated to point this out, if this behavior is expected and correct (CC @seanyoung please call me out if I'm wrong here). Or, we make sure the contract does not trap when there is nothing written into the output buffer.

xermicus:
Closed by 1454

PR #1454, opened 2023-07-21
Polkadot: Bubble up reverts

xermicus:
This is a continuation of #1415. To align Solang with `solc` further, any uncaught reverts must be bubbled up back to the caller. To do so, handling of the return code for `call`, `instantiate` and `transfer` runtime API calls is now done in codegen and no longer treated as a bool.

This PR also fixes try-catch in codegen resulting in fixing a bunch of deactivated tests.


src/codegen/statements.rs
```diff
@@ -1162,41 +1166,7 @@ fn try_catch(
...
-                    let decoded = abi_decode(&Codegen, buf, &func_returns, ns, vartab, cfg, None);
-                    for instruction in res.iter().zip(decoded).map(|(var, expr)| Instr::Set {
-                        loc: Codegen,
-                        res: *var,
-                        expr,
-                    }) {
-                        cfg.add(vartab, instruction)
-                    }
-                }
+                (polkadot_ret_switch(loc, success, cfg, vartab), func_returns)
```
LucasSte, line 1169: Shouldn't this call be within an if condition? Although Solana does not support try-catches, if, when building for Solana, we ever reach this function, there might be a problem in our code.

src/codegen/statements.rs
```diff
@@ -1289,10 +1307,45 @@ fn try_catch(
             _ => vartab.temp_anonymous(&Type::String),
         };
 
+        let buf = Expression::Variable {
+            loc: Codegen,
+            ty: Type::DynamicBytes,
+            var_no: error_ret_data_var,
+        };
+
+        // Expect the returned data to contain at least the selector + 1 byte of data
```
LucasSte, line 1316: Isn't this substrate specific code?
Wouldn't it be nice to guard it behind an if statement or perhaps a crate?

tests/codegen_testcases/solidity/unused_variable_elimination.sol
```diff
@@ -66,7 +66,7 @@ contract c {
         x = 102 + t*y/(t+5*y) + g + test3() - vec.push(2) + ct.sum(1, 2);
 		return 2;
 // CHECK: push array ty:int32[] value:int32 2
-// CHECK: _ = external call::regular address:%ct payload:%abi_encoded.temp.94 value:uint128 0 gas:uint64 0 accounts: seeds:
+// CHECK: return int32 2
```
LucasSte, line 69: Isn't the call here anymore?
The purpose of this check it make sure the external call is not eliminated, when simplifying the expression in line 66.
Verifying that it returns 2 is not right here.

src/codegen/statements.rs
xermicus: It is guarded behind an if. The code panics on any other target than Polkadot.

> Wouldn't it be nice to guard it behind an if statement or perhaps a trait?

Error selectors on Solidity are 4 bytes. Why does this need to go behind an if or trait?

The decoding of the error data should really go into a trait. But the PR isn't about fixing error data decoding, it's about bubbling up reverts.

src/codegen/statements.rs
xermicus: This is why I added

```rust
    if !ns.target.is_polkadot() {
        unimplemented!()
    }
```

at the top of this function

tests/codegen_testcases/solidity/unused_variable_elimination.sol
xermicus: Good catch, the call is still there, added the check.

src/codegen/expression.rs
```diff
@@ -3757,3 +3817,137 @@ fn add_prefix_and_delimiter_to_print(mut expr: Expression) -> Expression {
         }
     }
 }
+
+/// Handles all cases from the [RetBlock] accordingly.
+/// * On success, nothing is done and the execution continues at the success block.
+/// If the callee reverted and output was supplied, it will be bubble up.
+/// Otherwise, a revert without data will be inserted.
+fn polkadot_check_ext_call_ret(
```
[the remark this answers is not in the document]
xermicus: The error message is different, but I can change that to an argument to the function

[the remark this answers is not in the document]
xermicus: Refactored this code

src/codegen/statements.rs
xermicus: The range check I added is there because the ABI encoder returns empty error data if it doesn't have enough data. But if the error data is less than 4 bytes, there is no way it matches `Error(string)`. But I just realized there might be another problem, I'm gonna look into it again tomorrow)

xermicus: Yeah the code for generating try catch seems broken. Also having #1449 in `main` will really help writing test cases here

src/codegen/expression.rs
```diff
@@ -3757,3 +3828,186 @@ fn add_prefix_and_delimiter_to_print(mut expr: Expression) -> Expression {
...
+    let msg = "value transfer failure";
+    log_runtime_error(opt.log_runtime_errors, msg, *loc, cfg, vartab, ns);
+    cfg.add(vartab, Instr::AssertFailure { encoded_args: None });
+
+    cfg.set_basic_block(success_block);
+
+    None
+}
+
+pub(super) mod polkadot {
```
seanyoung, line 3886: Why do we need a new module here? I think it should be in a new file if it needs to be in a module.

seanyoung (approved):
Nice to see less code in emit and more in codegen

src/codegen/expression.rs
```diff
@@ -3757,3 +3828,186 @@ fn add_prefix_and_delimiter_to_print(mut expr: Expression) -> Expression {
         }
     }
 }
+
+/// Check the return code of `transfer`.
+///
+/// If `bubble_up` is set to true, this will revert the contract execution on failure.
+/// Otherwise, the comparison if the return code equals `0` is returned.
```
LucasSte, line 3835: This line is unclear.

src/codegen/expression.rs
LucasSte: A new file is a good idea.

LucasSte (approved):

xermicus:
Closes #1050

src/codegen/polkadot.rs
```diff
@@ -0,0 +1,179 @@
...
+            success_var: self.success_var,
+        }
+    }
+
+    pub(crate) fn msg(self, msg: &'static str) -> Self {
+        Self {
+            loc: self.loc,
+            msg,
+            success_var: self.success_var,
+        }
```
seanyoung, line 47:
```rust
self.msg = msg;
self
```
Seems less error prone (e.g. if we add a field)?

src/codegen/polkadot.rs
```diff
@@ -0,0 +1,179 @@
...
+    success_var: Success,
+}
+
+impl RetCodeCheckBuilder {
+    pub(crate) fn loc(self, loc: Loc) -> Self {
+        Self {
+            loc,
+            msg: self.msg,
+            success_var: self.success_var,
+        }
```
seanyoung, line 39: same as below

src/codegen/polkadot.rs
```diff
@@ -0,0 +1,179 @@
...
+pub(crate) struct RetCodeCheck {
+    pub success: usize,
+    pub revert: usize,
+    pub error_no_data: usize,
+    msg: &'static str,
+    loc: Loc,
+}
+
+#[derive(Default)]
+pub(crate) struct RetCodeCheckBuilder<Success = ()> {
```
seanyoung, line 27: Why the generic parameter when it is always going to be a `usize`?

src/sema/ast.rs
```diff
@@ -1695,7 +1695,7 @@ pub struct TryCatch {
     pub errors: Vec<(Option<usize>, Parameter, Vec<Statement>)>,
     pub catch_param: Option<Parameter>,
     pub catch_param_pos: Option<usize>,
-    pub catch_stmt: Vec<Statement>,
+    pub catch_stmt: Option<Vec<Statement>>,
```
seanyoung, line 1698: So we have three options:
1. No `catch (e) { }` statement: all options are `None`
2. catch without parameter `catch { }`: `catch_stmt`: `Some` and `catch_param(_pos)`: `None`
3. catch with parameter: `catch (bytes x) {}`: all `Some`

Maybe it would make more sense to have:
```rust
pub struct CatchCase {
    pub param: Option<Parameter>,
    pub param_pos: Option<usize>,
    pub stmt: Vec<Statement>,
}
```
and then `struct TryCatch` can contain:
```rust
      pub cases: Vec<CatchCase>,
      pub catch_all: Option<CatchCase>,
```

tests/create_me/flipper.json
```diff
@@ -0,0 +1,75 @@
+{
+  "version": "0.2.1",
+  "name": "flipper",
+  "instructions": [
+    {
+      "name": "new",
+      "docs": [
```
seanyoung, line 7: Remove this (should probably be in `.gitignore` or temp directory

seanyoung (changes requested):

src/codegen/polkadot.rs
xermicus: I don't think it matters as it's purely a question of style. I changed it for now (also this was already ✅ before and not related to what I changed since then)

src/codegen/polkadot.rs
xermicus: This is a typestate builder

tests/create_me/flipper.json
xermicus: added to gitignore

src/sema/ast.rs
xermicus: Yeah we could do something like this, it's more intuitive. I changed it

LucasSte (changes requested):

src/codegen/statements/try_catch.rs
```diff
@@ -0,0 +1,591 @@
...
+            callee_contract_no,
+            ns,
+            vartab,
+            loops,
+            placeholder,
+            return_override,
+            opt,
+        );
+
+        reachable = stmt.reachable();
```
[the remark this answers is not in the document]
seanyoung: `|=` would mean that any of the statements are reachable. We're really only interested if the last statement is reachable (reachable means "does the control flow continue _after_ this statement).

seanyoung (approved):

src/bin/languageserver/mod.rs
```diff
@@ -492,14 +492,20 @@ impl<'a> Builder<'a> {
             }
             ast::Statement::TryCatch(_, _, try_stmt) => {
                 self.expression(&try_stmt.expr, symtab);
-                for stmt in &try_stmt.catch_stmt {
-                    self.statement(stmt, symtab);
+                if let Some(stmts) = &try_stmt
+                    .catch_all
+                    .as_ref()
+                    .map(|clause| clause.stmt.clone())
```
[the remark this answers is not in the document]
xermicus: No

src/codegen/statements/try_catch.rs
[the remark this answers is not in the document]
xermicus: Exactly

src/sema/dotgraphviz.rs
```diff
@@ -1838,13 +1842,13 @@ impl Dot {
...
-                        &try_catch.catch_stmt,
-                        func,
-                        ns,
-                        parent,
-                        String::from("catch"),
-                    );
+                    if let Some(stmts) = try_catch
+                        .catch_all
+                        .as_ref()
+                        .map(|clause| clause.stmt.clone())
```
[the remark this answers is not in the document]
xermicus: No

src/sema/mutability.rs
```diff
@@ -278,10 +278,16 @@ fn recurse_statements(stmts: &[Statement], ns: &Namespace, state: &mut StateChec
...
                 recurse_statements(&try_catch.ok_stmt, ns, state);
-                for (_, _, s) in &try_catch.errors {
-                    recurse_statements(s, ns, state);
+                for clause in &try_catch.errors {
+                    recurse_statements(&clause.stmt, ns, state);
+                }
+                if let Some(stmts) = try_catch
+                    .catch_all
+                    .as_ref()
+                    .map(|clause| clause.stmt.clone())
```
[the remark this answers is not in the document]
xermicus: No

src/codegen/expression.rs
```diff
@@ -2766,14 +2796,30 @@ pub fn emit_function_call(
                 },
             );
 
-            vec![
+            let success = if ns.target.is_polkadot() {
+                let ret_code = Expression::Variable {
+                    loc: *loc,
+                    ty: Type::Uint(32),
+                    var_no: success,
```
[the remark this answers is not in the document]
xermicus: > Is there any test for an ExternalFunctionCallRaw?

Yes; It's not valid source code and `sema` refuses to compile that.

> success is declared as a boolean in line 2776. I don't think you can change the type here to Type::Uint(32) without a cast.

Good point. I didn't think too closely about it.

src/emit/polkadot/target.rs
```diff
@@ -1058,44 +1035,20 @@ impl<'a> TargetRuntime<'a> for PolkadotTarget {
...
-                .build_conditional_branch(is_success, success_block, bail_block);
-
-            binary.builder.position_at_end(bail_block);
-
-            binary.log_runtime_error(self, "external call failed".to_string(), Some(loc), ns);
-            self.assert_failure(binary, byte_ptr!().const_null(), i32_zero!());
-
-            binary.builder.position_at_end(success_block);
         }
+        .into();
```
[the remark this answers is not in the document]
xermicus: I mean it's possible if I a cast the ret from the regular call into a basic value enum

tests/codegen_testcases/solidity/unused_variable_elimination.sol
```diff
@@ -115,42 +122,42 @@ return 3;
...
+    // BEGIN-CHECK: c3::function::test10
     function test10() public view returns (int) {
         int f = 4;
 
-        int c = 32 +4 *(f = it1+it2);
-// CHECK: ty:int256 %c = (int256 32 + (sext int256 (int64 4 * (trunc int64 (%temp.126 + %temp.127)))))
-// NOT-CHECK: ty:int256 %f = (%temp.10 + %temp.11)
+        int c = 32 + 4 * (f = it1 + it2);
+        // CHECK: ty:int256 %c = (int256 32 + (sext int256 (int64 4 * (trunc int64 (%temp.130 + %temp.131)))))
+        // NOT-CHECK: ty:int256 %f = (%temp. + %temp.)
```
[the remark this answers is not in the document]
xermicus: Yeah good point. I mean this was very subtle before too. What do you think would be a better way for writing the NOT-CHECK so that temp var numbers don't play any role?

src/codegen/statements/try_catch.rs
```diff
@@ -0,0 +1,591 @@
...
+        kind: Builtin::ReadFromBuffer,
+        args: vec![buffer.clone(), offset],
+    };
+
+    cfg.set_basic_block(enough_data_block);
+    // Expect the returned data to match the 4 bytes function selector for "Error(string)"
+    let err_id = Expression::NumberLiteral {
+        loc: Codegen,
+        ty: Type::Bytes(4),
+        value: 0x08c3_79a0.into(),
```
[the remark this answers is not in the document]
xermicus: Yeah why not

src/bin/languageserver/mod.rs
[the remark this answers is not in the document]
xermicus: No

.gitignore
```diff
@@ -4,6 +4,7 @@ Cargo.lock
 **/*.rs.bk
 *.ll
 tests/.tmp*
+tests/create_me/flipper.json
```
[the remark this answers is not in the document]
xermicus: Yeah I think so

tests/codegen_testcases/solidity/unused_variable_elimination.sol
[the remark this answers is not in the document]
xermicus: I just removed the numbers so it just checks for the `f` var assignment

src/sema/dotgraphviz.rs
```diff
@@ -1806,24 +1806,28 @@ impl Dot {
...
 
-                        self.add_statement(stmt, func, ns, parent, String::from("error"));
+                        self.add_statement(&clause.stmt, func, ns, parent, String::from("error"));
                     }
 
-                    if let Some(param) = &try_catch.catch_param {
+                    if let Some(param) = try_catch
+                        .catch_all
+                        .as_ref()
+                        .and_then(|clause| clause.param.clone())
```
LucasSte, line 1829: Is the clone necessary?

src/codegen/expression.rs
```diff
@@ -2766,14 +2796,35 @@ pub fn emit_function_call(
...
+                let ret_ok = Expression::NumberLiteral {
+                    loc: *loc,
+                    ty: Type::Uint(32),
+                    value: 0.into(),
+                };
+                Expression::Equal {
+                    loc: *loc,
+                    left: ret_code.into(),
+                    right: ret_ok.into(),
+                }
```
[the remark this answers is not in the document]
xermicus: So on the polkadot target, `success` is an u32. But since you said earlier that on Solana this variable isn't used anyways. I just turned the `success` var into an u32 instead. WDYT @LucasSte

src/sema/dotgraphviz.rs
xermicus: The clone here is just for convenience, it can be written differently

src/codegen/expression.rs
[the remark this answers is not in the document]
LucasSte: That's a good idea. I see in all other places (where you check the target before creating a temporary) you are creating the `success` variable as a boolean. Should those be `Type::Uint(32)` too?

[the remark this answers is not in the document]
xermicus: Well they might be all bools is because that's what you get on EVM for raw calls. But we don't support EVM for now, so I guess yes

src/codegen/expression.rs
```diff
@@ -2766,14 +2796,30 @@ pub fn emit_function_call(
...
+            } else {
                 Expression::Variable {
                     loc: *loc,
                     ty: Type::Bool,
```
LucasSte, line 2818:
```suggestion
                    ty: Type::Uint(32),
```

src/codegen/expression.rs
```diff
@@ -2844,10 +2890,14 @@ pub fn emit_function_call(
                     .as_ref()
                     .map(|expr| expression(expr, cfg, caller_contract_no, func, ns, vartab, opt));
 
+                let success = ns
+                    .target
+                    .is_polkadot()
+                    .then(|| vartab.temp_name("success", &Type::Bool));
```
LucasSte, line 2896:
```suggestion
                    .then(|| vartab.temp_name("success", &Type::Uint(32)));
```

src/codegen/expression.rs
```diff
@@ -2917,11 +2976,14 @@ pub fn emit_function_call(
                     .flags
                     .as_ref()
                     .map(|expr| expression(expr, cfg, caller_contract_no, func, ns, vartab, opt));
-
+                let success = ns
+                    .target
+                    .is_polkadot()
+                    .then(|| vartab.temp_name("success", &Type::Bool));
```
LucasSte, line 2982:
```suggestion
                    .then(|| vartab.temp_name("success", &Type::Uint(32)));
```

tests/polkadot_tests/calls.rs
```diff
@@ -1027,3 +1028,178 @@ fn constructors_and_messages_distinct_in_dispatcher() {
...
+        true,
+    );
+
+    // Expect the contract to catch the reverting child constructor
+    runtime.function("test", 0u128.encode());
+    assert_eq!(runtime.output(), vec![0x41u8].encode());
+    assert!(runtime.debug_buffer().contains("seal_instantiate=2"));
+
+    // Trying to instantiate with value while having insufficient funds result in
+    // seal_instantiate faliing with transfer failed (return code 5).
```
LucasSte, line 1199:
```suggestion
    // seal_instantiate failing with transfer failed (return code 5).
```

LucasSte (approved):

src/codegen/polkadot.rs
xermicus: Well, the typestate enforces 1) that the `success` var was set, as the impl of <success = ()> doesn't implement `insert` to conclude the builder (only `<Success = usize> does implement it) and 2) allow exact behavior regarding to at what state which variable is allowed to be set, all at compile time. I implemented it this way originally because I thought it was necessary while it's actually not. But I'll leave it in place now, I mean it's all compile time and makes for an easily extendable interface.

PR #1455, opened 2023-07-23
Bugfix: Do not make assumptions about storage variable reads

xermicus:
Contracts can get upgraded so we must not make any assumptions whether a storage variable is read or not; writes to storage variables must happen regardless whether the variable is never read.


seanyoung (approved):
The test is a bit strange. Would it not be easier to have a contract which writes to an unread storage variable, and then ensure that the actual storage slot 0 contains the value?

xermicus:
Sure, I can add a direct check on the storage. I wrote the test like that because it also highlights why we want this behavior.

LucasSte:
How does this affect upgradable contracts?
If a variable is declared without a visibility identifier, it is `internal`, meaning it can only be accessed inside the contract and its derived ones.

LucasSte:
> How does this affect upgradable contracts?
> If a variable is declared without a visibility identifier, it is `internal`, meaning it can only be accessed inside the contract and its derived ones.

I think we only remove assignments if the variables are not public.

LucasSte:
Does this problem arise when we upgrade the contract and change the variable's visibility?

xermicus:
> Does this problem arise when we upgrade the contract and change the variable's visibility?

This is one case. The upgraded contract could also use the variable internally.

> I think we only remove assignments if the variables are not public.

`public` storage variables have getters, this is why

LucasSte (approved):

PR #1458, opened 2023-07-23
Polkadot does not yet support catching Panics or custom errors

xermicus:
This isn't implemented yet in codegen and it won't make it into the July release (planned for August).

### DIFF
--- a/tests/evm.rs
+++ b/tests/evm.rs
@@ -248,7 +248,7 @@ fn ethereum_solidity_tests() {
         })
         .sum();
 
-    assert_eq!(errors, 1030);
+    assert_eq!(errors, 1029);
 }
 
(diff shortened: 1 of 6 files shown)


tests/evm.rs
```diff
@@ -248,7 +248,7 @@ fn ethereum_solidity_tests() {
         })
         .sum();
 
-    assert_eq!(errors, 1029);
+    assert_eq!(errors, 1030);
```
seanyoung, line 251: That's a regression for evm

xermicus: Hmm yeah, fixed. It's weird that at the same time we support EVM target and we also don't. I keep forgetting about it.

LucasSte (approved):

Issue #1461, opened 2023-07-25
Accessing function selectors adds foreign contracts functions


xermicus:
If a contract function accesses the selector of another contracts function like so:

```solidity
contract C {
    function ext_func_call(uint128 amount) public payable {
        A a = new A();
        (bool ok, bytes b) = address(a).call(
            bytes4(A.a.selector) // Here
        );
    }
}

contract A {
    function a() public pure {}
}
```

This will add function `A::a` to contract `C`. The dispatcher dispatches it and it's present in the metadata, which means it gets erroneously added and treated as a contract function further up the compilation pipeline.

I can reproduce this back to Solang `v0.2.0`; earlier versions segfault in `emit`  so it's unclear when this regression was introduced (not recently).

xermicus:
I can reproduce it for Solana too:

> /home/cyrill/Downloads/solang-linux-x86-64 --version
> solang version v0.2.0


> /home/cyrill/Downloads/solang-linux-x86-64 compile --target bug.sol --emit cfg


```
# function dispatch_C public:false selector: nonpayable:false
# params: buffer_pointer,uint64
# returns: 
block0: # entry
	branchcond ((arg #1) >= uint64 4), block1, block2
block1: # switch
	switch (builtin ReadFromBuffer ((arg #0), uint32 0)):
		case uint32 2051782505: goto block #3
		case uint32 3369171261: goto block #8
		case uint32 526892557: goto block #11
		default: goto block #2
```

526892557 is the selector of `A::a`.

seanyoung:
Yes, you're right. That's not great. :grimacing:

seanyoung:
Fixed by #1462

PR #1462, opened 2023-07-26
a.b.selector adds b() as function to current contract

seanyoung:
Compile the Solidity below and we end up with the non-base function in our contract.

```bash
$ solang compile --target polkadot --emit cfg
# function C::A::function::a public:true selector:0dbe671f nonpayable:true
# params: 
# returns: 
block0: # entry
	return 
```

```solidity
contract C {
    function ext_func_call(uint128 amount) public payable {
        A a = new A();
        (bool ok, bytes b) = address(a).call(
            bytes4(A.a.selector)
        );
	b = abi.encodeCall(A.a);
    }
}

contract A {
    function a() public pure {}
}
```

### DIFF
--- a/src/codegen/external_functions.rs
+++ b/src/codegen/external_functions.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::sema::ast::{DestructureField, Expression, Namespace, Statement};
+use crate::sema::ast::{Builtin, DestructureField, Expression, Namespace, Statement};
 use crate::sema::Recurse;
 use indexmap::IndexSet;
 use solang_parser::pt;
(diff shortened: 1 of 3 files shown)


seanyoung:
Fixes #1461

src/codegen/external_functions.rs
```diff
@@ -98,6 +111,20 @@ fn check_expression(expr: &Expression, call_list: &mut CallList) -> bool {
         | Expression::InternalFunction { function_no, .. } => {
             call_list.solidity.insert(*function_no);
         }
+        Expression::Builtin {
```
xermicus, line 114: Good catch, I thought there might be more

xermicus (approved):
LGTM

src/codegen/external_functions.rs
seanyoung: That's why we have the assert

tests/codegen_testcases/solidity/selector_adds_non_base_function.sol
```diff
@@ -0,0 +1,19 @@
+// RUN: --target polkadot --emit cfg
+
+// This test makes sure that we're not including the function a in contract C
+
+// NOT-CHECK: C::A::function::a
```
LucasSte, line 5: This test does not fail if I comment the changes in `src/codegen/external_functions.rs`

seanyoung: Good point, fixed now

tests/codegen.rs
```diff
@@ -35,6 +35,7 @@ fn run_test_for_path(path: &str) {
 #[derive(Debug)]
 enum Test {
     Check(String),
+    CheckAbsent(String),
     NotCheck(String),
```
LucasSte, line 39: When should I use `CHECK-ABSENT` and when should I use `NOT-CHECK`?

LucasSte (approved):

tests/codegen.rs
seanyoung: I've added some comments

PR #1465, opened 2023-07-27
Relative imports are only done with relative paths and expose import_no

seanyoung:
When importing in Solidity, `import "foo";` does not check path relative to the currrent file, only `import "./foo";` does.

This work is also a re-write of https://github.com/hyperledger/solang/pull/1443

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to [Solang](https://github.com/hyperledger/solang/)
 will be documented here.
 
+## Unreleased
+
+### Fixed
+- **breaking** Resolving import paths now matches solc more closely, and only resolves relative
+  paths when specified as `./foo` or `../foo`. [seanyoung](https://github.com/seanyoung)
+
 ## v0.3.1 Göttingen
 
 ### Added
@@ -11,7 +17,7 @@ will be documented here.
 - The `wasm-opt` optimizer now optimizes the Wasm bytecode on the Substrate target. [xermicus](https://github.com/xermicus)
 - Call flags are now available for Substrate. [xermicus](https://github.com/xermicus)
 - Read compiler configurations from toml file. [salaheldinsoliman](https://github.com/salaheldinsoliman)
-- Accounts declared with `@payer(my_account)` can be accessed with the 
+- Accounts declared with `@payer(my_account)` can be accessed with the
   syntax `tx.accounts.my_account`. [LucasSte](https://github.com/LucasSte)
 - `delegatecall()` builtin has been added for Substrate. [xermicus](https://github.com/xermicus)
 - `get_contents_of_file_no` for Solang parser. [BenTheKush](https://github.com/BenTheKush)
@@ -52,7 +58,7 @@ will be documented here.
   contract MyContract {
     @payer(acc) // Declares account acc
     @space(2+3) // Only literals or constant expressions allowed
-    constructor(@seed bytes myseed) {} 
+    constructor(@seed bytes myseed) {}
     // When an annotations refers to a parameter, the former must appear right before the latter.
   }
   ```
--- a/src/file_resolver.rs
+++ b/src/file_resolver.rs
@@ -106,7 +106,7 @@ impl FileResolver {
     /// Atempt to resolve a file, either from the cache or from the filesystem.
     /// Returns Ok(Some(..)) if the file is found and loaded
     /// Returns Ok(None) if no file by this path can be found.
-    /// Returns Err(..) if a file could be found but not could not be read.
+    /// Returns Err(..) if a file was found but could not be read.
     fn try_file(
         &mut self,
         filename: &OsStr,
--- a/src/file_resolver.rs
+++ b/src/file_resolver.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::sema::ast;
+use normalize_path::NormalizePath;
 use solang_parser::pt::Loc;
 use std::collections::HashMap;
 use std::ffi::{OsStr, OsString};
@@ -10,13 +11,14 @@ use std::io::{prelude::*, Error, ErrorKind};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
+#[derive(Default)]
 pub struct FileResolver {
     /// Set of import paths search for imports
     import_paths: Vec<(Option<OsString>, PathBuf)>,
-    /// List file by import path
+    /// List file by path
     cached_paths: HashMap<PathBuf, usize>,
     /// The actual file contents
-    files: Vec<Arc<str>>,
+    files: Vec<ResolvedFile>,
 }
 
 /// When we resolve a file, we need to know its base compared to the import so
@@ -25,37 +27,17 @@ pub struct FileResolver {
 /// user exactly which file has errors/warnings.
 #[derive(Clone, Debug)]
 pub struct ResolvedFile {
+    /// Original name used on cli or import statement
+    pub path: OsString,
     /// Full path on the filesystem
     pub full_path: PathBuf,
     /// Which import path was used, if any
-    import_no: usize,
-    // Base part relative to import
-    base: PathBuf,
-}
-
-impl ResolvedFile {
-    /// Get the import number associated with this `ResolvedFile`
-    pub fn get_import_no(&self) -> usize {
-        self.import_no
-    }
-}
-
-impl Default for FileResolver {
-    fn default() -> Self {
-        FileResolver::new()
-    }
+    pub import_no: Option<usize>,
+    /// The actual file contents
+    pub contents: Arc<str>,
 }
 
 impl FileResolver {
-    /// Create a new file resolver
-    pub fn new() -> Self {
-        FileResolver {
-            import_paths: Vec::new(),
-            cached_paths: HashMap::new(),
-            files: Vec::new(),
-        }
-    }
-
     /// Add import path
     pub fn add_import_path(&mut self, path: &Path) -> io::Result<()> {
         self.import_paths.push((None, path.canonicalize()?));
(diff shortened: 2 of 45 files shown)


bkushigian:
> When importing in Solidity, `import "foo";` does not check path relative to the currrent file, only `import "./foo";` does.
> 
> This work is also a re-write of #1443

Yeah this looks good. Having `ResolvedFile` contain the contents is a nice solution, conceptually cleaner from my standpoint.

bkushigian:
Also, while we are poking around this area, there is a difference in how solang and solc handle imports.

In the following structure:
```
proj/
+- contracts/
   +- A.sol
   +- B.sol
```

where `A.sol` has `import "B.sol";`

Running `solc contracts/A.sol` from `proj` will break because `B.sol` is not on the base-path or the include-path, while `solang contracts/A.sol` will succeed because it looks for `B.sol` by appending the import path to `A.sol`'s base name (which results in `contracts/A.sol`).

This is [documented](https://solang.readthedocs.io/en/v0.3.1/language/imports.html):

> Solang will look for the file defines.sol in the same directory as the current file

but diverges from solc behavior, which defaults to an empty import path when no `--base-name` or `--include-path` is specified (I've verified that compiling this w/ solc fails if no `--base-name` is specified). It might be worth noting that the import semantics differ from solc's in the compilation docs? I imagine this could cause some weird errors if switching from one compiler to another.

seanyoung:
> Also, while we are poking around this area, there is a difference in how solang and solc handle imports.
> 
> In the following structure:
> 
> ```
> proj/
> +- contracts/
>    +- A.sol
>    +- B.sol
> ```
> 
> where `A.sol` has `import "B.sol";`
> 
> Running `solc contracts/A.sol` from `proj` will break because `B.sol` is not on the base-path or the include-path, while `solang contracts/A.sol` will succeed because it looks for `B.sol` by appending the import path to `A.sol`'s base name (which results in `contracts/A.sol`).

This PR fixes that (as in, the same behavior as solc)

> This is [documented](https://solang.readthedocs.io/en/v0.3.1/language/imports.html):
> 
> > Solang will look for the file defines.sol in the same directory as the current file

I've amended the documentation in this PR now.

> but diverges from solc behavior, which defaults to an empty import path when no `--base-name` or `--include-path` is specified (I've verified that compiling this w/ solc fails if no `--base-name` is specified). It might be worth noting that the import semantics differ from solc's in the compilation docs? I imagine this could cause some weird errors if switching from one compiler to another.

I've added a `--basepath` to solang in this PR.

Let us know what you think.

bkushigian:
> This PR fixes that (as in, the same behavior as solc)

Great!

> I've added a `--basepath` to solang in this PR.

Is `--basepath` different than `--importpath`? I like Solang's current interface import paths. From my reading of the solc docs `--base-path` and `--include-paths` are the exact same thing, and they are separate for historical/bad design reasons.

I'd argue that if `import "A.sol";` is treated as a direct import (resolved against import paths) and not as a relative import (resolved against parent's base path) then this will match solc's import semantics and also provide a much cleaner/less confusing interface.

bkushigian:
(it's also possible that I'm missing something about solc's `--base-path` vs `--include-path`, but I've spent quite a bit of time trying to find the difference and I can't)

bkushigian:
Oh, here's one test to see if imports work the same as solc (and also an illustration of how horrible solc's import system is). 

```
import "/absolute/path.sol";
````

with no basepath specified in solc will treat the absolute path as an absolute path. However, if we specify `--base-path .` this import will try to look up `./absolute/path.sol` and will not look at file location `/absolute/path.sol`.

This is cursed af, but it's what solc does, so 🤷

**Comment by seanyoung** [comment]
> > 
> 
> Is `--basepath` different than `--importpath`? I like Solang's current interface for import paths. From my reading of the solc docs `--base-path` and `--include-paths` are the exact same thing, and they are separate for historical/bad design reasons.

When you do `solang compile foo.sol`, then `.` is added as an import path otherwise `foo.sol` won't be found. `--basepath` adds a different path than `.`. This is not very clear at all, so we probably need something better.

>  However, if we specify --base-path . this import will try to look up ./absolute/path.sol and will not look at file location /absolute/path.sol.
>
> This is cursed af, but it's what solc does, so :shrug:

That is insane. `--allow-path` is also an abomination. Definitely don't want to go down that path.

**Comment by bkushigian** [comment]
> When you do `solang compile foo.sol`, then `.` is added as an import path otherwise `foo.sol` won't be found. `--basepath` adds a different path than `.`. This is not very clear at all, so we probably need something better.

Does `solang compile foo.sol` temporarily add `.` just long enough to resolve `foo.sol`? Or is `.` permanently added as an import path? I'm not actually sure how this should be handled. I wonder if there is a way to allow a FileResolver to resolve a canonical path without needing an import root? Because running `solang compile foo.sol` isn't really importing anything. But again, not sure, this is a thing that's caused me a bit of confusion.

> `--allow-path` is also an abomination`

Yeah, it's very bad. Its implementation is also broken and departs from the docs (at least in 7.x, haven't run the same experiment in 8.x). For instance, I have a build where you need to specify `--base-path . --allow-path .`, but according to the docs `--base-path .` should automatically allow `.`.

**Comment by seanyoung** [comment]
> > When you do `solang compile foo.sol`, then `.` is added as an import path otherwise `foo.sol` won't be found. `--basepath` adds a different path than `.`. This is not very clear at all, so we probably need something better.
> 
> Does `solang compile foo.sol` temporarily add `.` just long enough to resolve `foo.sol`? Or is `.` permanently added as an import path? I'm not actually sure how this should be handled. I wonder if there is a way to allow a FileResolver to resolve a canonical path without needing an import root? Because running `solang compile foo.sol` isn't really importing anything. But again, not sure, this is a thing that's caused me a bit of confusion.

That is a good point! There is no reason to do this. I've removed this totally so there is no `--basepath` and no import path `.` added. Should be more straightforward now.

File: CHANGELOG.md [hunk 1]
```diff
@@ -2,6 +2,12 @@
 All notable changes to [Solang](https://github.com/hyperledger/solang/)
 will be documented here.
 
+## Unreleased
+
+### Fixed
+- Resolving import paths now matches solc more closely, and only resolves relative
+  paths when specified as `./foo` or `../foo`. [seanyoung](https://github.com/seanyoung)
```
LucasSte, line 9: This is a breaking change, isn't it? Perhaps we should mark it with `**BREAKING**`

LucasSte (commented):
Do we have a test in which we set an import path and verify if the files listed as `import 'my_file.sol'` are correctly imported?

CHANGELOG.md
seanyoung: Yes, it is. Added.

seanyoung:
> Do we have a test in which we set an import path and verify if the files listed as `import 'my_file.sol'` are correctly imported?

We have https://github.com/hyperledger/solang/blob/main/tests/imports_testcases/import.sol which is run from https://github.com/hyperledger/solang/blob/main/tests/imports.rs#L90

LucasSte (approved):

src/file_resolver.rs
```diff
@@ -96,28 +78,66 @@ impl FileResolver {
...
     /// Get file with contents. This must be a file which was previously
     /// add to the cache
     pub fn get_file_contents_and_number(&self, file: &Path) -> (Arc<str>, usize) {
         let file_no = self.cached_paths[file];
 
-        (self.files[file_no].clone(), file_no)
+        (self.files[file_no].contents.clone(), file_no)
+    }
+
+    /// Atempt to resolve a file, either from the cache or from the filesystem
```
xermicus, line 106: Could explain what a `Ok(None)` return value means and in what cases an `Err` is returned instead

src/file_resolver.rs
```diff
@@ -156,107 +181,57 @@ impl FileResolver {
     ) -> Result<ResolvedFile, String> {
         let path = PathBuf::from(filename);
 
-        let path = if let Ok(m) = path.strip_prefix("./") {
-            m.to_path_buf()
-        } else {
-            path
-        };
+        // relative path
```
xermicus, line 184: This comment confuses me: Aren't relative paths all paths except ones starting with a slash `/`? Where `./` and `../` are just special cases of relative paths.

Also, what about other OS, does `PathBuf::from(filename)` normalize the path?

seanyoung: In Solidity, when you do `import "foo.sol";` then relative paths are not considered; that's one of things that this PR is fixing.

seanyoung: > Also, what about other OS, does PathBuf::from(filename) normalize the path?

That just creates a `PathBuf` from a `String` or `OsString`.

xermicus: > In Solidity, when you do import "foo.sol"; then relative paths are not considered; that's one of things that this PR is fixing.

Sure, that's why I thought the comment could explain that (or maybe even better in the doc comment), because this isn't apparent by the code or by the comment. Sorry I wasn't entirely clear about this first hand.

> That just creates a PathBuf from a String or OsString.

I think that on Windows, backslashes are used for those special relative paths. It's not obvious to me what should happen in that case. or are we just saying that the import path must always be a unix path?

seanyoung: Should all be fixed now

src/file_resolver.rs
seanyoung: Comment added

src/file_resolver.rs
```diff
@@ -96,28 +78,69 @@ impl FileResolver {
...
         let file_no = self.cached_paths[file];
 
-        (self.files[file_no].clone(), file_no)
+        (self.files[file_no].contents.clone(), file_no)
+    }
+
+    /// Atempt to resolve a file, either from the cache or from the filesystem.
+    /// Returns Ok(Some(..)) if the file is found and loaded
+    /// Returns Ok(None) if no file by this path can be found.
+    /// Returns Err(..) if a file could be found but not could not be read.
```
xermicus, line 109:
```suggestion
    /// Returns Err(..) if a file was found but could not be read.
```

xermicus (approved):

PR #1466, opened 2023-07-27
Ensure build works with latest contract-build and contract-metadata

seanyoung:
The build is failing with errors about missing argument. The argument is for a container image name which we don't support.


LucasSte (approved):

seanyoung:
@xermicus I am going to merge this since the build is broken. Let me know if you have any concerns and we can address them later.

Issue #1430, opened 2023-07-06
Represent Solana's contracts by their program id


LucasSte:
This feature implies the following changes:

- [x] `Type::Contract` on Solana saves a program id under the hood, instead of a data account.
- [x] `this` must return the contract's program id.
- [x] `tx.program_id` can be eliminated.
- [x] Variables of type external function save the contract's program id and a discriminator.
- [x] The builtin `tx.accounts.dataAccount` (or another name) is always available. If it is used, the data account is going to be added to the IDL. It shouldn't be allowed in pure functions.
- [x] If we read or modify a state variable, the data account is going to be automatically added to the IDL.
- [x] Update the documentation and examples accordingly.
- [x] New syntax for representing contracts (they are not available as types anymore).

PR #1468, opened 2023-07-28
Decouple data account from contract in the Solana mock VM

LucasSte:
As we start representing contracts by their program id (#1430), we bump into an incompatibility in our VM. The program id and the contract's data account are tightly coupled!

In order for us to decouple them from Solang as whole, we need to refactor the VM so that it accepts contracts without a data account.

### DIFF
--- a/tests/solana.rs
+++ b/tests/solana.rs
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::borsh_encoding::{decode_at_offset, encode_arguments, BorshToken};
-use anchor_syn::idl::Idl;
+use anchor_syn::idl::{Idl, IdlAccountItem};
 use base58::{FromBase58, ToBase58};
 use byteorder::{ByteOrder, LittleEndian, WriteBytesExt};
+use itertools::Itertools;
 use libc::c_char;
 use rand::Rng;
 use serde::{Deserialize, Serialize};
@@ -17,12 +18,9 @@ use solana_rbpf::{
     verifier::{RequisiteVerifier, TautologyVerifier},
     vm::{BuiltinProgram, Config, ContextObject, EbpfVm, ProgramResult, StableResult},
 };
+use solang::abi::anchor::discriminator;
 use solang::{
-    abi::anchor::{discriminator, generate_anchor_idl},
-    compile,
-    file_resolver::FileResolver,
-    sema::ast,
-    Target,
+    abi::anchor::generate_anchor_idl, compile, file_resolver::FileResolver, sema::ast, Target,
 };
 use std::{
     cell::{RefCell, RefMut},
@@ -80,9 +78,8 @@ type CallParametersCheck = fn(vm: &VirtualMachine, instr: &Instruction, pda: &[P
 
 struct VirtualMachine {
     account_data: HashMap<Account, AccountState>,
-    origin: Account,
-    programs: Vec<Contract>,
-    stack: Vec<Contract>,
+    programs: Vec<Program>,
+    stack: Vec<Program>,
     logs: String,
     events: Vec<Vec<Vec<u8>>>,
     return_data: Option<(Account, Vec<u8>)>,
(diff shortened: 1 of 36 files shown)


tests/solana.rs
```diff
@@ -1757,3 +1600,149 @@ pub fn parse_and_resolve(src: &'static str, target: Target) -> ast::Namespace {
...
     solang::parse_and_resolve(OsStr::new("test.sol"), &mut cache, target)
 }
+
+struct VmFunction<'a, 'b> {
+    vm: &'a mut VirtualMachine,
+    idx: usize,
+    _expected: u64,
+    _accounts: Vec<AccountMeta>,
+    has_remaining: bool,
+    _arguments: Option<&'b [BorshToken]>,
```
seanyoung, line 1610: Why  the ugly underscores?

tests/solana.rs
```diff
@@ -36,7 +34,7 @@ use std::{
 };
 use tiny_keccak::{Hasher, Keccak};
 
-mod borsh_encoding;
+pub mod borsh_encoding;
```
seanyoung, line 37: Is this necessary? Works fine without `pub` here

tests/solana.rs
```diff
@@ -80,20 +78,18 @@ type CallParametersCheck = fn(vm: &VirtualMachine, instr: &Instruction, pda: &[P
...
     events: Vec<Vec<Vec<u8>>>,
     return_data: Option<(Account, Vec<u8>)>,
     call_params_check: HashMap<Pubkey, CallParametersCheck>,
 }
 
 #[derive(Clone)]
-struct Contract {
-    program: Account,
+struct Program {
+    id: Account,
```
seanyoung, line 91: I am not keen on `id`. I don't like `program_id` either, I think it should have been called `program_account` or so. But we're stuck with it, so I guess we just have to live with `id`. :cry:

tests/solana.rs
```diff
@@ -1748,6 +1577,20 @@ impl VirtualMachine {
...
         }
     }
+
+    fn initialize_data_account(&mut self) -> Account {
+        let address = account_new();
+        self.account_data.insert(
+            address,
+            AccountState {
+                lamports: 0,
+                data: [0; 4096].to_vec(),
```
seanyoung, line 1587: `vec![0; 4096]`

tests/solana.rs
```diff
@@ -1757,3 +1600,149 @@ pub fn parse_and_resolve(src: &'static str, target: Target) -> ast::Namespace {
...
+        }
+
+        let return_data = if let Some((_, return_data)) = &self.vm.return_data {
+            return_data.as_slice()
+        } else {
+            &[]
+        };
+
+        if let Some(ret) = &idl_instr.returns {
+            let mut offset: usize = 0;
```
seanyoung, line 1707: surely rustc can infer `: usize`

tests/solana.rs
LucasSte: You're right. That should not be there.

tests/solana.rs
LucasSte: My inspiration was `program_id`, but I later noticed the code does not look so cool with `id` either. Do you have any other suggestion for naming this field?

tests/solana.rs
seanyoung: Why is this error prone? Rust allows struct members and functions to have the same name.

tests/solana.rs
seanyoung: No, I don't have any ideas. I just don't like it very much - pointless comment

tests/solana.rs
LucasSte: > Why is this error prone? Rust allows struct members and functions to have the same name.

If you do `vm.function("new").accounts` the code compiles without error, but that is not what you want.

LucasSte: I've removed the underscores.

seanyoung (approved):

PR #1469, opened 2023-07-31
Add optimizations test

smoelius:
This PR adds a test to do the following, with various Solidity programs and associated test vectors:
1. Compile the program with optimizations enabled.
1. Run the resulting binary on the test vectors.
1. Compile the same program with optimizations disabled.
1. Run the resulting binary on the same set of test vectors.
1. Verify that the two binaries’ outputs are equal.

Nits are welcome.


seanyoung (commented):
Looks great, thank you. I don't have any comments other than that we have a merge conflict.

tests/solana_tests/optimizations.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    if let Ok(testname) = var("TESTNAME") {
+        run_tests(std::iter::once(calls.join(testname).with_extension("json")));
+    } else {
+        run_tests(read_dir(calls).unwrap().map(|entry| entry.unwrap().path()));
+    }
+}
+
+fn run_tests(iter: impl Iterator<Item = PathBuf>) {
+    for path in iter {
```
xermicus, line 45: These are all independent test and can be run in parallel right? With rayon this probably is a one line change here

tests/solana_tests/optimizations.rs
```diff
@@ -0,0 +1,138 @@
...
+        let file_stem = path.file_stem().unwrap();
+
+        // Known problematic test.
+        if file_stem == "a467d917a5158b613f94954d4a558a5cb61e6773" {
+            continue;
+        }
+
+        dbg!(file_stem);
+
+        let file = File::open(&path).unwrap();
```
xermicus, line 55: This should be an `expect(...)` explaining why this can never fail, otherwise print a more friendly error message (when this test panics here it'll be a very annoying "Os Error" on the open which is super annoying to debug).

tests/solana_tests/optimizations.rs
```diff
@@ -0,0 +1,138 @@
...
+fn run_tests(iter: impl Iterator<Item = PathBuf>) {
+    for path in iter {
+        let file_stem = path.file_stem().unwrap();
+
+        // Known problematic test.
+        if file_stem == "a467d917a5158b613f94954d4a558a5cb61e6773" {
+            continue;
+        }
+
+        dbg!(file_stem);
```
xermicus, line 53: Why is this a `dbg!` and not `println!`? Also without context, this print is useless (should tell "this file is a known failure and should be fixed: {}")

tests/solana_tests/optimizations.rs
```diff
@@ -0,0 +1,138 @@
...
+            [Options::default(), NO_OPTIMIZATIONS.clone()],
+        );
+    }
+}
+
+fn run_one_test(program: &str, calls: &Calls, opts: impl IntoIterator<Item = Options>) {
+    let mut results_prev: Option<Vec<Result<Option<BorshToken>, u64>>> = None;
+
+    for (i, opts) in opts.into_iter().enumerate() {
+        dbg!(i);
```
xermicus, line 76: Same, should be `println!`

xermicus (changes requested):

xermicus:
@smoelius very nice, we need such tests, thanks!

tests/solana_tests/optimizations.rs
```diff
@@ -0,0 +1,138 @@
...
+            if let Some(idl) = &vm.stack[0].idl {
+                let mut idl = idl.clone();
+
+                idl.instructions.push(IdlInstruction {
+                    name: name.clone(),
+                    docs: None,
+                    accounts: vec![],
+                    args: vec![],
+                    returns: None,
+                });
```
LucasSte, line 101: The IDL is generated when you call `build`. Why did you need to push the instructions again here?

tests/optimization_testcases/programs/006a330e8155de6efb7596b31f48735cc28cbcd0.sol
```diff
@@ -0,0 +1,6 @@
+
+        contract test {
```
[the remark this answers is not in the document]
smoelius: I ran prettier on all of the test programs and recomputed their hashes.

tests/borsh_encoding.rs
```diff
@@ -369,3 +370,58 @@ fn integer_byte_width(ty: &IdlType) -> usize {
...
+    }
+    fn visit_array(&mut self, v: &mut Vec<BorshToken>) {
+        visit_array(self, v)
+    }
+    fn visit_tuple(&mut self, v: &mut Vec<BorshToken>) {
+        visit_tuple(self, v)
+    }
+}
+
+pub fn visit_mut(visitor: &mut (impl VisitorMut + ?Sized), token: &mut BorshToken) {
```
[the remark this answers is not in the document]
smoelius: I changed these to use generic type parameters.

tests/solana_tests/optimizations.rs
smoelius: I changed this to `unwrap_or_else(|error| panic!(...))`.

tests/solana_tests/optimizations.rs
smoelius: I changed this to `println!`.

But to answer your question `dbg!` provides source location information, which I (personally) like for tests.

tests/solana.rs
```diff
@@ -1659,6 +1685,13 @@ impl<'a, 'b> VmFunction<'a, 'b> {
     }
 
     fn call(&mut self) -> Option<BorshToken> {
+        match self.call_with_error_code() {
+            Ok(output) => output,
+            Err(num) => panic!("unexpected return {num:#x}"),
+        }
+    }
+
+    fn call_with_error_code(&mut self) -> Result<Option<BorshToken>, u64> {
```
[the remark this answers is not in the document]
smoelius: It's needed because we don't know ahead of time whether a call will succeed or fail.

The test performs each call using `call_with_error_code` and records the call's result. The test does this with and without optimizations enabled. The test then verifies that the two calls produced equivalent results (e.g., both `Ok(...)` or both `Err(...)`). The test doesn't actually check whether they are `Ok(...)` or `Err(...)`, just that they are the same. (Please let me know if this is not clear.)

tests/solana_tests/optimizations.rs
smoelius: To be honest, I don't really understand what is going on there.

I was imitating this example: https://github.com/hyperledger/solang/blob/373cff71f27515cba98f16c2f7632c2092c95041/tests/solana_tests/balance.rs#L384-L396

If you comment out that `if` block you cited in `optimizations.rs` and run:
```
TESTNAME=c3c446fbffb01f22ace1e7311385624ff08b40ae cargo test --test solana optimizations
```
you get `panicked at 'Function not found'`.

For convenience, here is [c3c446fbffb01f22ace1e7311385624ff08b40ae.sol](https://github.com/hyperledger/solang/blob/b1b2ac7d93a8c7b251e98e8a3ed6324be1d1516b/tests/optimization_testcases/programs/c3c446fbffb01f22ace1e7311385624ff08b40ae.sol).

smoelius: To be sure I understand the changes requested: remove the `if let Some(idl) ...` block and all testcases that produce "Function not found" errors?

smoelius: > all testcases that produce "Function not found" errors

It looks like  `c3c446fbffb01f22ace1e7311385624ff08b40ae` was the only one.

LucasSte: > To be sure I understand the changes requested: remove the `if let Some(idl) ...` block and all testcases that produce "Function not found" errors?

Yes, exactly. `c3c446fbffb01f22ace1e7311385624ff08b40ae` must be the only case in which this error happens.

tests/borsh_encoding.rs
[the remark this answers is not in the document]
xermicus: Nice, thanks

tests/solana_tests/optimizations.rs
xermicus: perfect, thanks

tests/solana_tests/optimizations.rs
xermicus: Thanks! Sure, but source code location is provided anyways when some assertion fails in the test. Knowing where the print was happening is probably not helpful in fixing broken test cases :slightly_smiling_face:

tests/solana_tests/optimizations.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    let path = Path::new("tests/optimization_testcases/programs")
+        .join(file_stem)
+        .with_extension("sol");
+    let program = read_to_string(path).unwrap();
+
+    run_test_with_opts(
+        &program,
+        &calls,
+        [Options::default(), NO_OPTIMIZATIONS.clone()],
```
xermicus, line 71: Would be good to have a config corresponding to what `compile --release` uses; I don't think `Options::default()` is necessarily the same.

xermicus (approved):
Thanks @smoelius

tests/solana_tests/optimizations.rs
smoelius: Perhaps that could be a separate PR?

LucasSte (approved):

seanyoung (approved):

seanyoung:
@smoelius would you mind updating your branch (as in, merge the latest main)? We cannot merge without that. Thank you!

smoelius:
> @smoelius would you mind updating your branch (as in, merge the latest main)? We cannot merge without that. Thank you!

Sure. Would you like me to squash at all? And, if so, could you provide some guidance on what or how much to squash?

seanyoung:
That's fine, thank you very much @smoelius, great stuff.

PR #1470, opened 2023-07-31
Improve function names and other minor changes (nfc)

LucasSte:
This PR has general fixes in Solang, namely:

1. Added an assertion at `fn must_fail` to make sure the called function returns an error.
2. Rename `edges` to `successors`, as the function returns the successors of a block.
3. Rename `clone_for_parent_block` to `deep_clone`.
4. Rename `set_maxi_signed` to `get_max_signed`
5. Rename `set_max_unsigned` to `set_max_unsigned`
6. Fix typos in comments and in function names.

Edit: I slipped in another change.
7. Use `.expect("reason")` instead of `if {} else { unreachable!()}` at `dead_storage.rs`.

### DIFF
--- a/tests/solana.rs
+++ b/tests/solana.rs
@@ -1743,6 +1743,10 @@ impl<'a, 'b> VmFunction<'a, 'b> {
                 .validate_account_data_heap(&self.accounts[idx].pubkey);
         }
 
+        if let ProgramResult::Ok(num) = result {
+            assert_ne!(num, 0);
+        }
+
         result
     }
 }
--- a/tests/solana_tests/primitives.rs
+++ b/tests/solana_tests/primitives.rs
@@ -967,7 +967,7 @@ fn test_overflow_boundaries() {
 
         let lower_second_op = lower_boundary_minus_two.div(2);
 
-        let res = contract
+        contract
             .function("mul")
             .arguments(&[
                 BorshToken::Int {
@@ -982,9 +982,7 @@ fn test_overflow_boundaries() {
             .accounts(vec![("dataAccount", data_account)])
             .must_fail();
 
-        assert_ne!(res.unwrap(), 0);
-
-        let res = contract
+        contract
             .function("mul")
             .arguments(&[
                 BorshToken::Int {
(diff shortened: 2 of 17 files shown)


tests/solana_tests/create_contract.rs
```diff
@@ -367,7 +367,7 @@ fn missing_contract() {
                 is_signer: false,
             },
         ])
-        .must_fail();
+        .may_fail();
```
seanyoung, line 370: Missing check on result, This is exactly why `fn must_fail` should check the result itself

tests/solana.rs
```diff
@@ -1719,7 +1719,7 @@ impl<'a, 'b> VmFunction<'a, 'b> {
         }
     }
 
-    fn must_fail(&mut self) -> ProgramResult {
+    fn may_fail(&mut self) -> ProgramResult {
```
seanyoung, line 1722: This is seems quite error prone. Would it not be better to retain the name `must_fail` and assert that the result is non-zero?

seanyoung (changes requested):
This PR needs a much better title

tests/solana.rs
LucasSte: Fixed

LucasSte:
I slipped another change in `dead_storage.rs`.

tests/solana_tests/primitives.rs
```diff
@@ -1275,7 +1261,7 @@ fn test_overflow_detect_unsigned() {
             // Calculate a number that when multiplied by first_operand_rand, the result will overflow N bits
             let second_operand_rand = rng.gen_biguint_range(&BigUint::from(2usize), &limit);
 
-            let res = contract
+            let _ = contract
```
seanyoung, line 1264: Does this and other lines need `let _ = `?

seanyoung (commented):
How about "Improve function names and other minor changes (nfc)"

tests/solana_tests/primitives.rs
LucasSte: Removed.

seanyoung (approved):
Very nice, thanks

PR #1471, opened 2023-07-31
Automatically apply suggestions of clippy pedantic

LucasSte:
I've run `cargo clippy --fix --bin "solang" -- -W clippy::pedantic`.

### DIFF
--- a/src/bin/languageserver/mod.rs
+++ b/src/bin/languageserver/mod.rs
@@ -14,7 +14,21 @@ use solang::{
 use solang_parser::pt;
 use std::{collections::HashMap, ffi::OsString, path::PathBuf};
 use tokio::sync::Mutex;
-use tower_lsp::{jsonrpc::Result, lsp_types::*, Client, LanguageServer, LspService, Server};
+use tower_lsp::{
+    jsonrpc::Result,
+    lsp_types::{
+        CompletionOptions, CompletionParams, CompletionResponse, Diagnostic,
+        DiagnosticRelatedInformation, DiagnosticSeverity, DidChangeConfigurationParams,
+        DidChangeTextDocumentParams, DidChangeWatchedFilesParams, DidChangeWorkspaceFoldersParams,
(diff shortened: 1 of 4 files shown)


src/bin/languageserver/mod.rs
```diff
@@ -14,7 +14,21 @@ use solang::{
...
+    lsp_types::{
+        CompletionOptions, CompletionParams, CompletionResponse, Diagnostic,
+        DiagnosticRelatedInformation, DiagnosticSeverity, DidChangeConfigurationParams,
+        DidChangeTextDocumentParams, DidChangeWatchedFilesParams, DidChangeWorkspaceFoldersParams,
+        DidCloseTextDocumentParams, DidOpenTextDocumentParams, DidSaveTextDocumentParams,
+        ExecuteCommandOptions, ExecuteCommandParams, Hover, HoverContents, HoverParams,
+        HoverProviderCapability, InitializeParams, InitializeResult, InitializedParams, Location,
+        MarkedString, MessageType, OneOf, Position, Range, ServerCapabilities,
+        SignatureHelpOptions, TextDocumentContentChangeEvent, TextDocumentSyncCapability,
+        TextDocumentSyncKind, Url, WorkspaceFoldersServerCapabilities, WorkspaceServerCapabilities,
```
seanyoung, line 28: This is not better

seanyoung (commented):
Cc: @chioni16

src/bin/languageserver/mod.rs
LucasSte: Clippy does not like importing everything from a module, because it increases the likelihood of name collisions both from structures in our file and from other imported libraries.

I can revert this change, no problem.

What do you think?

seanyoung: I get the reasoning, it's still ugly.

However, it's much easier to just live with clippy opinions than fight against it. Let's just go with it.

seanyoung (approved):

PR #1472, opened 2023-07-31
Minor fixes to the Solana library

LucasSte:
I fixed an incorrect comment and a wrong index when encoding data.

### DIFF
--- a/integration/solana/simple_collectible.spec.ts
+++ b/integration/solana/simple_collectible.spec.ts
@@ -76,7 +76,7 @@ describe('Simple collectible', function () {
             .remainingAccounts([
                 { pubkey: new_owner_token_account.address, isSigner: false, isWritable: true },
                 { pubkey: owner_token_account.address, isSigner: false, isWritable: true },
-                { pubkey: nft_owner.publicKey, isSigner: true, isWritable: true },
+                { pubkey: nft_owner.publicKey, isSigner: true, isWritable: false },
             ])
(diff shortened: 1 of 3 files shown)


seanyoung (changes requested):
What about the other problems in this file? Why only fix two issues?

LucasSte:
> What about the other problems in this file? Why only fix two issues?

The other problems require a separate PR with proper testing. These two were quick to fix.

seanyoung:
Line 279 does not need a test

LucasSte:
> Line 279 does not need a test

Is this the line you are talking about?
https://github.com/hyperledger/solang/blob/9ce88f2bb97aae9339b14715cc4b53b16fe31f90/solana-library/spl_token.sol#L279

seanyoung:
> @seanyoung I've updated the writer flag in the transfer function as well. I am quite sure you suggested me a way to test this, but I can't recall all the details. Can you refresh my memory?

You've already done this by changing the `is_writable` flag in the `token.spec.ts`.

The burn function also wants `is_writable: true` for the `is_signer: true`. Is that correct?

LucasSte:
> The burn function also wants `is_writable: true` for the `is_signer: true`. Is that correct?

Now that you asked, I decided to re-check every instruction in SPL-Token. Both `burn` and `mint_to` had incorrect writable booleans.

seanyoung (approved):

PR #1478, opened 2023-08-07
Fix incorrect deserialization

LucasSte:
The deserialization of `struct TokenAccountData` in the Solana library is incorrect. This PR fixes the issue and tests the deserialization of both `struct TokenAccountData` and `struct MintAccountData`.

### DIFF
--- a/integration/solana/account_data.spec.ts
+++ b/integration/solana/account_data.spec.ts
@@ -7,39 +7,15 @@ import {
     approveChecked,
     AuthorityType,
     createMint,
-    getAccount,
+    getAccount, getMint,
     getOrCreateAssociatedTokenAccount, mintTo,
     setAuthority
 } from "@solana/spl-token";
-import { publicKey, u64, bool } from '@solana/buffer-layout-utils';
-import { u32, u8, struct } from '@solana/buffer-layout';
 import expect from "expect";
 import {Program} from "@coral-xyz/anchor";
-import assert from "assert";
-import exp from "constants";
 
 describe('Deserialize account data', function () {
     this.timeout(500000);
-
-    interface RawMint {
-        mintAuthorityOption: 1 | 0;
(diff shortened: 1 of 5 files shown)


seanyoung (commented):
This really needs a test against a real spl-token to make sure it is correct.

LucasSte:
> This really needs a test against a real spl-token to make sure it is correct.

We already use both `TokenAccountData` and `MintAccountData` in `simple_collectible.sol` during integration tests, although we do not read all of their fields.

What kind of test do you have in your mind? Checking all fields of the structs?

seanyoung:
> What kind of test do you have in your mind? Checking all fields of the structs?

Ideally, yes. We don't know if it works unless we test against the real thing.

xermicus (approved):
LGTM but I agree with Sean in that an integration test or similar would be great.

integration/solana/account_data.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+    const MintLayout = struct<RawMint>([
+        u32('mintAuthorityOption'),
+        publicKey('mintAuthority'),
+        u64('supply'),
+        u8('decimals'),
+        bool('isInitialized'),
+        u32('freezeAuthorityOption'),
+        publicKey('freezeAuthority'),
+    ]);
+
```
seanyoung, line 43: Why are these definitions here? They're from `@solana/spl-token` (where they belong). Can't you use `getMint()` so we don't have to copy and paste this?

LucasSte: I found a guide on the internet about the Solana library that misled me and made have these structures. I should have checked deeper the library before following the guide.

LucasSte: Fixed now.

seanyoung (approved):
Very nice, thanks @LucasSte

PR #1483, opened 2023-08-10
Use Ubuntu 20.04 rather than 22.04

seanyoung:
We've had a few users reporting that solang does not work in their environment; usually this is ubuntu 20.04 WSL.


seanyoung:
Approving this also approves https://github.com/hyperledger/solang-llvm/pull/24

seanyoung:
Approving this also approves https://github.com/hyperledger/solang-llvm/pull/24

LucasSte (approved):

xermicus (approved):

PR #1485, opened 2023-08-14
Avoid overflow in transfer

LucasSte:
`address.transfer` and `address.send` utilize `void sol_transfer` from `solana.c` under the hood, which was saving the overflowed valued to the account balance. This PR fixes such an issue.

### DIFF
--- a/docs/language/managing_values.rst
+++ b/docs/language/managing_values.rst
@@ -95,8 +95,7 @@ Here is an example:
     function to be called, use ``address.call{value: 100}("")`` instead.
 
 .. note::
-    On Solana, these functions manage native token assets from a contract's data account when
-    invoked from a contract variable. Normally, funds are not stored in a contract's account, as
-    one can set a separate payer account to pay for all the transactions. In this case, it is more
-    straightforward to use the :ref:`system instruction library <system_instruction_library>` to transfer
-    native tokens between accounts.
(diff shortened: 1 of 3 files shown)


docs/language/managing_values.rst
```diff
@@ -88,6 +88,10 @@ Here is an example:
         }
     }
 
+On Solana, only the owner can change the balance of its accounts. As a corollary, the functionality of ``send()`` and
+``transfer`` is limited. They only work when the contract's program account owns both the sender and the
+recipient accounts.
```
seanyoung, line 93: I think it is clearer to first state the limitation, and then the cause. Also, there are many superfluous words. The paragraph should be a note, and possibly merged with the last note in this file. How about:

On Solana, ``send()`` and ``transfer()`` can only transfer native tokens between accounts that are both own by
the current contract's program account, since only the owner can modify account balances. Use the 
:ref:`system instruction library <system_instruction_library>`_ `transfer()` function to avoid this problem.

seanyoung (changes requested):

docs/language/managing_values.rst
LucasSte: Fixed

seanyoung (approved):

xermicus (approved):

PR #1486, opened 2023-08-14
Use named constants instead of magic numbers

LucasSte:
Using magic numbers degrades code readability.


LucasSte:
I attempted to use both `solana-sdk` and `solana-program` crates, but they BLAKE3 library it requires is incompatible with LLVM implementation on Linux Arm. Please, check my first commit to know it was wired up.

seanyoung (approved):

PR #1487, opened 2023-08-15
Bring import resolution in line w/ Solc's

BenTheKush:
Solang's import resolution departs from solc's in several key ways. I've spent quite a bit of time trying to understand how solc's import system works (it's very bad), and how Solang differs. This has been both through reading code and through experimenting with both systems.

I've compiled a set of tests [here](https://github.com/BenTheKush/solc_remapping_behavior_test). These tests show differences in behavior between the import resolution (this includes filename resolution specified from CLI, remapping behavior, import path behavior, etc etc).

This PR is attempting to bring solang's implementation in line with my understanding of the solidity documentation. It's currently a draft and is rough around the edges (including debug println statements so I can figure out control flow, etc).

### 1. Remap Targets
Solang and Solc treat the target of a remapping `map=target` very differently.

+ **Solc:** In solc, when an import statement of the form `import "map/path/to/Foo.sol";` is encountered, solc will attempt to remplace `map` with `target` and then resolve this new replaced version of the import path against the base-path and any specified include-paths.

  **Note:** from the current working director, the path specified by `target` DOES NOT NEED TO EXIST!
  
+ **Solang:** Solang [canonicalizes the target](https://github.com/hyperledger/solang/blob/90c687ecb52dd4c75adda3c52443c0b5725ab16c/src/file_resolver.rs#L43); if the target does not exist compilation immediately fails. Worse, this canonicalization may be valid but point to a _different_ target than what was expected by the developers. This could silently lead to incorrect files being imported, leading to potential security risks.

The fix is to remove the call to `.canonicalize()`

Another bug is that when a remapping applies then it _must_ be applied.

Consider the following setup:`Contract.sol` imports `import "lib/A.sol";`, which exists in `resources/node_modules/lib/A.sol`. 

+ **Solc:** Compiling with `solc Contract.sol lib=resources/node_modules/lib --base-path node_modules` fails, _even though the imported path `lib/A.sol` can be appended to the base path `resoucres/node_modules` to get the correct path `resources/node_modules/lib/A.sol`_. This is because the remapping `lib=node_modules/lib` __modifies__ the imported path `lib/A.sol` to `node_modules/lib/A.sol`, and then _this path_, not the original imported path, is used to resolve against the base-path.

+ **Solang:** Compiling with `solang --target solana Contract.sol -m lib=resources/node_modules/lib -Inode_modules` succeeds.

The fix is to apply a remapping when possible and not allow the original imported path to be used for resolution.
**EDIT:** After some experimenting with solc it appears that _the last remapping_ is always the one that is applied!
Note that these are not cumulative: each successive remapping is applied to the original import path, replacing any other remappings that applied:

```
solc a=good C.sol                 # good
solc a=good a=bad C.sol           # bad
solc a=bad a=good C.sol           # good
solc a=good a=bad a=good C.sol    # good
```

### 2. Incorrect Direct Imports

A _direct import_ is any import that doesn't start with `"./"` or `"../"` (these are relative imports). 

+ **Solc:** resolves direct imports against the `--base-path` and any specified `--include-path`s

+ **Solang:** currently attempts to [resolve an import path by canonicalizing it in the host filesystem](https://github.com/hyperledger/solang/blob/90c687ecb52dd4c75adda3c52443c0b5725ab16c/src/file_resolver.rs#L209-L213)
  ```rust
        if let Some(file) = self.try_file(filename, &path, None)? {
            return Ok(file);
        } else if path.is_absolute() {
            return Err(format!("file not found '{}'", filename.to_string_lossy()));
        }
  ```
  
  This is correct behavior only when this is a file specified from CLI, which is true iff parent is `None` (at least, according to the current implementation). The easiest solution is to change the above lines to:
  
  ```rust
    if parent.is_none(){
        if let Some(file) = self.try_file(filename, &path, None)? {
            return Ok(file);
        }
        return Err(format!("file not found '{}'", filename.to_string_lossy()));
    }
  ```
  
  **Note:** when parent is none and try_file failed we fail (as opposed to only when it was an absolute path). Why? Because a file specified from CLI is a path in the _host file system_, not the _virtual file system_. This means that if `canonicalize()` fails, it's a bad filename.

### 3. Ambiguous Imports should Fail

An ambiguous import (i.e., an import path that can be resolved multiple ways given the configuration) should fail. This is specified in solidity docs. It's also undesirable to have import map order not commute (i.e., `-I a -I b` results in a different binary than `-I b -I a`). 

The fix for this is to create a list of resolved files and then, if there is more than one, report an error and list all resolved file paths. Example from solc where I specified the same import root twice, causing multiple resolutions, once for each redundant occurance of the `.` import root:

Screen Shot 2023-08-14 at 9 51 33 PM

### 4. Remapping's maps should support multi-path segments

A remap like `a/b/c=target` is valid according to solidity docs/solc. I've replaced the [previous `iter()`/`first()` based implementation](https://github.com/hyperledger/solang/blob/90c687ecb52dd4c75adda3c52443c0b5725ab16c/src/file_resolver.rs#L216-L218):

```rust
        let mut iter = path.iter();
        if let Some(first_part) = iter.next() {
            let relpath: &PathBuf = &iter.collect();
```

with a `strip_prefix()` based implementation.

### 5. Solc's import path specification is _non-monotonic_

By this I mean that you can add an import path (specifically `--base-path`), and previously valid imports will now become _invalid_. YUCK.

This is because the default `--base-path` is `""`, and this gets _replaced_ when you specify an import path. This is gross and leads to disgusting edge cases. I'll expand more on this later

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,7 +4,3 @@
 [submodule "integration/polkadot/tornado/tornado-cli"]
 	path = integration/polkadot/tornado/tornado-cli
 	url = https://github.com/tornadocash/tornado-cli.git
-[submodule "testdata/solang_import_resolution_tests"]
-	path = testdata/solang_import_resolution_tests
-  url = https://github.com/bkushigian/solang_import_resolution_tests
-  branch = v1.0.0
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "integration/polkadot/tornado/tornado-cli"]
 	path = integration/polkadot/tornado/tornado-cli
 	url = https://github.com/tornadocash/tornado-cli.git
+[submodule "testdata/solang_import_resolution_tests"]
+	path = testdata/solang_import_resolution_tests
+	url = https://github.com/bkushigian/solang_import_resolution_tests
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,4 +6,5 @@
 	url = https://github.com/tornadocash/tornado-cli.git
 [submodule "testdata/solang_import_resolution_tests"]
 	path = testdata/solang_import_resolution_tests
-	url = https://github.com/bkushigian/solang_import_resolution_tests
+  url = https://github.com/bkushigian/solang_import_resolution_tests
+  branch = v1.0.0
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -515,11 +515,11 @@ impl PackageTrait for DocPackage {
 pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     let mut resolver = FileResolver::default();
 
-    for filename in package.get_input() {
-        if let Ok(path) = PathBuf::from(filename).canonicalize() {
-            let _ = resolver.add_import_path(path.parent().unwrap());
-        }
-    }
+    // for filename in package.get_input() {
+    //     if let Ok(path) = PathBuf::from(filename).canonicalize() {
+    //         let _ = resolver.add_import_path(path.parent().unwrap());
+    //     }
+    // }
 
     if let Some(paths) = package.get_import_path() {
         for path in paths {
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -527,18 +527,20 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     }
 
     if let Some(maps) = package.get_import_map() {
-        let mut existing_mapped_values = resolver
-            .get_import_paths()
-            .iter()
-            .filter_map(|x| x.0.clone())
-            .collect::<Vec<OsString>>();
         for (map, path) in maps {
-            let map2 = OsString::from(map);
-            if existing_mapped_values.contains(&map2) {
-                eprintln!("warning: mapping {} is overwritten", map)
+            let os_map = OsString::from(map);
+            if let Some((_, existing_path)) = resolver
+                .get_import_paths()
+                .iter()
+                .find(|(m, _)| *m == Some(os_map.clone()))
+            {
+                eprintln!(
+                    "warning: mapping '{}' to '{}' is overwritten",
+                    map,
+                    existing_path.display()
+                )
             }
-            existing_mapped_values.push(map2.clone());
-            resolver.add_import_map(map2, path.clone());
+            resolver.add_import_map(os_map, path.clone());
         }
     }
 
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -527,7 +527,7 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     }
 
     if let Some(maps) = package.get_import_map() {
-        let existing_mapped_values = resolver
+        let mut existing_mapped_values = resolver
             .get_import_paths()
             .iter()
             .filter_map(|x| x.0.clone())
@@ -535,9 +535,9 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
         for (map, path) in maps {
             let map2 = OsString::from(map);
             if existing_mapped_values.contains(&map2) {
-                // TODO: This doesn't print, how should we handle this?
                 eprintln!("warning: mapping {} is overwritten", map)
             }
+            existing_mapped_values.push(map2.clone());
             resolver.add_import_map(map2, path.clone());
         }
     }
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -515,7 +515,6 @@ impl PackageTrait for DocPackage {
 pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     let mut resolver = FileResolver::default();
 
-    let mut added_path = false;
     if let Some(paths) = package.get_import_path() {
         let dups: Vec<_> = paths.iter().duplicates().collect();
 
@@ -529,11 +528,7 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
 
         for path in paths {
             resolver.add_import_path(path);
-            added_path = true;
         }
-    }
-    if !added_path {
-        resolver.add_import_path(&PathBuf::from(""));
     }
 
     if let Some(maps) = package.get_import_map() {
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -521,9 +521,8 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
             if let Err(e) = resolver.add_import_path(path) {
                 eprintln!("error: import path '{}': {}", path.to_string_lossy(), e);
                 exit(1);
-            } else {
-                added_path = true;
             }
+            added_path = true;
         }
     }
     if !added_path {
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -8,15 +8,15 @@ use clap_complete::Shell;
 #[cfg(feature = "wasm_opt")]
 use contract_build::OptimizationPasses;
 
+use itertools::Itertools;
 use semver::Version;
 use serde::Deserialize;
-use std::{ffi::OsString, path::PathBuf, process::exit};
-
 use solang::{
     codegen::{OptimizationLevel, Options},
     file_resolver::FileResolver,
     Target,
 };
+use std::{ffi::OsString, path::PathBuf, process::exit};
 
 mod test;
 #[derive(Parser)]
@@ -517,6 +517,16 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
 
     let mut added_path = false;
     if let Some(paths) = package.get_import_path() {
+        let dups: Vec<_> = paths.iter().duplicates().collect();
+
+        if !dups.is_empty() {
+            eprintln!(
+                "error: import paths {} specifed more than once",
+                dups.iter().map(|p| format!("'{}'", p.display())).join(", ")
+            );
+            exit(1);
+        }
+
         for path in paths {
             resolver.add_import_path(path);
             added_path = true;
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -515,18 +515,21 @@ impl PackageTrait for DocPackage {
 pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     let mut resolver = FileResolver::default();
 
-    // for filename in package.get_input() {
-    //     if let Ok(path) = PathBuf::from(filename).canonicalize() {
-    //         let _ = resolver.add_import_path(path.parent().unwrap());
-    //     }
-    // }
-
+    let mut added_path = false;
     if let Some(paths) = package.get_import_path() {
         for path in paths {
             if let Err(e) = resolver.add_import_path(path) {
                 eprintln!("error: import path '{}': {}", path.to_string_lossy(), e);
                 exit(1);
+            } else {
+                added_path = true;
             }
+        }
+    }
+    if !added_path {
+        if let Err(e) = resolver.add_import_path(&PathBuf::from("")) {
+            eprintln!("error: could not add default import path '': {}", e);
+            exit(1);
         }
     }
 
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -527,8 +527,18 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     }
 
     if let Some(maps) = package.get_import_map() {
+        let existing_mapped_values = resolver
+            .get_import_paths()
+            .iter()
+            .filter_map(|x| x.0.clone())
+            .collect::<Vec<OsString>>();
         for (map, path) in maps {
-            resolver.add_import_map(OsString::from(map), path.clone());
+            let map2 = OsString::from(map);
+            if existing_mapped_values.contains(&map2) {
+                // TODO: This doesn't print, how should we handle this?
+                eprintln!("warning: mapping {} is overwritten", map)
+            }
+            resolver.add_import_map(map2, path.clone());
         }
     }
 
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -518,26 +518,17 @@ pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     let mut added_path = false;
     if let Some(paths) = package.get_import_path() {
         for path in paths {
-            if let Err(e) = resolver.add_import_path(path) {
-                eprintln!("error: import path '{}': {}", path.to_string_lossy(), e);
-                exit(1);
-            }
+            resolver.add_import_path(path);
             added_path = true;
         }
     }
     if !added_path {
-        if let Err(e) = resolver.add_import_path(&PathBuf::from("")) {
-            eprintln!("error: could not add default import path '': {}", e);
-            exit(1);
-        }
+        resolver.add_import_path(&PathBuf::from(""));
     }
 
     if let Some(maps) = package.get_import_map() {
         for (map, path) in maps {
-            if let Err(e) = resolver.add_import_map(OsString::from(map), path.clone()) {
-                eprintln!("error: import path '{}': {}", path.display(), e);
-                exit(1);
-            }
+            resolver.add_import_map(OsString::from(map), path.clone());
         }
     }
 
--- a/src/bin/solang.rs
+++ b/src/bin/solang.rs
@@ -186,7 +186,7 @@ fn compile(compile_args: &Compile) {
     for filename in compile_args.package.get_input() {
         // TODO: this could be parallelized using e.g. rayon
         let ns = process_file(
-            &filename.canonicalize().unwrap(),
+            &filename,
             &mut resolver,
             target,
             &compile_args.compiler_output,
--- a/src/bin/solang.rs
+++ b/src/bin/solang.rs
@@ -186,7 +186,7 @@ fn compile(compile_args: &Compile) {
     for filename in compile_args.package.get_input() {
         // TODO: this could be parallelized using e.g. rayon
         let ns = process_file(
-            filename,
+            &filename.canonicalize().unwrap(),
             &mut resolver,
             target,
             &compile_args.compiler_output,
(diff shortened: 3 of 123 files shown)


bkushigian:
Sorry I thought I set this up as a draft, looks like it's a regular PR

LucasSte:
> Sorry I thought I set this up as a draft, looks like it's a regular PR

If it is still a draft, you can click `convert to draft` right below the reviewers list.

BenTheKush:
Okay, the current import resolution matches solc's import semantics. I wrote a number of tests of both correct and incorrect invocations of solc and mapped these invocations to solang. I then ran both compilers and ensured that the exit codes were the same. Her are the results of solang 0.3.0:

Screen Shot 2023-08-15 at 8 01 45 PM


And here are the results after incorporating this PR:

Screen Shot 2023-08-15 at 8 01 32 PM

I think it is okay to depart from solc's semantics, but this should be done deliberately.

Also, code quality isn't up to snuff, but I want to get feedback before I start polishing too much as to features that we do/do not want.

src/bin/cli/mod.rs
```diff
@@ -515,20 +515,23 @@ impl PackageTrait for DocPackage {
 pub fn imports_arg<T: PackageTrait>(package: &T) -> FileResolver {
     let mut resolver = FileResolver::default();
 
-    for filename in package.get_input() {
-        if let Ok(path) = PathBuf::from(filename).canonicalize() {
-            let _ = resolver.add_import_path(path.parent().unwrap());
-        }
-    }
-
```
BenTheKush, line 523: Adding the parent of files as a basepath is incorrect IMO, and is not what solc does. I think we should remove this

src/bin/cli/mod.rs
```diff
@@ -515,20 +515,23 @@ impl PackageTrait for DocPackage {
...
+    let mut added_path = false;
```
BenTheKush, line 518: This is kludgy, and we can improve the implementation. The problem is, there isn't an easy way to tell if import paths are empty (querying `file_resolver.get_import_path(0)` might work, but it's not very clear, I prefer this stylistically).

Happy to fix this if there is an API change

src/bin/solang.rs
```diff
@@ -186,7 +186,7 @@ fn compile(compile_args: &Compile) {
     for filename in compile_args.package.get_input() {
         // TODO: this could be parallelized using e.g. rayon
         let ns = process_file(
-            filename,
+            &filename.canonicalize().unwrap(),
```
BenTheKush, line 189: Rather than canonicalizing the import paths and adjoining the filename to the import paths, it's better to canonicalize the filenames we _know_ exist in our host filesystem, and those are precisely the file names passed in by command line

src/file_resolver.rs
```diff
@@ -40,25 +40,14 @@ pub struct ResolvedFile {
 impl FileResolver {
     /// Add import path
     pub fn add_import_path(&mut self, path: &Path) -> io::Result<()> {
-        self.import_paths.push((None, path.canonicalize()?));
+        self.import_paths.push((None, path.to_path_buf()));
```
BenTheKush, line 43: To match solc's semantics we don't want to fail if an import path doesn't exist.

I'm happy to depart from this, and I'm also happy to try to canonicalize and just not push a new path when we fail to canonicalize. No strong feelings.

src/file_resolver.rs
```diff
@@ -40,25 +40,14 @@ pub struct ResolvedFile {
...
-        {
-            Err(Error::new(
-                ErrorKind::Other,
-                format!("duplicate mapping for '{}'", map.to_string_lossy()),
-            ))
-        } else {
-            self.import_paths.push((Some(map), path.canonicalize()?));
-            Ok(())
-        }
+        self.import_paths.push((Some(map), path));
```
BenTheKush, line 49: solc allows duplicate mappings and just uses the last one. I think this is desirable and correct (e.g., this is easier for debugging large generated compilation configs)

src/file_resolver.rs
```diff
@@ -184,6 +173,8 @@ impl FileResolver {
     ) -> Result<ResolvedFile, String> {
         let path = PathBuf::from(filename);
 
+        let mut result: Vec<Result<ResolvedFile, String>> = vec![];
+
```
BenTheKush, line 177: Solc fails when an ambiguous import exists. Rather than returning the first resolved import path, we track _all_ possible import paths and fail with an error if there isn't a canonical choice (e.g., not exactly 1).

src/file_resolver.rs
```diff
@@ -199,49 +190,69 @@ impl FileResolver {
...
-        if let Some(file) = self.try_file(filename, &path, None)? {
-            return Ok(file);
-        } else if path.is_absolute() {
-            return Err(format!("file not found '{}'", filename.to_string_lossy()));
+        if parent.is_none() {
+            if let Some(file) = self.try_file(filename, &path, None)? {
+                return Ok(file);
+            } else if path.is_absolute() {
+                return Err(format!("file not found '{}'", filename.to_string_lossy()));
+            }
```
BenTheKush, line 206: We only want to use the actual filename (i.e., not try it against import maps) when `parent` is `None`. This is because `parent.is_none()` if and only if the file was passed in from command line (i.e., is not an import).

src/file_resolver.rs
```diff
@@ -199,49 +190,69 @@ impl FileResolver {
...
-                if let (Some(mapping), import_path) = &self.import_paths[import_no] {
-                    if first_part == mapping {
-                        let path = import_path.join(relpath);
-
-                        if let Some(file) = self.try_file(filename, &path, Some(import_no))? {
-                            return Ok(file);
-                        }
-                    }
+
+        let mut remapped = path.clone();
```
BenTheKush, line 211: We could just make `path` mutable, can clean this up if it's desirable.

Basically, what we are doing here is iterating over remappings and applying them if possible. This leaves us with the final applicable remapping being applied, which matches Solc's semantics

src/file_resolver.rs
seanyoung: I don't follow why allowing duplicate mappings is desirable. The entries before the last one get silently dropped.

How about warning that a mapping is being overwritten by another one by the same name?

src/file_resolver.rs
seanyoung: The idea here is that if an import path does not exist, then it unusable. So, it's like an early warning system: your import path is no good, which might help the user if they're getting "file not found" errors. At least, that's the idea.

src/bin/solang.rs
seanyoung: We cannot call `unwrap()` or `canonicalize()` because the file might not exist, it would at least need some better error handling.

src/bin/cli/mod.rs
seanyoung: Yes, I agree. This should be removed. However, for some reason this is causing test failures in the `ethereum_solidity_tests` which exist exactly to compare behavior with solc. That needs debugging.

src/file_resolver.rs
seanyoung: Very nice :+1:

seanyoung:
It would be fantastic to have your tests merged into solang if you agree with that

src/bin/solang.rs
bkushigian: I agree on the unwrap, sorry sloppy coding on my end. We were calling canonicalize before I think, just somewhere else? My idea was to bubble up any canonicalization to locations where paths represent host file system paths rather than virtual file system paths.

src/file_resolver.rs
bkushigian: The more I think about it the more I like being permissive (but I think we should add a warning).

If humans were the only ones to use compilers I think that a buggy configuration should be a hard fail. But many (most?) compilation invocations originate from a build system, and if there is a bug in the build system and we hard fail when an extra path is passed in, then the user is just stuck w/ no recourse. I think that maybe a good philosophy is be strict with human-facing interfaces and more permissive with black box-facing interfaces?

I'm in a PL lab and I just sent out the bat signal to all the compiler peeps to get their hot takes on best practices for compiler interface behavior

src/file_resolver.rs
bkushigian: Yeah, I'm happy to add a warning. I definitely think it's better to succeed than to fail (I'll again appeal to the build system argument from above).

bkushigian:
> @BenTheKush this is some great analysis, and it works look good. It could really do with some further tests and all the tests should pass before it can be merged

Awesome! I'll look into why tests are failing.

Suggestions on how to simulate a CLI invocation? I would really like to port the tests I wrote up in the other repo I linked, which test behavior on a bunch of edge cases, into solang.

BenTheKush:
@seanyoung I'm not sure what's causing the tests to fail. From what I can tell only `parse_and_resolve` is getting called, and every single file is failing. Many of the errors don't seem to have anything to do with imports?

```
file: /Users/benku/Solang/fork/solang/testdata/solidity/test/libsolidity/semanticTests/smoke/fallback.sol name:/Users/benku/Solang/fork/solang/testdata/solidity/test/libsolidity/semanticTests/smoke/fallback.sol
error: fallback function must not be declare payable, use 'receive() external payable' instead
  ┌─ /Users/benku/Solang/fork/solang/testdata/solidity/test/libsolidity/semanticTests/smoke/fallback.sol:5:5
  │
5 │     fallback() external payable {
  │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

I've poked around for a bit but don't have any obvious next steps. Thoughts?

BenTheKush:
This is also the failure message that comes up when I run brew-installed solang on the same file:

Screen Shot 2023-08-18 at 1 39 01 PM

src/bin/solang.rs
BenTheKush: Wasn't sure what the best way to handle this was (since we don't return a Result). Just removed canonicalization since `process_file` tries to canonicalize anyway.

src/file_resolver.rs
BenTheKush: General consensus is "do what solc does".

What is best practice to give a warning? Just `eprintln!()`?

BenTheKush:
> It would be fantastic to have your tests merged into solang if you agree with that

I could use some direction on how to test the CLI. We probably want to build up a `Compile` struct. Is there a simple way to do that, e.g. something like `let compile = parse::<Compile>(vec!["--target", "evm", "-I", path.parent().to_str(), path.to_str(), ...])`

BenTheKush:
Also @seanyoung I'm modifying `src/tests/imports.rs` which is failuring (expected). However, we are comparing against the stdout, and even after modifying the expected stdout I'm getting test failures triggered by formatting:

```
stderr: error: file not found 'foo.sol'


thread 'import_map_dup' panicked at 'assertion failed: `(left == right)`
  left: `"\u{1b}[0m\u{1b}[1m\u{1b}[38;5;9merror\u{1b}[0m\u{1b}[1m: file not found 'foo.sol'\u{1b}[0m\n\n"`,
 right: `"error: file not found 'foo.sol'\n"`', tests/imports.rs:27:5
test import_map_dup ... FAILED
```

I'm not sure how this ever passed, since the previous error was also formatted.

I could use a hand figuring out the testing infrastructure (looks like you were the author).

src/bin/cli/mod.rs
seanyoung: I don't understand what an empty `""` path means. What does an import path of nothing do? I suspect it is the same as `"."`. We should be permitting an empty path, let alone adding one.

solc does not allow it either.
```bash
$ solc --include-path "" 
--include-path option requires a non-empty base path.
```

src/bin/cli/mod.rs
```diff
@@ -515,27 +515,20 @@ impl PackageTrait for DocPackage {
...
-                eprintln!("error: import path '{}': {}", path.to_string_lossy(), e);
-                exit(1);
-            }
+            resolver.add_import_path(path);
+            added_path = true;
         }
     }
+    if !added_path {
+        resolver.add_import_path(&PathBuf::from(""));
+    }
```
seanyoung, line 527: This does not make much sense.

src/bin/cli/mod.rs
```diff
@@ -515,27 +515,20 @@ impl PackageTrait for DocPackage {
...
 
     if let Some(maps) = package.get_import_map() {
         for (map, path) in maps {
-            if let Err(e) = resolver.add_import_map(OsString::from(map), path.clone()) {
-                eprintln!("error: import path '{}': {}", path.display(), e);
-                exit(1);
-            }
+            resolver.add_import_map(OsString::from(map), path.clone());
```
seanyoung, line 531: We should warn if we're overwriting a previous set mapping.

src/file_resolver.rs
seanyoung: You can't do a `eprintln!()` in the solang crate, you have to do this in `src/bin/solang.rs`. 

Having said that, `clang` does not complain if a path does not exist either.

src/file_resolver.rs
```diff
@@ -184,6 +170,22 @@ impl FileResolver {
...
+        // with caching.
+        //
+        // If this file has already been cashed in the VFS, return it.
+        if let Some(import_no) = self.cached_paths.get(&path) {
+            if let Some(res) = self.files.get(*import_no) {
+                return Ok((*res).clone());
+            } else {
+                return Err("".to_string());
+            }
+        }
```
seanyoung, line 185: We need a better solution to this, I'm afraid

.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "integration/polkadot/tornado/tornado-cli"]
 	path = integration/polkadot/tornado/tornado-cli
 	url = https://github.com/tornadocash/tornado-cli.git
+[submodule "testdata/solang_import_resolution_tests"]
+	path = testdata/solang_import_resolution_tests
+  url = https://github.com/bkushigian/solang_import_resolution_tests
+  branch = v1.0.0
```
seanyoung, line 10: Could we import these tests rather using submodules please?

seanyoung (changes requested):
Looking great, just some minor issues

src/bin/cli/mod.rs
BenTheKush: We had discussed this before. an empty base-path (which is different than an empty include path, which can't exist), is the same as `"."` except for when you import an absolute path `import "/usr/node_modules/Contract.sol";`. If base-path is `"."` then this import will be expanded to `"./usr/node_modules/Contract.sol"`. It is horrible and cursed, but it's the solidity semantics.

As for solc, your example has two errors:
1. you can never specify an include path when you have an empty base path (which you have by default).
2. you can never specify an empty include path:

```
$ solc --base-path ""
No input files given. If you wish to use the standard input please specify "-" explicitly.
$ solc --base-path . --include-path ""
Empty values are not allowed in --include-path.
```

[Solidity docs](https://docs.soliditylang.org/en/v0.8.17/path-resolution.html#base-path-and-include-paths) are explicit:
> _"By default the base path is empty, which leaves the source unit name unchanged. When the source unit name is a relative path, this results in the file being looked up in the directory the compiler has been invoked from. It is also the only value that results in absolute paths in source unit names being actually interpreted as absolute paths on disk. If the base path itself is relative, it is interpreted as relative to the current working directory of the compiler."_

BenTheKush: See also item 5 in my writeup at the start of this PR

src/bin/cli/mod.rs
BenTheKush: See above comment. This is from [the docs](https://docs.soliditylang.org/en/v0.8.17/path-resolution.html#base-path-and-include-paths)

BenTheKush: And I agree, it sucks but it's also what solc does, so 🤷

src/bin/cli/mod.rs
BenTheKush: As you [mentioned before](https://github.com/hyperledger/solang/pull/1487#discussion_r1304103625), we will need to warn from solang.rs, not from here. Not sure how to bubble this data up to solang.rs, maybe do a separate check there?

FWIW Solc doesn't issue a warning, so if it's a headache to implement maybe better to not?

BenTheKush: Okay, I tried to add an `eprintln` and it's not showing up on stderr. Your call on how you want to implement this.

src/bin/cli/mod.rs
```diff
@@ -515,27 +515,32 @@ impl PackageTrait for DocPackage {
...
         for path in paths {
-            if let Err(e) = resolver.add_import_path(path) {
-                eprintln!("error: import path '{}': {}", path.to_string_lossy(), e);
-                exit(1);
-            }
+            resolver.add_import_path(path);
+            added_path = true;
         }
     }
+    if !added_path {
```
xermicus, line 525: This deserves a comment. Even better if the function had a Doc comment explaining it all and linking to https://docs.soliditylang.org/en/v0.8.17/path-resolution.html#base-path-and-include-paths

src/file_resolver.rs
```diff
@@ -39,25 +39,18 @@ pub struct ResolvedFile {
 
 impl FileResolver {
     /// Add import path
-    pub fn add_import_path(&mut self, path: &Path) -> io::Result<()> {
-        self.import_paths.push((None, path.canonicalize()?));
-        Ok(())
+    pub fn add_import_path(&mut self, path: &Path) {
+        self.import_paths.push((None, path.to_path_buf()));
```
xermicus, line 43: The struct fields doc comment says that `import_paths` is a set of path. Instead, it's just a list; what happens if duplicate paths are added? Does it matter? If yes, is it handled or prevented at an earlier stage?

seanyoung: Good point. If you add the same path twice, import path resolution will detect that it can find the same file twice - via exactly the same path. That will be confusing, so we should detect this problem earlier and error about it.

seanyoung: I've added code for this

xermicus: Nice 👍 Total nitpick: `import_paths` could be a HashSet then right?

LucasSte (approved):

testdata/solang_import_resolution_tests/.gitignore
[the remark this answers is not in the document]
seanyoung: Do you mean that the top-level directory `testdata` should be renamed to `test_data` for clarity?

Maybe it is a little clearer, but only by a tiny amount. I am not sure it is worth it, also considering that this would require everyone to update their git submodules.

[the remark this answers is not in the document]
LucasSte: > Do you mean that the top-level directory `testdata` should be renamed to `test_data` for clarity?

Exactly.

src/file_resolver.rs
BenTheKush: Also, fwiw, the simple fact that import path order doesn't matter is one of the reasons I like solang's interface so much more than solc's. Solc has --base-path and --include-path, and weird rules about when you can use each of them. But if you just use Solang's --import_path instead you get the exact same behavior.

xermicus (approved):

Issue #1484, opened 2023-08-13
No exception is thrown on overflow during testing via Anchor Framework


YanhuiJessica:
**Describe the bug**
I started a project following [Getting started with Solang](https://solana.com/developers/guides/solang/getting-started). I have tried some math calculations that would overflow, but found that all tests passed.

**To Reproduce**
1. `anchor init example --solidity`
2. Overwrite the `flip()` function with the following lines (`example/solidity/example.sol`)

```js
function flip() public {
   uint32 x = 2147483648;
   uint32 y = 2147483648;
   uint32 z = x + y;
   print("z: {}".format(z));
   value = !value;
}
```

3. `anchor test` All tests passed
4. `example/.anchor/program-logs/F1ipperKF9EfD821ZbbYjS319LXYiBmjhzkkf5a26rC.example.log`

```
Streaming transaction logs mentioning F1ipperKF9EfD821ZbbYjS319LXYiBmjhzkkf5a26rC. Confirmed commitment
Transaction executed in slot 2:
  Signature: N3wupwfD5JcQLAVznxcqLXYB5wqqzoGGzmNyqMRwSqMjVsXJt8FZqu6ZjUg51pqdcMRMgBuut2kUvHwGGt5g97K
  Status: Ok
  Log Messages:
    ...
    Program log: Hello, World!
    ...
Transaction executed in slot 3:
  Signature: 4pL9WzzwpVp8CVbzFarEgcyXhCAUW7b661VGqWeUGHbwD8M4XT6ccprzu6nKXcCv7vCrS883jaBaptJ5Yg56orJJ
  Status: Ok
  Log Messages:
    Program F1ipperKF9EfD821ZbbYjS319LXYiBmjhzkkf5a26rC invoke [1]
    Program log: z: 0
    ...
```

**Expected behavior**
Throws an exception on overflow.

**Hyperledger Solang version**
v0.3.1

seanyoung:
The const folding pass is at fault here. The addition is done at compile time, but the overflow is not detected.

PR #1488, opened 2023-08-15
Produce error diagnostics if overflow occurs in constant folding

seanyoung:
The negate operator does not check for overflow, add this too.

Fixes https://github.com/hyperledger/solang/issues/1484

### DIFF
--- a/src/codegen/constant_folding.rs
+++ b/src/codegen/constant_folding.rs
@@ -14,8 +14,10 @@ use std::ops::{Add, BitAnd, BitOr, BitXor, Div, Mul, Rem, Shl, Shr, Sub};
 use tiny_keccak::{Hasher, Keccak};
 
 /// Constant folding pass on the given cfg. During constant folding, we may find issues
-/// like divide by zero, so this function may add diagnostics to the namespace.
-pub fn constant_folding(cfg: &mut ControlFlowGraph, ns: &mut Namespace) {
+/// like divide by zero, so this function may add diagnostics to the namespace. If dry_run
+/// is true, then diagnostics are generated but the CFG is not modified. This
+///
+pub fn constant_folding(cfg: &mut ControlFlowGraph, dry_run: bool, ns: &mut Namespace) {
     // for each block, instruction
     for block_no in 0..cfg.blocks.len() {
         let mut vars = cfg.blocks[block_no].defs.clone();
@@ -29,11 +31,13 @@ pub fn constant_folding(cfg: &mut ControlFlowGraph, ns: &mut Namespace) {
                         ns.var_constants.insert(*loc, expr.clone());
                     }
 
-                    cfg.blocks[block_no].instr[instr_no] = Instr::Set {
-                        loc: *loc,
-                        res: *res,
-                        expr,
-                    };
+                    if !dry_run {
+                        cfg.blocks[block_no].instr[instr_no] = Instr::Set {
+                            loc: *loc,
+                            res: *res,
+                            expr,
+                        };
+                    }
                 }
                 Instr::Call {
                     res,
@@ -46,12 +50,14 @@ pub fn constant_folding(cfg: &mut ControlFlowGraph, ns: &mut Namespace) {
                         .map(|e| expression(e, Some(&vars), cfg, ns).0)
                         .collect();
 
-                    cfg.blocks[block_no].instr[instr_no] = Instr::Call {
-                        res: res.clone(),
-                        call: call.clone(),
-                        args,
-                        return_tys: return_tys.clone(),
-                    };
+                    if !dry_run {
+                        cfg.blocks[block_no].instr[instr_no] = Instr::Call {
+                            res: res.clone(),
+                            call: call.clone(),
+                            args,
+                            return_tys: return_tys.clone(),
+                        };
+                    }
                 }
                 Instr::Return { value } => {
                     let value = value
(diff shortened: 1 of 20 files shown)


src/codegen/expression.rs
```diff
@@ -43,7 +43,18 @@ pub fn expression(
...
+    let constant_expr;
+
+    let expr =
+        if let (Some(expr), _) = eval_constants_in_expression(expr, &mut Diagnostics::default()) {
+            constant_expr = expr;
+
+            &constant_expr
+        } else {
+            expr
+        };
```
LucasSte, line 55:
```suggestion
    let evaluated = eval_constants_in_expression(expr, &mut Diagnostics::default());
    let expr = evaluated.0.as_ref().unwrap_or(expr);
```

tests/contract_testcases/solana/expressions/overflow_constant_propagation.sol
```diff
@@ -0,0 +1,62 @@
...
+            print("c3: {}".format(c3));
+
+            int16 a4 = -0x8000;
+            int16 c4 = -a4;
+            print("c4: {}".format(c4));
+        }
+    }
+}
+// ---- Expect: diagnostics ----
+// error: 6:20-25: arithmetic overflow, 4294967296 does not fit into uint32
```
LucasSte, line 58:
```suggestion
// error: 6:20-25: arithmetic overflow: 4294967296 does not fit into uint32
```

src/codegen/expression.rs
seanyoung: That's much better, thanks!

tests/contract_testcases/solana/expressions/overflow_constant_propagation.sol
seanyoung: Fixed

xermicus (approved):

src/codegen/constant_folding.rs
```diff
@@ -604,37 +658,82 @@ fn expression(
...
+                            "arithmetic overflow: {value} does not fit into {}",
+                            ty.to_string(ns)
+                        ),
+                    ));
+                }
+                let (_, mut bs) = value.to_bytes_le();
+                bs.truncate(*bits as usize / 8);
+
+                BigInt::from_bytes_le(Sign::Plus, &bs)
+            } else if value.bits() > *bits as u64 {
```
LucasSte, line 698: This branch is never reached, because the condition is the same as in the previous.

src/codegen/constant_folding.rs
```diff
@@ -604,37 +658,82 @@ fn expression(
...
+                            "arithmetic overflow: {value} does not fit into {}",
+                            ty.to_string(ns)
+                        ),
+                    ));
+                }
+                let mut bs = value.to_signed_bytes_le();
+                bs.resize(*bits as usize / 8, 0xff);
+
+                BigInt::from_bytes_le(Sign::Plus, &bs)
+            } else if value.bits() > *bits as u64 || value.sign() == Sign::Minus {
```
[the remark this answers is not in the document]
seanyoung: Good point, fixed

src/codegen/constant_folding.rs
seanyoung: Good point, removed

tests/contract_testcases/solana/expressions/overflow_constant_propagation.sol
```diff
@@ -0,0 +1,62 @@
...
+            print("c3: {}".format(c3));
+
+            int16 a4 = -0x8000;
+            int16 c4 = -a4;
+            print("c4: {}".format(c4));
+        }
+    }
+}
+// ---- Expect: diagnostics ----
+// error: 6:20-25: arithmetic overflow: 4294967296 does not fit into uint32
```
LucasSte, line 58: Are you sure that printing the number is a good idea?
```solidity
contract Test {
        function foo() public pure returns (uint256) {
                uint256 a = 1 << 255;
                uint64 b = 1 << 3;
                return a**b;
        }
}
```

```
error: arithmetic overflow: 126238304966058622268417487065116999845484776053576109500509161826268184136202698801551568013761380717534054534851164138648904527931605160527688095259563605939964364716019515983399209962459578542172100149937763938581219604072733422507180056009672540900709554109516816573779593326332288314873251559077853068444977864803391962580800682760017849589281937637993445539366428356761821065267423102149447628375691862210717202025241630303118559188678304314076943801692528246980959705901641444238894928620825482303431806955690226308773426829503900930529395181208739591967195841536053143145775307050594328881077553168201547776 does not fit into uint256
```

LucasSte: Also, trying to calculate can be very slow:
 ```solidity
contract Test {
        function foo() public pure returns (uint256) {
                uint256 a = 1 << 255;
                uint64 b = 1 << 31;
                return a**b;
        }
}
```

LucasSte: > Also, trying to calculate can be very slow:
> 
> ```solidity
> contract Test {
>        function foo() public pure returns (uint256) {
>                uint256 a = 1 << 255;
>                uint64 b = 1 << 31;
>                return a**b;
>        }
> }
> ```

This calculation will require too much memory as well.

seanyoung: Nice catch. I've re-written this and added these examples to the test

LucasSte: I couldn't find the examples

LucasSte (approved):
Looks good. I just couldn't find the examples I pointed out.

PR #1490, opened 2023-08-16
Implement goto-references, goto-implementations, goto-type-definitions

chioni16:


src/bin/languageserver/mod.rs
```diff
@@ -252,10 +274,42 @@ impl SolangServer {
...
             self.definitions.lock().await.extend(definitions);
+            self.types.lock().await.extend(types);
 
             res.await;
         }
     }
+
+    // Used for goto-{definitions, implementations, declarations, type_definitions}
+
+    async fn get_reference_from_params(
```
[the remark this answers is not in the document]
chioni16: Is it possible to add a non-trait method within a block for trait implementation? 
Yeah, I will change the comment.

src/bin/languageserver/mod.rs
```diff
@@ -1525,7 +1592,7 @@ impl<'a> Builder<'a> {
             );
         }
 
-        for lookup in &mut builder.hovers {
+        for lookup in builder.hovers.iter_mut() {
```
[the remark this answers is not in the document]
chioni16: ah, there is no reason. I am just used to using `iter_mut` in these situations and I must have used it when refactoring things.

src/bin/languageserver/mod.rs
```diff
@@ -1806,45 +1902,154 @@ impl LanguageServer for SolangServer {
...
+            | DefinitionType::Field(_, _) => {
+                if let Some(def) = types.get(&reference) {
+                    def
+                } else {
+                    return Ok(None);
+                }
+            }
+            // DefinitionIndex::Variant(enum_id, _) => {
+            //     &DefinitionIndex::Enum(*enum_id)
+            // }
```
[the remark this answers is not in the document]
chioni16: I wanted to know what would be the right behaviour when `goto-type-def` is used on an enum variant. All enum variants are internally represented as integers. So, should I just return the enum type when this happens or should I return the integer type (in this case, as the integer type is not defined by the user, `None` is returned).

src/bin/languageserver/mod.rs
```diff
@@ -114,6 +134,7 @@ pub struct SolangServer {
     importmaps: Vec<(String, PathBuf)>,
     files: Mutex<Files>,
     definitions: Mutex<Definitions>,
+    types: Mutex<Types>,
```
seanyoung, line 137: These two can be protected by the same mutex, right?

src/bin/languageserver/mod.rs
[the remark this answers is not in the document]
chioni16: The comment has been changed. 
Unfortunately non-trait methods can't be added to the `impl` block for a trait. 
So the function `get_reference_from_params` will have to be present in a different`impl` block

src/bin/languageserver/mod.rs
chioni16: yes, I have made changes so that both `definitions` and `types` use a single `Mutex`. 
I have named the field `global_cache` as it caches things required by the language server "globally" (as opposed to the per-file caches represented by the `files` field). 
Please let me know if you have another name in mind.

seanyoung (approved):
@chioni16 please remove the empty `refs.sol` file

src/bin/languageserver/mod.rs
```diff
@@ -1217,25 +1276,29 @@ impl<'a> Builder<'a> {
...
+        };
+        self.definitions
+            .insert(di.clone(), loc_to_range(&field.loc, file));
+        if let Some(dt) = get_type_definition(&field.ty) {
+            self.types.push((di, dt.into()));
+        }
     }
 
     // Traverses namespace to extract information used later by the language server
     // This includes hover messages, locations where code objects are declared and used
```
xermicus, line 1294: Why is this not a doc comment?

src/bin/languageserver/mod.rs
```diff
@@ -1793,45 +1899,158 @@ impl LanguageServer for SolangServer {
...
-            {
-                let di = &reference.val;
-                if let Some(range) = self.definitions.lock().await.get(di) {
-                    let uri = Url::from_file_path(&di.def_path).unwrap();
-                    let ret = Ok(Some(GotoDefinitionResponse::Scalar(Location {
-                        uri,
-                        range: *range,
-                    })));
-                    return ret;
+    async fn goto_type_definition(
```
xermicus, line 1916: Needs a doc comment explaining the return type and what the function does on a high level

src/bin/languageserver/mod.rs
```diff
@@ -1793,45 +1899,158 @@ impl LanguageServer for SolangServer {
...
+            .map(|range| {
+                let uri = Url::from_file_path(&di.def_path).unwrap();
+                Location { uri, range: *range }
+            })
+            .map(GotoTypeDefinitionResponse::Scalar);
+
+        Ok(location)
+    }
+
+    async fn goto_implementation(
```
xermicus, line 1956: Needs a doc comment explaining the return type and what the function does on a high level

src/bin/languageserver/mod.rs
```diff
@@ -1793,45 +1899,158 @@ impl LanguageServer for SolangServer {
...
+                        })
+                    })
+                    .collect()
+            })
+            .map(GotoImplementationResponse::Array);
+
+        Ok(impls)
+    }
+
+    async fn references(&self, params: ReferenceParams) -> Result<Option<Vec<Location>>> {
```
xermicus, line 2001: Needs a doc comment explaining the return type and what the function does on a high level

xermicus (commented):
LGTM, just a few missing doc comments

src/bin/languageserver/mod.rs
```diff
@@ -107,13 +127,18 @@ struct Files {
 //
 // More information can be found here: https://github.com/hyperledger/solang/pull/1411
 
+struct GlobalCache {
```
LucasSte, line 130: This struct deserves a doc comment.

LucasSte (approved):

src/bin/languageserver/mod.rs
chioni16: Hello! 
The functions `references`, `goto_type_definition` and `goto_implementation` are part of the standard interface that a Language server is expected to implement. 

As a result, the parameter and return types are documented in the `lsp_types` crate, which is where we import them from. One can see the documentation for the types by hovering over them (provided by a language server like rust-analyser) or by visiting the crate code. So, I believe addition of comments to describe the types may not add much value in this context. 

However, I do think it's worthwhile to add a comment describing the behaviour of the `goto_implementation` function as its behaviour is tailored to Solidity (expected to provide a list of the methods implemented for a given contract) and may be different from what one expects coming from a different language. 

As for the the other two functions, the behaviour remains consistent with the language server implementations for most languages and hence a comment may not be strictly necessary, but I will be happy to provide one, just in case. 

Thanks!

src/bin/languageserver/mod.rs
chioni16: Changed it into doc comments

src/bin/languageserver/mod.rs
chioni16: Added a doc comment that describes this struct

src/bin/languageserver/mod.rs
seanyoung: A comment is helpful to someone who is not familiar with language servers. You could explain:
- When do you expect this function to be called (user clicks "find all references", I believe)
- What parameters it receives (text position, file + line + column)
- What the function does with this (get reference from position, find all locations that have reference, etc)
- What is returned

src/bin/languageserver/mod.rs
seanyoung: "stores information" is not descriptive. Why do you have a global cache and not a per-file cache? This needs to be explained

xermicus: @chioni16 FYI, doc comments like that usually explain the higher level semantics of that struct. For example, why does it exist, how is supposed to be used, when is it useful, what pitfalls exists when using it. The question Sean suggested is a great example.

This helps other developers reading through the code base to get a much quicker understanding of how everything works together.

src/bin/languageserver/mod.rs
xermicus: @chioni16 still missing on the return type.

`Result<Option<GotoTypeDefinitionResponse>>`

Why / when can this function fail and return `Err`? What does a return value of `Ok(None)` exactly mean? Why can there be cases where the function succeeds but nothing was found (_assuming_ that this is what `Ok(None)` means)?

src/bin/languageserver/mod.rs
chioni16: Hello @seanyoung @xermicus @LucasSte !

I have tried to improve the doc comments as per your suggestions. The changes can be found here - https://github.com/hyperledger/solang/pull/1490/commits/1eadcad7c97611031eded5ba913d9ff976918d5c
Please let me know if something can be improved further. 

I really appreciate the suggestions and your time. 
Thanks!

xermicus (approved):
Thanks for adding the comments, looks great

PR #1497, opened 2023-08-22
Solana: Allow address and bytes to be used as @seed

seanyoung:
Allows `address` or `bytesN` to have the @seed annotation on constructors.

        contract bar {
            @payer(wallet)
            constructor(@seed address seed, @seed bytes2 seed2, @bump byte b) {}
        }

### DIFF
--- a/src/emit/expression.rs
+++ b/src/emit/expression.rs
@@ -2003,6 +2003,83 @@ fn runtime_cast<'a>(
 
             bin.builder.build_load(slice_ty, slice, "slice")
         }
+        (Type::Address(_), Type::Slice(_)) => {
+            let slice_ty = bin.llvm_type(to, ns);
+            let slice = bin.build_alloca(function, slice_ty, "slice");
+            let address = bin.build_alloca(function, bin.llvm_type(from, ns), "address");
+
+            bin.builder.build_store(address, val);
+
+            let data_ptr = bin
+                .builder
+                .build_struct_gep(slice_ty, slice, 0, "data")
+                .unwrap();
+
+            bin.builder.build_store(data_ptr, address);
+
+            let len = bin
+                .context
+                .i64_type()
+                .const_int(ns.address_length as u64, false);
+
+            let len_ptr = bin
+                .builder
+                .build_struct_gep(slice_ty, slice, 1, "len")
+                .unwrap();
+
+            bin.builder.build_store(len_ptr, len);
+
+            bin.builder.build_load(slice_ty, slice, "slice")
+        }
+        (Type::Bytes(bytes_length), Type::Slice(_)) => {
+            let llvm_ty = bin.llvm_type(from, ns);
+            let src = bin.build_alloca(function, llvm_ty, "src");
+
+            bin.builder.build_store(src, val.into_int_value());
+
+            let dest = bin.build_alloca(
+                function,
+                bin.context.i8_type().array_type((*bytes_length).into()),
+                "dest",
+            );
+
+            bin.builder.build_call(
+                bin.module.get_function("__leNtobeN").unwrap(),
+                &[
+                    src.into(),
+                    dest.into(),
+                    bin.context
+                        .i32_type()
+                        .const_int((*bytes_length).into(), false)
+                        .into(),
+                ],
+                "",
+            );
+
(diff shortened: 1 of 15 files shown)


src/emit/expression.rs
```diff
@@ -1904,106 +1904,182 @@ fn runtime_cast<'a>(
...
-    {
-        // no conversion needed
-        val
-    } else if let Type::Address(_) = to {
-        let llvm_ty = bin.llvm_type(from, ns);
+    match (from, to) {
+        (Type::Address(_) | Type::Contract(_), Type::Address(_) | Type::Contract(_)) => {
+            // no conversion needed
+            val
+        }
```
xermicus, line 1911: Why this match arm when the catch-all does the same?

seanyoung: You're right, it's not necessary. Removed.

xermicus (approved):
LGTM

src/emit/expression.rs
```diff
@@ -1904,106 +1904,178 @@ fn runtime_cast<'a>(
...
 ) -> BasicValueEnum<'a> {
-    if matches!(from, Type::Address(_) | Type::Contract(_))
-        && matches!(to, Type::Address(_) | Type::Contract(_))
-    {
-        // no conversion needed
-        val
-    } else if let Type::Address(_) = to {
-        let llvm_ty = bin.llvm_type(from, ns);
+    match (from, to) {
+        (Type::Bytes(_) | Type::Int(_) | Type::Uint(_) | Type::Value, Type::Address(_)) => {
```
LucasSte, line 1908: I think there is no more `Type::Value` after codegen.

src/emit/expression.rs
```diff
@@ -1904,106 +1904,178 @@ fn runtime_cast<'a>(
...
+            let len_ptr = bin
+                .builder
+                .build_struct_gep(slice_ty, slice, 1, "len")
+                .unwrap();
+
+            bin.builder.build_store(len_ptr, len);
+
+            bin.builder.build_load(slice_ty, slice, "slice")
+        }
+        _ => val,
```
LucasSte, line 2079: Please, specify when you want the runtime cast to just return the same thing. As you are matching all possible cases here, this could lead to future bugs.

src/emit/expression.rs
LucasSte, line 1: I think a more complete test for these changes would be to make an external call from Solidity that uses seeds.

tests/solana_tests/create_contract.rs
```diff
@@ -566,10 +566,9 @@ fn account_with_seed_bump() {
     let mut vm = build_solidity(
         r#"
         contract bar {
-
             @space(511 + 102)
             @payer(payer)
-            constructor(@seed bytes seed, @bump byte b) {}
+            constructor(@seed address seed, @seed bytes2 seed2, @bump byte b) {}
```
LucasSte, line 571: When creating an account with seeds, the runtime checks if the account address can be derived with the passed seeds and bump. Our virtual machine does not do so, does it? This may mask problems in our test.

seanyoung: The tests are specifically written to test that the correct PDA is created.

src/emit/expression.rs
seanyoung: We need the higher-level type information after codegen, in order to generate the correct debug information. Also, in program runtime v2 the debug information becomes an essential part of the interface (via BTF). We can't drop higher-level type information.

src/emit/expression.rs
seanyoung: Although this is existing behaviour, I've fixed  this and fixed a bunch of bugs I found.

src/emit/expression.rs
seanyoung: Good point, tests added

tests/solana_tests/create_contract.rs
LucasSte: From what I read in `tests/solana.rs` we do not check if the PDA being created matches the address derived with the seeds. Can you please pass in the wrong seed in this test to check if it fails?

I don't have my computer with me, and I don't want to delay this PR because of this.

LucasSte (approved):

PR #1500, opened 2023-08-23
Polkadot: Do not emit the storage initializer twice 

xermicus:
There is no need to deliberately emit the storage initializer CFG a second time. It just ends up as dead code because it's  being emitted anyways.

### DIFF
--- a/src/emit/polkadot/mod.rs
+++ b/src/emit/polkadot/mod.rs
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use std::ffi::CString;
+
 use crate::codegen::Options;
 use crate::sema::ast::{Contract, Namespace};
 use inkwell::context::Context;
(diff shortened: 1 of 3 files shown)


src/emit/polkadot/mod.rs
```diff
@@ -144,7 +146,16 @@ impl PolkadotTarget {
 
         emit_functions(&mut target, &mut binary, contract, ns);
 
-        let storage_initializer = emit_initializer(&mut target, &mut binary, contract, ns);
+        let function_name = CString::new(format!("{}:storage_initializer", contract.name)).unwrap();
```
seanyoung, line 149: The rest of the functions use `::` as separators. In fact, the contract name is not necessary at all, I think.

seanyoung (approved):

src/emit/polkadot/mod.rs
seanyoung: very nice!

PR #1501, opened 2023-08-23
Polkadot: Prevent storage initializers overwriting the contract input in scratch buf

xermicus:
~~Store the contract input on the stack instead of in the scratch buffer to avoid overwriting it in the storage initializer.~~

The call to the storage initializer must happen before reading the input into the scratch buffer. Otherwise, the storage initializers can overwrite the input data in the scratch buffer.

### DIFF
--- a/src/emit/polkadot/mod.rs
+++ b/src/emit/polkadot/mod.rs
@@ -189,7 +189,7 @@ impl PolkadotTarget {
     fn public_function_prelude<'a>(
         &self,
         binary: &Binary<'a>,
-        function: FunctionValue,
+        function: FunctionValue<'a>,
     ) -> (PointerValue<'a>, IntValue<'a>) {
         let entry = binary.context.append_basic_block(function, "entry");
 
@@ -200,12 +200,13 @@ impl PolkadotTarget {
             .builder
             .build_call(binary.module.get_function("__init_heap").unwrap(), &[], "");
 
-        let scratch_buf = binary.scratch.unwrap().as_pointer_value();
-        let scratch_len = binary.scratch_len.unwrap().as_pointer_value();
+        let input_buf_ty = binary.context.i8_type().array_type(SCRATCH_SIZE);
+        let input_buf = binary.build_alloca(function, input_buf_ty, "input_buf");
+        let input_len = binary.scratch_len.unwrap().as_pointer_value();
 
         // copy arguments from input buffer
         binary.builder.build_store(
-            scratch_len,
+            input_len,
             binary
                 .context
                 .i32_type()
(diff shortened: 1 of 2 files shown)


src/emit/polkadot/mod.rs
```diff
@@ -200,12 +200,13 @@ impl PolkadotTarget {
             .builder
             .build_call(binary.module.get_function("__init_heap").unwrap(), &[], "");
 
-        let scratch_buf = binary.scratch.unwrap().as_pointer_value();
-        let scratch_len = binary.scratch_len.unwrap().as_pointer_value();
+        let input_buf_ty = binary.context.i8_type().array_type(SCRATCH_SIZE);
```
seanyoung, line 203: I don't think this is a good idea. You're allocating 32KiB on the stack, and this is never freed. So, for the majority of the execution this memory is just wasted.

Also, after your change, whenever `msg.data`  is accessed, the input buffer is copied again, even though there already is a copy on the stack.

You could use another static buffer (just like the scratch buffer) for the input, and avoid the allocation cost and it can re-used whenever `msg.data` or `msg.sig` is used.

seanyoung: Alternatively, `msg.sender` could use the stack rather than the scratch buffer

xermicus: > I don't think this is a good idea. You're allocating 32KiB on the stack, and this is never freed. So, for the majority of the execution this memory is just wasted.

How is using the scratch buffer different?

> Also, after your change, whenever msg.data is accessed, the input buffer is copied again, even though there already is a copy on the stack.
>
> You could use another static buffer (just like the scratch buffer) for the input, and avoid the allocation cost and it can re-used whenever msg.data or msg.sig is used.

I think a dedicated buffer for the input is a good idea. However I'm not concerned about the cost of the stack allocation. I didn't check but I'd be very surprised if its expensive (even more if it was so expensive that it's a concern; we use it for API calls which are super costly anyways). Ideally we can just :axe: the scratch buffer completely, it's causing multiple problems like this.

xermicus: Or, as discussed, we can move the call to the storage initializer before the input reading. This will avoid the problem too.

xermicus: @seanyoung you were right, I solved it differently. WDYT?

src/emit/polkadot/mod.rs
```diff
@@ -200,7 +200,8 @@ impl PolkadotTarget {
     fn public_function_prelude<'a>(
         &self,
         binary: &Binary<'a>,
-        function: FunctionValue,
+        function: FunctionValue<'a>,
+        init: Option<FunctionValue>,
```
seanyoung, line 204: Call it `storage_initializer`

seanyoung (approved):

src/emit/polkadot/mod.rs
```diff
@@ -341,19 +347,15 @@ impl PolkadotTarget {
         let ty = bin.context.void_type().fn_type(&[], false);
         let export_name = if init.is_some() { "deploy" } else { "call" };
         let func = bin.module.add_function(export_name, ty, None);
-        let (input, input_length) = self.public_function_prelude(bin, func);
+        let (input, input_length) = self.public_function_prelude(bin, func, init);
```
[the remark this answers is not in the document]
xermicus: changed both

PR #1503, opened 2023-08-24
Bugfix: Incorrect size width calculation of dynamic primitive arrays in SCALE

### LINKED ISSUE #1502: Incorrect encoding of integer arrays
**Describe the bug**
Integer arrays returned by a Solidity smart contract are incorrectly Scale encoded when the contract is compiled for the Substrate compilation target.

**To Reproduce**
Compile the following contract via Solang using `solang compile --target substrate ...`
```solidity
//SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.13;

contract Test {
    function test(
        
    )
        external view
        returns (uint256[] memory _amounts)
    {
        _amounts = new uint256[](2);
        _amounts[0] = 0x201f1e1d1c1b1a191817161514131211100f0e0d0c0b0a090807060504030201;
        _amounts[1] = 0x201f1e1d1c1b1a191817161514131211100f0e0d0c0b0a090807060504030201;
    }
}
```
Then instantiate the wasm smart contract on a Substrate chain and send the message `test`. This message will return the following binary Scale encoded value consisting of 66 bytes:
```
0x080102030405060708090a0b0c0d0e0f...1415161718191a1b1c1d1e1f2000
```

However, the correct encoding of the vector should be the 65 bytes (1 byte for the initial `0x08` being the compact encoding of the number 2 for the length of the vector and then 32 bytes for each entry):
```
0x080102030405060708090a0b0c0d0e0f...1415161718191a1b1c1d1e1f20
```
i.e., there is an incorrect additional trailing `00`-byte.

When sending the message `test` via the npm module `@polkadot/api`, then it will show the following error message:
```
createType(Vec<u256>):: Vec<u256>:: Decoded input doesn't match input, received 0x080102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f…02030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f2000 (66 bytes), created 0x080102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f…0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20 (65 bytes)
```

**Similar errors**
The contract will generate similar errors when using other `uint` data types whose number of bytes is not a power of two. For example, consider the following smart contract:
```solidity
//SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.13;

contract Test {
    function test(
        
    )
        external view
        returns (uint24[] memory _amounts)
    {
        _amounts = new uint24[](2);
        _amounts[0] = 0x030201;
        _amounts[1] = 0x030201;
    }
}
```
The ABI of the compiled smart contract specifies the return type of `test` as `Vec<u32>`. However, it returns `0x08010203000102` instead of the correct `0x080102030001020300`. The library `@polkadot/api` will show the following error message:

```
createType(Vec<u32>):: Vec<u32>:: Decoded input doesn't match input, received 0x08010203000102 (7 bytes), created 0x080102030001020000 (9 bytes)
```

**Potential further problem**
I didn't test but find it plausible that also the decoding of message arguments that are integer arrays is not working properly.

**Solang version**
solang version v0.3.1

xermicus:
Calculate the size width of dynamic primitive arrays based on the array length instead of the array size.

Partially fixes #1502

### DIFF
--- a/tests/polkadot_tests/arrays.rs
+++ b/tests/polkadot_tests/arrays.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use parity_scale_codec::{Decode, Encode};
+use primitive_types::U256;
 use rand::Rng;
 
 use crate::build_solidity;
@@ -1411,6 +1412,26 @@ fn abi_encode_dynamic_array4() {
         ])
         .encode()
     );
+}
+
+#[test]
+fn abi_encode_dynamic_array5() {
+    let uint_str = "201f1e1d1c1b1a191817161514131211100f0e0d0c0b0a090807060504030201";
+    let uint_val = U256::from_str_radix(uint_str, 16).unwrap();
+
+    let mut runtime = build_solidity(&format!(
+        "contract Test {{
+            function test() external view returns (uint256[] memory _amounts) {{
+                _amounts = new uint256[](2);
+                _amounts[0] = 0x{};
+                _amounts[1] = 0x{};
+            }}
+        }}",
+        uint_str, uint_str
+    ));
+
+    runtime.function("test", Vec::new());
+    assert_eq!(runtime.output(), vec![uint_val, uint_val].encode())
 }
 
 #[test]
--- a/src/codegen/encoding/mod.rs
+++ b/src/codegen/encoding/mod.rs
@@ -1515,6 +1515,8 @@ pub(crate) trait AbiEncoding {
                 };
             }
 
+            let size_width = self.size_width(&size, vartab, cfg);
+
             let type_size = Expression::NumberLiteral {
                 loc: Codegen,
                 ty: Uint(32),
@@ -1544,7 +1546,6 @@ pub(crate) trait AbiEncoding {
             if self.is_packed() || !matches!(&dims.last().unwrap(), ArrayLength::Dynamic) {
                 return size_var;
             }
-            let size_width = self.size_width(&size_var, vartab, cfg);
             Expression::Add {
                 loc: Codegen,
                 ty: Uint(32),


tests/polkadot_tests/arrays.rs
```diff
@@ -1413,6 +1414,26 @@ fn abi_encode_dynamic_array4() {
...
+
+    let mut runtime = build_solidity(&format!(
+        "contract Test {{
+            function test() external view returns (uint256[] memory _amounts) {{
+                _amounts = new uint256[](2);
+                _amounts[0] = 0x{};
+                _amounts[1] = 0x{};
+            }}
+        }}",
+        uint_str, uint_str
```
seanyoung, line 1430: I think a slightly better test would have two different values - or even, two random values.

seanyoung (approved):

tests/polkadot_tests/arrays.rs
xermicus: Please no random values, they make debugging failing tests painful :monkey: 

But I agree about using two different values, gonna change it before merging :+1:

seanyoung: I think using random values is great, it gives much better coverage. You might get random failures but those identify real issues.

xermicus: Ok, so TIL about [monkey testing](https://en.wikipedia.org/wiki/Monkey_testing) actually being a thing.

Inspired by this I wrote another test, which monkey tests (using random values each time) the encoding and decoding of the array for _every_ array length between 0 and 128.

Issue #1495, opened 2023-08-20
Unexpected variable type in function name mangling


YanhuiJessica:
```js
contract example {
    function foo() public {}
    function foo(address) public {}
}
```
`foo(address)` is mangled to `foo_bytes32`. The build will fail if the function `foo(bytes32)` also exists.

**Hyperledger Solang version**
v0.3.1

seanyoung:
Please tell us what the unexpected behavior is. I guess this is what you mean:


```solidity
contract example {
    function foo() public {}
    function foo(address) public {}
    function foo(bytes32) public {}
}
```
Results in:
```bash
$ cargo run compile --target solana f.sol
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/solang compile --target solana f.sol`
error: function 'foo' overrides function in same contract
  ┌─ /home/sean/solang/release/f.sol:4:5
  │
3 │     function foo(address) public {}
  │     ---------------------------- previous definition of 'foo'
4 │     function foo(bytes32) public {}
  │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Is that what you mean?

YanhuiJessica:
Yes. In Solidity, `foo(address)` and `foo(bytes32)` can co-exist. `foo(address)` is expected to be mangled to `foo_address`.

PR #1505, opened 2023-08-25
Fix address name mangling

seanyoung:
There was some old code for when we were using eth abis on Solana.

Fixes https://github.com/hyperledger/solang/issues/1495

### DIFF
--- a/tests/solana_tests/simple.rs
+++ b/tests/solana_tests/simple.rs
@@ -581,3 +581,110 @@ contract test3 {
         );
     }
 }
+
+#[test]
+fn overloading() {
+    let mut vm = build_solidity(
+        r#"
+        contract c {
+            function foo() public returns (uint32) { return 1; }
+            function foo_bar() public returns (uint32) { return 0; }
+            function foo(address) public returns (uint32) { return 2; }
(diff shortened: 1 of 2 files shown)


tests/solana_tests/simple.rs
```diff
@@ -581,3 +581,110 @@ contract test3 {
...
+        }"#,
+    );
+
+    let data_account = vm.initialize_data_account();
+    vm.function("new")
+        .accounts(vec![("dataAccount", data_account)])
+        .call();
+
+    let returns = vm
+        .function("foo_")
```
LucasSte, line 605: What happens when I call only `foo` (without the underscore)?

seanyoung: That does not work, the mangling appends `_` followed by the argument types for overloaded functions. In  this case, there are no arguments so the function is called `foo_`.  See https://github.com/hyperledger/solang/blob/main/src/sema/ast.rs#L422

seanyoung: So `"("` => `"_"` and `")"` => '"", that's how we end up there.

LucasSte (approved):

PR #1218, opened 2023-03-10
improve resolve pragma

ilyar:

PR #1511, opened 2023-08-31
improve resolve pragma

ilyar:
reopen https://github.com/hyperledger/solang/pull/1218


seanyoung:
This code does not compile, as you can see in CI. No attempt has been made to fix it.

PR #1513, opened 2023-09-04
Reduce CI costs

seanyoung:
cargo clippy and cargo doc need to do the same work as cargo build, so put them in the linux job, and reduce the duplicate effort.

In addition, use the free windows runner as the cost for solang-windows-latest is much too high. @ryjones 

Before this PR, we would running clippy with the latest stable. After this PR, we run clippy with 1.70.0 (or the current version we're using for building, rather than latest).


ryjones (approved):

xermicus (approved):

seanyoung:
> This PR was merged before I could submit my review.

For future reference, @ryjones in our project, the PR author merges once *all* the maintainers have reviewed the change and given their  :heavy_check_mark: approval

PR #1515, opened 2023-09-06
Add Polkadot Node CLI Interactions

tareknaser:
- [ ] add a test case (upload -> remove)
- [ ] add documentation

### DIFF
--- a/src/bin/cli/mod.rs
+++ b/src/bin/cli/mod.rs
@@ -16,9 +16,15 @@ use solang::{
     file_resolver::FileResolver,
     Target,
 };
+use std::fs::File;
+use std::io::Read;
 use std::{ffi::OsString, path::PathBuf, process::exit};
-
+mod polkadot;
 mod test;
+
+use polkadot::{
+    PolkadotCallCommand, PolkadotInstantiateCommand, PolkadotRemoveCommand, PolkadotUploadCommand,
+};
 #[derive(Parser)]
 #[command(author = env!("CARGO_PKG_AUTHORS"), version = concat!("version ", env!("SOLANG_VERSION")), about = env!("CARGO_PKG_DESCRIPTION"), subcommand_required = true)]
 pub struct Cli {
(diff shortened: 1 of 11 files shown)


src/bin/cli/mod.rs
```diff
@@ -682,3 +702,37 @@ fn explicit_args(matches: &ArgMatches) -> Vec<&Id> {
         })
         .collect()
 }
+
+/// A helper function to check if the target name provided by the user matches the target name in solang.toml
+fn check_target_match(target_name: &str) {
```
xermicus, line 707: This isn't tested anywhere.

src/bin/cli/polkadot/upload.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn to_json(&self) -> Result<String> {
+        Ok(serde_json::to_string_pretty(self)?)
+    }
+
+    pub fn print(&self) {
+        name_value_println!("Result", self.result);
+        name_value_println!("Code hash", format!("{:?}", self.code_hash));
+        name_value_println!("Deposit", format!("{:?}", self.deposit));
+    }
+}
```
xermicus, line 132: This code is stolen from [cargo-contract](https://github.com/paritytech/cargo-contract/blob/632959c5e37e7d86bf09e1966e8c6637321cd9aa/crates/cargo-contract/src/cmd/upload.rs#L123C1-L123C1).

src/bin/cli/polkadot/call.rs
```diff
@@ -0,0 +1,219 @@
+// SPDX-License-Identifier: Apache-2.0
```
xermicus, line 1: Code stolen from [cargo contract](https://github.com/paritytech/cargo-contract/blob/632959c5e37e7d86bf09e1966e8c6637321cd9aa/crates/cargo-contract/src/cmd/call.rs).

I'm no expert but strictly this might even constitute a GPL violations, as Solang is not GPL licensed.

xermicus (approved):

xermicus:
The entire PR is mostly a blatant copy of the code in cargo contract. Without even crediting the source. Which is unacceptable and might even be subject to license violations.

xermicus:
In the review I Just misclicked, I do NOT approve these changes.

tareknaser:
Could you elaborate more on how to do the PR?
My intention was to replicate `cargo-contract`'s functionality to meet `solang`’s needs.

To clarify, are there specific functions or aspects of `cargo-contract` code that we should implement differently in `solang`? 
I'd like to ensure we align with `solang`’s goals while respecting licensing.


I'll redo the PR from scratch and appreciate your guidance.

xermicus:
> Could you elaborate more on how to do the PR?

You copied vast amount of code from cargo contract, without crediting the source, and without respecting the license.  Which constitutes an unacceptable contribution. This is why I closed the PR.

> My intention was to replicate cargo-contract's functionality to meet solang’s needs.
> To clarify, are there specific functions or aspects of cargo-contract code that we should implement differently in solang?
> I'd like to ensure we align with solang’s goals while respecting licensing.

Copy + Pasting code is almost never the right approach. If you feel like cargo contract already has got everything you'd need, you should think about how you can make it a re-usable crate instead.

Issue #698, opened 2022-03-14
Support custom errors


seanyoung:
Ethereum Solidity 0.8 introduces a new syntax for custom errors.

sema changes:
- Using the `error Foo(int,string)` type syntax, custom errors can be declared. These need resolving in sema, which should be identical to event definition parsing
- Resolve `revert Foo(12, "bar")` type syntax. Again this should be very similar to `emit` resolving.
- The error defined in another contract be used using `revert ContractX.Foo(12, "bar")`
- The error can be imported from another contract via an import type:
```
// foo.sol
contract X  { error Foo(int,string); }
}
// bar.sol
import "foo.sol" as I;

contract A {
    function x() public { revert I.X.Foo(1, ""); }
}
```
- import types can imported again, so it's possible to have:
```
contract A {
    function x() public { revert I.a.b.c.X.Foo(1, ""); }
}
```

codegen changes:
- The should be a new Revert instruction for the cfg which looks like emit
- All the passes should be able to use expressions in the revert instruction, just like emit

emit changes:
 - Generate llvm for new revert instruction. This should be similar to emit, but now terminates the contract

try-catch support for custom errors:
 - The try-catch syntax can catch any number of custom errors. This needs support too

xermicus:
Found elsewhere but seems related here: 

```solidity
pragma solidity ^0.8.0;

contract Implementation {
    function foo(bool ok, string s) public pure {
        require(ok, s);
    }
}

contract A {
    address init;

    constructor(address init_addr) {
        init = init_addr;
    }

    function f(bytes calldata _calldata) public {
        (bool success, bytes memory error) = init.delegatecall(_calldata);
        if (!success) {
            if (error.length > 0) {
                revert(string(error));
            } else {
                revert("aa");
            }
        }
    }
}
```

This isn't displayed correctly:

```
runtime_error: 56a5f41bf5df82020a2efe5ee1894169a0f555d980038a809ed1eb469b8d64a4000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000.... 
```

Also this makes a nice testcase.

xermicus:
Closed by #1516

PR #1516, opened 2023-09-07
Polkadot: Support errors according to `solc`

xermicus:
Implement `Panic` and custom errors for the Polkadot target:
- Allow catching `Panic` in sema
- Allow custom errors in sema
- Implement selector  calculation for custom errors
- Refactor try-catch in codegen to allow catch clauses on `Panic` errors in
- Add any custom errors in the metadata


xermicus:
Closes #698

docs/language/statements.rst
```diff
@@ -145,3 +142,10 @@ This might be useful when no error string is expected, and will generate shorter
 
 .. include:: ../examples/polkadot/statement_try_catch_no_error_handling.sol
   :code: solidity
+
+.. note::
+
+    At the moment, only catching ``Error`` and ``Panic`` errors with an explicit catch clause is supported.
```
LucasSte, line 148:
```suggestion
    At the moment, only catching ``Error`` and ``Panic`` errors with an explicit catch clause are supported.
```

src/codegen/revert.rs
```diff
@@ -57,11 +64,20 @@ impl SolidityError {
...
             Self::String(_) => ERROR_SELECTOR,
             Self::Panic(_) => PANIC_SELECTOR,
+            Self::Custom { error_no, .. } => {
+                let mut buf = [0u8; 32];
+                let mut hasher = Keccak::v256();
+                let signature =
+                    ns.signature(&ns.errors[*error_no].name, &ns.errors[*error_no].fields);
+                hasher.update(signature.as_bytes());
+                hasher.finalize(&mut buf);
+                u32::from_be_bytes([buf[0], buf[1], buf[2], buf[3]])
```
LucasSte, line 79: What about `&buf[0..4]`?

src/codegen/revert.rs
```diff
@@ -294,50 +329,57 @@ pub(super) fn revert(
...
-        } else {
-            log_runtime_error(
-                opt.log_runtime_errors,
-                "revert encountered",
-                *loc,
-                cfg,
-                vartab,
-                ns,
-            )
+        match (error_no, &expr) {
```
LucasSte, line 332: There is no need to mach `error_no`. You can simplify this statement by using only a `if let Some(expr) = expr { ... } else { ... }`

src/codegen/revert.rs
```diff
@@ -402,14 +448,58 @@ mod tests {
...
+            SolidityError::String(Expression::Poison).selector(&ns),
         );
         assert_eq!(
             PANIC_SELECTOR,
-            SolidityError::Panic(PanicCode::Generic).selector(),
+            SolidityError::Panic(PanicCode::Generic).selector(&ns),
         );
     }
+
+    /// Tests the error selector calculation to be correct against two examples:
```
LucasSte, line 463:
```suggestion
    /// Tests if the error selector calculation is correct against two examples:
```

src/codegen/revert.rs
```diff
@@ -402,14 +448,58 @@ mod tests {
...
+    /// - `error ERC20InsufficientBalance(address sender, uint256 balance, uint256 needed);`
+    /// - `error Unauthorized();`
+    ///
+    /// Error selector calculation uses the same signature algorithm used for message selectors.
+    /// Which is expected to already have sufficient test coverage.
```
LucasSte, line 468:
```suggestion
    /// Error selector calculation uses the same signature algorithm used for message selectors,
    /// which is expected to already have sufficient test coverage.
```

LucasSte: Can you point where the expected tests are (in your comment)?

src/codegen/statements/try_catch.rs
```diff
@@ -366,7 +366,9 @@ fn insert_success_code_block(
     }
 }
 
-/// Insert all catch cases into the CFG.
+/// Insert all catch cases into the CFG (error selectors and catch-all clause).
+/// Currently, only catching "Error" and "Panic" errors are supported.
```
LucasSte, line 370:
```suggestion
/// Currently, only catching "Error" and "Panic" errors is supported.
```

tests/contract_testcases/polkadot/calls/try_catch_external_calls_04.sol
```diff
@@ -21,4 +21,4 @@
         }
         
 // ---- Expect: diagnostics ----
-// error: 8:25-28: only catch 'Error' is supported, not 'Foo'
+// error: 8:25-28: only catch 'Error' or 'Panic' is supported, not 'Foo'
```
LucasSte, line 24:
```suggestion
// error: 8:25-28: only catch 'Error' and 'Panic' are supported, not 'Foo'
```

src/codegen/revert.rs
xermicus: `u32::from_be_bytes` needs a `[u8; 4]` and not a `&[u8]`

src/codegen/revert.rs
xermicus: No, we only want to print for `Error(String)` which has an `expr` but no `error_no`

src/codegen/revert.rs
xermicus: I made this comment to justify as to why this test is not a test for our signature generation algorithm. There are a many different tests for the signature algorithm across the codebase, for example in sema directly, with function overloading, in the ABI encoding, in the integration tests, in the unit tests, and maybe other places I can't immediately think of. How would adding this list improve this comment?

src/codegen/revert.rs
LucasSte: You also match for `_`, which includes the case for when `error_no` is not none.

src/codegen/revert.rs
LucasSte: You wrote that this is `expected` to have tests elsewhere, but it is not clear for me where they are.

src/codegen/revert.rs
xermicus: If this was an `if let Some(expr) = expr` as you suggest. All custom errors with data would end in the `if let` branch. Which is doesn't work because they might contain data that is formattable as a string.

src/sema/ast.rs
```diff
@@ -60,6 +60,7 @@ pub enum Type {
     UserType(usize),
     /// There is no way to declare value in Solidity (should there be?)
     Value,
+    #[default]
```
seanyoung, line 63: I think this is a bit strange. Why is `Void` the default? I am not sure `Type` should have a default

src/codegen/revert.rs
```diff
@@ -38,15 +39,21 @@ pub enum SolidityError {
     String(Expression),
```
seanyoung, line 39: Why is this called `String` and not `Error`?

src/codegen/statements/try_catch.rs
```diff
@@ -439,47 +433,93 @@ fn insert_catch_clauses(
...
+            kind: Builtin::ReadFromBuffer,
+            args: vec![buffer.clone(), offset],
+        };
+        let cond = Expression::Equal {
+            loc: Codegen,
+            left: err_selector.into(),
+            right: err_id.into(),
+        };
+        let match_err_id = cfg.new_basic_block("match_err_id".into());
+        let instruction = Instr::BranchCond {
```
seanyoung, line 478: This would a lot more efficient with `Instr::Switch`.

For every error case you're reading the value from the buffer, rather than keeping the value in a `Type::Uint(32)` or `Type::Bytes(4)` and switching on that.

src/codegen/revert.rs
```diff
@@ -38,15 +39,21 @@ pub enum SolidityError {
...
     /// The `Panic(uint256)` selector
     Panic(PanicCode),
+    /// User defined errors
+    Custom {
+        error_no: usize,
+        expr: Option<Expression>,
```
seanyoung, line 45: I would expect this to be `Vec<Expression>` because a custom error can have any number of arguments.

seanyoung (changes requested):

src/sema/ast.rs
xermicus: It's very useful in tests. Without this, deriving default of other structs containing a `Type` isn't possible.

> Why is Void the default? 

Well, I had to pick one. I also saw `Unreachable` as a candidate, but it doesn't have any doc-comment so I wasn't too sure.

src/codegen/revert.rs
xermicus: If you write `revert("foo")` then this is just a string. However in the catch clause it's `Error(string)`. I'm not really sure which is ultimately the correct one. What I'm sure about is that I don't care how this is called, I can change it to `Error` if you want.

src/codegen/statements/try_catch.rs
xermicus: There are a maximum of two possible switch cases. So I don't think it's a lot more efficient even in the worst case.

This was done with a loop before, so I just adapted it. However I agree using a switch is cleaner, I can change that.

> For every error case you're reading the value from the buffer

Good catch, this is indeed unnecessary

src/codegen/revert.rs
xermicus: Ah right, this was a bit confusing, because the statement can be `revert("foo")` but also `revert Foo()`, I assumed in the case of a custom error it would capture the error (as a composite) instead individual arguments. Which is not the case, thanks for pointing this out

docs/language/statements.rst
```diff
@@ -145,3 +142,10 @@ This might be useful when no error string is expected, and will generate shorter
 
 .. include:: ../examples/polkadot/statement_try_catch_no_error_handling.sol
   :code: solidity
+
+.. note::
+
+    At the moment, only catching ``Error`` and ``Panic`` errors with an explicit catch clause are supported.
```
seanyoung, line 148: Remove "At the moment". 

You're not planing to implement this, are you?

src/codegen/revert.rs
```diff
@@ -57,11 +64,20 @@ impl SolidityError {
     }
 
     /// Return the selector of the error.
-    pub fn selector(&self) -> u32 {
+    pub fn selector(&self, ns: &Namespace) -> u32 {
```
seanyoung, line 67: Using `u32` has endianness problems. Why not `[u8; 4]` or `Vec<u8>`

src/codegen/revert.rs
```diff
@@ -77,8 +93,28 @@ impl SolidityError {
...
-            Self::String(data) => {
-                let args = vec![self.selector_expression(), data.clone()];
+            Self::Error(expr) => {
+                let args = vec![self.selector_expression(ns), expr.clone()];
+                create_encoder(ns, false)
+                    .const_encode(&args)
+                    .map(|bytes| {
+                        let size = Expression::NumberLiteral {
+                            loc: Codegen,
+                            ty: Type::Uint(32),
```
seanyoung, line 103: `Type::Bytes(4)` would be much nicer imho

src/codegen/revert.rs
```diff
@@ -35,18 +36,24 @@ pub enum SolidityError {
...
     Empty,
     /// The `Error(string)` selector
-    String(Expression),
+    Error(Expression),
     /// The `Panic(uint256)` selector
     Panic(PanicCode),
+    /// User defined errors
+    Custom {
+        error_no: usize,
+        expr: Vec<Expression>,
```
seanyoung, line 45: `exprs`? We usually use `expr` for singular.

src/codegen/revert.rs
```diff
@@ -290,54 +327,62 @@ pub(super) fn revert(
...
+                                loc: Codegen,
+                                ty: Type::Bytes(error_string.as_bytes().len() as u8),
+                                value: error_string.as_bytes().to_vec(),
+                            },
+                        ),
+                    ],
+                };
+                cfg.add(vartab, Instr::Print { expr: print_expr });
+            }
+            // Else: Not all fields might be formattable, just print the error type
```
seanyoung, line 367: Why not print the fields if they are formattable? Seems very useful

src/codegen/statements/try_catch.rs
```diff
@@ -459,31 +512,17 @@ fn insert_catch_clauses(
...
-        loc: Codegen,
-        left: err_selector.into(),
-        right: err_id.into(),
-    };
-    let match_err_id = cfg.new_basic_block("match_err_id".into());
-    let instruction = Instr::BranchCond {
-        cond,
-        true_block: match_err_id,
-        false_block: no_match_err_id,
+    let instruction = Instr::Switch {
```
seanyoung, line 515: The switch looks very nice here, great! :heart_eyes:

src/sema/ast.rs
```diff
@@ -61,6 +61,7 @@ pub enum Type {
     /// There is no way to declare value in Solidity (should there be?)
     Value,
     Void,
+    #[default]
```
seanyoung, line 64: I don't think `Type` should have a default value. Just does not make any sense.

seanyoung (changes requested):

src/sema/ast.rs
xermicus: Why do you think that it doesn't make any sense for types to be `Unreachable` by default? I argue the opposite: It makes sense because in tests, I need to work with data structures containing `Type`s, but I don't care about that part. So I want to leave them uninitialized.

We could also pick `Unresolved` by default. We just say that, an uninitiated type ("default") is unresolved or unreachable. To me it makes sense to declare either as the default value. And I'd expect our codegen to refuse CFGs containing `Unreachable` or `Unresolved` types anyways. So what are problems introduced with doing that you're afraid of?

src/codegen/statements/try_catch.rs
xermicus: Thanks!

src/codegen/revert.rs
xermicus: Changed that

src/codegen/revert.rs
xermicus: I'm not sure. Why do you think that the buffer size not be an uint?

src/codegen/revert.rs
xermicus: > Using u32 has endianness problems

This u32 is only used internally in the compiler. Surely the Rust language does handle that correctly regardless for all plattforms right?

However, it's still a good point, it should be `[u8; 4]` and not `u32` as we never do arithmetics on selectors. It does also make it easier to work with in many places. Thanks!

docs/language/statements.rst
xermicus: Yes and no. My plan is to implement this if either `solc` supports it as well or it's requested by our users.

seanyoung: So that is true for every possible Solidity feature: we'll implement if it is requested by users or implemented by solc.

seanyoung: - It's making solang look unfinished
- it does not add anything, users don't need to be told "at the moment"
- I think it looks unprofessional

src/sema/ast.rs
seanyoung: Can you point me to code that is made easier by having a default for `Type`?

docs/language/statements.rst
xermicus: FYI I already removed "At the moment" as you suggested :slightly_smiling_face: 

> So that is true for every possible Solidity feature: we'll implement if it is requested by users or implemented by solc.

This would justify leaving "At the moment", because at this moment we don't support, but at the moment it's requested we would support it, wouldn't it? However, it's not important to me.

src/sema/ast.rs
xermicus: [Sure](https://github.com/hyperledger/solang/pull/1516/files#diff-65527b752e88b0b1678c19e4d555d0db9e9f07c8c60de9050ef8c5e5086ebbdaR476-R499)

And this is just one place, there are more existing tests like that, and more to be written in the future.

seanyoung: You're trying to do:
```rust
                 fields: vec![
                     Parameter {
                         ty: Type::Address(false),
                         ..Default::default()
                     },
                     Parameter {
                         ty: Type::Uint(256),
                         ..Default::default()
                     },
                     Parameter {
                         ty: Type::Uint(256),
                         ..Default::default()
                     },
                 ],
```
I think this can be better achieved with a function

```rust
impl Parameter {
    fn default(ty: Type) -> Self {
        Parameter {
            ty,
           indexed: false,
           ...
 
        }
    }
}
```
Now your test code is even shorter:

```rust
                 fields: vec![
                     Parameter::default(Type::Address(false)),
                     Parameter::default(Type::Uint(256)),
                     Parameter::default(Type::Uint(256)),
                 ],
```

xermicus: I think this is a bad idea.

The `Default` trait is used extensively across many rust crates.

Why do you desperately want to work around that?

xermicus: Actually with builders, we could do something close to what you suggested but without having `Type` to implement default, but in a more idiomatic manner.

However I still don't see why `Type` can not implement `Default`

seanyoung: Because does not have an natural default. You keep on changing your mind on which one should be default - that's a clear indicator that there is no good default value.

Writing code is about not doing anything surprising/unexpected. I think making some random type the default is pretty much the definition of doing the unexpected. If you had this enum:

It's just like:
```rust
enum Coin {
    Penny,
    Nickel,
    Dime,
    Quarter,
}
```
What default value would you pick?

xermicus: > Because does not have an natural default.

Rusts [Default](https://doc.rust-lang.org/std/default/trait.Default.html) has **nothing** to whether there is a natural default. I quote:

> Sometimes, you want to fall back to some kind of default value, and don’t particularly care what it is.

And this is also _exactly_ why I want to have it implement `Default`.

>  You keep on changing your mind on which one should be default - that's a clear indicator that there is no good default value.
> Writing code is about not doing anything surprising/unexpected. I think making some random type the default is pretty much the definition of doing the unexpected. If you had this enum

It doesn't matter which value we pick by _definition_ of the `Default` trait. But choosing `Unreachable` or `Unresolved` should give us an extra safeguard for bugs in codegen.

I still don't see why this can't implement `Default`

LucasSte: The impression I have about setting a default for `Type` is that either there is a default type in Solidity or the compiler has an internal default type. Neither option is true. We must declare a specific type in Solidity and the runtime needs to know the type of a value in order to operate it.

Types must be explicit in a compiler, so I do not like the idea of choosing a default for it. As Sean pointed, there is no default type in Solidity nor in a compiler. We should enforce everyone to explicitly choose one.

LucasSte: I also believe `Type` should not have a fallback option, since types should be explicit. When you are executing a program in your CPU, it does not assume a fallback type in case it is not specified. It does not know what to do.

xermicus: Ok, if it's too concerning even if `Unreachable` was picked. I changed it so it's only possible to use that in the tests. And then we can solve the problem outside tests with builders. WDYT

src/codegen/revert.rs
```diff
@@ -402,14 +452,58 @@ mod tests {
...
+            SolidityError::Panic(PanicCode::Generic).selector(&ns),
         );
     }
+
+    /// Tests if the error selector calculation is correct against two examples:
+    /// - `error ERC20InsufficientBalance(address sender, uint256 balance, uint256 needed);`
+    /// - `error Unauthorized();`
+    ///
+    /// Error selector calculation uses the same signature algorithm used for message selectors,
+    /// which is expected to already have sufficient test coverage.
```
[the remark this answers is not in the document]
xermicus: I already changed it but didn't save so it didn't commit sorry

src/sema/ast.rs
xermicus: > The impression I have about setting a default for Type is that either there is a default type in Solidity or the compiler has an internal default type. Neither option is true.
> We must declare a specific type in Solidity and the runtime needs to know the type of a value in order to operate it.
> I also believe Type should not have a fallback option

This doesn't seem accurate to me. The existence of the `Unresolved` and `Unreachable` `Type` variants directly contradict the entire reasoning.

src/codegen/revert.rs
seanyoung: Sorry, don't know what I was thinking. I was confused with the selector here, ignore me

src/sema/ast.rs
seanyoung: > Ok, if it's too concerning even if `Unreachable` was picked. I changed it so it's only possible to use that in the tests. And then we can solve the problem outside tests with builders. WDYT

This makes the code even worse.

You're probably thinking, who cares what the default is, I just want to be able to `#![derive(Default)]`. I just don't agree with that, some things don't have a default value and should not have one.

seanyoung: `Unreachable` has a very specific meaning - control flow should never reach this point
`Unresolved` should never occur in codegen

xermicus: > This makes the code even worse.

Why? It makes writing tests easier, making the code arguably better.

> some things don't have a default value

Again, the idea of the `Default` trait is not to assign a "meaningful" default (for example `bool` and numeric values have a default value too). It's just a way to tell that it's possible to use this type without specifying what the value it is. And we need _exactly_ that in the tests. Now the concern is that implementing this makes it easier to accidentally use the value of `Type::default()` implicitly. Which, by your own reasoning, would be mitigated by picking `Unreachable` or `Unresolved` as the default value. And further giving in, by only implementing it for tests, it's not even possible.

However this is going nowhere, and blocking off the entire contribution. I just tried to make writing unit tests easier using some Rust Language intrinsic. Unfortunately this is being opposed to the point it's starting to get very time consuming is discouraging. I removed the `Default` implementation on type and implemented a helper on `Parameter`, which you liked better, instead

src/sema/ast.rs
```diff
@@ -24,6 +24,7 @@ use std::{
 };
 use tiny_keccak::{Hasher, Keccak};
 
+#[cfg_attr(test, derive(Default))]
```
LucasSte, line 27: Now that you are not using the default for `Type` anymore, is this attribute still necessary?

tests/polkadot_tests/calls.rs
```diff
@@ -1203,3 +1221,102 @@ fn try_catch_transfer_fail() {
...
+    let mut runtime = build_solidity(
+        r#"contract A {
+        function a(uint div) public payable returns (uint) {
+            B b = new B();
+            try b.b(div) returns (uint) {
+                return 3;
+            } catch Error(string) {
+                return 1;
+            } catch Panic(uint) {
+                return 0;
```
LucasSte, line 1262: I think it is good to ensure that the string and the integer are properly decoded.

src/sema/ast.rs
xermicus: No. I must have forgotten to `git push` before I left yesterday, sorry

integration/polkadot/try_catch.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    function test(uint128 div) public payable returns (uint128) {
+        TryCatchCallee instance = new TryCatchCallee();
+
+        try instance.test(div) returns (uint128) {
+            return 4;
+        } catch Error(string) {
+            return 1;
+        } catch Panic(uint) {
+            return 0;
```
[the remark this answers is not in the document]
xermicus: Yes, in other places, but I added the assert anyways

tests/polkadot_tests/calls.rs
xermicus: It's tested in other places, but I added the assert anyways

seanyoung (approved):
Very nice, thank you for tolerating my complaints about `#[default]` on `Type`.

src/sema/statements.rs
```diff
@@ -2520,30 +2520,17 @@ fn try_catch(
...
-                    }
+                if id.name != "Error" && id.name != "Panic" {
+                    let message = format!(
+                        "only catch 'Error' and 'Panic' are supported, not '{}'",
+                        id.name
+                    );
+                    ns.diagnostics.push(Diagnostic::error(id.loc, message));
+                    return Err(());
+                }
+                if let Some(annotation) = &param.annotation {
```
LucasSte, line 2531:
```suggestion
                } else if let Some(annotation) = &param.annotation {
```

PR #1517, opened 2023-09-07
New syntax for contracts on Solana

LucasSte:
Now that we represent contracts by their program id on Solana, we decided to elaborate a new syntax to handle them. Contracts cannot be a type in Solidity, consequently, they cannot be function arguments, function returns nor variables.


docs/language/contracts.rst
```diff
@@ -31,10 +31,22 @@ Instantiation using new
 _______________________
 
 Contracts can be created using the ``new`` keyword. The contract that is being created might have
-constructor arguments, which need to be provided.
+constructor arguments, which need to be provided. On Solana, contracts cannot exist as a variable, so it is not
+possible to instantiate a contract with the new keyword. Instead, a function ``new`` is made available to call its
```
seanyoung, line 35: "On Solana, contracts cannot exist as a variable, so it is not possible to instantiate a contract with the new keyword. " That does not follow.

docs/language/contracts.rst
```diff
@@ -87,15 +99,16 @@ can use. gas is a ``uint64``.
...
-Instantiating a contract on Solana
-__________________________________
+Solana constructors
+___________________
 
-On Solana, the contract being created must have the ``@program_id()`` annotation that specifies the program account to
-which the contract code has been deployed. This account holds only the contract's executable binary.
-When calling a constructor only once from an external function, no call arguments are needed. The data account
-necessary to initialize the contract should be present in the IDL and is identified as ``contractName_dataAccount``.
+Solidiy contracts are coupled to a data account, which stores the contract's state variables on the blockchain.
```
seanyoung, line 108: Solidity

docs/language/contracts.rst
```diff
@@ -125,6 +138,25 @@ The sequence of the accounts in the ``AccountMeta`` array matters and must follo
 :ref:`IDL ordering <account_management>`.
 
 
+.. _solana_contract_call:
+
+Calling a contract on Solana
+____________________________
+
+On Solana, contracts are deployed to a program account, which holds only the contract's executable binary. For this
```
seanyoung, line 146: "On Solana, contracts are deployed to a program account, which holds only the contract's executable binary. "

This is not clear. Where does the data go then?

docs/language/contracts.rst
```diff
@@ -125,6 +138,25 @@ The sequence of the accounts in the ``AccountMeta`` array matters and must follo
...
+reason, a Solidity contract on Solana is represented by its program account address. When making an external call,
```
seanyoung, line 147: Remove "for this reason"

docs/language/contracts.rst
```diff
@@ -125,6 +138,25 @@ The sequence of the accounts in the ``AccountMeta`` array matters and must follo
...
+the contract definition serves only as a interface to provide the functions' signatures. As the program account address,
```
seanyoung, line 148: "When making an external call, the contract definition serves only as a interface to provide the functions' signatures"

I have no idea what this means.

docs/language/contracts.rst
```diff
@@ -125,6 +138,25 @@ The sequence of the accounts in the ``AccountMeta`` array matters and must follo
...
+also known as program id, is often determined beforehand, it can be a compile-time constant when declared above a
```
seanyoung, line 149: "determined beforehand" needs explanation

src/sema/expression/constructor.rs
```diff
@@ -577,29 +603,46 @@ pub(super) fn deprecated_constructor_arguments(
...
-                "in order to instantiate contract '{}', a @program_id is required on contract '{}'",
+                "cannot construct '{}' of type '{}'",
                 ns.contracts[constructor_contract_no].name,
-                ns.contracts[constructor_contract_no].name
+                ns.contracts[constructor_contract_no].ty
             ),
         ));
     }
 
+    if let Some(context_contract) = context.contract_no {
```
seanyoung, line 625: Not sure why this check is needed, and why it is not needed when called from a freestanding function.

seanyoung (changes requested):

docs/language/contracts.rst
LucasSte: The explanation about the data account is on line 108. I did not think it was necessary to repeat it here, because data account are not necessary for calling a contract (from now on, they are not necessary if we do not modify any contract variable, regardless of the function mutability). Perhaps we should add a section explaining how contracts are laid out on Solana, so we do not need to repeat the inner principle everywhere. What do you think?

seanyoung: This whole paragraph can be written better. The heading is "Calling a contract on Solana". First explain how to make a call on Solana, and then possibly give some background information.

https://solang.readthedocs.io/en/v0.3.2/language/types.html#contract-types needs updating to reflect that contract types are not allowed.

LucasSte: > https://solang.readthedocs.io/en/v0.3.2/language/types.html#contract-types needs updating to reflect that contract types are not allowed.

Good point. I had forgotten about that.

src/sema/expression/constructor.rs
LucasSte: Good point. In fact, we do not allow constructors in standalone functions.

Having said that, I did not cover such a case in my syntax refactoring.

src/sema/namespace.rs
```diff
@@ -25,6 +25,14 @@ use solang_parser::{
 };
 use std::collections::HashMap;
 
+/// This struct provides context information for the `resolve_type` function.
```
xermicus, line 28:
```suggestion
/// Provides context information for the `resolve_type` function.
```

It's an enum, but saying this is not necessary anyways

tests/codegen_testcases/solidity/borsh_decoding_simple_types.sol
```diff
@@ -33,17 +30,17 @@ contract Testing {
...
 
         // CHECK: block1: # inbounds
-	    // CHECK: ty:uint8 %temp.68 = (builtin ReadFromBuffer ((arg #0), uint32 0))
-	    // CHECK: ty:uint16 %temp.69 = (builtin ReadFromBuffer ((arg #0), uint32 1))
-	    // CHECK: ty:uint32 %temp.70 = (builtin ReadFromBuffer ((arg #0), uint32 3))
-	    // CHECK: ty:uint64 %temp.71 = (builtin ReadFromBuffer ((arg #0), uint32 7))
-	    // CHECK: ty:uint128 %temp.72 = (builtin ReadFromBuffer ((arg #0), uint32 15))
-	    // CHECK: ty:uint256 %temp.73 = (builtin ReadFromBuffer ((arg #0), uint32 31))
-	    // CHECK: branchcond (unsigned less uint32 63 < %temp.67), block3, block4
+	    // CHECK: ty:uint8 %temp.64 = (builtin ReadFromBuffer ((arg #0), uint32 0))
```
xermicus, line 37: We should think about how we can improve our codegen tests. They are hard to reason about and too brittle. Often, a single change in one place requires updating hundreds of unrelated lines. Sometimes even completely unrelated tests fail after a chage in codegen.

xermicus (approved):
LGTM code-wise and on a high level (I'm not too familiar with the Solana runtime)

src/sema/expression/constructor.rs
```diff
@@ -577,29 +602,51 @@ pub(super) fn deprecated_constructor_arguments(
...
+        {
+            ns.contracts[context_contract]
+                .creates
+                .push(constructor_contract_no);
+        }
+    } else {
+        diagnostics.push(Diagnostic::error(
+            *loc,
+            "constructors not allowed in standalone functions".to_string(),
+        ));
```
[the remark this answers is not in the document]
LucasSte: > As you say, we don't allow them though (in polkadot or solana)

Should I allow constructors in standalone functions for Solana? I think I can do this in this PR, as I do not expect many changes. My original concern was to maintain coherence with what we already have.

tests/solana_tests/using.rs
```diff
@@ -2,98 +2,6 @@
...
-        .function("test")
-        .accounts(vec![
-            ("I_programId", program_id),
-            ("systemProgram", [0; 32]),
-        ])
-        .call();
-
-    assert_eq!(runtime.logs, "X libX contractx:2");
-}
-
```
seanyoung, line 96: I think you've deleted the only runtime test for `using` on Solana. It should be re-written

PR #1519, opened 2023-09-12
Verify magic number in fallback

LucasSte:
Presently, we do not check the magic number when we enter a fallback function. If it modifies or read the data account, a malefactor can forge an account so that the function would read or write malicious data. This PR fixes such an issue.

### DIFF
--- a/src/codegen/dispatch/solana.rs
+++ b/src/codegen/dispatch/solana.rs
@@ -222,7 +222,19 @@ pub(crate) fn function_dispatch(
     }
 
     match fallback {
-        Some((cfg_no, _)) => {
+        Some((cfg_no, fallback_cfg)) => {
+            let ASTFunction::SolidityFunction(ast_func_no) = fallback_cfg.function_no else {
+                unreachable!("fallback must be a Solidity function");
+            };
+
+            if ns.functions[ast_func_no]
+                .solana_accounts
+                .borrow()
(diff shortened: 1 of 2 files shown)


src/codegen/dispatch/solana.rs
```diff
@@ -222,7 +222,21 @@ pub(crate) fn function_dispatch(
     }
 
     match fallback {
-        Some((cfg_no, _)) => {
+        Some((cfg_no, fallback_cfg)) => {
+            let ast_func_no = if let ASTFunction::SolidityFunction(no) = fallback_cfg.function_no {
+                no
+            } else {
+                unreachable!("fallback must be a Solidity function");
+            };
```
xermicus, line 230: Can be simplified with a `let-else`

```suggestion
            let ASTFunction::SolidityFunction(ast_func_no) = fallback_cfg.function_no else {
                unreachable!("fallback must be a Solidity function");
            };
```

xermicus (approved):
Solidity also has `receive`, is this already done properly there?

LucasSte:
> Solidity also has `receive`, is this already done properly there?

`receive` are not permitted on Solana: https://solang.readthedocs.io/en/v0.3.2/targets/solana.html#receive-function

seanyoung (approved):

Issue #1520, opened 2023-09-13
Polkadot abi generation overflows its stack


LucasSte:
When compiling this contract, the polkadot ABI generation overflows its stack.
```solidity
contract hatchling {
    struct A {
        A [][2][1] b;
    }

    A private n1;
    constructor() {}

    function foo(uint a, uint b) public {

    }
}
```

PR #1522, opened 2023-09-13
Fix recursive struct issue

LucasSte:
Structs that contains themselves in a dynamic array are not recursive and do not have infinite size, as in the example below:
```solidity
struct C {
    C [2][3][][2] m;
}
```

### DIFF
--- a/src/sema/types.rs
+++ b/src/sema/types.rs
@@ -20,9 +20,11 @@ use num_traits::{One, Zero};
 use petgraph::algo::{all_simple_paths, tarjan_scc};
 use petgraph::stable_graph::IndexType;
 use petgraph::Directed;
+use phf::{phf_set, Set};
 use solang_parser::diagnostics::Note;
 use solang_parser::{doccomment::DocComment, pt, pt::CodeLocation};
 use std::collections::HashSet;
+use std::ops::MulAssign;
 use std::{fmt::Write, ops::Mul};
 
 type Graph = petgraph::Graph<(), usize, Directed, usize>;
@@ -288,7 +290,8 @@ fn check_infinite_struct_size(graph: &Graph, nodes: Vec<usize>, ns: &mut Namespa
         let mut infinite_edge = false;
         for edge in graph.edges_connecting(a.into(), b.into()) {
             match &ns.structs[a].fields[*edge.weight()].ty {
-                Type::Array(_, dims) if dims.first() != Some(&ArrayLength::Dynamic) => {}
+                Type::Array(_, dims) if dims.contains(&ArrayLength::Dynamic) => continue,
+                Type::Array(_, _) => {}
                 Type::Struct(StructType::UserDefined(_)) => {}
                 _ => continue,
             }
@@ -367,6 +370,7 @@ fn find_struct_recursion(ns: &mut Namespace) {
             check_recursive_struct_field(n.index(), &graph, ns);
         }
     }
+
     for n in 0..ns.structs.len() {
         let mut notes = vec![];
         for field in ns.structs[n].fields.iter().filter(|f| f.infinite_size) {
@@ -1579,7 +1583,11 @@ impl Type {
     /// which is not accounted for.
     pub fn solana_storage_size(&self, ns: &Namespace) -> BigInt {
         match self {
-            Type::Array(_, dims) if dims.first() == Some(&ArrayLength::Dynamic) => 4.into(),
+            Type::Array(_, dims) if dims.contains(&ArrayLength::Dynamic) => {
+                let size = dynamic_array_size(dims);
+                // A pointer is four bytes on Solana
+                size * 4
+            }
             Type::Array(ty, dims) => ty.solana_storage_size(ns).mul(
                 dims.iter()
                     .map(|d| match d {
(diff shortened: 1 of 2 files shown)


LucasSte:
@xermicus I implemented the fix for Polkadot as well, but I could not write tests for it because I've hit a problem in ABI generation (#1520). Can you advise if I should maintain the changes I made in Polkadot or revert them?

src/sema/types.rs
```diff
@@ -1556,7 +1562,12 @@ impl Type {
     /// which is not accounted for.
     pub fn solana_storage_size(&self, ns: &Namespace) -> BigInt {
         match self {
-            Type::Array(_, dims) if dims.first() == Some(&ArrayLength::Dynamic) => 4.into(),
+            Type::Array(_, dims) if dims.iter().any(|dim| *dim == ArrayLength::Dynamic) => {
```
seanyoung, line 1565: You can write `.contains(&ArrayLength::Dynamic)`

src/sema/types.rs
```diff
@@ -1684,8 +1695,12 @@ impl Type {
                 Type::Value => BigInt::from(ns.value_length),
                 Type::Uint(n) | Type::Int(n) => BigInt::from(n / 8),
                 Type::Rational => unreachable!(),
-                Type::Array(_, dims) if dims.first() == Some(&ArrayLength::Dynamic) => {
-                    BigInt::from(4)
+                Type::Array(_, dims) if dims.iter().any(|dim| *dim == ArrayLength::Dynamic) => {
```
seanyoung, line 1698: same here

src/sema/types.rs
```diff
@@ -2151,10 +2168,28 @@ impl Type {
 
 /// These names cannot be used on Windows, even with an extension.
 /// shamelessly stolen from cargo
+static WINDOWS_KEYWORDS: Set<&'static str> = phf_set! {
+     "con", "prn", "aux", "nul", "com1", "com2", "com3", "com4", "com5", "com6", "com7", "com8",
+         "com9", "lpt1", "lpt2", "lpt3", "lpt4", "lpt5", "lpt6", "lpt7", "lpt8", "lpt9",
+};
```
seanyoung, line 2174: They're not keywords, they're device names

seanyoung: reserved file names?
how about `WINDOWS_RESERVED`

src/sema/types.rs
```diff
@@ -1556,7 +1562,12 @@ impl Type {
     /// which is not accounted for.
     pub fn solana_storage_size(&self, ns: &Namespace) -> BigInt {
         match self {
-            Type::Array(_, dims) if dims.first() == Some(&ArrayLength::Dynamic) => 4.into(),
+            Type::Array(_, dims) if dims.contains(&ArrayLength::Dynamic) => {
+                let mut size = dynamic_array_size(dims);
+                // A pointer is four bytes on Solana
+                size.mul_assign(4);
+                size
```
seanyoung, line 1569: size * 4?

seanyoung (approved):
LGTM!

src/sema/types.rs
```diff
@@ -1684,8 +1692,12 @@ impl Type {
...
                 Type::Uint(n) | Type::Int(n) => BigInt::from(n / 8),
                 Type::Rational => unreachable!(),
-                Type::Array(_, dims) if dims.first() == Some(&ArrayLength::Dynamic) => {
-                    BigInt::from(4)
+                Type::Array(_, dims) if dims.contains(&ArrayLength::Dynamic) => {
+                    let mut size = dynamic_array_size(dims);
+
+                    // A pointer is four bytes on Solana
+                    size.mul_assign(4);
+                    size
```
seanyoung, line 1700: size + 4?

src/sema/types.rs
```diff
@@ -1556,7 +1560,11 @@ impl Type {
     /// which is not accounted for.
     pub fn solana_storage_size(&self, ns: &Namespace) -> BigInt {
         match self {
-            Type::Array(_, dims) if dims.first() == Some(&ArrayLength::Dynamic) => 4.into(),
+            Type::Array(_, dims) if dims.contains(&ArrayLength::Dynamic) => {
+                let size = dynamic_array_size(dims);
+                // A pointer is four bytes on Solana
+                size * 4
```
xermicus, line 1566: Wouldn't this be the size + 4 times the number of dynamic arrays?

LucasSte: As I understand, an array `uint8 [2][][3]` means a fixed-length array of pointers that contains three elements. Each pointer points to `uint8[2]`. The size will only be `3*pointer_size`. Now if the type was `uint8 [2][][][3]`, we'll still have a fixed-array of pointers that contains three elements. Each pointer points to a `uint8[2][]`. The size will continue to be `3*pointer_size`.

The function calculates the minimum size we need for storage. As soon as we allocate elements, more space will be necessary.

src/sema/types.rs
LucasSte: > size + 4?

`size * 4`, right? The operation is `size.mul_assign(4)`.

src/sema/types.rs
xermicus: I see, thanks.

What you explained here would look good in a comment to the source code (or even in the doc comment of this function)

xermicus (commented):
There ar _so_ many issues with struct

src/sema/types.rs
LucasSte: This is explained right above `dynamic_array_size` between lines 2172 and 2176.

xermicus (approved):

Issue #1523, opened 2023-09-14
Error: `type(int256).min` is too large when in comparison


BenTheKush:
**Describe the bug**
The expression `type(int256).min` crashes during parse_and_resolve().

**To Reproduce**
I've added this to [this repo](https://github.com/BenTheKush/solang-op-assign-bug), but I'll repeat here.

Here is the code that triggers the bug:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.0;

library Issue1523 {
    function i1523_fail1(int256 x) internal pure returns (bool) {
        return x <= type(int256).min;
    }

    function i1523_fail2(int256 x) internal pure returns (bool) {
        return x <= type(int256).min + 1;
    }

    function i1523_fail3(int256 x) internal pure returns (bool) {
        // Actual min value inlined
        return x <= -57896044618658097711785492504343953926634992332820282019728792003956564819968;
    }

    function i1523_fail4(int256 x) internal pure returns (bool) {
        // Actual min value + 1
        return x <= -57896044618658097711785492504343953926634992332820282019728792003956564819967;
    }

    function i1523_pass1() internal pure returns (int256) {
        return type(int256).min + 1;
    }

    function i1523_pass2() internal pure returns (int256) {
        return type(int256).min;
    }
}
```

If I run:
```
solang compile --target solana bug.sol
```

error: -57896044618658097711785492504343953926634992332820282019728792003956564819968 is too large
  ┌─ /Users/benku/Playground/issue1523.sol:6:21
  │
6 │         return x <= type(int256).min;
  │                     ^^^^^^^^^^^^^^^^

error: -57896044618658097711785492504343953926634992332820282019728792003956564819968 is too large
   ┌─ /Users/benku/Playground/issue1523.sol:10:21
   │
10 │         return x <= type(int256).min + 1;
   │                     ^^^^^^^^^^^^^^^^

error: -57896044618658097711785492504343953926634992332820282019728792003956564819968 is too large
   ┌─ /Users/benku/Playground/issue1523.sol:14:21
   │
14 │         return x <= -57896044618658097711785492504343953926634992332820282019728792003956564819968;
   │  

Note that the first three functions fail. The fourth function I've added  to the min value literal, and this passes. This sounds like we are trying to create the value `57896044618658097711785492504343953926634992332820282019728792003956564819968` as an `int256` and then apply a negative unary operator to it, but only in certain contexts? I

**Expected behavior**
This compiles w/ solc.

[...]

PR #1527, opened 2023-09-14
Fix type(int256).min in comparisons

seanyoung:
When resolving comparisons, we don't know what the types of the operands are so we resolve with `ResolveTo::Unknown`. In this path, there is an mistake in the bounds check for negative integers.

Fixes https://github.com/hyperledger/solang/issues/1523

### DIFF
--- a/src/sema/expression/integers.rs
+++ b/src/sema/expression/integers.rs
@@ -4,10 +4,9 @@ use crate::sema::ast::{Expression, Namespace, Type};
 use crate::sema::diagnostics::Diagnostics;
 use crate::sema::expression::ResolveTo;
 use num_bigint::{BigInt, Sign};
-use num_traits::{One, Zero};
+use num_traits::Zero;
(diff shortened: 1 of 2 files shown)


src/sema/expression/integers.rs
```diff
@@ -208,6 +207,14 @@ pub fn bigint_to_expression(
         }
     }
 
+    // BigInt bits() returns the bits used without the sign. Negative value is allowed to be one
+    // larger than positive value, e.g int8 has inclusive range -128 to 127.
+    let bits = if n.sign() == Sign::Minus {
+        n.add(BigInt::one()).bits()
```
LucasSte, line 213: Nitpick: Why not `n.bits() + 1`?

LucasSte (approved):

src/sema/expression/integers.rs
seanyoung: Good point, fixed

xermicus (approved):

Issue #1525, opened 2023-09-14
Error with Library's `using {f} for ...`


BenTheKush:
**Describe the bug**
There is a compilation error/parse error for code that compiles for `solc`

**To Reproduce**
I've added a minimized version of this bug to [this repository](https://github.com/BenTheKush/solang-op-assign-bug). I'm reproducing it here for convenience.

`issue1525.sol`:
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.0;

library Lib {
    using {f} for int256;

    function f(int256 a) internal pure returns (int256) {
        return a;
    }
}
```
Running
```
$ solang compile --target solana issue1525.sol
error: 'f' not found
  ┌─ /Users/benku/Playground/issue1525.sol:5:12
  │
5 │     using {f} for int256;
  │            ^

$ solc issue1525.sol
Compiler run successful, no output requested.
```

Ran with solang 0.3.0

PR #1528, opened 2023-09-14
Allow using {func} for type to use library functions

seanyoung:
Fixes https://github.com/hyperledger/solang/issues/1525

### DIFF
--- a/src/sema/namespace.rs
+++ b/src/sema/namespace.rs
@@ -364,9 +364,10 @@ impl Namespace {
     }
 
     /// Resolve a free function name with namespace
-    pub(super) fn resolve_free_function_with_namespace(
+    pub(super) fn resolve_function_with_namespace(
         &mut self,
         file_no: usize,
+        contract_no: Option<usize>,
         name: &pt::IdentifierPath,
         diagnostics: &mut Diagnostics,
     ) -> Result<Vec<(pt::Loc, usize)>, ()> {
@@ -376,7 +377,7 @@ impl Namespace {
             .map(|(id, namespace)| (id, namespace.iter().collect()))
             .unwrap();
 
-        let s = self.resolve_namespace(namespace, file_no, None, id, diagnostics)?;
+        let s = self.resolve_namespace(namespace, file_no, contract_no, id, diagnostics)?;
 
         if let Some(Symbol::Function(list)) = s {
             Ok(list.clone())
@@ -1335,6 +1336,7 @@ impl Namespace {
                         ));
                         return Err(());
                     };
+                    namespace.clear();
                     Some(*n)
                 }
                 Some(Symbol::Function(_)) => {
@@ -1388,6 +1390,10 @@ impl Namespace {
                 }
                 Some(Symbol::Import(..)) => unreachable!(),
             };
+        }
+
+        if !namespace.is_empty() {
+            return Ok(None);
         }
 
         let mut s = self
--- a/src/sema/namespace.rs
+++ b/src/sema/namespace.rs
@@ -377,12 +377,12 @@ impl Namespace {
             .map(|(id, namespace)| (id, namespace.iter().collect()))
             .unwrap();
 
-        let s = self.resolve_namespace(namespace, file_no, contract_no, id, diagnostics)?;
+        let symbol = self.resolve_namespace(namespace, file_no, contract_no, id, diagnostics)?;
 
-        if let Some(Symbol::Function(list)) = s {
+        if let Some(Symbol::Function(list)) = symbol {
             Ok(list.clone())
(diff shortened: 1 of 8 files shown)


tests/contract_testcases/solana/using_functions.sol
```diff
@@ -0,0 +1,31 @@
...
+library Lib {
+	function baz(int256 a, bool b) internal pure returns (int256) {
+		if (b) {
+			return 1;
+		} else {
+			return a;
+		}
+	}
+	using {L.bar, baz} for int256;
+}
```
LucasSte, line 22: I think it is good to write a call to these functions so that we check if they are properly wired after sema. Doing something like `uint256 a=1; a.baz(true); a.L.bar;`

seanyoung: You are right. Actually I was thinking about this last night and then forgot about it again.

tests/solana_tests/using.rs
```diff
@@ -38,6 +38,40 @@ fn using_for_contracts() {
...
+            function f(int) external;
+        }
+
+        library L {
+            function F(I i, bool b, int n) public {
+                if (b) {
+                    print("Hello");
+                }
+            }
+        }
```
LucasSte, line 53: Would you mind not creating a function that receives an interface?
This test will stop working because of #1517.

tests/solana_tests/using.rs
```diff
@@ -92,6 +126,67 @@ fn using_for_contracts() {
...
+                i.f1(2);
+            }
+
+            function X(I i) external {
+                print("X lib");
+            }
+
+            function Y(I i) external {
+                i.X();
+            }
```
LucasSte, line 150: Would you mind not passing the interface as a parameter?
#1517 (awaiting your review) will make this stop working.

tests/solana_tests/using.rs
LucasSte: Please, do not create a giant monolithic test. It is better to create a new function whose name explains exactly what you are testing.

tests/solana_tests/using.rs
LucasSte: The function `using_for_contracts` is completely removed in that PR because it is not possible to test libraries for interfaces.

seanyoung: That's not good

LucasSte: > That's not good

Well, if an interface cannot be a type, then there cannot be a library for it.

seanyoung: Let's deal with #1517  before we deal with this

tests/polkadot_tests/libraries.rs
```diff
@@ -74,7 +74,7 @@ fn using() {
     let mut runtime = build_solidity(
         r##"
         contract test {
-            using ints for uint32;
```
xermicus, line 77: This potentially reduces test coverage, or is it tested in another place anyways?

src/sema/namespace.rs
```diff
@@ -376,7 +377,7 @@ impl Namespace {
             .map(|(id, namespace)| (id, namespace.iter().collect()))
             .unwrap();
 
-        let s = self.resolve_namespace(namespace, file_no, None, id, diagnostics)?;
+        let s = self.resolve_namespace(namespace, file_no, contract_no, id, diagnostics)?;
```
xermicus, line 380: Not to be pedantic, but just `s` is a bad variable name

LucasSte: > Not to be pedantic, but just `s` is a bad variable name

I agree

src/sema/using.rs
```diff
@@ -251,7 +264,22 @@ pub(crate) fn using_decl(
...
+
+                            if dummy
+                                .cast(
+                                    &loc,
+                                    &func.params[0].ty,
+                                    true,
+                                    ns,
+                                    &mut Diagnostics::default(),
+                                )
+                                .is_err()
```
LucasSte, line 281: Is there any test in which we are casting the first argument for the library call?

LucasSte (approved):

tests/polkadot_tests/libraries.rs
seanyoung: This is already tested above (line 50 to 70)

src/sema/using.rs
seanyoung: The polkadot test does this

src/sema/namespace.rs
seanyoung: Renamed to symbol

xermicus (approved):
LGTM

Issue #1524, opened 2023-09-14
NatSpec annotations not supported


BenTheKush:
**Describe the bug**
The [Solidity docs describe NatSpec](https://docs.soliditylang.org/en/v0.8.17/natspec-format.html) which allows annotations of the form `"@custom:<name>"`. 

I've documented this bug in [this repository](https://github.com/BenTheKush/solang-op-assign-bug), but here are the details:

`issue1524.sol`:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.0;

/** 
 * The following example is taken from solidity docs:
 * https://docs.soliditylang.org/en/v0.8.17/natspec-format.html#documentation-example
 */

/// @title A simulator for trees
/// @author Larry A. Gardner
/// @notice You can use this contract for only the most basic simulation
/// @dev All function calls are currently implemented without side effects
/// @custom:experimental This is an experimental contract.
library Issue1524 {
}
```

Running solang fails

```bash
$  solang compile --target solana issue1524.sol 
error: tag '@custom:experimental' is not valid for contract
   ┌─ /Users/benku/Playground/issue1524.sol:13:6
   │
13 │ /// @custom:experimental This is an experimental contract.
   │      ^^^^^^^^^^^^^^^^^^^^
```

solang version v0.3.0
revision: "70af554c42748009e414e6263dd4607fb380e5dc"

PR #1529, opened 2023-09-14
Add support for custom tags

seanyoung:
Fixes https://github.com/hyperledger/solang/issues/1524

### DIFF
--- a/src/sema/tags.rs
+++ b/src/sema/tags.rs
@@ -26,7 +26,7 @@ pub fn resolve_tags(
 
         match c.tag.as_str() {
             "notice" | "author" | "title" | "dev" => {
-                fold_tag(loc, &mut res, c);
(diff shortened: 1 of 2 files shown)


src/sema/tags.rs
```diff
@@ -167,17 +159,43 @@ pub fn resolve_tags(
...
+                    ));
+                }
             }
         }
     }
 
     res
 }
 
+fn fold_tag(loc: pt::Loc, res: &mut Vec<Tag>, doc_comment: &DocCommentTag) {
```
LucasSte, line 184: How about adding a doc comment here?

LucasSte (approved):

src/sema/tags.rs
seanyoung: Good point, added

xermicus (approved):

Issue #1526, opened 2023-09-14
`assembly("memory-safe")` causes failure during parse_and_resolve


BenTheKush:
**Describe the bug**
When I run `parse_and_resolve` on a file with `assembly("memory-safe")` in it I get the following error:

```
error: flag 'memory-safe' not supported
  ┌─ /Users/benku/Playground/1526.sol:6:19
  │
6 │         assembly ("memory-safe") {
  │                   ^^^^^^^^^^^^^
```

It's understandable to not support a feature, but I think that this should not affect the AST.

PR #1530, opened 2023-09-14
Permit memory-safe flag on assembly

seanyoung:
Fixes https://github.com/hyperledger/solang/issues/1526

### DIFF
--- a/src/sema/statements.rs
+++ b/src/sema/statements.rs
@@ -888,17 +888,38 @@ fn statement(
                 return Err(());
             }
 
+            let mut memory_safe = None;
+
             if let Some(flags) = flags {
                 for flag in flags {
-                    ns.diagnostics.push(Diagnostic::error(
-                        flag.loc,
-                        format!("flag '{}' not supported", flag.string),
-                    ));
+                    if flag.string == "memory-safe" && ns.target == Target::EVM {
+                        if let Some(prev) = &memory_safe {
+                            ns.diagnostics.push(Diagnostic::error_with_note(
+                                flag.loc,
+                                format!("flag '{}' already specified", flag.string),
+                                *prev,
+                                "previous location".into(),
+                            ));
+                        } else {
+                            memory_safe = Some(flag.loc);
+                        }
+                    } else {
+                        ns.diagnostics.push(Diagnostic::error(
+                            flag.loc,
+                            format!("flag '{}' not supported", flag.string),
+                        ));
+                    }
                 }
             }
 
-            let resolved_asm =
-                resolve_inline_assembly(loc, &block.statements, context, symtable, ns);
+            let resolved_asm = resolve_inline_assembly(
+                loc,
+                memory_safe.is_some(),
+                &block.statements,
+                context,
+                symtable,
+                ns,
+            );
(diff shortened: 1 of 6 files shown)


LucasSte (approved):

src/sema/statements.rs
```diff
@@ -888,17 +888,38 @@ fn statement(
...
 
+            let mut memory_safe = None;
+
             if let Some(flags) = flags {
                 for flag in flags {
-                    ns.diagnostics.push(Diagnostic::error(
-                        flag.loc,
-                        format!("flag '{}' not supported", flag.string),
-                    ));
+                    if flag.string == "memory-safe" && ns.target == Target::EVM {
```
xermicus, line 895: Why is there no enum and the comparison is on `flag.string`?

src/sema/yul/ast.rs
```diff
@@ -12,6 +12,7 @@ use std::sync::Arc;
 #[derive(Debug, Clone)]
 pub struct InlineAssembly {
     pub loc: pt::Loc,
+    pub memory_safe: bool,
```
xermicus, line 15: I'm a bit confused. We add this `memory_safe` field here. But I never see this being actually used. So why bother adding it?

Alternatively we should have a comment explaining what's going on here

src/sema/statements.rs
seanyoung: assembly statements can have flags, like so:
```solidity
function foo() { 
     assembly ("foo", "bar", "baz") {
         // ..
     }
}
```
according to the solidity parser/grammer. Our parser just parses it into a list of flags (string literals).

However there is only one flag defined which is `"memory-safe"`. The change is to parse this for EVM and add it to the AST, we don't allow the flag for any other target and we don't use it on EVM either - as our EVM target is incomplete.

src/sema/yul/ast.rs
seanyoung: I've added a comment to this struct.

Do you think this clear enough?

src/sema/statements.rs
xermicus: > Our parser just parses it into a list of flags (string literals).

So in sema, we can parse this list of string literals parsed into a `HashSet` or`BTreeSet` of an `AssemblyFlag` enum right? This will make using it in later compile stages in the future easier.

src/sema/yul/ast.rs
xermicus: Whlie that certainly helps, I meant that the comment should mention as to why this field is present but not (yet?) used any further

xermicus (approved):
I mean functional wise looks good, I left some suggestions for improvements in the comments. I'm heading off now, I'll leave it up to you

src/sema/statements.rs
seanyoung: Right now there is only one valid flag: `memory-safe`. I don't think it's worth creating a `HashSet` for that yet. We can implement that should more flags be added.

src/sema/yul/ast.rs
seanyoung: Good point. I've explained it a bit more.

PR #1531, opened 2023-09-14
Verify account space specified in annotation

LucasSte:
On Solana, we can specify the space necessary for a contract's data account in the constructor with an annotation. We do not verify, however, if the declared space is enough to hold all the contract variables. As a consequence, we could call the constructor without any issue using a very small space value and hit other runtime errors later.

PS: I found it unpleasant to keep iterating over the constructor annotations to find the one we were looking for, so I replaced the vector by a struct to hold the permitted annotations.

### DIFF
--- a/src/codegen/mod.rs
+++ b/src/codegen/mod.rs
@@ -42,12 +42,15 @@ use std::cmp::Ordering;
 use crate::codegen::cfg::ASTFunction;
 use crate::codegen::solana_accounts::account_management::manage_contract_accounts;
 use crate::codegen::yul::generate_yul_function_cfg;
+use crate::sema::diagnostics::Diagnostics;
+use crate::sema::eval::eval_const_number;
 use crate::sema::Recurse;
 #[cfg(feature = "wasm_opt")]
 use contract_build::OptimizationPasses;
 use num_bigint::{BigInt, Sign};
 use num_rational::BigRational;
 use num_traits::{FromPrimitive, Zero};
+use solang_parser::diagnostics::Diagnostic;
 use solang_parser::{pt, pt::CodeLocation};
 
 // The sizeof(struct account_data_header)
@@ -55,6 +58,10 @@ pub const SOLANA_FIRST_OFFSET: u64 = 16;
 
 /// Name of the storage initializer function
 pub const STORAGE_INITIALIZER: &str = "storage_initializer";
+
+/// Maximum permitted size of account data (10 MiB).
+/// https://github.com/solana-labs/solana/blob/08aba38d3507c8cb66f85074d8f1249d43e64a75/sdk/program/src/system_instruction.rs#L85
+pub const MAXIMUM_ACCOUNT_SIZE: u64 = 10 * 1024 * 1024;
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub enum OptimizationLevel {
@@ -326,6 +333,27 @@ fn layout(contract_no: usize, ns: &mut Namespace) {
                 });
 
                 slot += ty.storage_slots(ns);
+            }
+        }
+    }
+
+    let constructors = ns.contracts[contract_no].constructors(ns);
+    if !constructors.is_empty() {
+        if let Some((_, exp)) = &ns.functions[constructors[0]].annotations.space {
+            // This code path is only reachable on Solana
+            assert_eq!(ns.target, Target::Solana);
+            if let Ok((_, value)) = eval_const_number(exp, ns, &mut Diagnostics::default()) {
+                if slot > value {
+                    ns.diagnostics.push(Diagnostic::error(
+                        exp.loc(),
+                        format!("contract requires at least {} bytes of space", slot),
+                    ));
+                } else if value > BigInt::from(MAXIMUM_ACCOUNT_SIZE) {
+                    ns.diagnostics.push(Diagnostic::error(
+                        exp.loc(),
+                        "Solana's runtime does not permit accounts larger than 10 MB".to_string(),
+                    ));
+                }
             }
         }
     }
--- a/src/codegen/solana_deploy.rs
+++ b/src/codegen/solana_deploy.rs
@@ -4,13 +4,15 @@ use super::{
     cfg::ReturnCode, expression, Builtin, ControlFlowGraph, Expression, Instr, Options, Type,
     Vartable,
 };
+use crate::codegen::revert::string_to_expr;
 use crate::codegen::solana_accounts::account_management::{
     account_meta_literal, retrieve_key_from_account_info,
 };
 use crate::sema::ast::{
-    self, ArrayLength, CallTy, ConstructorAnnotation, Function, FunctionAttributes, Namespace,
-    StructType,
+    self, ArrayLength, CallTy, Function, FunctionAttributes, Namespace, StructType,
 };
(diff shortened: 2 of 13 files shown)


src/codegen/mod.rs
```diff
@@ -330,6 +331,25 @@ fn layout(contract_no: usize, ns: &mut Namespace) {
         }
     }
 
+    // FIXME: PR #1517 introduces the 'constructors' function that simplifies this
```
seanyoung, line 334: might as well wait for #1517 to be merged before this one

src/sema/ast.rs
```diff
@@ -347,7 +347,7 @@ pub struct Function {
     /// For overloaded functions this is the mangled (unique) name.
     pub mangled_name: String,
     /// Solana constructors may have seeds specified using @seed tags
-    pub annotations: Vec<ConstructorAnnotation>,
+    pub new_annotations: ConstructorAnnotations,
```
seanyoung, line 350: `new_` ?

src/sema/ast.rs
```diff
@@ -366,26 +366,17 @@ pub struct SolanaAccount {
...
 
-#[derive(Debug)]
-pub enum ConstructorAnnotation {
-    Seed(Expression),
-    Payer(pt::Loc, String),
-    Space(Expression),
-    Bump(Expression),
+#[derive(Debug, Default)]
+pub struct ConstructorAnnotations {
+    pub seeds: Vec<Expression>,
```
seanyoung, line 371: no location information for seeds?

tests/contract_testcases/solana/annotations/bad_accounts.sol
```diff
@@ -75,7 +75,7 @@ contract OldAnnotationSyntax {
 // error: 39:12-29: 'SysvarInstruction' is a reserved account name
 // error: 47:12-18: account 'solang' already defined
 // 	note 46:5-19: previous definition
-// error: 58:11-18: '@seed' annotation on top of a constructor only accepts literals
-// error: 59:12-20: '@space' annotation on top of a constructor only accepts literals
-// error: 60:11-18: '@bump' annotation on top of a constructor only accepts literals
+// error: 58:11-18: expression not allowed in constant number expression
+// error: 59:12-20: expression not allowed in constant number expression
+// error: 60:11-18: expression not allowed in constant number expression
```
seanyoung, line 80: this is worse

src/codegen/solana_deploy.rs
```diff
@@ -314,13 +302,57 @@ pub(super) fn solana_deploy(
...
-            .annotations
-            .iter()
-            .find(|tag| matches!(tag, ConstructorAnnotation::Space(..)))
-        {
-            let space_var = vartab.temp_name("space", &Type::Uint(64));
+        let (space, lamports) = if let Some((_, space_expr)) = &func.new_annotations.space {
             let expr = expression(space_expr, cfg, contract_no, None, ns, vartab, opt);
+            // If the space has not been resolved into a literal in sema,
+            // we must verify if we are allocating enough space during runtime.
+            if !matches!(expr, Expression::NumberLiteral { .. }) {
```
seanyoung, line 309: so what about when it's a constant expression?

seanyoung (changes requested):

src/sema/ast.rs
LucasSte: I use the location information to detect duplicate annotations. This is the case for space and bump. There can be multiple seeds, on the other hand, so I need not track their locations. Shall I add them?

src/sema/ast.rs
LucasSte: That is a remainder from my development. I will remove it. 🫣

tests/contract_testcases/solana/annotations/bad_accounts.sol
LucasSte: That is the error message that stems from `eval_constant_expression`, when it does not encounter a constant expression. I thought the code would be simplified if I used the same error we already have, but I can revert it back to `annotation only supports literals`.

src/codegen/solana_deploy.rs
LucasSte: If look carefully at `src/sema/function_annotation.rs`, between there can be only four options when resolving an annotation:
1. It is a repeated one, in which case we error out.
2. It is a literal, in which case we save it.
3. It is resolved by `eval_const_number`, which returns a `BigInt`. We save this big it in a number literal.
4. `eval_const_number` errors out, in which case we discard the annotation and never enter codegen.

By the paths no. 2 and no. 3, I am faithfully assuming we are not going to have constant expressions in codegen.

src/sema/ast.rs
seanyoung: I think this is useful information to have in the ast.

src/codegen/solana_deploy.rs
seanyoung: The ast should be a representation of the source code, and not have constant expressions folded.

LucasSte: Fixed

src/sema/ast.rs
LucasSte: Fixed

tests/contract_testcases/solana/annotations/invalid_space.sol
```diff
@@ -0,0 +1,14 @@
...
+    @payer(acc)
+    @space(5<<64)
+    constructor(address c, address d) {
+        a = c;
+        b = d;
+    }
+}
+
+// ---- Expect: diagnostics ----
+// error: codegen: value 92233720368547758208 does not fit into type uint64.
```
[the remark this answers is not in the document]
LucasSte: That is a good idea. I'll add such a check.

src/codegen/mod.rs
```diff
@@ -330,6 +333,27 @@ fn layout(contract_no: usize, ns: &mut Namespace) {
...
+        if let Some((_, exp)) = &ns.functions[constructors[0]].annotations.space {
+            // This code path is only reachable on Solana
+            assert_eq!(ns.target, Target::Solana);
+            if let Ok((_, value)) = eval_const_number(exp, ns, &mut Diagnostics::default()) {
+                if slot > value {
+                    ns.diagnostics.push(Diagnostic::error(
+                        exp.loc(),
+                        format!("contract requires at least {} bytes of space", slot),
+                    ));
+                } else if value > BigInt::from(10_485_760u64) {
```
seanyoung, line 347: This deserves a comment with a link to the solana repo where the constant is defined

src/codegen/mod.rs
```diff
@@ -330,6 +333,27 @@ fn layout(contract_no: usize, ns: &mut Namespace) {
...
+                    ns.diagnostics.push(Diagnostic::error(
+                        exp.loc(),
+                        "an account can be no larger than 10 MB".to_string(),
```
seanyoung, line 350: this should make it clear this is restrictions of the solana runtime, not solang

src/codegen/mod.rs
LucasSte: As soon as we migrate to the new LLVM branch, I can just import Solana's crate and use the constant directly from there.

xermicus (approved):
LGTM

seanyoung (approved):

Issue #1533, opened 2023-09-15
Incorrect annotations should not kill parse/compile


bkushigian:
**Describe the bug**
Having an incorrect annotation bricks parse. In a vacuum I think it makes sense for a compiler to be strict about this sort of thing, but in an ecosystem where the main compiler (solc) is too permissive this means there are tons of libraries that exist that don't follow these conventions. For instance, a client is using the prb-math library which has the [following code](https://github.com/PaulRBerg/prb-math/blob/5959ef59f906d689c2472ed08797872a1cc00644/src/ud60x18/Math.sol#L65):

```solidity
/// @notice Yields the smallest whole number greater than or equal to x.
///
/// @dev This is optimized for fractional value inputs, because for every whole value there are (1e18 - 1) fractional
/// counterparts. See https://en.wikipedia.org/wiki/Floor_and_ceiling_functions.
///
/// Requirements:
/// - x must be less than or equal to `MAX_WHOLE_UD60x18`.
///
/// @param x The UD60x18 number to ceil.
/// @param result The smallest whole number greater than or equal to x, as a UD60x18 number.
/// @custom:smtchecker abstract-function-nondet
function ceil(UD60x18 x) pure returns (UD60x18 result) {
   ...
}
```

Using Solang to parse this gives an error (note: this example is slightly different than the code I linked above because the source I found is a different version than what is cached in node_modules for me, but they are morally equivalent):

```

error: tag '@param' no field 'result'
   ┌─ /Users/benku/Certora/gmx-synthetics_all/node_modules/prb-math/contracts/PRBMathUD60x18.sol:52:16
   │
52 │     /// @param result The least integer greater than or equal to x, as an unsigned 60.18-decimal fixed-point number.
   │                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Now I _think_ you actually caught a bug in prb-math's documentation, which is good! But it unfortunately hard-bricks solang's parser, and it is out of our client's control to fix this bug.

Is it possible to make annotation enforcement optional? Probably best to implement solc's behavior by default and then have a `--strict_checks` flag to enable features that solc should actually do, but I don't know how possible this is?

seanyoung:
[...]

PR #1539, opened 2023-09-16
solc allows @param tag for return values

seanyoung:
@return should be used, warn about this.

https://github.com/hyperledger/solang/issues/1533

### DIFF
--- a/tests/solana_tests/tags.rs
+++ b/tests/solana_tests/tags.rs
@@ -211,6 +211,40 @@ fn functions() {
     assert_eq!(func.tags[2].tag, "inheritdoc");
     assert_eq!(func.tags[2].value, "b");
     assert_eq!(func.tags[2].no, 0);
+
+    let ns = parse_and_resolve(
+        r#"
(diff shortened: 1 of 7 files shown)


tests/solana_tests/tags.rs
```diff
@@ -211,6 +211,40 @@ fn functions() {
...
+
+        contract b {}"#,
+        Target::Solana,
+    );
+
+    assert_eq!(ns.diagnostics.len(), 4);
+
+    assert_eq!(
+        ns.diagnostics.first_error(),
+        "tag '@return' no matching return value 'x'"
```
LucasSte, line 232: This error sentence is a little unclear. How about writing `could not find return value 'x' to match with tag`?

seanyoung: I've reworded it

xermicus (approved):
I like seeing all this PRs bringing the the EVM test failures number down :+1:

LucasSte (approved):

PR #1540, opened 2023-09-17
Implement Go To Declaration

chioni16:
-  gives a list of contract methods that the given contract method overrides.
-  only returns the methods belonging to the immediate parent contracts.
-  handles multiple inheritance.


src/sema/ast.rs
```diff
@@ -768,7 +768,7 @@ pub struct Contract {
     pub fixed_layout_size: BigInt,
     pub functions: Vec<usize>,
     pub all_functions: BTreeMap<usize, usize>,
-    pub virtual_functions: HashMap<String, usize>,
+    pub virtual_functions: HashMap<String, Vec<usize>>,
```
seanyoung, line 771: This needs a very clear comment explaining what this `HashMap<..>` does and what the `Vec<usize>` means

seanyoung (approved):

src/sema/ast.rs
```diff
@@ -768,7 +768,10 @@ pub struct Contract {
     pub fixed_layout_size: BigInt,
     pub functions: Vec<usize>,
     pub all_functions: BTreeMap<usize, usize>,
-    pub virtual_functions: HashMap<String, usize>,
+    // maps the name of virtual functions to a vector of overriden functions.
```
xermicus, line 771: Why is this not a doc comment?

src/bin/languageserver/mod.rs
```diff
@@ -1318,14 +1332,23 @@ impl<'a> Builder<'a> {
...
-    fn build(ns: &ast::Namespace) -> (Vec<FileCache>, Definitions, Types, Implementations) {
+    fn build(
+        ns: &ast::Namespace,
+    ) -> (
+        Vec<FileCache>,
+        Definitions,
+        Types,
+        Declarations,
+        Implementations,
+    ) {
```
xermicus, line 1343: On a quick glance, seems like a weird API. Usually, a `build` function is used on a builder, made available after configuring the builder.

The method doesn't take `self` but it also doesn't return `Self`, instead it returns some other types. Without reading through some example code where this is actually being used, it's very unclear how and when this should be used.

xermicus (approved):

src/sema/ast.rs
chioni16: Changed it into doc comments. Thanks!

src/bin/languageserver/mod.rs
```diff
@@ -1318,14 +1332,23 @@ impl<'a> Builder<'a> {
...
     /// This includes hover messages, locations where code objects are declared and used
-    fn build(ns: &ast::Namespace) -> (Vec<FileCache>, Definitions, Types, Implementations) {
+    fn build(
+        ns: &ast::Namespace,
+    ) -> (
+        Vec<FileCache>,
+        Definitions,
+        Types,
+        Declarations,
+        Implementations,
```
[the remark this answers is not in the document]
chioni16: I am now returning the `GlobalCache` instead of individual fields. Thanks!

src/bin/languageserver/mod.rs
chioni16: The interface has now been changed. 
The `Builder` type now has a `new` method defined which returns a new `Builder`. We don't have any need to configure the `Builder` as of yet, so no provision has been made for that. 
Then the `build` method on `Builder` consumes the `Builder` and returns `(FileCache, GlobalCache)`, which represent the things we have extracted from `Namespace` that are needed by `SolangServer` to service the requests from languageclient later on. The `build` method doesn't return a new `SolangServer` as we need to combine information from the current run of `build` method with information we already had from the previous run of `build` method.

vscode/src/test/suite/extension.test.ts
```diff
@@ -201,15 +208,68 @@ async function testtypedefs(docUri: vscode.Uri) {
...
+async function testdecls(docUri: vscode.Uri) {
+  await activate(docUri);
+
+  const pos0 = new vscode.Position(6, 14);
+  const actualdecl0 = (await vscode.commands.executeCommand(
+    'vscode.executeDeclarationProvider',
+    docUri,
+    pos0,
+  )) as vscode.Location[];
+  assert.strictEqual(actualdecl0.length, 2);
```
seanyoung, line 229: Why does this return two locations, and not the function `foo` in `a` and `b3`?

src/bin/languageserver/mod.rs
```diff
@@ -281,19 +302,20 @@ impl SolangServer {
 
             let res = self.client.publish_diagnostics(uri, diags, None);
 
-            let (caches, definitions, types, implementations) = Builder::build(&ns);
+            let (file_caches, global_cache) = {
+                let builder = Builder::new(&ns);
+                builder.build()
+            };
```
seanyoung, line 308: This looks great! I think you could shorten it to:
```rust
 let (file_caches, global_cache) = Builder::new().build();
```

LucasSte (approved):

vscode/src/test/suite/extension.test.ts
chioni16: The requirement was to return the overridden functions that belonged to immediate parent contracts. 

So, in the following scenario where `A -> B` means `A` extends `B`, i.e, `A` is a child of `B`
```
     A -> B -> C 
```
when `goto-declaration` is invoked on a function in `A`, only the overridden function in `B` will be returned and not the one in `C`.  

If there are multiple immediate parents, then the overridden functions from all the immediate parents are returned.

seanyoung:
```
 function foo() internal override(b1, b2, b3) returns (uint64) {
```
This overrides three functions in three contracts.

xermicus (approved):

seanyoung (approved):

Issue #1536, opened 2023-09-16
Parse and resolve error: unknown function or type 'ecrecover'


bkushigian:
I believe the `ecrecover` function should be supported, at least in parsing? I don't have a minimal example, but if you clone [Solemate](https://github.com/transmissions11/solmate) and try to parse and resolve the file `src/utils/SafeTransferLib.sol` with import path `src`, then you will get the following error:

```
error: unknown function or type 'ecrecover'
    ┌─ /Users/benku/Certora/tests/solmate/src/tokens/ERC20.sol:130:40
    │
130 │             address recoveredAddress = ecrecover(
    │                                        ^^^^^^^^^

```

I used the `solang_parser` binary produced by [this testing repo](https://github.com/bkushigian/SolangParse4Debugging) and ran the following command:

```
solang_parser src/utils/SafeTransferLib.sol --import_paths src
```

PR #1543, opened 2023-09-17
Add ecrecover() builtin for EVM

seanyoung:
Fixes https://github.com/hyperledger/solang/issues/1536

### DIFF
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -2186,6 +2186,66 @@ fn expr_builtin(
                 expr: Box::new(codegen_expr),
             }
         }
+        ast::Builtin::ECRecover => {
+            // TODO:
+            // EVM: call precompile 1 (code below is untested)
+            // let args = args
+            //     .iter()
+            //     .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt))
+            //     .collect::<Vec<Expression>>();
+            //
+            // let payload = abi_encode(loc, args, ns, vartab, cfg, false).0;
+            //
+            // let instr = Instr::ExternalCall {
+            //     loc: *loc,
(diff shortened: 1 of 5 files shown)


LucasSte:
Is there a test for this change?

seanyoung:
> Is there a test for this change?

Now there is :blush:

src/codegen/expression.rs
```diff
@@ -2186,6 +2186,56 @@ fn expr_builtin(
                 expr: Box::new(codegen_expr),
             }
         }
+        ast::Builtin::ECRecover => {
```
LucasSte, line 2189: Is it possible to use `Instr::Unimplemented` here?

seanyoung: Why `Instr::Unimplemented` preferable over an actual implementation?

seanyoung: Which test coverage numbers?

xermicus: Yes, yolo in code in which we can't test isn't a good practice

seanyoung: ok, I've been outvoted, it's gone

xermicus: I mean it could also be commented out, saying that this is how we think it should be implemented, but it's commented out and an `Unreachable` instead until we can test it?

xermicus (approved):
LGTM. Gonna implement it for Polakdot soon, thanks!

src/codegen/expression.rs
seanyoung: Good idea. I've added the code I wrote as a comment

LucasSte (approved):

PR #1545, opened 2023-09-19
add hint to install ninja

nadeemb53:


seanyoung (approved):

docs/installing.rst
```diff
@@ -167,7 +167,7 @@ Building LLVM from source
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The LLVM project itself has a guide to `installing from source <http://www.llvm.org/docs/CMake.html>`_ which
-you may need to consult. First of all clone our LLVM repository:
+you may need to consult. You might need to `install Ninja <https://ninja-build.org>`_. First of all clone our LLVM repository:
```
LucasSte, line 170: The installation command uses Ninja, so I would not use `might` here. A better wording would be:

`Ninja is necessary for building LLVM from source.`

xermicus (approved):

LucasSte (approved):

PR #1546, opened 2023-09-20
Update and fix target config file templates

xermicus:
We don't want to maintain the same file twice. One version was outdated. The other one contained an invalid config for  `wasm-opt`.


LucasSte (approved):
It looks good to me, but you just need to investigate why the Mac Arm test is failing.

xermicus:
Sure, added the feature flag to the test fixing it

seanyoung (approved):

Issue #1433, opened 2023-07-07
Solana: it should be possible to pass seeds into function


seanyoung:
This program causes a panic in solang. 

```solidity
import "solana-library/spl_token.sol";

contract c {

    // Invoke the token program to mint tokens to a token account, using a PDA as the mint authority
    function mintTo(address mint, address account, address authority, uint64 amount, bytes[] seeds) internal {
        // Prepare instruction data
        bytes instructionData = new bytes(9);
        instructionData[0] = uint8(7); // MintTo instruction index
        instructionData.writeUint64LE(amount, 1); // Amount to mint

        // Prepare accounts required by instruction
        AccountMeta[3] metas = [
            AccountMeta({pubkey: mint, is_writable: true, is_signer: false}),
            AccountMeta({pubkey: account, is_writable: true, is_signer: false}),
            AccountMeta({pubkey: authority, is_writable: true, is_signer: true})
        ];

        // Invoke the token program with prepared accounts and instruction data
        SplToken.tokenProgramId.call{accounts: metas, seeds: seeds}(instructionData);
    }
}
```

LucasSte:
Should we have a specific syntax for slices? `bytes[]` is a vector of `bytes`.

LucasSte:
This is the solution of the problem described in #1028

PR #1547, opened 2023-09-21
Allow seeds to be passed into functions as string[] or string[][]

seanyoung:
Fixes https://github.com/hyperledger/solang/issues/1433


src/emit/expression.rs
```diff
@@ -2121,3 +2123,216 @@ fn runtime_cast<'a>(
...
+
+            let to = bin.llvm_type(to, ns);
+
+            let length = match dims.last().unwrap() {
+                ArrayLength::Dynamic => bin.vector_len(val),
+                ArrayLength::Fixed(len) => i32_const!(len.to_u64().unwrap()),
+                _ => unreachable!(),
+            };
+
+            // In Program Runtime v1, we can't do dynamic alloca. Remove the malloc once we move to
```
LucasSte, line 2277:
```suggestion
            // FIXME: In Program Runtime v1, we can't do dynamic alloca. Remove the malloc once we move to
```

src/emit/instructions.rs
```diff
@@ -867,69 +869,13 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
...
             };
 
-            let seeds = if let Some(seeds) = seeds {
-                let len = seeds.ty().array_length().unwrap().to_u64().unwrap();
-                let seeds_ty = bin.llvm_type(
-                    &Type::Slice(Box::new(Type::Slice(Box::new(Type::Bytes(1))))),
-                    ns,
-                );
+            let seeds_ty =
+                Type::Slice(Type::Slice(Type::Slice(Type::Bytes(1).into()).into()).into());
```
LucasSte, line 873: Why did the type go from slice of slice (`Type::Slice(Type::Slice(..))`) to slice of slice of slice (`Type::Slice(Type::Slice(Type::Slice(..)))`)?

LucasSte: Doesn't `invoke_signed_c` require a slice of slice?

src/sema/expression/literals.rs
```diff
@@ -799,6 +805,18 @@ pub(super) fn array_literal(
     }
 }
 
+/// When converting an array literal to slice of slices, what is the slice element of
+/// the array? (or array of array, or array of array of array etc).
```
LucasSte, line 809: This is not clear to me. Can you give an example?

tests/contract_testcases/solana/bad_seeds_on_external_call.sol
```diff
@@ -26,10 +26,10 @@ contract c {
 
 // ---- Expect: diagnostics ----
 // error: 13:4-7: type bytes1 found where array of slices expected
-// error: 15:4-5: expected 'bytes[]', found integer
+// error: 15:4-5: expected 'bytes slice[]', found integer
```
LucasSte, line 29: What is `bytes slice[]` supposed to be in Solidity?
I think the error message should point to something one can write in Solidity. You are providing a hint that does not work.

LucasSte:
```solidity
import {AccountMeta} from 'solana';
            contract pda {
                address constant tokenProgramId = address"TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
                address constant SYSVAR_RENT_PUBKEY = address"SysvarRent111111111111111111111111111111111";
                function test_bytes4(string[][] seeds) public {
                    bytes instr = new bytes(1);
                    instr[0] = 0x95;
                    AccountMeta[1] metas = [
                        AccountMeta({pubkey: SYSVAR_RENT_PUBKEY, is_writable: false, is_signer: false})
                    ];
                    tokenProgramId.call{seeds: seeds, accounts: metas}(instr);
                }
            }
```

Is this expected?
```
error: conversion from string[][] to bytes slice slice not possible
   ┌─ /Users/lucasste/Documents/solang/examples/test.sol:11:48
   │
11 │                     tokenProgramId.call{seeds: seeds, accounts: metas}(instr);
   │
```

LucasSte:
```solidity
            import {AccountMeta} from 'solana';
            contract pda {
                address constant tokenProgramId = address"TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
                address constant SYSVAR_RENT_PUBKEY = address"SysvarRent111111111111111111111111111111111";
                function test(bytes[] dyn, address[] addr, bytes5[] b5) public {
                    bytes instr = new bytes(1);
                    instr[0] = 0x95;
                    AccountMeta[1] metas = [
                        AccountMeta({pubkey: SYSVAR_RENT_PUBKEY, is_writable: false, is_signer: false})
                    ];
                    bytes3 foo = "foo";
                    string foo2 = "123456";

                    tokenProgramId.call{seeds: [ dyn, addr, b5, [foo, foo2] ], accounts: metas}(instr);
                }
            }
```

Is this expected (part 2)?

```
error: conversion from string to bytes not possible
   ┌─ /Users/lucasste/Documents/solang/examples/test.sol:15:71
   │
15 │                     tokenProgramId.call{seeds: [ dyn, addr, b5, [foo, foo2] ], accounts: metas}(instr);
   │
```

tests/solana_tests/call.rs
```diff
@@ -733,3 +733,252 @@ fn pda() {
...
+                    instr[0] = 0x95;
+
+                    AccountMeta[1] metas = [
+                        AccountMeta({pubkey: SYSVAR_RENT_PUBKEY, is_writable: false, is_signer: false})
+                    ];
+
+                    tokenProgramId.call{seeds: seeds, accounts: metas}(instr);
+                }
+
+                function test_addr(address[][] seeds) public {
```
LucasSte, line 855: If change the type to `address[][2]`, I get this error:
```
Assertion failed: (Ty && "Invalid GetElementPtrInst indices for type!"), function checkGEPType, file Instructions.h, line 922.
zsh: abort      ./target/debug/solang compile --target solana ../solang/examples/test.sol
```

If the type is `address[2][]`, the error is:
```
thread 'main' panicked at 'Found ArrayValue(ArrayValue { name: "elem21", address: 0x6000011e1820, is_const: false, is_const_array: false, is_const_data_array: false, is_null: false, llvm_value: "  %elem21 = load [2 x [32 x i8]], ptr %index_access20, align 1", llvm_type: ArrayType { array_type: Type { address: 0x15b075078, llvm_type: "[2 x [32 x i8]]" } } }) but expected PointerValue variant', /Users/lucasste/.cargo/registry/src/index.crates.io-6f17d22bba15001f/inkwell-0.2.0/src/values/enums.rs:302:13
```

Now, if it is `address[2][4]`, the error is:
```
thread 'main' panicked at 'Found ArrayValue(ArrayValue { name: "elem17", address: 0x6000002a54c0, is_const: false, is_const_array: false, is_const_data_array: false, is_null: false, llvm_value: "  %elem17 = load [2 x [32 x i8]], ptr %index_access16, align 1", llvm_type: ArrayType { array_type: Type { address: 0x15485be78, llvm_type: "[2 x [32 x i8]]" } } }) but expected PointerValue variant', /Users/lucasste/.cargo/registry/src/index.crates.io-6f17d22bba15001f/inkwell-0.2.0/src/values/enums.rs:302:13
```

These errors go away when I comment the external call.

src/emit/instructions.rs
seanyoung: A single seed is a slice of bytes.
There are multiple seeds allowed for an address (essentially hash of (program_id + seeds1 + seed2)
A single call can sign for multiple addresses.
That's three slices.

I'll add a comment.

tests/solana_tests/call.rs
seanyoung: You're right, I forgot to test fixed length arrays.

tests/contract_testcases/solana/bad_seeds_on_external_call.sol
seanyoung: I've removed `slice` from diagnostics and now it looks like solidity types.

src/sema/expression/literals.rs
seanyoung: You're right, this is totally not clear. I've re-written this and it's not necessary.

tests/contract_testcases/solana/bad_seeds_on_external_call.sol
```diff
@@ -26,10 +26,10 @@ contract c {
...
 // error: 16:4-5: type string found where array of slices expected
 // error: 17:4-13: type bytes2 found where array of slices expected
 // error: 18:4-5: type bytes found where array of slices expected
-// error: 20:5-6: expected 'bytes', found integer
+// error: 20:5-6: expected 'bytes slice', found integer
```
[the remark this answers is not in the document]
seanyoung: Also fixed

seanyoung:
Type `string` is now allowed as a seed

LucasSte:
```solidity
import {AccountMeta} from 'solana';
            contract pda {
                address constant tokenProgramId = address"TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
                address constant SYSVAR_RENT_PUBKEY = address"SysvarRent111111111111111111111111111111111";
                function test(bytes[] dyn, address[] addr, bytes5[] b5, string f) public {
                    bytes instr = new bytes(1);
                    instr[0] = 0x95;
                    AccountMeta[1] metas = [
                        AccountMeta({pubkey: SYSVAR_RENT_PUBKEY, is_writable: false, is_signer: false})
                    ];
                    bytes3 foo = "foo";
                    tokenProgramId.call{seeds: [ dyn, addr, b5, [f] ], accounts: metas}(instr);
                }
            }
```

Is this correct?
```
thread 'main' panicked at 'internal error: entered unreachable code', src/emit/expression.rs:2123:14
```

LucasSte:
I also tested the following example, but I am not confident it is going to be a common construct, so perhaps we should not worry about it.
```solidity
import {AccountMeta} from 'solana';
            contract pda {
                address constant tokenProgramId = address"TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
                address constant SYSVAR_RENT_PUBKEY = address"SysvarRent111111111111111111111111111111111";
                function test_bytes(bytes[][] seeds) public {
                    bytes instr = new bytes(1);
                    instr[0] = 0x95;
                    AccountMeta[1] metas = [
                        AccountMeta({pubkey: SYSVAR_RENT_PUBKEY, is_writable: false, is_signer: false})
                    ];
                    tokenProgramId.call{seeds: [seeds[0], seeds[1]], accounts: metas}(instr);
                }
            }

```

```
error: type bytes[] found where array expected
   ┌─ /Users/lucasste/Documents/solang/examples/test.sol:13:49
   │
13 │                     tokenProgramId.call{seeds: [seeds[0], seeds[1]], accounts: metas}(instr);
   │                                                 ^^^^^^^^

error: type bytes[] found where array expected
   ┌─ /Users/lucasste/Documents/solang/examples/test.sol:13:59
   │
13 │                     tokenProgramId.call{seeds: [seeds[0], seeds[1]], accounts: metas}(instr);
   │                                                           ^^^^^^^^

```

src/emit/expression.rs
```diff
@@ -2121,3 +2123,227 @@ fn runtime_cast<'a>(
...
+
+            bin.builder.build_store(src, val.into_int_value());
+
+            let bytes_length: u64 = (*bytes_length).into();
+
+            let dest = call!("__malloc", &[i32_const!(bytes_length).into()])
+                .try_as_basic_value()
+                .left()
+                .unwrap()
+                .into_pointer_value();
```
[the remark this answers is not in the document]
seanyoung: The other allocation is on the stack and that is a single allocation. This malloc allocation is per-slice.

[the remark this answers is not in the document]
LucasSte: You do not need to allocate space in the stack and in the heap. I can just allocated the memory in one place (heap or stack(, store the integer there, change the endianness in place and return the pointer to that memory region.

src/emit/binary.rs
```diff
@@ -805,33 +805,23 @@ impl<'a> Binary<'a> {
...
+                                aty = self.module.get_struct_type("struct.vector").unwrap().into();
                             }
                             ArrayLength::AnyFixed => {
                                 unreachable!()
                             }
                         }
                     }
 
-                    BasicTypeEnum::ArrayType(aty)
+                    aty
```
xermicus, line 824: Can be simplified (unless the loop introduces side-effects, which would rather surprise me):  

```rust
match dims.last() {
  Some(ArrayLength::Fixed(d)) => self.llvm_field_ty(base_ty, ns).array_type(d.to_u32().unwrap()).into(),
  Some(ArrayLength::Dynamic) => self.module.get_struct_type("struct.vector").unwrap().into(),
  None => self.llvm_field_ty(base_ty, ns),
  _ => unreachable!()
}
```

src/emit/expression.rs
```diff
@@ -2121,3 +2123,227 @@ fn runtime_cast<'a>(
...
+        Expression::AllocDynamicBytes {
+            initializer: Some(initializer),
+            ..
+        } => {
+            let ptr = bin.emit_global_string("slice_constant", initializer, true);
+            let len = i64_const!(initializer.len() as u64);
+
+            return (ptr, len);
+        }
+        _ => (),
```
xermicus, line 2200: This is a bit confusing, I expected an `unreachable!()` instead. Is this enforced in an earlier stage? Maybe a comment explaining why this is fine here would be nice

src/emit/binary.rs
seanyoung: We want to walk all the dimensions (e.g. `int32[4][3][2]` needs to be the same in llvm). However you do have a point, it could be simpler. I've re-written it using `.fold()`. Hope this is better - it's certainly shorter.

src/emit/expression.rs
seanyoung: You're right, the logic is strange here. I've re-written it, let me know what you think.

src/emit/polkadot/mod.rs
```diff
@@ -43,6 +43,20 @@ macro_rules! emit_context {
...
+            };
+        }
+
+        #[allow(unused_macros)]
+        macro_rules! i64_zero {
+            () => {
+                $binary.context.i64_type().const_zero()
+            };
+        }
+
```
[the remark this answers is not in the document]
seanyoung: They're moved.

src/emit/expression.rs
xermicus: Ah I see. Yes that's much shorter and quicker to grasp.

However, I think I overlooked the early return in case of `ArrayLength::Dynamic`, that might actually play a role here?

src/emit/binary.rs
xermicus: Ah I see. Yes that's much shorter and quicker to grasp.

However, I think I overlooked the early return in case of` ArrayLength::Dynamic` in the original version. That might actually play a role here?

src/emit/expression.rs
xermicus: Easy to grasp now, thanks!

src/emit/binary.rs
xermicus: Very well then

seanyoung:
@LucasSte I've added the two cases you found as a test cases

src/emit/expression.rs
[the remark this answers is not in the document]
seanyoung: This might be something for another PR. We do all our endian-ness swaps like this.

In fact, I think `bytesN` should be represented by a byte array and then all these swaps would go away.

src/emit/expression.rs
```diff
@@ -2120,3 +2122,227 @@ fn runtime_cast<'a>(
         _ => unreachable!(),
     }
 }
+
+/// Emit expression into a slice
```
xermicus, line 2126: Nitpick: Superfluous. A more useful comment could for example say when this is needed or explain the return type (which I'd guess is a pointer to the slice and the length of the slice)?

src/emit/expression.rs
```diff
@@ -2120,3 +2122,227 @@ fn runtime_cast<'a>(
...
+                bin.builder
+                    .build_load(bin.llvm_field_ty(&from_elem, ns), input_elem, "elem")
+            } else {
+                input_elem.into()
+            };
+
+            let (data, len) =
+                basic_value_to_slice(bin, input_elem, &from_elem, &to_elem, function, ns);
+
+            let output_data = unsafe {
```
xermicus, line 2324: [Each unsafe block should have a `SAFETY:` comment ](https://std-dev-guide.rust-lang.org/policy/safety-comments.html)

src/emit/expression.rs
```diff
@@ -2120,3 +2122,227 @@ fn runtime_cast<'a>(
...
+                bin.builder
+                    .build_gep(to, output, &[index, i32_zero!()], "output_data")
+            };
+
+            bin.builder.build_store(output_data, data);
+
+            let output_len = unsafe {
```
xermicus, line 2331: Missing `SAFETY:` comment

src/emit/expression.rs
```diff
@@ -2120,3 +2122,227 @@ fn runtime_cast<'a>(
...
+                    target,
+                    bin,
+                    &values[i as usize],
+                    to_elem_ty,
+                    vartab,
+                    function,
+                    ns,
+                );
+
+                let output_ptr = unsafe {
```
xermicus, line 2165: Missing `SAFETY:` comment

src/emit/expression.rs
```diff
@@ -2120,3 +2122,227 @@ fn runtime_cast<'a>(
...
+                        llvm_to,
+                        output,
+                        &[i32_const!(i.into()), i32_zero!()],
+                        "output_ptr",
+                    )
+                };
+
+                bin.builder.build_store(output_ptr, ptr);
+
+                let output_len = unsafe {
```
xermicus, line 2176: Missing `SAFETY:` comment

xermicus (commented):
LGTM overall, just some cosmetics

src/emit/expression.rs
seanyoung: You're right, that comment is terrible.

src/emit/expression.rs
seanyoung: In emit we have `unsafe {}` blocks for all calls to `build_gep()`. This is because llvm can crash if the arguments are not correct, and the "safe" inkwell wrappers do not check for this (in fact I don't think it's possible to fix in inkwell, this should be fixed in llvm).

So I think all the `unsafe {}` blocks in emit are for `build_gep()`. Do we really require `SAFETY: ` comment for each and every one with the same comment (83 occurrences)?

I am not against documenting this - it's certainly useful information - I am just not sure where it should go.

xermicus: Yeah I see. What if we just start doing this from now on?

seanyoung: Ok I've added comments. I think it clutters the code

xermicus: `build_gep() is unsafe since it can crash deep in llvm`

This is not the point of the comment. The comment should say why we think this is still safe, despite the `unsafe`. E.g. a proof that the arguments are correct in that case.

>  I think it clutters the code

I got used to it. To the extend I already expect it :grin:

xermicus (approved):
As discussed in the standup, it would be great to have those `SAFETY` comments, explaining our thoughts around why all those arguments to the LLVM GEP builder are correct. But it's not only in this place, I'll leave it up to you.

seanyoung:
> As discussed in the standup, it would be great to have those `SAFETY` comments, explaining our thoughts around why all those arguments to the LLVM GEP builder are correct. But it's not only in this place, I'll leave it up to you.

What do you think of the SAFETY comments I've just added?

LucasSte (commented):
I tested it locally and couldn't find any problem. Codewise, everything looks good. I have only one last question.

Should there be an integration test for this feature?

seanyoung:
> I tested it locally and couldn't find any problem. Codewise, everything looks good. I have only one last question.
> 
> Should there be an integration test for this feature?

I've added a simple test.

This will need re-writing for your branch.

There is also a fix for the "blockhash not found" issue (at least it fixes it locally)

LucasSte (approved):

PR #1548, opened 2023-09-22
Add code coverage bot 🤖

LucasSte:
Coverage reports will be stored [here](https://app.codecov.io/gh/hyperledger/solang).

According to Codecov's documentation,
> Once merged to the default branch, subsequent pull requests will have checks and report comment.

This is an example of how comments look like: https://github.com/solana-labs/solana/pull/33363#issuecomment-1730528544

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -416,3 +416,28 @@ jobs:
           make test
           ./test
         working-directory: ./stdlib
+
+  coverage:
+    runs-on: ubuntu-latest
+    container: ghcr.io/hyperledger/solang-llvm:ci-5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools
+      - name: cargo install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: cargo llvm-cov
+        run: cargo llvm-cov --all-features --lcov --output-path lcov.info
+        env:
+          CARGO_HUSKY_DONT_INSTALL_HOOKS: true
+      - name: Upload coverage report to codecov.io
+        uses: codecov/codecov-action@v3
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        with:
+          fail_ci_if_error: true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,7 +90,10 @@ jobs:
       if: always()
       run: cargo llvm-cov --all-features --workspace --no-report
     - name: Compress test coverage files
-      run: tar -czvf rust-tests.tar.gz *.profraw ./llvm-cov-target/*.profraw
+      run: |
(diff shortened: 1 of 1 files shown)


seanyoung:
This does not give a good representation of our coverage; the integration tests aren't run under `llvm-cov`. For example it looks like the language server looks entirely without coverage; there are no rust tests for the language server, since there is no rust client for a language server.

.github/workflows/test.yml
```diff
@@ -417,3 +417,28 @@ jobs:
           make test
           ./test
         working-directory: ./stdlib
+
+  coverage:
+    runs-on: ubuntu-latest
+    container: ghcr.io/hyperledger/solang-llvm:ci-5
```
xermicus, line 423: Meanwhile we bumped this to `ci-6`, don't forget to update the image before merging tag

LucasSte: That has not been merged.

.github/workflows/test.yml
```diff
@@ -64,16 +78,29 @@ jobs:
       run: cargo doc --workspace --bins
     - name: Run cargo fmt
       run: cargo fmt --all -- --check
-    - name: Build
-      if: always()
```
seanyoung, line 68: any reason ` if: always()` has been removed?

The idea was that if clippy/rustfmt fails, then we can still try to build/test it (is this useful?)

.github/workflows/test.yml
```diff
@@ -416,3 +478,67 @@ jobs:
...
+      CARGO_LLVM_COV_TARGET_DIR: ${{ github.workspace }}/target
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools
```
seanyoung, line 501: You're not specify the toolchain version - this may cause problems. i.e. build with version X, but now the prof files won't work with X + 1

.github/workflows/test.yml
```diff
@@ -47,13 +47,27 @@ jobs:
     name: Linux x86-64
     runs-on: solang-ubuntu-latest
     container: ghcr.io/hyperledger/solang-llvm:ci-5
+    env:
+      RUSTFLAGS: -C instrument-coverage -C llvm-args=--instrprof-atomic-counter-update-all --cfg=coverage --cfg=trybuild_no_target
+      LLVM_PROFILE_FILE: ${{ github.workspace }}/target/solang-%p-%10m.profraw
+      CARGO_INCREMENTAL: 0
+      CARGO_LLVM_COV: 1
+      CARGO_LLVM_COV_TARGET_DIR: ${{ github.workspace }}/target
```
seanyoung, line 55: Rather than repeating the env for each job, the env can be set on the entire workflow: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env

seanyoung (approved):
Great, fantastic feature for the project

.github/workflows/test.yml
LucasSte: That was an oversight. I'll add that again.

.github/workflows/test.yml
LucasSte: Good catch.

xermicus (approved):
Cool, this is great to have, thanks!

PR #1552, opened 2023-09-27
Bugfix storage variable return param

xermicus:
Variables declared as return parameters can have a storage location too.

### DIFF
--- a/src/sema/statements.rs
+++ b/src/sema/statements.rs
@@ -247,7 +247,7 @@ pub fn resolve_function_body(
                 ns,
                 VariableInitializer::Solidity(None),
                 VariableUsage::ReturnVariable,
-                None,
+                p.1.as_ref().unwrap().storage.clone(),
             ) {
                 ns.check_shadowing(file_no, contract_no, name);
                 symtable.returns.push(pos);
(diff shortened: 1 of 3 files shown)


seanyoung (approved):
This PR should be called:

 Bugfix: Named variables declared in return parameters can have a storage location

xermicus:
It's also for [unnamed variables](https://github.com/hyperledger/solang/pull/1552/files#diff-4a712d7a3bd2c623e9ae5bcb7bed4928ed2227fea3d57b0768921e124de66eceR273)?

seanyoung:
> It's also for [unnamed variables](https://github.com/hyperledger/solang/pull/1552/files#diff-4a712d7a3bd2c623e9ae5bcb7bed4928ed2227fea3d57b0768921e124de66eceR273)?

Yes, you're right. I don't get how this could have worked before, we have been returning storage slots from functions.

Anyway looks good

xermicus:
@seanyoung I think in practice this isn't noticeable on unnamed variables declared in the return parameters, because in that case the variable doesn't _really_ exist until it's returned. It's sort of "shadowed" by something else, until the corresponding `return` statement is reached. And by then, the function already returned, so nothing can for example access the `slot`, where it would play a role. So it can't have consumers in the function local scope, hence we probably never noticed.

However, I agree, technically, it's still correct to set the storage location.

PR #1557, opened 2023-10-03
Add function annotations 🚀

LucasSte:
This PR introduces account annotations for developers to declare accounts in a Solidity function. This is one of the task items in #1251.

The new syntax tremendously facilitates writing contracts and is less error prone as we do not need to both pass the account address as an argument and in the accounts field.

I wrote some tests where I noticed they were necessary and I attempted to refactor all existing Solana Solidity to use the annotations, so that my implementation would be stress tested in all those cases.

There are a few items this PR did not accomplish, but are going to be implemented in future changes:

1. I did not validate accounts during runtime, i.e. I am not checking if the bits are properly set before dispatching the function, but I wonder if this is really needed.
2. When generating the interface for an IDL, I am not creating the account annotations, but this is an amazing feature to be added soon.
3. The `address.balance` function is useless on Solana, but has not been eliminated yet. We can now just do `tx.accounts.myAccount.lamports`.
4. The `address.send` and `address.transfer` are also not necessary, but I did not purge the compiler of them.

### DIFF
--- a/docs/language/functions.rst
+++ b/docs/language/functions.rst
@@ -325,6 +325,16 @@ calling.
         (bool success, bytes rawresult) = foo.call{value: 102, gas: 1000}(rawcalldata);
     }
 
+External calls with the ``call()`` method on Solana must have the ``accounts`` call argument, regardless of the
+callee function visibility, because the compiler has no information about the caller function to generate the
+``AccountMeta`` array automatically.
+
+.. code-block:: solidity
+
+    function test(address foo, bytes rawcalldata) public {
+        (bool success, bytes rawresult) = foo.call{accounts: []}(rawcalldata);
+    }
+
 .. _fallback_receive:
 
 Calling an external function using ``delegatecall``
--- a/integration/solana/calls.spec.ts
+++ b/integration/solana/calls.spec.ts
@@ -18,57 +18,58 @@ describe('Testing calls', function () {
 
         await callee.program.methods.setX(new BN(102))
             .accounts({ dataAccount: callee.storage.publicKey })
-            .rpc();
+            .rpc({commitment: "confirmed"});
 
         let res = await callee.program.methods.getX()
             .accounts({ dataAccount: callee.storage.publicKey })
-            .view();
+            .view({commitment: "confirmed"});
 
         expect(res).toEqual(new BN(102));
 
         res = await caller.program.methods.whoAmI()
-            .view();
+            .view({commitment: "confirmed"});
 
         expect(res).toStrictEqual(caller.program_key);
 
-        await caller.program.methods.doCall(callee.program_key, new BN(13123))
+        await caller.program.methods.doCall(new BN(13123))
             .accounts({
                 callee_dataAccount: callee.storage.publicKey,
-                callee_programId: callee.program_key,
+                callee_pid: callee.program_key,
                 })
-            .rpc();
+            .rpc({commitment: "confirmed"});
 
         res = await callee.program.methods.getX()
             .accounts({ dataAccount: callee.storage.publicKey })
-            .view();
+            .view({commitment: "confirmed"});
 
         expect(res).toEqual(new BN(13123));
 
-        res = await caller.program.methods.doCall2(callee.program_key, new BN(20000))
+        res = await caller.program.methods.doCall2(new BN(20000))
             .accounts({
                 callee_dataAccount: callee.storage.publicKey,
-                callee_programId: callee.program_key,
+                callee_pid: callee.program_key,
             })
-            .view();
+            .view({commitment: "confirmed"});
 
         expect(res).toEqual(new BN(33123));
 
-        res = await caller.program.methods.doCall3(callee.program_key, callee2.program_key, [new BN(3), new BN(5), new BN(7), new BN(9)], "yo")
+        res = await caller.program.methods.doCall3([new BN(3), new BN(5), new BN(7), new BN(9)], "yo")
             .accounts({
-                callee2_programId: callee2.program_key,
-                callee_programId: callee.program_key,
+                callee2_pid: callee2.program_key,
+                callee_pid: callee.program_key,
             })
-            .view();
+            .view({commitment: "confirmed"});
 
         expect(res.return0).toEqual(new BN(24));
         expect(res.return1).toBe("my name is callee");
 
-        res = await caller.program.methods.doCall4(callee.program_key, callee2.program_key, [new BN(1), new BN(2), new BN(3), new BN(4)], "asda")
+        res = await caller.program.methods.doCall4([new BN(1), new BN(2), new BN(3), new BN(4)], "asda")
             .accounts({
-                callee2_programId: callee2.program_key,
-                callee_programId: callee.program_key,
+                callee2_pid: callee2.program_key,
+                callee_pid: callee.program_key,
+                other_callee2: callee2.program_key,
             })
-            .view();
+            .view({commitment: "confirmed"});
 
         expect(res.return0).toEqual(new BN(10));
         expect(res.return1).toBe("x:asda");
--- a/solang-parser/src/solidity.lalrpop
+++ b/solang-parser/src/solidity.lalrpop
@@ -442,7 +442,7 @@ NamedArgument: NamedArgument = {
         let name = Identifier { loc:  Loc::File(file_no, l, ar), name: "address".into() };
 
         NamedArgument{ loc: Loc::File(file_no, l, r), name, expr }
-    }
+    },
 }
 
 FunctionCall: Expression = {
@@ -488,7 +488,7 @@ NoFunctionTyPrecedence0: Expression = {
             Identifier { loc: Loc::File(file_no, al, b), name: "new".to_string() })
     },
     <l:@L> <ty:NoFunctionType> <r:@R> => Expression::Type(Loc::File(file_no, l, r), ty),
-    <a:@L> "[" <v:CommaOne<Expression>> "]" <b:@R> => {
+    <a:@L> "[" <v:Comma<Expression>> "]" <b:@R> => {
         Expression::ArrayLiteral(Loc::File(file_no, a, b), v)
     },
     <SolNoRevertIdentifier> => Expression::Variable(<>),
--- a/src/emit/binary.rs
+++ b/src/emit/binary.rs
@@ -126,6 +126,13 @@ macro_rules! emit_context {
                 )
             };
         }
+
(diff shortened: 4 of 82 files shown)


src/emit/binary.rs
```diff
@@ -868,10 +868,34 @@ impl<'a> Binary<'a> {
...
-                .get_struct_type("struct.SolAccountInfo")
-                .unwrap()
-                .into();
+                .context
+                .struct_type(
+                    &[
+                        self.context
+                            .i8_type()
+                            .ptr_type(AddressSpace::default())
+                            .as_basic_type_enum(), // SolPubkey *
```
seanyoung, line 877: `byte_ptr!()`?

src/emit/binary.rs
```diff
@@ -868,10 +868,34 @@ impl<'a> Binary<'a> {
...
+                        self.context
+                            .i64_type()
+                            .ptr_type(AddressSpace::default())
+                            .as_basic_type_enum(), // uint64_t *
+                        self.context.i64_type().as_basic_type_enum(), // uint64_t
+                        self.context
+                            .i8_type()
+                            .ptr_type(AddressSpace::default())
+                            .as_basic_type_enum(), // uint8_t *
```
seanyoung, line 886: `byte_ptr!()`

src/emit/binary.rs
```diff
@@ -868,10 +868,34 @@ impl<'a> Binary<'a> {
...
+                        self.context
+                            .i8_type()
+                            .ptr_type(AddressSpace::default())
+                            .as_basic_type_enum(), // SolPubkey *
+                        self.context.i64_type().as_basic_type_enum(), // uint64_t
+                        self.context.i8_type().as_basic_type_enum(),  // bool
+                        self.context.i8_type().as_basic_type_enum(),  // bool
+                        self.context.i8_type().as_basic_type_enum(),  // bool
```
seanyoung, line 894: Maybe we should have macros for these too, all this is a bit long

src/emit/solana/mod.rs
```diff
@@ -1208,6 +1175,30 @@ impl SolanaTarget {
...
+                    .ptr_type(AddressSpace::default())
+                    .into(),
+                binary.context.i32_type().into(),
+            ],
+            false,
+        );
+
+        binary
+            .module
+            .add_function("sol_invoke_signed_c", fn_type, None);
```
seanyoung, line 1200: This belongs in `fn declare_externals()`

src/emit/solana/mod.rs
```diff
@@ -1112,12 +1058,33 @@ impl SolanaTarget {
...
+                    binary
+                        .context
+                        .i8_type()
+                        .ptr_type(AddressSpace::default())
+                        .as_basic_type_enum(),
+                    binary.context.i64_type().as_basic_type_enum(),
+                ],
+                false,
+            )
+            .as_basic_type_enum();
```
seanyoung, line 1087: declare_externals?

integration/solana/calls.spec.ts
```diff
@@ -18,57 +18,58 @@ describe('Testing calls', function () {
...
+        res = await caller.program.methods.doCall4([new BN(1), new BN(2), new BN(3), new BN(4)], "asda")
             .accounts({
-                callee2_programId: callee2.program_key,
-                callee_programId: callee.program_key,
+                callee2_pid: callee2.program_key,
+                callee_pid: callee.program_key,
+                other_callee2: callee2.program_key,
             })
-            .view();
+            .view({commitment: "confirmed"});
```
seanyoung, line 72: does this make any sense on view function?

docs/language/functions.rst
```diff
@@ -325,6 +325,16 @@ calling.
         (bool success, bytes rawresult) = foo.call{value: 102, gas: 1000}(rawcalldata);
     }
 
+External calls with the ``call()`` method on Solana must receive the ``accounts`` call argument, regardless of the
```
seanyoung, line 328: receive -> have

seanyoung (changes requested):
How the annotations work with constructors and virtual functions?

solang-parser/src/solidity.lalrpop
```diff
@@ -442,6 +442,12 @@ NamedArgument: NamedArgument = {
         let name = Identifier { loc:  Loc::File(file_no, l, ar), name: "address".into() };
 
         NamedArgument{ loc: Loc::File(file_no, l, r), name, expr }
+    },
+
+    <l:@L> <name:SolIdentifier> ":" <al:@L> "[" "]" <r:@R> => {
+        NamedArgument{loc: Loc::File(file_no, l, r), name,
+            expr: Expression::ArrayLiteral(Loc::File(file_no, al, r), vec![]),
+        }
```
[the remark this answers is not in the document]
LucasSte: Empty arrays are not allowed in Solidity. The production you are suggesting allows them and we would need to treat such cases in sema.

src/emit/binary.rs
LucasSte: Well, any type of pointer works, because they are opaque.

LucasSte: The pointer here is pointing to `SolPubkey`, but I did not want to insert the struct definition just to retrieve a pointer to it.

integration/solana/calls.spec.ts
LucasSte: Without this, I get the `BlockHash not found` error locally.

LucasSte:
> How the annotations work with constructors and virtual functions?

As constructors are just like any other function on Solana, the annotations work as in an external function. A test case for constructors is missing, though.

I missed the case about virtual functions. We should either disallow annotations or required that the overriding function repeat the account declaration (I'd extend these rules to interfaces). Do you have any opinion on the subject?

solang-parser/src/solidity.lalrpop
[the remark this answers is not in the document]
seanyoung: Yes, we should parse them and give appropriate error messages in sema, rather than parse error. That is what solc does too.

src/emit/binary.rs
seanyoung: I'm just suggesting to replace `self.context
                            .i8_type()
                            .ptr_type(AddressSpace::default())
                            .as_basic_type_enum()` with `byte_ptr!()`

integration/solana/calls.spec.ts
seanyoung: ah interesting :pray:

src/sema/ast.rs
```diff
@@ -1216,12 +1217,69 @@ pub struct CallArgs {
...
+    NoAccount,
+    #[default]
+    AbsentArgument,
+    Present(T),
+}
+
+impl<T> ExternalCallAccounts<T> {
+    /// Is the accounts call argument missing?
+    pub fn is_absent(&self) -> bool {
+        matches!(self, ExternalCallAccounts::AbsentArgument)
```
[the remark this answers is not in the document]
LucasSte: That is only possible if `T` implements PartialEq.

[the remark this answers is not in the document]
LucasSte: I did not want to restrict the data structure to that requirement.

src/emit/solana/mod.rs
LucasSte: I don't think there is a way to add struct definitions to the module.

LucasSte:
@seanyoung I am requiring now that overriding functions declared the same accounts as their respective virtual ones. I may not have conceived the most ideal error message, though. Can you please check if you like it?

integration/solana/calls.spec.ts
```diff
@@ -18,57 +18,58 @@ describe('Testing calls', function () {
...
-                callee_programId: callee.program_key,
+                callee_pid: callee.program_key,
                 })
-            .rpc();
+            .rpc({commitment: "confirmed"});
 
         res = await callee.program.methods.getX()
             .accounts({ dataAccount: callee.storage.publicKey })
-            .view();
+            .view({commitment: "confirmed"});
```
seanyoung, line 43: Rather than providing this for each call, I think this can be set once for the connection as a second argument to `AnchorProvider.local()` in `setup.ts`.

src/sema/contracts.rs
```diff
@@ -833,6 +834,81 @@ fn base_function_compatible(
...
+/// Checks if the accounts from the virtual function and the overriding one are compatible
+fn verify_accounts_correspondence<'a>(
+    func_accounts: &'a IndexMap<String, SolanaAccount>,
+    other_accounts: &'a IndexMap<String, SolanaAccount>,
+    no_correspondence: &mut Vec<(pt::Loc, &'a String)>,
+    incorrect_flag: &mut IndexSet<(pt::Loc, pt::Loc, &'a String)>,
+    reverse: bool,
+) {
+    for (account_name, account_flags) in other_accounts {
+        if let Some(other_account) = func_accounts.get(account_name) {
```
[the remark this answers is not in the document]
LucasSte: Good point. The order matters if one is manually creating the AccountMeta array for the call. When we call the function using Anchor's interface, its already deals with the order for you.

Having said that, I'd rather check the order.

src/sema/contracts.rs
```diff
@@ -833,6 +834,81 @@ fn base_function_compatible(
...
+        &mut incorrect_flag,
+        true,
+    );
+
+    if !no_correspondence.is_empty() {
+        let notes = no_correspondence
+            .iter()
+            .map(|(loc, account_name)| Note {
+                loc: *loc,
+                message: format!("a correspondence for account '{}' is missing", account_name),
```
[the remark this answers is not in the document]
LucasSte: According to the [dictionary](https://www.collinsdictionary.com/dictionary/english/correspondence),

> If there is a correspondence between two things, there is a [close](https://www.collinsdictionary.com/dictionary/english/close) similarity or [connection](https://www.collinsdictionary.com/dictionary/english/connection) between them.

I do believe, though, that `corresponding` account sounds better.

integration/solana/calls.spec.ts
LucasSte: Adding `{commitment: "confirmed"}` to `AnchorProvider.local` causes all tests to fail.

src/emit/solana/mod.rs
seanyoung: There must be a way to types to a module, how else would you build a complete IR for a file. 

Having said that, I don't know how it is done. I can't find it either

seanyoung (approved):

xermicus (approved):
LGTM

PR #1559, opened 2023-10-05
Fix clippy 1.73 warnings

LucasSte:
I updated Rust in my computer and fixed the errors from clippy.

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -657,10 +657,10 @@ impl ControlFlowGraph {
                 dimensions, values, ..
             } => format!(
                 "constant {} [ {} ]",
-                dimensions
-                    .iter()
-                    .map(|d| format!("[{d}]"))
-                    .collect::<String>(),
+                dimensions.iter().fold(String::new(), |mut output, d| {
+                    write!(output, "[{d}]").unwrap();
+                    output
+                }),
                 values
                     .iter()
                     .map(|e| self.expr_to_string(contract, ns, e))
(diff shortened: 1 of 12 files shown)


src/codegen/cfg.rs
```diff
@@ -657,10 +657,10 @@ impl ControlFlowGraph {
                 dimensions, values, ..
             } => format!(
                 "constant {} [ {} ]",
-                dimensions
-                    .iter()
-                    .map(|d| format!("[{d}]"))
-                    .collect::<String>(),
+                dimensions.iter().fold(String::new(), |mut output, d| {
+                    let _ = write!(output, "[{d}]");
```
seanyoung, line 661: is `let _  = ` necessary?

src/codegen/cfg.rs
```diff
@@ -671,10 +671,10 @@ impl ControlFlowGraph {
                 dimensions, values, ..
             } => format!(
                 "{} [ {} ]",
-                dimensions
-                    .iter()
-                    .map(|d| format!("[{d}]"))
-                    .collect::<String>(),
+                dimensions.iter().fold(String::new(), |mut output, d| {
+                    let _ = write!(output, "[{d}]");
```
seanyoung, line 675: is `let _  = ` necessary?

seanyoung (approved):

src/codegen/cfg.rs
LucasSte: It is:
```
warning: unused `std::result::Result` that must be used
   --> src/codegen/cfg.rs:661:21
    |
661 |                     write!(output, "[{d}]");
    |                     ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this `Result` may be an `Err` variant, which should be handled
    = note: `#[warn(unused_must_use)]` on by default
    = note: this warning originates in the macro `write` (in Nightly builds, run with -Z macro-backtrace for more info)
```

seanyoung: So your `let _ = ` swallows errors, that's not great. It should be use `.unwrap()` or `.expect()`.

LucasSte: The `let _ = ` was clippy's suggestion. https://rust-lang.github.io/rust-clippy/master/index.html#/format_collect
I'll change it to unwrap.

PR #1560, opened 2023-10-06
Language server code complete

chioni16:


src/sema/symtable.rs
```diff
@@ -78,12 +78,12 @@ pub enum VariableUsage {
...
 
 #[derive(Debug, Clone)]
-struct VarScope(HashMap<String, usize>, Option<HashSet<usize>>);
+pub struct VarScope(pub HashMap<String, usize>, Option<HashSet<usize>>);
 
 #[derive(Default, Debug, Clone)]
 pub struct Symtable {
     pub vars: IndexMap<usize, Variable>,
-    scopes: Vec<VarScope>,
+    pub scopes: Vec<VarScope>,
```
seanyoung, line 86: Surely this vector is empty after resolving a function, all the scopes have been popped off again.

src/sema/symtable.rs
```diff
@@ -187,8 +187,19 @@ impl Symtable {
...
+            let curr_scope = curr_scope
+                .0
+                .values()
+                .filter_map(|pos| {
+                    self.vars
+                        .get(pos)
+                        .map(|var| (var.id.name.clone(), var.ty.clone()))
+                })
+                .collect();
+            ns.scopes.push((loc, curr_scope));
```
seanyoung, line 201: You are storing a scope with the name of all the local variables and their types, along with the location in the global namespace. This is not good, this is only useful to the language server.

Function scopes should be stored in `struct Function` (or `struct Symtable`). Please refer to variables by their id, not their string name. The string name is not unique, one scope can shadow another variable, so the ast needs precise information - the variable id. Using the string name + type is duplicating information, waste of memory and not precise.

src/sema/ast.rs
```diff
@@ -1871,6 +1877,28 @@ impl Statement {
         matches!(&self, Statement::Underscore(_))
     }
 
+    pub fn loc(&self) -> pt::Loc {
```
seanyoung, line 1880: already implemented on line 1537

src/sema/ast.rs
```diff
@@ -1871,6 +1877,28 @@ impl Statement {
...
+            | Statement::Delete(loc, _, _)
+            | Statement::Destructure(loc, _, _)
+            | Statement::Continue(loc)
+            | Statement::Break(loc)
+            | Statement::Return(loc, _)
+            | Statement::Revert { loc, .. }
+            | Statement::Emit { loc, .. }
+            | Statement::TryCatch(loc, _, _)
+            | Statement::Underscore(loc) => *loc,
+            Statement::Assembly(ia, _) => ia.loc,
```
seanyoung, line 1898: although this particular line is better than the old impl

src/sema/ast.rs
```diff
@@ -694,6 +694,12 @@ pub struct Namespace {
     pub var_constants: HashMap<pt::Loc, codegen::Expression>,
     /// Overrides for hover in the language server
     pub hover_overrides: HashMap<pt::Loc, String>,
+    /// Inner scopes of all functions in the namespace.
+    /// Inner scopes of a function includes all block scopes in a function
+    /// excluding the outermost scope that includes named function params,
+    /// return values and variables defined outside any of the inner scopes.
+    /// Used by the language server.
+    pub scopes: Vec<(pt::Loc, Vec<(String, Type)>)>,
```
seanyoung, line 702: The ast should not store this by string/type. This is just useful for the language server.

src/sema/symtable.rs
chioni16: The outermost scope of a function (which includes the arguments and named return values along with the local variables declared in the outermost scope) is not popped. 
So, currently I am using the scope that's left behind in the `scopes` field in `languageserver` module. 
Do you think there is a better place where I can do this? 

@seanyoung

src/sema/symtable.rs
chioni16: I was thinking of introducing a new field in `struct Symtable` for storing the popped scopes. 
This way I can extract the variable name, type and location in the `languageserver` module and we can do away with the new `scopes` field in `struct Namespace`.

src/sema/symtable.rs
seanyoung: If the last scope is not popped then that's a bug. Also, that is not enough information, you need all the scopes, right?

chioni16: > If the last scope is not popped then that's a bug.
I just printed the `Namespace` struct from within the `languageserver` module. I can see that there is always 1 `VarScope` element in the `scopes` field. 

> you need all the scopes, right?
Yes, I need information about all the different lexical scopes of a function and the variables declared within each of the scopes.

seanyoung: A new field with all the scopes might be better.

src/sema/ast.rs
chioni16: Removed the `scopes` field from `struct Namespace`.

src/sema/symtable.rs
chioni16: I have made some changes. Now I have introduced a new field `popped_fields` (looking for a better name). The scopes popped by calling `leave_scope` method are now added to the `popped_scopes` - https://github.com/hyperledger/solang/pull/1560/files#diff-88601e831b596c4462675b562c54089bbcf53d14b625c8fb8a955eed6d6bb08aR194

All the lexical scopes but the outermost scope of a function are added to `popped_scopes` by the time `languageserver` module can use `Namespace` struct. 

I make use of these `popped_scopes` along with the outermost scope that has been left behind in the `languageserver` module. - https://github.com/hyperledger/solang/pull/1560/files#diff-bf2200d9cb4eee660f8ff53c8416979a1f0ebe5d5856609e7d7f46fe0c8e9846R1619-R1652

src/sema/ast.rs
```diff
@@ -319,6 +319,7 @@ impl fmt::Display for Mutability {
 pub struct Function {
     pub tags: Vec<Tag>,
     /// The location of the prototype (not body)
+    pub loc_prototype: pt::Loc,
```
seanyoung, line 322: Fantastic :tada:

src/sema/symtable.rs
seanyoung: #1560 has been merged, please see if you can the `active_scopes` into `ExprContext`. :pray:

src/sema/symtable.rs
```diff
@@ -78,23 +79,25 @@ pub enum VariableUsage {
...
 
 #[derive(Debug, Clone)]
-struct VarScope(HashMap<String, usize>, Option<HashSet<usize>>);
+pub struct VarScope(pub HashMap<String, usize>, pub Option<HashSet<usize>>);
 
 #[derive(Default, Debug, Clone)]
 pub struct Symtable {
     pub vars: IndexMap<usize, Variable>,
-    scopes: Vec<VarScope>,
+    // pub active_scopes: Vec<VarScope>,
```
seanyoung, line 87: remove

src/sema/symtable.rs
```diff
@@ -78,23 +79,25 @@ pub enum VariableUsage {
...
     pub arguments: Vec<Option<usize>>,
     pub returns: Vec<usize>,
+    pub popped_scopes: Vec<(pt::Loc, VarScope)>,
```
seanyoung, line 90: it should be called scopes

src/sema/symtable.rs
```diff
@@ -78,23 +79,25 @@ pub enum VariableUsage {
...
 }
 
 impl Symtable {
     pub fn new() -> Self {
         Symtable {
             vars: IndexMap::new(),
-            scopes: vec![VarScope(HashMap::new(), None)],
+            // active_scopes: vec![VarScope(HashMap::new(), None)],
```
seanyoung, line 97: remove commented out code

src/sema/expression/mod.rs
```diff
@@ -68,6 +70,17 @@ pub struct ExprContext {
...
+}
+
+impl ExprContext {
+    pub fn new() -> Self {
+        Self {
+            active_scopes: vec![VarScope(HashMap::new(), None)],
+            ..Default::default()
+        }
+    }
 }
```
seanyoung, line 84: Having both `fn new()` and `#[derive(Default)]` is confusing. A better solution would be to not `#[derive(Default)]` and do `impl Default for ExprContext`.

Having said all that, how about using the existing `#[derive(Default)]` and do a `enter_scope()` to create the first scope.

src/sema/symtable.rs
```diff
@@ -183,12 +197,14 @@ impl Symtable {
         None
     }
 
-    pub fn enter_scope(&mut self) {
-        self.scopes.push(VarScope(HashMap::new(), None));
+    pub fn enter_scope(&mut self, context: &mut ExprContext) {
+        context.active_scopes.push(VarScope(HashMap::new(), None));
```
seanyoung, line 201: This does not touch `self` so it should be method on `ExprContext`.

src/sema/symtable.rs
```diff
@@ -183,12 +197,14 @@ impl Symtable {
...
     }
 
-    pub fn leave_scope(&mut self) {
-        self.scopes.pop();
+    pub fn leave_scope(&mut self, context: &mut ExprContext, loc: pt::Loc) {
+        if let Some(curr_scope) = context.active_scopes.pop() {
+            self.popped_scopes.push((loc, curr_scope));
+        }
```
seanyoung, line 207: How about making this a method on `ExprContext` and passing in `&mut Symtable`.

seanyoung (changes requested):
Getting there!

src/bin/languageserver/mod.rs
```diff
@@ -1677,6 +1758,89 @@ impl<'a> Builder<'a> {
                 });
 
             self.declarations.extend(decls);
+
+            //////////// Code objects defined within the contract ////////////
```
seanyoung, line 1762:
```suggestion
            // Code objects defined within the contract
```
No ascii art please :)

src/sema/variables.rs
```diff
@@ -781,8 +799,9 @@ pub fn resolve_initializers(
         let mut context = ExprContext {
             file_no,
             contract_no: Some(*contract_no),
-            ..Default::default()
+            ..ExprContext::default()
```
seanyoung, line 802: Why this change - and the same change all over the tree? I don't see how this makes it better.

seanyoung (changes requested):

src/sema/variables.rs
chioni16: I had previously changed all `Default::default()` to `ExprContext::new()` because of the `active_scopes` field. But now we are calling `enter_scope` explicitly, so the previous change was no longer necessary. When going back from `ExprContext::new()`, the lsp changed it to `ExprContext::default()` and not `Default::default()`.

seanyoung (approved):

src/sema/symtable.rs
```diff
@@ -78,23 +79,23 @@ pub enum VariableUsage {
...
-struct VarScope(HashMap<String, usize>, Option<HashSet<usize>>);
+pub struct VarScope(pub HashMap<String, usize>, pub Option<HashSet<usize>>);
 
 #[derive(Default, Debug, Clone)]
 pub struct Symtable {
     pub vars: IndexMap<usize, Variable>,
-    scopes: Vec<VarScope>,
     pub arguments: Vec<Option<usize>>,
     pub returns: Vec<usize>,
+    pub scopes: Vec<(pt::Loc, VarScope)>,
```
xermicus, line 89: `Vec<(pt::Loc, (pub HashMap<String, usize>, pub Option<HashSet<usize>>))>` is a bit hairy. Is there a reason why we no named fields in `VarScope`? Maybe could even add that as an `Option<pt::Loc>` to the variable scope.

xermicus (approved):
LGTM overall, but there are quite a few warnings regarding untested code sections

src/sema/symtable.rs
chioni16: I have added more tests for completion feature. 

I have now added named fields to `VarScope` struct. Because the second field previously present (Option<HashSet<usize>>) wasn't used, I have removed it. @seanyoung

src/sema/builtin_structs.rs
```diff
@@ -4,222 +4,236 @@ use crate::sema::ast::{Namespace, Parameter, StructDecl, StructType, Type};
 use once_cell::sync::Lazy;
 use solang_parser::pt;
 
-static BUILTIN_STRUCTS: Lazy<[StructDecl; 3]> = Lazy::new(|| {
+pub struct BuiltinStructDeclaration {
```
LucasSte, line 7: This struct deserves a doc comment.

LucasSte (approved):

PR #1566, opened 2023-10-16
Update library documentation for clarity and correctness

0xArkonomi:
This Pull Request enhances the clarity and correctness of the documentation regarding libraries in Solidity. It addresses the restrictions applied to libraries, emphasizing their inability to modify the state and the necessity of using "view" or "pure" functions. These changes aim to provide more accurate guidance to users of the documentation.


seanyoung (commented):
I don't know that this is true. 

```solidity
library L {
	function foo(int[2] storage a) internal {
		a[0] = 1;
		a[1] = 2;
	}
}
```
solc also allow this. Am I missing something?

Issue #1450, opened 2023-07-20
Run optimization passes multiple times


xermicus:
We need to change the current model so that optimization passes can run multiple times. Extracted from a discussion in the chat:

```
What I can say about constant folding is that it is not complete.
E.g.
int a = 1;
int b = 4;
int c = a+b;
int d = c+1;

After constant folding you have:
int c = 1+4;
int d = c+1;
Ideally we would run the optmization many times to eliminate all the constants. 
```

My opinion is that any optimization pass should be able to be run as many times as we want. The only thing that should happen that the code might get further optimized with more iterations of the same pass.

Compile time ABI encoding depends on this. Some optimization passes in can also benefit mutually from having multiple pass runs, for example constant folding and CSE. I think we should run optimizations until the CFG doesn't change after running them any more or a limit (maximum runs) is reached.

chioni16:
```
contract A {
    function h() public pure returns (uint) {
        uint a = 1;
        uint b = 4;
        uint c = a+b;
        uint d = c+1;
        return d;
    }
}
```

This is the CFG emitted when the above code is compiled
```
# function A::A::function::h public:true selector:50dd0ebade8d719b nonpayable:true
# params: 
# returns: uint256
block0: # entry
    ty:uint256 %a = uint256 1
    ty:uint256 %b = uint256 4
    ty:uint256 %c = uint256 5
    ty:uint256 %d = uint256 6
    return uint256 6
```

It looks like the constant folding produced the desired output, but unused variables were not removed. 
An explicit optimisation pass that takes care of removing unused variables may be useful here.

xermicus:
Yes, as discussed in the chat; In my opinion the proper way to fix that is to turn unused variable elimination into an optimization pass, and then run all passes passes multiple times until the code converges or a limit is reached.  This is also yet another good example why we want to run all optimization passes multiple times. Thanks!

PR #1569, opened 2023-10-17
Bugfix: Propagate RHS type of resolved assignments

xermicus:
When resolving an assignment, the resolved variable should inherit the RHS type to avoid various issues in later compilation stages (for example with assignment chains).

### DIFF
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -2579,9 +2579,9 @@ pub fn assign_single(
         }
         _ => {
             let left_ty = left.ty();
-            let ty = left_ty.deref_memory();
+            let ty = cfg_right.ty();
 
-            let pos = vartab.temp_anonymous(ty);
+            let pos = vartab.temp_anonymous(&ty);
 
             // Set a subscript in storage bytes needs special handling
             let set_storage_bytes = if let ast::Expression::Subscript { array_ty, .. } = &left {
@@ -2667,7 +2667,7 @@ pub fn assign_single(
 
             Expression::Variable {
                 loc: left.loc(),
-                ty: ty.deref_storage().clone(), // LHS value propagation is always in memory
+                ty,
                 var_no: pos,
             }
         }
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -2667,7 +2667,7 @@ pub fn assign_single(
 
             Expression::Variable {
                 loc: left.loc(),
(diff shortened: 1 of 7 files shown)


src/codegen/expression.rs
```diff
@@ -2667,7 +2667,7 @@ pub fn assign_single(
 
             Expression::Variable {
                 loc: left.loc(),
-                ty: ty.clone(),
+                ty: ty.deref_storage().clone(), // LHS value propagation is always in memory
                 var_no: pos,
```
seanyoung, line 2671: Although this works there is a mismatch between how variable `pos` is used, and how it is created on line 2584.

`--emit cfg --no-constant-folding --no-cse`
```
# function C::C::function::f__uint64 public:true selector:70e8dbcb nonpayable:true
# params: uint64 x
# returns: uint64
block0: # entry
	ty:uint64 %x = (arg #0)
	ty:uint64 storage %temp.11 = %x
	store storage slot(uint256 1) ty:uint64 = %temp.11
	ty:uint64 storage %temp.12 = %temp.11
	store storage slot(uint256 0) ty:uint64 = %temp.12
	return %temp.12
```
See how `%temp.11` and `%temp.12` are used as `uint64` but the variable type is `uint64 storage`.

xermicus: Hmm yeah, I already tried to switch to the RHS type, but that lead to a bunch of codegen test failures. Maybe those are wrong too

seanyoung: I noticed that too, I think the codegen tests will need fixing up

xermicus:
I changed the fix to use the RHS type of the assignment. Which seems to fix a bunch of cases (see codegen tests).

The `FIXME` added to the constant folding codegen test should go away after #1450. Unless there is easy fix for this, I think it deserves a dedicated fix (in the sense of a unused variable pass after codegen / until it converges). Otherwise the problem might only be partially fixed.

src/sema/types.rs
```diff
@@ -2075,6 +2075,14 @@ impl Type {
         }
     }
 
+    /// If the type is StorageRef, get the underlying type
+    pub fn deref_storage(&self) -> &Self {
```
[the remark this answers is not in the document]
xermicus: Good catch, I've removed it, if it is needed in the future can just add it again

seanyoung (approved):

LucasSte (approved):

PR #1572, opened 2023-10-19
Fix slice length

LucasSte:
When converting [this example](https://github.com/solana-developers/program-examples/blob/main/tokens/pda-mint-authority/solang/solidity/pda-mint-authority.sol) to the new Solidity annotation syntax, I uncovered a small bug in slice creation, in which we were not casting the length to `uint64`, leading to an `InvalidLength` error in Solana's runtime.


seanyoung (approved):
Good catch!

xermicus (approved):

PR #1575, opened 2023-10-25
Fix common lsp panics

seanyoung:
These lsp crashes are concurrency issues, so writing tests from them is hard.

If we could hand-craft lsp json-rpc messages then this would be easier, but we only test the extension by controlling vscode.

### DIFF
--- a/src/bin/languageserver/mod.rs
+++ b/src/bin/languageserver/mod.rs
@@ -906,7 +906,7 @@ impl<'a> Builder<'a> {
             }
             ast::Expression::ConstantVariable { loc, ty, contract_no, var_no } => {
                 let (contract, name) = if let Some(contract_no) = contract_no {
-                    let contract = format!("{}.", self.ns.contracts[*contract_no].name);
+                    let contract = format!("{}.", self.ns.contracts[*contract_no].id);
                     let name = &self.ns.contracts[*contract_no].variables[*var_no].name;
                     (contract, name)
                 } else {
@@ -945,7 +945,7 @@ impl<'a> Builder<'a> {
             ast::Expression::StorageVariable { loc, ty, contract_no, var_no } => {
                 let contract = &self.ns.contracts[*contract_no];
                 let name = &contract.variables[*var_no].name;
-                let val = format!("{} {}.{}", ty.to_string(self.ns), contract.name, name);
+                let val = format!("{} {}.{}", ty.to_string(self.ns), contract.id, name);
                 self.hovers.push((
                     loc.file_no(),
                     HoverEntry {
@@ -1083,7 +1083,7 @@ impl<'a> Builder<'a> {
                         msg
                     }).join(", ");
 
-                let contract = fnc.contract_no.map(|contract_no| format!("{}.", self.ns.contracts[contract_no].name)).unwrap_or_default();
+                let contract = fnc.contract_no.map(|contract_no| format!("{}.", self.ns.contracts[contract_no].id)).unwrap_or_default();
 
                 let val = format!("{} {}{}({}) returns ({})\n", fnc.ty, contract, fnc.name, params, rets);
 
@@ -1141,7 +1141,7 @@ impl<'a> Builder<'a> {
                         msg
                     }).join(", ");
 
-                let contract = fnc.contract_no.map(|contract_no| format!("{}.", self.ns.contracts[contract_no].name)).unwrap_or_default();
+                let contract = fnc.contract_no.map(|contract_no| format!("{}.", self.ns.contracts[contract_no].id)).unwrap_or_default();
 
                 let val = format!("{} {}{}({}) returns ({})\n", fnc.ty, contract, fnc.name, params, rets);
 
@@ -1563,7 +1563,7 @@ impl<'a> Builder<'a> {
                         stop: base.loc.exclusive_end(),
                         val: make_code_block(format!(
                             "contract {}",
-                            self.ns.contracts[base.contract_no].name
+                            self.ns.contracts[base.contract_no].id
                         )),
                     },
                 ));
@@ -1590,8 +1590,8 @@ impl<'a> Builder<'a> {
             self.hovers.push((
                 file_no,
                 HoverEntry {
-                    start: contract.loc.start(),
-                    stop: contract.loc.start() + contract.name.len(),
+                    start: contract.id.loc.start(),
+                    stop: contract.id.loc.exclusive_end(),
                     val: render(&contract.tags[..]),
                 },
             ));
@@ -1602,7 +1602,7 @@ impl<'a> Builder<'a> {
             };
 
             self.definitions
-                .insert(cdi.clone(), loc_to_range(&contract.loc, file));
+                .insert(cdi.clone(), loc_to_range(&contract.id.loc, file));
 
             let impls = contract
                 .functions
--- a/src/bin/languageserver/mod.rs
+++ b/src/bin/languageserver/mod.rs
@@ -334,16 +334,17 @@ impl SolangServer {
         let files = self.files.lock().await;
         if let Some(cache) = files.caches.get(&path) {
             let f = &cache.file;
-            let offset = f.get_offset(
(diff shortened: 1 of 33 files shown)


seanyoung:
Cc @chioni16

src/bin/languageserver/mod.rs
```diff
@@ -1761,7 +1765,9 @@ impl<'a> Builder<'a> {
                         .filter(|h| h.0 == i)
                         .map(|(_, i)| {
                             let mut i = i.clone();
-                            i.val.def_path = defs_to_files[&i.val.def_type].clone();
+                            if let Some(path) = defs_to_files.get(&i.val.def_type) {
+                                i.val.def_path = path.clone();
+                            }
                             i
```
chioni16, line 1771: There's one more place where `defs_to_files` is used. https://github.com/hyperledger/solang/blob/main/src/bin/languageserver/mod.rs#L1722

As part of https://github.com/hyperledger/solang/pull/1560, 
I have made similar changes (for example, https://github.com/hyperledger/solang/pull/1560/files#diff-bf2200d9cb4eee660f8ff53c8416979a1f0ebe5d5856609e7d7f46fe0c8e9846R1935).

src/bin/languageserver/mod.rs
```diff
@@ -1990,24 +1998,25 @@ impl LanguageServer for SolangServer {
...
+                    {
+                        let range = get_range_exclusive(hover.start, hover.stop, &cache.file);
+
+                        return Ok(Some(Hover {
+                            contents: HoverContents::Scalar(MarkedString::from_markdown(
+                                hover.val.to_string(),
+                            )),
+                            range: Some(range),
+                        }));
+                    }
```
LucasSte, line 2019: The `Ok(None)` code path is never tested (This is why we have this alert from the bot).

src/bin/languageserver/mod.rs
```diff
@@ -334,16 +334,17 @@ impl SolangServer {
...
-            {
-                return Ok(Some(reference.val.clone()));
+            ) {
+                if let Some(reference) = cache
+                    .references
+                    .find(offset, offset + 1)
+                    .min_by(|a, b| (a.stop - a.start).cmp(&(b.stop - b.start)))
+                {
+                    return Ok(Some(reference.val.clone()));
+                }
```
LucasSte, line 347: The `Ok(None)` code path is never tested (This is why we have this alert from the bot).

src/bin/languageserver/mod.rs
chioni16: Hello @LucasSte ! 
Does the bot only consider tests written in rust? Because currently we test the language server code primarily through tests written in typescript.

seanyoung: The bot considers all code paths. Ideally we would have a vscode test which checks this path.

xermicus (approved):

src/bin/languageserver/mod.rs
seanyoung: This is tested now.

src/bin/languageserver/mod.rs
seanyoung: @chioni16 I tried adding a test to go through this code path, and it does not work. If you request the references for `}` then I get a range of the whole file, see https://github.com/hyperledger/solang/pull/1575/files#diff-5710c64b02f305009aefd0e96860a2a6281dc9a411a25d6bb4fc6e7ef1e873ddR439

Any ideas what is going on here?

chioni16: This is because we look for the smallest range that encompasses the location where the request to get the references was made - https://github.com/hyperledger/solang/blob/main/src/bin/languageserver/mod.rs#L341-L347

But as of now, the "ranges" that I mentioned above represent the whole structs, enums etc with all their fields, variants etc. Ideally, this should just represent the name of the struct/enum. 
This can be changed once the "Identifier" locations are added to the parse tree, which is something that is missing at the moment. 

regarding testing the `Ok(None)` case, I believe it should be possible by invoking `GotoReference` at the end of the file, outside all the structs, enums, contracts. 

For example, 
Screenshot 2023-10-29 at 10 02 20 AM

@seanyoung

seanyoung: > But as of now, the "ranges" that I mentioned above represent the whole structs, enums etc with all their fields, variants etc. Ideally, this should just represent the name of the struct/enum.

I don't understand that means that the entire file range is returned. Since the ast has not been fixed yet, the precise location of the enum/struct identifier is not used, but the whole definition in including the fields etc. 

The entire file range is a bug.

> regarding testing the Ok(None) case, I believe it should be possible by invoking GotoReference at the end of the file, outside all the structs, enums, contracts.

When I do go to references in the same location as the test case in vscode, I also get the same "No references found" 

Please debug, thank you

chioni16: > I don't understand that means that the entire file range is returned.
The position (21, 6) (https://github.com/hyperledger/solang/pull/1575/files#diff-5710c64b02f305009aefd0e96860a2a6281dc9a411a25d6bb4fc6e7ef1e873ddR431) for which `GotoReference` request is made - doesn't coincide with the closing brace `}` of `enum value`. The closing brace is at position (21, 5). As such, the position in the test case (21, 6) is right next to the `}` but is not considered to be part of the `enum value`. 
So, the smallest enclosing scope here is the `contract deck`, **which happens to cover the entire file**. 
So, when the request is made for position (21, 6), it's considered to be part of `contract deck`, due to the current limitations in AST mentioned in the previous comment. 
So, the current code looks for the uses of this contract in the file. There are no uses found in the file. But the request also sends `true` for  the `include_declaration` field in the request. As a result, I return the location range for the contract definition, which again, happens to cover the whole file range. 

I have attached a screenshot that shows what happens when the request is made. I have highlighted the relevant parts. 

Screenshot 2023-10-29 at 3 08 00 PM
1. The position for which the request was made.  `(21, 6)`
2. What it was identified as.  `Contract(0)` aka `contract deck`
3. Return the location of the references (which is the contract definition in our case) `(0,0) - (44, 1)`

Even when I added another function at the end of the file, after the contract, repeating the above steps gave me the same response. This shows that the returned location range has very little to do with the file range. This is just an unfortunate coincidence. 

Making the request for the position (21, 5) ( which is the position of `}` of `enum value`) gives different results. 
(I have attached a similar screenshot as above, sorry if the text is too small to read) 

Screenshot 2023-10-29 at 3 30 02 PM

1. The position for which the request was made.  `(21, 5)`
2. What it was identified as.  `Enum(1)` aka `enum value`
3. Return the location of the enum definition (references include definitions as mentioned above)  `(7,4) - (21, 5)` 
4. Return the location of reference `(23, 8) - (23, 13)`

> When I do go to references in the same location as the test case in vscode, I also get the same "No references found"
I will have to look into this. It'd be helpful if you can share a screenshot of the position from where the `GotoReference` request was made. One more thing I'd like to add here is - one position right / left of the intended position can give different results as seen above. Also, when making `GotoReference` request for a position where no text is present (for example, (21, 6)), I have seen that it's hard to make the request exactly for that position. This is something I have seen when debugging this. When I think I have placed my cursor at (21, 6) and made a request, I have seen (21, 5) in the logs. Again, this should not matter when we specify the location explicitly as we have done in the test case.

chioni16: TLDR; the returned location range doesn't correspond to the file but to `contract deck`, which in this case, happens to cover the whole file.

seanyoung: I see, you are right. That means that go to references always defaults the current contract, which is awful.

I've fix the ast to have a loc for the contract name.

LucasSte (approved):
There many warnings of lines not covered by tests in this PR. I'm overlooking them by now, but we should be more careful with them.

seanyoung:
> There many warnings of lines not covered by tests in this PR. I'm overlooking them by now, but we should be more careful with them.

I agree. I have added some tests for those.

PR #1579, opened 2023-10-29
Update tests to solc v0.8.22 and fix a few tests

seanyoung:
- Do not modify `ns.diagnostics` directly - ever. We may want to resolve an expression and drop all the diagnostics for it.
- Ensure all variable errors are surfaced
- Update solc testdata to `v0.8.22`


seanyoung:
I am not sure why the coverage bot is complaining. I get an overall failure but in all the files the coverage has gone up or stayed the same.

xermicus (approved):

LucasSte:
> I am not sure why the coverage bot is complaining. I get an overall failure but in all the files the coverage has gone up or stayed the same.

In `src/sema/tags.rs`, lines 101-104 were previously covered by test, but are not covered anymore. Please, check the indirect changes in the coverage report.

seanyoung:
> In `src/sema/tags.rs`, lines 101-104 were previously covered by test, but are not covered anymore. Please, check the indirect changes in the coverage report.

Good point. I've added a bunch of test and now the coverage does not decrease (in fact, it increases). 

There are some areas which are still uncovered, but I have not decided how to fix that. For example for the dot tests I think we need a solidity file with all the syntax in it (or just a minimal set of files). That's not in this PR.

LucasSte (approved):

PR #1580, opened 2023-10-30
Add missing Identifier locations to AST

chioni16:
Adds missing Identifier locations to AST

### DIFF
--- a/src/bin/languageserver/mod.rs
+++ b/src/bin/languageserver/mod.rs
@@ -1329,14 +1329,14 @@ impl<'a> Builder<'a> {
                 ));
             }
         }
-
-        let file_no = field.loc.file_no();
+        let loc = field.id.as_ref().map(|id| &id.loc).unwrap_or(&field.loc);
+        let file_no = loc.file_no();
         let file = &self.ns.files[file_no];
         self.hovers.push((
             file_no,
             HoverEntry {
-                start: field.loc.start(),
-                stop: field.loc.exclusive_end(),
+                start: loc.start(),
+                stop: loc.exclusive_end(),
                 val: make_code_block(format!(
                     "{} {}",
                     field.ty.to_string(self.ns),
@@ -1353,7 +1353,7 @@ impl<'a> Builder<'a> {
             ),
         };
         self.definitions
-            .insert(di.clone(), loc_to_range(&field.loc, file));
+            .insert(di.clone(), loc_to_range(loc, file));
         if let Some(dt) = get_type_definition(&field.ty) {
             self.types.insert(di, dt.into());
         }
--- a/src/bin/languageserver/mod.rs
+++ b/src/bin/languageserver/mod.rs
@@ -1373,7 +1373,7 @@ impl<'a> Builder<'a> {
                         stop: loc.exclusive_end(),
                         val: make_code_block(format!(
                             "enum {}.{} {}",
-                            enum_decl.name, nam, discriminant
+                            enum_decl.id, nam, discriminant
                         )),
                     },
                 ));
@@ -1388,13 +1388,13 @@ impl<'a> Builder<'a> {
                 self.types.insert(di, dt.into());
             }
 
-            let file_no = enum_decl.loc.file_no();
+            let file_no = enum_decl.id.loc.file_no();
             let file = &self.ns.files[file_no];
             self.hovers.push((
                 file_no,
                 HoverEntry {
-                    start: enum_decl.loc.start(),
-                    stop: enum_decl.loc.start() + enum_decl.name.len(),
+                    start: enum_decl.id.loc.start(),
+                    stop: enum_decl.id.loc.exclusive_end(),
                     val: render(&enum_decl.tags[..]),
                 },
             ));
@@ -1403,7 +1403,7 @@ impl<'a> Builder<'a> {
                     def_path: file.path.clone(),
                     def_type: DefinitionType::Enum(ei),
                 },
-                loc_to_range(&enum_decl.loc, file),
+                loc_to_range(&enum_decl.id.loc, file),
             );
         }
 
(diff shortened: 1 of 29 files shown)


seanyoung:
An internal function call can have too parts: base contract and function name.

```
abstract contract a {
	function foo() internal virtual returns (int) {
		return 1;
	}
}

contract b is a {
	function foo() internal override returns (int) {
		return 2;
	}

	function bar() public returns (int x) {
		x = a.foo();
	}
}
```

Here `a.foo` is `ast::Expression::InternalFunction`. In fact, you can also import a function from another file via an import object. So, the `id` field should be an `IdentifierPath`.

chioni16:
> An internal function call can have too parts: base contract and function name.
> 
> ```
> abstract contract a {
> 	function foo() internal virtual returns (int) {
> 		return 1;
> 	}
> }
> 
> contract b is a {
> 	function foo() internal override returns (int) {
> 		return 2;
> 	}
> 
> 	function bar() public returns (int x) {
> 		x = a.foo();
> 	}
> }
> ```
> 
> Here `a.foo` is `ast::Expression::InternalFunction`. In fact, you can also import a function from another file via an import object. So, the `id` field should be an `IdentifierPath`.

Do we have something that can also hold the types of individual parts of the `IdentifierPath`? 
This is because when the `rename` functionality is invoked on `contract a` in your example, we would also like to rename the `a` in `a.foo()`, which is currently not possible. 
One more example that I can think of is enum literals of the form `enum.variant`. We would like to treat this as two separate parts `enum` and `variant`. But currently the whole thing evaluates to `NumLiteral`, which is not ideal when `rename` / `references` / any other functionality is invoked.  
@seanyoung

seanyoung:
> > Here `a.foo` is `ast::Expression::InternalFunction`. In fact, you can also import a function from another file via an import object. So, the `id` field should be an `IdentifierPath`.
> 
> Do we have something that can also hold the types of individual parts of the `IdentifierPath`? This is because when the `rename` functionality is invoked on `contract a` in your example, we would also like to rename the `a` in `a.foo()`, which is currently not possible. 

We don't have that right now, but in this case you can assume that the last part if the function name.

> One more example that I can think of is enum literals of the form `enum.variant`. We would like to treat this as two separate parts `enum` and `variant`.

enum variants are not stored as `IdentifierPath`.

> But currently the whole thing evaluates to `NumLiteral`, which is not ideal when `rename` / `references` / any other functionality is invoked.

This is a different problem than function names, 

> I'm thinking that part of this type resolution will happen when the language server functionality is invoked/ we are preparing things for the language server. And some type information may not be possible to be derived in the language server and hence we will have to make some changes in `sema`. These are just my initial thoughts though.

We'll have to deal with each individual case here.

src/sema/ast.rs
```diff
@@ -838,8 +838,9 @@ pub enum Expression {
     },
     StructLiteral {
         loc: pt::Loc,
+        id: pt::IdentifierPath,
         ty: Type,
-        values: Vec<Expression>,
+        values: Vec<(Option<pt::Loc>, Expression)>,
```
seanyoung, line 843: Needs a comment explaining that the `pt::Loc` is the location of the field name, if the literal was specified with field names.

src/sema/ast.rs
```diff
@@ -14,7 +14,7 @@ use num_rational::BigRational;
 use once_cell::unsync::OnceCell;
 pub use solang_parser::diagnostics::*;
 use solang_parser::pt;
-use solang_parser::pt::{CodeLocation, FunctionTy, OptionalCodeLocation};
+use solang_parser::pt::{CodeLocation, FunctionTy, Identifier, OptionalCodeLocation};
```
seanyoung, line 17: The problem with this is, now we have both `pt::Identifier` in this file and `Identifier`. Please make the file consistently use one or the other.

src/sema/ast.rs
seanyoung: It might make more sense to have `Vec<(Option<pt::Identifier>, Expression)>`, that is much clearer

seanyoung (changes requested):

src/sema/ast.rs
```diff
@@ -402,7 +402,7 @@ impl FunctionAttributes for Function {
 impl Function {
     pub fn new(
         loc: pt::Loc,
-        name: String,
+        name: pt::Identifier,
```
seanyoung, line 405: This function looks better if you name this `id: pt::Identifier`

seanyoung (changes requested):

seanyoung (approved):

src/bin/languageserver/mod.rs
```diff
@@ -1135,7 +1153,7 @@ impl<'a> Builder<'a> {
 
                 let contract = fnc.contract_no.map(|contract_no| format!("{}.", self.ns.contracts[contract_no].id)).unwrap_or_default();
 
-                let val = format!("{} {}{}({}) returns ({})\n", fnc.ty, contract, fnc.name, params, rets);
+                let val = format!("{} {}{}({}) returns ({})\n", fnc.ty, contract, fnc.id, params, rets);
```
LucasSte, line 1156: Can you add a test for this line?

src/sema/ast.rs
```diff
@@ -838,8 +838,10 @@ pub enum Expression {
     },
     StructLiteral {
         loc: pt::Loc,
+        id: pt::IdentifierPath,
         ty: Type,
-        values: Vec<Expression>,
+        // pt::Identifier represents the field name
```
xermicus, line 843: I think this should be a doc-comment, it's quite relevant information that should show up in help texts from the analyzer when using this Enum

xermicus (approved):

LucasSte (approved):

PR #1586, opened 2023-11-01
Match solc diagnostic levels

seanyoung:
The following errors are warnings in solc

 - unreachable code should be warning
 - function types with parameter/return names should be warning
 - unknown assembly flags should be warnings

### DIFF
--- a/src/sema/namespace.rs
+++ b/src/sema/namespace.rs
@@ -1134,7 +1134,7 @@ impl Namespace {
                         .into_iter()
                         .map(|p| {
                             if let Some(name) = p.id {
-                                diagnostics.push(Diagnostic::error(
+                                diagnostics.push(Diagnostic::warning(
                                     name.loc,
                                     "function type parameters cannot be named".to_string(),
                                 ));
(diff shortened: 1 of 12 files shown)


src/sema/namespace.rs
```diff
@@ -1134,7 +1134,7 @@ impl Namespace {
                         .into_iter()
                         .map(|p| {
                             if let Some(name) = p.id {
-                                diagnostics.push(Diagnostic::error(
+                                diagnostics.push(Diagnostic::warning(
```
LucasSte, line 1137: Should success still be false here even if this is not an error anymore?

src/sema/namespace.rs
```diff
@@ -1148,7 +1148,7 @@ impl Namespace {
...
```
LucasSte, line 1151: Should success still be false here even if this is not an error anymore?

src/sema/namespace.rs
seanyoung: Good point, fixed

src/sema/namespace.rs
seanyoung: Good point, fixed

LucasSte (approved):

xermicus (approved):

Issue #1534, opened 2023-09-15
Parse and resolve fails on overloaded functions depending on order of declaration


bkushigian:
**Describe the bug**
Overloaded function resolution can trigger a failure depending on order of function declarations. I have a minimal reproducible version of this bug [here](https://github.com/bkushigian/SolangParse4Debugging). To reproduce, run `cargo run -- testcase/overloaded_order_matters/*.sol`.

For instance, the following code works fine for solc and solang's parse_and_resolve:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract Overloaded {
    // Version 1
    function f(uint n) internal pure returns (uint) {
        return n;
    }

    // Version 2
    function f(uint n, uint x) internal pure returns (uint) {
        return n + x;
    }

    // Version 3
    function f(int n, uint x) internal pure returns (int) {
        uint result = f(uint256(n), x);
        return n > 0 ? int(result) : -int(result);
    }
}

```

However, parse_and_resolve fails to handle this case, where the final two functions are reversed:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract Overloaded {
    // Version 1
    function f(uint n) internal pure returns (uint) {
        return n;
    }

    // Version 3
    function f(int n, uint x) internal pure returns (int) {
        uint result = f(uint256(n), x);
        return n > 0 ? int(result) : -int(result);
    }

    // Version 2
    function f(uint n, uint x) internal pure returns (uint) {
        return n + x;
    }
}
```

Running solang here causes the following error:

```
=====  Error: testcase/overload_order_matters/Overloaded3.sol  =====
error: function expects 1 arguments, 2 provided
   ┌─ /Users/benku/solang_parse/testcase/overload_order_matters/Overloaded3.sol:12:23
   │
12 │         uint result = f(uint256(n), x);
   │                       ^^^^^^^^^^^^^^^^
```

I tried all 6 possible orders of function declarations and this is the only one that fails 🤔. I also wasn't able to make this trigger with only two functions, so I _think_ this is a minimal test case?

seanyoung:
This is a bug in our overloaded function resolution. It really needs to be re-written.

#707 is related

seanyoung:
Probably need to wait for #1517 to be merged before we can tackle this

Issue #707, opened 2022-03-23
compiler message for overloaded functions is not great


seanyoung:
Given this solidity:
```
contract C {
	function foo(int a) public {}
	function foo(int a, int b) public {}
	function foo(int a, int b, int c) public {}
	function bar() public {
		foo(true);
	}
	function bar2() public {
		foo({a:true});
	}
}
```
The error message you get is:

```
/home/sean/git/solang/overloaded.sol:6:3-12: error: cannot find overloaded function which matches signature
Line 6:
	foo(true);
	^^^^^^^^^
/home/sean/git/solang/overloaded.sol:9:3-16: error: cannot find overloaded function which matches signature
Line 9:
	foo({a:true});
	^^^^^^^^^^^^^
```
That is not useful. The error message should list the overloaded functions which were tried, and the error message for each of them.

The same handling should be used for overloaded constructors when deploying a new contract.

devratapuri:
Is this still open ?

seanyoung:
yes, the errors are still like above

PR #1592, opened 2023-11-10
Improve overloaded function call diagnostics

seanyoung:
From Solidity 0.6 onwards, when an overloaded function/event resolves to multiple candidates, this is an error. In earlier versions, the first result is used. So, use the `pragma solidity` version to decide whether to error or not.

Fixes: https://github.com/hyperledger/solang/issues/1534
Fixes: https://github.com/hyperledger/solang/issues/707


src/sema/expression/constructor.rs
```diff
@@ -108,84 +108,99 @@ pub fn match_constructor_to_args(
...
+
+            Err(())
         }
+        1 => Ok(resolved_calls.remove(0)),
         _ => {
-            diagnostics.push(Diagnostic::error(
+            diagnostics.push(Diagnostic::error_with_notes(
                 *loc,
-                "cannot find overloaded constructor which matches signature".to_string(),
+                "constructor can be resolved to multiple functions".into(),
```
LucasSte, line 188: There is no test for this.

src/sema/expression/constructor.rs
```diff
@@ -344,86 +386,107 @@ pub fn constructor_named_args(
...
+            // check if arguments can be implicitly casted
+            for i in 0..params_len {
+                let param = ns.functions[*function_no].params[i].clone();
+
+                let arg = match arguments.get(param.name_as_str()) {
+                    Some(a) => a,
+                    None => {
+                        candidate_diagnostics.push(Diagnostic::cast_error_with_note(
+                            *loc,
+                            format!("missing argument '{}' to constructor", param.name_as_str()),
```
LucasSte, line 399: No test for this message.

src/sema/expression/constructor.rs
```diff
@@ -344,86 +386,107 @@ pub fn constructor_named_args(
...
         }),
-        0 | 1 => {
-            diagnostics.extend(errors);
+        0 => {
+            diagnostics.extend(call_diagnostics);
+
+            if function_nos.is_empty() {
+                diagnostics.push(Diagnostic::error(
+                    *loc,
+                    "cannot find matching constructor".into(),
```
LucasSte, line 464: No test.

src/sema/expression/function_call.rs
```diff
@@ -2709,67 +2817,91 @@ fn contract_call_named_args(
...
-                    ns.contracts[external_contract_no].id, func_name.name
-                ),
-            ));
+            diagnostics.extend(call_diagnostics);
+
+            if name_matches.is_empty() {
+                diagnostics.push(Diagnostic::error(
+                    *loc,
+                    format!(
+                        "contract '{}' does not have function '{}'",
```
LucasSte, line 2877: Code path not tested.

src/sema/expression/function_call.rs
```diff
@@ -2805,106 +2937,140 @@ fn contract_call_pos_args(
...
-        }
 
-        if matches {
-            let resolved_call = contract_call_match(
+            if candidate_diagnostics.any_errors() {
+                if name_matches.len() != 1 {
+                    // will be de-duped
+                    candidate_diagnostics.push(Diagnostic::error(
+                        *loc,
+                        "cannot find overloaded function which matches signature".into(),
```
LucasSte, line 2986: Code path not tested.

src/sema/expression/function_call.rs
```diff
@@ -2805,106 +2937,140 @@ fn contract_call_pos_args(
...
-        diagnostics.extend(errors);
-    } else if name_matches.len() != 1 {
-        diagnostics.push(Diagnostic::error(
-            *loc,
-            "cannot find overloaded function which matches signature".to_string(),
-        ));
-    }
+            diagnostics.push(Diagnostic::error_with_notes(
+                *loc,
+                "function call can be resolved to multiple functions".into(),
```
LucasSte, line 3057: This error message is not tested.

src/sema/pragma.rs
```diff
@@ -0,0 +1,264 @@
...
+                    res.push(v);
+                }
+
+                ns.pragmas.push(ast::Pragma::SolidityVersion {
+                    loc: *loc,
+                    versions: res,
+                });
+            }
+        }
+        pt::PragmaDirective::Identifier { .. } => (),
```
LucasSte, line 51: Is this case supposed to be an error?

seanyoung:
@LucasSte Thanks for your "there is no test for this" comments. I am aware of this. Not exactly a useful review.

src/sema/pragma.rs
seanyoung: no

xermicus (commented):
LGTM overall. How / where are we going to document this?

src/sema/pragma.rs
```diff
@@ -0,0 +1,264 @@
...
+            *loc,
+            "experimental solidity features are not supported".to_string(),
+        ));
+    } else if name == "abicoder" && value == "v2" {
+        ns.diagnostics.push(ast::Diagnostic::debug(
+            *loc,
+            "pragma 'abicoder' with value 'v2' is ignored".to_string(),
+        ));
+    } else {
+        ns.diagnostics.push(ast::Diagnostic::warning(
```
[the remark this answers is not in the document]
seanyoung: oh nice, that reduces the failed ethereum solidity tests even more :rocket:

seanyoung:
> How / where are we going to document this?

I had not thought about this. Do you have any suggestions?

xermicus:
> > How / where are we going to document this?
> 
> I had not thought about this. Do you have any suggestions?

I guess it depends on the general direction we want to go with this. Currently, we just say in the docs that we ignore the version pragma, which is kind of no longer true after this PR.

xermicus (commented):
If I got the following source:

```solidity
// a.sol
pragma solidity ^0.7;

import "b.sol";

contract C {
        function c() public payable {
                require(msg.value > SERVICE_FEE); 
        }
}
```

```solidity
// b.sol
pragma solidity ^0.8;

uint128 constant SERVICE_FEE = 123;
```

Shouldn't Solang then refuse to compile that?

xermicus:
Maybe it's a sensible strategy to have each Solang release being compatible with a specific  `solc` release. I think that would clarify a lot of things.

seanyoung:
> I guess it depends on the general direction we want to go with this. Currently, we just say in the docs that we ignore the version pragma, which is kind of no longer true after this PR.

There are some differences between versions, but nothing huge. The idea here is that we switch behaviour based on the solidity pragma, so that - ultimately - we can compile any Solidity. A single solidity compiler for all versions.

seanyoung:
> If I got the following source:
> 
> ```solidity
> // a.sol
> pragma solidity ^0.7;
> 
> import "b.sol";
> 
> contract C {
>         function c() public payable {
>                 require(msg.value > SERVICE_FEE); 
>         }
> }
> ```
> 
> ```solidity
> // b.sol
> pragma solidity ^0.8;
> 
> uint128 constant SERVICE_FEE = 123;
> ```
> 
> Shouldn't Solang then refuse to compile that?

No, why? Just because solc only support one version does not mean we have to be limited to one version too.

seanyoung:
> Maybe it's a sensible strategy to have each Solang release being compatible with a specific `solc` release. I think that would clarify a lot of things.

That would mean supporting lots of different versions of solang, one for each version of Solidity. That's too much work.

src/sema/expression/function_call.rs
```diff
@@ -2805,106 +2937,141 @@ fn contract_call_pos_args(
...
 
-    let mut errors = Diagnostics::default();
+    let mut call_diagnostics = Diagnostics::default();
+    let mut resolved_calls = Vec::new();
+
+    // try to resolve the arguments, give up if there are any errors
+    if args.iter().fold(false, |acc, arg| {
+        acc | expression(arg, context, ns, symtable, diagnostics, ResolveTo::Unknown).is_err()
+    }) {
+        return Err(());
```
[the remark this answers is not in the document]
seanyoung: The test coverage is already much better than what it was. I have added a test case.

xermicus:
I agree that'd be a lot of work.

What about a warning if there are conflicting version pragma found in the source. This way, the user at least gets warned that he is trying to compile "incompatible" sources.

tests/contract_testcases/evm/builtin_overloaded.sol
```diff
@@ -0,0 +1,17 @@
...
+	}
+
+	function test2() public {
+		bs.readInt8(1, "foo");
+	}
+}
+
+// ---- Expect: diagnostics ----
+// error: 3:20-34: cannot call function in constant expression
+// error: 6:15-20: implicit conversion to uint32 from bytes3 not allowed
```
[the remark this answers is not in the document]
seanyoung: All our conversion errors and warning say "to X from Y". I am not changing this in this PR.

seanyoung:
> I agree that'd be a lot of work.
> 
> What about a warning if there are conflicting version pragma found in the source. This way, the user at least gets warned that he is trying to compile "incompatible" sources.

The problem is, how do you determine that two pragmas are conflicting? We might have in one file:
`pragma solidity 0.5 - 0.5.16;`
and another:
`pragma solidity <=0.6;`

In the solidity pragma parsing, all we have right now in solang is what the highest supported Solidity version. In this case we have `0.5.16`  and `0.6`. Is that a conflict? If you were to compile with solc 0.5.15 then it works fine, but if you compile with solc 0.6.1 then it is not fine.

I think this is something for another PR.

xermicus:
By conflicting I meant there is no suitable compiler, (e.g. `pragma solidity ^0.5.16;` and `pragma solidity ^0.6.16`)

tests/contract_testcases/evm/overloaded.sol
```diff
@@ -0,0 +1,112 @@
...
+// error: 6:3-23: cannot find overloaded function which matches signature
+// 	note 2:2-29: candidate function
+// error: 6:3-23: cannot find overloaded function which matches signature
+// 	note 3:2-39: candidate function
+// error: 6:3-23: cannot find overloaded function which matches signature
+// 	note 4:2-39: candidate function
+// error: 6:3-23: function cannot be called with named arguments as 1 of its parameters do not have names
+// 	note 2:2-26: definition of foo
+// 	note 2:2-29: candidate function
+// error: 6:11-15: conversion from bool to int256 not possible
```
LucasSte, line 56: > All our conversion errors and warning say "to X from Y".

@seanyoung This is not true.

xermicus (approved):
> I don't understand why that is relevant to this PR. Is there some sort of regression?

No, it's not a regression. I still think it's relevant. As the PR description says:

> From Solidity 0.6 onwards, when an overloaded function/event resolves to multiple candidates, this is an error. In earlier versions, the first result is used. So, use the pragma solidity version to decide whether to error or not.

Which means Solang will _silently_ compile identical source codes differently, depending on the version pragma. I just thought it would be nice to inform our users when this happens. It might not be what they expected.  I _think_ it would be nice to emit some warnings if the compiler

- Encounters `version` pragmas that are mutually exclusive (meaning there would be no `solc` version that can compile this)
- Is compiling the same code differently due to different version pragmas

tests/contract_testcases/evm/overloaded.sol
seanyoung: @LucasSte as always, pedantic and not constructive.

xermicus:
Sure, can be done later.

> warning: in later versions of solidity constructor visibility is ignored but if you remove it your contract will no longer build due to the pragma solidity 0.5;

No, that wouldn't be useful. What I meant was:

```
pragma solidity 0.6;

contract Fibonacci {
    constructor() {}
}
```

This should produce a warning like follows:

`Warning: Solidity 0.6 requires constructors to have a visibility.`

Which would remind the user that he is doing something that was not intended in Ethereum Solidity but works anyways in Solang.

seanyoung:
It does make sense. Also, inth

After some more testing I've re-jigged the code to match the actual behaviour better, and also better error messages which shows the actual change in behaviour between versions which helps the developer.

seanyoung:
let me know what you think

xermicus:
Great! 

> make it clear to the developer what has changed and how it affects their code - in case there are any potential pitfalls.

Yeah this is exactly what I meant :)

PR #1594, opened 2023-11-14
Fix broken build

seanyoung:
For some reason, cargo is not pick the latest version of ethers-core. This breaks the build. Add a line to Cargo.toml fixes the problem.


xermicus (approved):

LucasSte (approved):

seanyoung:
Fixed by https://github.com/hyperledger/solang/pull/1595

PR #1596, opened 2023-11-16
Move LoopScopes into ExprContext

seanyoung:
This simplifies the code, and makes it to use ExprContext for other purposes (for example variable scoping).


seanyoung:
Merging for #1560

Issue #1600, opened 2023-11-25
`_;` gets parsed as an `Expression::Variable`


alexfertel:
### Problem

tl;dr `_;` gets parsed as an `Expression::Variable`, which means that when the parse tree is formatted using `Display`, it gets emitted as `_`, without the trailing semicolon. Since https://github.com/ethereum/solidity/releases/tag/v0.4.0, `_` must be followed by a `;`.

### Repro

Given this contract:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.0;

contract Foo {
    modifier bar() {
        _;
    }

    function baz() external bar {}
}
```

The formatted representation of the parse tree is:

```solidity
pragma solidity 0.8.0;
contract Foo {modifier bar() {_} function baz() external bar {}}
```

### Solution

1. Parse it as a `Statement::VariableDefinition`, like so:

```rust
                let variable = Statement::VariableDefinition(
                    variable_loc,
                    VariableDeclaration {
                        loc: variable_loc,
                        ty: Expression::Variable(Identifier {
                            loc: variable_loc,
                            name: "_".to_owned(),
                        }),
                        storage: None,
                        name: None,
                    },
                    None,
                );
```

since the `Display` implementation of this node does append a `;`.

2. Add a new `Statement::Placeholder` node, which represents a `_;` statement.
3. Special case `_` in the display implementation.

seanyoung:
The parser is correct here: it cannot know that `_` is a placeholder statement. In regular functions, you can have a variable called `_`. The problem here is that the code formatter does follow an expression statement with a `;`, fixed in #1601

alexfertel:
>  In regular functions, you can have a variable called _.

I tried declaring one in Remix, but couldn't:

![Screenshot 2023-11-26 at 11 01 16](https://github.com/hyperledger/solang/assets/22298999/ebfb82ba-7c82-4c00-a350-cb25f182f9fa)
![Screenshot 2023-11-26 at 11 00 49](https://github.com/hyperledger/solang/assets/22298999/319244e8-3d36-409d-a1d5-f4d86a628e58)

You can see on the second screenshot that `_` is a reserved name, so I don't think you can have variables named like that.

> This is determined during semantic analysis.

Could you expand on this? Do you mean that the semantic analysis phase is what determines if `_` is a placeholder statement or not?

> The problem here is that the code formatter does follow an expression statement with a ;, fixed in https://github.com/hyperledger/solang/pull/1601

[...]

PR #1601, opened 2023-11-26
An expression statement should be followed by ;

seanyoung:
Fixes https://github.com/hyperledger/solang/issues/1600

### DIFF
--- a/solang-parser/src/helpers/fmt.rs
+++ b/solang-parser/src/helpers/fmt.rs
@@ -932,7 +932,10 @@ impl Display for pt::Statement {
                 f.write_str(") ")?;
                 block.fmt(f)
             }
-            Self::Expression(_, expr) => expr.fmt(f),
+            Self::Expression(_, expr) => {
+                expr.fmt(f)?;
+                f.write_char(';')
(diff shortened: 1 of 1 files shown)


seanyoung:
The coverage report says this is not tested, but it is. It works when I run `cargo llvm-cov --all-features --workspace` locally, even if I remove the temporary files. I would not be surprised if codecov cannot deal with crates properly.

seanyoung:
> I feel like it's becoming inevitable to honor solc version `pragma`s

Actually https://github.com/hyperledger/solang/pull/1592 switches behaviour based on the version `pragma`

LucasSte:
> The coverage report says this is not tested, but it is. It works when I run `cargo llvm-cov --all-features --workspace` locally, even if I remove the temporary files. I would not be surprised if codecov cannot deal with crates properly.

I think the `cargo llvm-cov` is misconfigured. Solana's repository consists of multiple crates and codecov works properly. Please, note that codecov is just a tool to visualize and store the coverage reports. If it is showing wrong results, it is because we are sending an incorrect report to it.

PR #1604, opened 2023-11-30
No version of Solidity supports octal numbers

seanyoung:


tests/contract_testcases/evm/octal.sol
```diff
@@ -0,0 +1,18 @@
+contract C {
+	int public test1 = 00;
+	int public test2 = 0;
+	int public test3 = 01e1;
+	int public test4 = 09.1 * 10;
+	int public test5 = -00;
+	int public test6 = -0;
+	int public test7 = -01e1;
+	int public test8 = -09.1 * 10;
```
LucasSte, line 9: Why are these numbers called octal?
If I am understanding this correctly, octals can only have digits between 0 and 7.
https://en.wikipedia.org/wiki/Octal

seanyoung: They're using the octal notation

LucasSte: > They're using the octal notation

There cannot be a `9` digit in octal.

LucasSte: The solidity documentation says:
>  Octal literals do not exist in Solidity and leading zeros are invalid.

A number with leading zeros is not an octal. All the numbers you wrote here are decimals, not octals. This is the definition of an octal: https://en.wikipedia.org/wiki/Octal

LucasSte: I think the message should be say that leading zeros are not permitted. Even though C considers numbers with leading zeros as octals, this is not a patterns across programming languages. See Python:

```
>>> a = 012
  File "<stdin>", line 1
    a = 012
        ^
SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers
```

seanyoung: > A number with leading zeros is not an octal.

A number of with a leading zero is an octal in many of not most programming languages.

seanyoung: I've updated the message. @LucasSte is probably not the only one who does not know octal notation.

LucasSte: > > A number with leading zeros is not an octal.
> 
> A number of with a leading zero is an octal in many if not most programming languages.

It is NOT in Rust, Python, Zig, OCaml, Ruby.
It is in C/C++, JavaScript/Typecript, Java, Golang, PHP.

We have a tie!

LucasSte (approved):

xermicus (approved):

PR #1609, opened 2023-12-12
Represent type(f) correctly in the ast

seanyoung:
This makes the ast correct, and also the following code is now parsed correctly:
    
            function foo() {
                    type(int);
            }

 `type(enum-type).min` or `type(enum).min` is now supported. 

This fixes all the issues wrt `type(T)`  in the solc tests.


seanyoung:
codecov says `src/sema/eval.rs:232` is not covered, but we have an explicit test for it. If I change the line to panic, the test panics. I don't know what's up with codecov, I don't have the time to investigate.

src/codegen/expression.rs
```diff
@@ -2238,6 +2250,45 @@ fn expr_builtin(
...
+                unreachable!();
+            };
+
+            interfaceid(ns, *contract_no, loc)
+        }
+        ast::Builtin::TypeRuntimeCode | ast::Builtin::TypeCreatorCode => {
+            let ast::Expression::TypeOperator {
+                ty: Type::Contract(contract_no),
+                ..
+            } = &args[0]
```
LucasSte, line 2285: Instead of a builtin, wouldn't it be nicer of item we access from a type operator was a member access? This way you could define the possible members for the operator.

I found it somewhat uncomfortable that you require the argument to be `TypeOperator` in all cases with an unreachable.

LucasSte: As there is a constraint for the argument, perhaps it could be enforced in another way that avoids the unreachable.

src/sema/expression/constructor.rs
```diff
@@ -673,6 +675,48 @@ pub(super) fn deprecated_constructor_arguments(
     Ok(())
 }
 
+/// Check that we are not creating contracts where we cannot
+pub(crate) fn check_circular_reference(contract_no: usize, ns: &mut Namespace) {
+    let mut creates = Vec::new();
+    let mut diagnostics = Diagnostics::default();
+
+    swap(&mut creates, &mut ns.contracts[contract_no].creates);
```
LucasSte, line 683: Why is the swap necessary?

seanyoung: borrow checker

src/codegen/expression.rs
seanyoung: What do you mean by member access? `Expression::StructMember`? That expression is not suitable, it takes a field number and it has "Struct" in its name. 

You may want the location of `type(..)` in `type(int)  .    min` and `type(int)` can be free standing too, so a separate `Expression::TypeOperator` enum makes sense to me.

Please make suggestions because I have no idea.

seanyoung:
If there are no further comments, I'll merge

PR #1611, opened 2023-12-16
Various fallback improvements

seanyoung:
Allow the following:

	contract A {
		// Take raw calldata as argument and return return data
		fallback(bytes calldata input) external payable (bytes memory) {}
	}

Allow a function to be called fallback:

	contract A {
		// Will give a warning that's not a fallback function,
		// just a regular functon
		function fallback() public {}
	}

TODO:

 - docs
 - tests


docs/language/functions.rst
```diff
@@ -384,6 +384,11 @@ is executed. This made clear in the declarations; ``receive()`` must be declared
 with value and no ``receive()`` function is defined, then the call reverts, likewise if
 call is made without value and no ``fallback()`` is defined, then the call also reverts.
 
+The fallback function can defined in two ways. First, it can have no parameters or return
+values. Alternatively, it must have a ``bytes`` parameter and ``bytes`` return value. In this
+case, the parameter contains the undecoded input (also known as calldata or instruction data),
```
LucasSte, line 389: undecoded = encoded?

LucasSte (approved):

docs/language/functions.rst
seanyoung: The wording is not great. What I mean is the raw bytes data before going through borsh/scale/eth decoding.

xermicus (approved):

xermicus:
@seanyoung I'll be preparing a release in the upcoming days, I think this would be nice to have in :)

PR #1615, opened 2024-01-02
Do not build multi arch container images

seanyoung:
Building a multiarch container image is fragile and depends on the linux-arm64 runner having access to a remote amd64 docker daemon.

As far as I know this is not ever used.


LucasSte (approved):

seanyoung:
Merging so the CI tests pass again

PR #1617, opened 2024-01-10
Bump polkadot node version

xermicus:
Upgrade the Polkadot node. In recent versions, the node (in debug mode) logs any API calls and their results to debug buffer on its own. Because this feature was only relevant in the Polkadot target, I :axe: it.


LucasSte (approved):

integration/polkadot/balances.spec.ts
```diff
@@ -3,6 +3,7 @@ import { weight, createConnection, deploy, transaction, aliceKeypair, daveKeypai
 import { ContractPromise } from '@polkadot/api-contract';
 import { ApiPromise } from '@polkadot/api';
 
+
```
seanyoung, line 6: random line add

src/emit/polkadot/storage.rs
```diff
@@ -66,9 +64,7 @@ impl StorageSlot for PolkadotTarget {
             scratch_len.into()
         );
 
-        log_return_code(binary, "seal_get_storage", exists);
-
-        let exists = binary.builder.build_int_compare(
+        let condition = binary.builder.build_int_compare(
```
seanyoung, line 67: `exists` was supposed to mean "does the key exist". `condition` is not very clear

src/emit/polkadot/target.rs
```diff
@@ -166,9 +158,7 @@ impl<'a> TargetRuntime<'a> for PolkadotTarget {
             scratch_len.into()
         );
 
-        log_return_code(binary, "seal_get_storage: ", exists);
-
-        let exists = binary.builder.build_int_compare(
+        let condition = binary.builder.build_int_compare(
```
seanyoung, line 161: `exists` was supposed to mean "does the key exist". `condition` is not very clear

src/emit/polkadot/target.rs
```diff
@@ -223,9 +213,7 @@ impl<'a> TargetRuntime<'a> for PolkadotTarget {
...
```
seanyoung, line 216: `exists` was supposed to mean "does the key exist". `condition` is not very clear

seanyoung (approved):

integration/polkadot/balances.spec.ts
xermicus: removed

src/emit/polkadot/target.rs
xermicus: Its the condition to the select, changed to `exists_is_zero`.

Issue #1618, opened 2024-01-22
Bug: Invalid codegen for structs


xermicus:
Found a case where Solang (latest commit on `main`) generates invalid code, failing an LLVM assertion. [Code is from here](https://github.com/adlerjohn/inflate-sol/blob/master/contracts/InflateLib.sol) but I extracted a minimal reproducer below which is enough to trigger the bug.

```bash
$ solang compile --target polkadot --release reproducer.sol

thread 'main' panicked at /home/glow/.cargo/registry/src/index.crates.io-6f17d22bba15001f/inkwell-0.2.0/src/values/enums.rs:286:13:
Found PointerValue(PointerValue { ptr_value: Value { name: "struct member14", address: 0x5c58cccf2980, is_const: false, is_null: false, is_undef: false, llvm_value: "  %\"struct member14\" = getelementptr inbounds { ptr, i256, i256, i256 }, ptr %0, i32 0, i32 3", llvm_type: "ptr" } }) but expected the IntValue variant
```

## Minimal reproducer

```solidity
contract Bug {
    struct State {
        bytes input;
        uint256 incnt;
        uint256 bitbuf;
        uint256 bitcnt;
    }

    function bits(State memory s) public pure {
        uint256 val = s.bitbuf;
        while (s.bitcnt < 1) {
            val |= uint256(uint8(s.input[s.incnt++])) << s.bitcnt;
        }

        s.bitbuf = val >> 1;
    }
}
```

yp945:
I submitted a PR to fix this issue, looking forward to your review. 😄

xermicus:
@yp945 thanks for the fix

PR #1627, opened 2024-03-01
Fix panics caused by shifting(#1618)

yp945:
Fixed the bug where there was no load when the right expression of the shift calculation is a structure member.

### DIFF
--- a/src/sema/expression/arithmetic.rs
+++ b/src/sema/expression/arithmetic.rs
@@ -227,12 +227,19 @@ pub(super) fn shift_left(
     let (right_length, _) = type_bits_and_sign(&right.ty(), &r.loc(), false, ns, diagnostics)?;
 
     let left_type = left.ty().deref_any().clone();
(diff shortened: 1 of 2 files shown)


seanyoung (commented):
This could really use a test case

yp945:
> This could really use a test case

Ok, I will add a test case.

yp945:
I  have successfully run the `cargo test --workspace` command locally, but it failed on CI. I'm not sure if it's due to recent changes.

seanyoung:
> I have successfully run the `cargo test --workspace` command locally, but it failed on CI. I'm not sure if it's due to recent changes.

It was an intermittent failure in the integration tests. Rerunning it fixed the problem.

seanyoung (approved):
Looks good!

LucasSte (approved):

xermicus (approved):
Thanks

PR #1639, opened 2024-04-12
Release 0.3.4 Athens

xermicus:
This release brings various improvements and bug fixes to our stable compilation targets. Notably, the Polkadot target is now compatible with `ink!` version 5.0 metadata.


seanyoung (approved):
Looks good 👍

CHANGELOG.md
```diff
@@ -4,15 +4,44 @@ will be documented here.
...
 ### Added
+- **Experimental**: Three address code format IR for the CFG has been implemented (disabled by default).
+  [fanyi-zhao](https://github.com/fanyi-zhao).
 - **Soroban** Work on adding support for [Stellar's Soroban](https://soroban.stellar.org/docs) contracts platforms started, by adding a skeleton that supports the Soroban runtime. [Salaheldin Soliman](https://github.com/salaheldinsoliman)
-
 - The `string.concat()` and `bytes.concat()` builtin functions are supported. [seanyoung](https://github.com/seanyoung)
+- **Polkadot**: Compatibility with ink! v5.0 metadata and substrate-contracts-node v0.39.0
+- Implement event selectors [seanyoung](https://github.com/seanyoung)
+- Add a feature flag for compiling language server [xermicus](https://github.com/xermicus)
+- Parsing the pragma solidity version numbers making this information available in the downstream
```
LucasSte, line 17:
```suggestion
- Parsing the pragma solidity version numbers, making this information available in the downstream 
```

LucasSte (approved):

xermicus:
Unfortunately `cargo publish` will fail, as we depend on `forge-fmt` which depends on solang-parser `=0.3.2`.

While pinned dependencies in libraries is definitely annoying; however I think depending on our own crate via a third party would block us anyways, for example what if we wanted to release a major version?

seanyoung:
> Unfortunately `cargo publish` will fail, as we depend on `forge-fmt` which depends on solang-parser `=0.3.2`.

I think `forge-fmt` should copied into the solang repo. The published crate is not being updated.

> While pinned dependencies in libraries is definitely annoying; I think depending on our own crate via a third party would block us anyways, for example what if we wanted to release a major version?

Yep, that's not good. Copying it into our repo would solve that problem, and we can make sure the formatter is up to date with the latest syntax too - right now, it's out of date and cannot parse everything that solang can.

xermicus:
Right, given the licenses, both Apache 2.0, are compatible, I think this should work.

The other solution coming to my mind would be to factor the language server out into a dedicated project (I think just outside the solang worspace would do too)?

seanyoung:
> Right, given the licenses, both Apache 2.0, are compatible, I think this should work.
> 
> The other solution coming to my mind would be to factor the language server out into a dedicated project (I think just outside the solang worspace would do too)?

That doesn't solve the problem, as the language server would have the same issue.

On the up-side, the solang crate would not have any problems :)

I've had a quick stab at copying `forge-fmt` into the solang repo. It needs some work, most notably forge-fmt needs to updated to use the latest parse tree.

xermicus:
> That doesn't solve the problem, as the language server would have the same issue.

I see, so the solution would rather be factoring `solang-parser` out of the workspace I guess.

> I've had a quick stab at copying forge-fmt into the solang repo. It needs some work, most notably forge-fmt needs to updated to use the latest parse tree.

Any idea why foundry themselves are not doing it? You are right, the last release of `forge-fmt` was 9 months back which is quite a while

seanyoung:
> Any idea why foundry themselves are not doing it? You are right, the last release of `forge-fmt` was 9 months back which is quite a while

From what I can see, foundry don't update their crates on crates.io - they published `forge-fmt` after I asked them.

PR #1642, opened 2024-04-29
Add GitHub ARM64 runners

ryjones:
Use github provided runners rather than our own.


seanyoung (approved):

seanyoung:
> Why would docker not be installed? We run our tests in container which contains all the build tools (llvm, rust, solana, and others)?

Looks like the vm is a minimal ubuntu server install. We have to install the tools ourselves - hence the `apt-get install`.

LucasSte (approved):

PR #1645, opened 2024-05-18
Soroban counter.sol example

salaheldinsoliman:
This PR aims to make Solang support a simple counter.sol example on Soroban, where a storage variable is instatiated, modified and retrieved. The counter contract is only limited to `uint64` data types, and only supports `instance` soroban storage.

This can be considered a "skeleton" for supporting more data and storage types, as well as more host function invokations. 

- [x] Support Soroban storage function calls `put_contract_data`, `get_contract_data` and `has_contract_data`
- [x] Implement a wrapper `init` for `storage_initializer` 
- [x] Implement wrappers for public functions
- [x] Insert decoding/encoding instructions into the wrapper functions 
- [x] Soroban doesn't have function return codes. This needs to be handled all over emit
- [x] Add integration tests and MockVm tests


src/bin/cli/mod.rs
```diff
@@ -321,6 +321,15 @@ pub struct CompilePackage {
     #[arg(name = "VERSION", help = "specify contracts version", long = "version", num_args = 1, value_parser = ValueParser::new(parse_version))]
     #[serde(default, deserialize_with = "deserialize_version")]
     pub version: Option<String>,
+
+    #[arg(
+        name = "SOROBAN-VERSION",
+        help = "specify soroban contracts pre-release number",
+        short = 's',
+        long = "soroban-version",
+        num_args = 1
```
seanyoung, line 330: How about:
```
value_parser = value_parser!(u32).range(0..1000),
```
If there is a known range for this number

src/codegen/dispatch/mod.rs
```diff
@@ -17,6 +18,7 @@ pub(super) fn function_dispatch(
         Target::Polkadot { .. } | Target::EVM => {
             polkadot::function_dispatch(contract_no, all_cfg, ns, opt)
         }
-        Target::Soroban => vec![],
+        //Target::Soroban => vec![],
```
seanyoung, line 21: Please remove this commented out lube

src/codegen/dispatch/soroban.rs
```diff
@@ -0,0 +1,147 @@
...
+                cfg.name.clone()
+            }
+        };
+
+        let mut wrapper_cfg = ControlFlowGraph::new(wrapper_name.to_string(), ASTFunction::None);
+
+        wrapper_cfg.params = function.params.clone();
+        //wrapper_cfg.returns = function.returns.clone();
+
+        //let returns = Vec::new();
```
seanyoung, line 54: remove

src/codegen/dispatch/soroban.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    let mut wrapper_cfgs = Vec::new();
+
+    for cfg in all_cfg.iter_mut() {
+        if cfg.function_no == ASTFunction::None {
+            continue;
+        }
+
+        let function = match &cfg.function_no {
+            ASTFunction::SolidityFunction(no) | ASTFunction::YulFunction(no) => &ns.functions[*no],
```
seanyoung, line 33: yul functions are never public (I think)

src/codegen/dispatch/soroban.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        let wrapper_name = {
+            if cfg.public {
+                if function.mangled_name_contracts.contains(&contract_no) {
+                    function.mangled_name.clone()
+                } else {
+                    function.id.name.clone()
+                }
+            } else {
+                cfg.name.clone()
```
seanyoung, line 45: do we need a wrapper for non-public functions?

src/emit/instructions.rs
```diff
@@ -484,54 +486,59 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
                 .build_call(bin.functions[cfg_no], &parms, "")
                 .unwrap()
                 .try_as_basic_value()
-                .left()
-                .unwrap();
+                .left();
+            //.unwrap();
```
seanyoung, line 490: remove

src/emit/instructions.rs
```diff
@@ -484,54 +486,59 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
...
-                    "success",
-                )
-                .unwrap();
-
-            let success_block = bin.context.append_basic_block(function, "success");
-            let bail_block = bin.context.append_basic_block(function, "bail");
-            bin.builder
-                .build_conditional_branch(success, success_block, bail_block)
-                .unwrap();
+            // Soroban doesnt have return codes, and only returns a single i64 value
```
seanyoung, line 492: I was wondering if this should be enforced in sema; for Soroban, we could simply give an error diagnostic if a function returns anything else but an `int64`.

seanyoung (commented):
Looks great

src/emit/instructions.rs
salaheldinsoliman: What is reffered to here is the `wasm` functions, not Solidity functions. That is, if a Solidity function returns nothing, the `wasm` function returns an `i64` value to be XDR decoded as `Void`.

src/codegen/dispatch/soroban.rs
salaheldinsoliman: I also think so yes

src/emit/instructions.rs
seanyoung: Right, the comment could with explaining that

src/codegen/dispatch/soroban.rs
```diff
@@ -0,0 +1,145 @@
...
+) -> Vec<ControlFlowGraph> {
+    // For each function in all_cfg, we will generate a wrapper function that will call the function
+    // The wrapper function will call abi_encode to encode the arguments, and then call the function
+
+    let mut wrapper_cfgs = Vec::new();
+
+    for cfg in all_cfg.iter_mut() {
+        if cfg.function_no == ASTFunction::None {
+            continue;
+        }
```
seanyoung, line 30: This case is already handled on line 34

seanyoung (approved):

src/emit/instructions.rs
```diff
@@ -484,54 +486,58 @@ pub(super) fn process_instruction<'a, T: TargetRuntime<'a> + ?Sized>(
                 .build_call(bin.functions[cfg_no], &parms, "")
                 .unwrap()
                 .try_as_basic_value()
-                .left()
-                .unwrap();
+                .left();
```
LucasSte, line 489: Why did you remove the unwrap from here if you are unwrapping the value everywhere else? Is there a case when this call is going to fail?

src/emit/soroban/mod.rs
```diff
@@ -68,69 +86,68 @@ impl SorobanTarget {
...
         }
 
+        let init_type = context.i64_type().fn_type(&[], false);
+        binary
+            .module
+            .add_function("storage_initializer", init_type, None);
+
         for (func_decl, cfg) in defines {
             emit_cfg(&mut SorobanTarget, binary, contract, cfg, func_decl, ns);
+            //Self::emit_function_dispatcher(binary, cfg, func_decl, &dispatcher_name, ns, &contract);
```
LucasSte, line 127: Please, remove commented code.

src/emit/soroban/mod.rs
```diff
@@ -68,69 +86,68 @@ impl SorobanTarget {
...
+            // if func is a default constructor, then the function name is the contract name
 
             let linkage = if cfg.public {
+                let name = if cfg.name.contains("::") {
+                    // get the third part of the name which is the function name
+                    cfg.name.split("::").collect::<Vec<&str>>()[2]
+                } else {
+                    &cfg.name
+                };
+                Self::emit_function_spec_entry(context, cfg.clone(), name.to_string(), binary);
```
LucasSte, line 99: Cloning the CFG may be expensive.

src/emit/soroban/mod.rs
```diff
@@ -68,69 +86,68 @@ impl SorobanTarget {
...
         context: &'a Context,
-        cfg: &'a ControlFlowGraph,
+        cfg: ControlFlowGraph,
         name: String,
         binary: &mut Binary<'a>,
     ) {
         if cfg.public && !cfg.is_placeholder() {
-            // TODO: Emit custom type spec entries.
+            // TODO: Emit custom type spec entries
+            //let outputs = ScSpecTypeDef::Vec(());
```
LucasSte, line 149: Please, remove unused code.

src/emit/soroban/target.rs
```diff
@@ -51,7 +72,28 @@ impl<'a> TargetRuntime<'a> for SorobanTarget {
         function: FunctionValue<'a>,
         ns: &ast::Namespace,
     ) {
-        unimplemented!()
+        emit_context!(binary);
```
LucasSte, line 75: Why is `emit_context` necessary here?

src/emit/soroban/target.rs
```diff
@@ -193,9 +235,7 @@ impl<'a> TargetRuntime<'a> for SorobanTarget {
     }
 
     /// Prints a string
-    fn print(&self, bin: &Binary, string: PointerValue, length: IntValue) {
-        unimplemented!()
-    }
+    fn print(&self, bin: &Binary, string: PointerValue, length: IntValue) {}
```
LucasSte, line 238: Should this function be doing nothing?

src/codegen/dispatch/soroban.rs
```diff
@@ -0,0 +1,145 @@
...
+};
+
+pub fn function_dispatch(
+    contract_no: usize,
+    all_cfg: &mut [ControlFlowGraph],
+    ns: &mut Namespace,
+    _opt: &Options,
+) -> Vec<ControlFlowGraph> {
+    // For each function in all_cfg, we will generate a wrapper function that will call the function
+    // The wrapper function will call abi_encode to encode the arguments, and then call the function
```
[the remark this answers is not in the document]
salaheldinsoliman: Soroban doesn't have a `dispatch` function convention. Instead, soroban contracts directly export all public functions, thus we need a wrapper for each one.

src/emit/soroban/target.rs
salaheldinsoliman: When there is a runtime error, an overflow maybe, a call to this function is made. The ideal scenario is to make a call to Soroban's host function responsible for logging, but since this is out of scope for this PR, I left it out. 

A `//TODO` might be better here

src/codegen/dispatch/soroban.rs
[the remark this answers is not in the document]
LucasSte: Are you allowed to export the function without a wrapper?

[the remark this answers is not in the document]
salaheldinsoliman: Nope. The wrapper's functionality is to decode the arguments and encode the returns. Basically, there is a dispatcher per each public function.

src/emit/soroban/target.rs
salaheldinsoliman: It's for `i64_const!(2).into(),` in this [line](https://github.com/salaheldinsoliman/solang/blob/84d5f7db6b5d3010b062a075b4c50ea43036f610/src/emit/soroban/target.rs#L88). 
I can use:
`binary.context.i64_type().const_int(2, false).into()` though

src/emit/instructions.rs
salaheldinsoliman: Great question, thanks
In Soroban, there is no return codes for functions or a pointer to returned data.
Instead, the actual data is returned as is. so `ret` might be `None`.
Therefore, in this [line](https://github.com/salaheldinsoliman/solang/blob/84d5f7db6b5d3010b062a075b4c50ea43036f610/src/emit/instructions.rs#L538-L539)
`} else if let Some(value) = ret {
   w.vars.get_mut(&res[0]).unwrap().value = value;
 }`
`ret` is not always `unwrapped`
 
 Does that properly answer your question?

PR #1649, opened 2024-06-06
Vendor forge-fmt into solang repo

seanyoung:
The published forge-fmt is out date, not maintained and depends on an old version of solang-parser.

This is a copy of forge-fmt:

 https://github.com/foundry-rs/foundry/tree/master/crates/fmt

commit: 00854b602ef0e67379a2027ccc5d0aad553e5333


seanyoung:
> I don't know if this is much of an effort, but have you considered adding forge-fmt as a submodule or as a git subtree?

It needs modifications to make it work, and I don't know that that is possible with submodule or subtree

fmt/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "forge-fmt"
+version = "0.2.0"
```
xermicus, line 3: Is this still accurate?

fmt/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "forge-fmt"
```
xermicus, line 2: I'm thinking, should we name it `solang-fmt` instead? Or do we just copy it over from time to time leaving as much as possible as it is?

fmt/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+version = "0.2.0"
+edition = "2021"
+authors = ["Foundry Contributors"]
+license = "MIT OR Apache-2.0"
```
xermicus, line 6: Is this compatible with our license (looks like we have Apache only)?

fmt/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+homepage = "https://github.com/foundry-rs/foundry"
+repository = "https://github.com/foundry-rs/foundry"
```
xermicus, line 8: Is this intentional? Might be misleading for example if we want to publish it on crates.io but I guess this isn't supposed to be published?

xermicus (approved):
LGTM

xermicus:
Yeah I quickly had the same thought as @LucasSte, however we'd have to modify it anyways and it wouldn't be just `forge-fmt` but the whole foundry repo.

xermicus:
@seanyoung also noticed the PR title could be renamed to "Vendor forge-fmt into solang" as it would explain the reasoning of why we need this :)

fmt/Cargo.toml
seanyoung: It's a pretty useful crate TBH. We might want to publish it in the future.

So I agree, why not rename it to `solang-fmt`.

fmt/Cargo.toml
seanyoung: What version should it have if we rename it to `solang-fmt`?

xermicus: If we are going to publish it, intuitively should be the same version as solang-parser and solang?

xermicus: However if we just want to vendor it we should keep the original version I think

seanyoung: How about we just vendor it for now. If we want to publish it, we can always rename it and give it a new version.

seanyoung:
The windows failure is unrelated to this PR

fmt/Cargo.toml
seanyoung: I've changed my mind. Let's just vendor it for now.

fmt/Cargo.toml
seanyoung: As we're vendoring for now, let's leave it as is

PR #1650, opened 2024-06-07
Evaluate Comparison Expressions with Constant Operands during Compilation

### LINKED ISSUE #755: Comparisons should support constants
There is no support for constants in comparisons. So things like `0 < 1` do not work yet. Ideally, this can be created in `src/sema` and the comparison can be evaluated in compile time.

Nirhar:
This patch supports evaluation of comparison expressions with constants as operands during compile time.

Fixes #755

### DIFF
--- a/src/sema/eval.rs
+++ b/src/sema/eval.rs
@@ -729,6 +729,150 @@ pub(crate) fn eval_constants_in_expression(
                 (None, true)
             }
         }
+        Expression::More {
+            loc,
+            left,
+            right,
+        } => {
+            let left = eval_constants_in_expression(left, diagnostics).0;
+            let right = eval_constants_in_expression(right, diagnostics).0;
+
+            if let (
+                Some(Expression::NumberLiteral { value: left, .. }),
+                Some(Expression::NumberLiteral { value: right, .. }),
+            ) = (&left, &right)
+            {
+                (
+                    Some(Expression::BoolLiteral { 
+                        loc: *loc,
+                        value: left > right
+                    }),
+                    true,
+                )
+            } else {
+                (None, true)
+            }
+        }
+        Expression::Less {
+            loc,
+            left,
+            right,
+        } => {
+            let left = eval_constants_in_expression(left, diagnostics).0;
+            let right = eval_constants_in_expression(right, diagnostics).0;
+
+            if let (
+                Some(Expression::NumberLiteral { value: left, .. }),
+                Some(Expression::NumberLiteral { value: right, .. }),
+            ) = (&left, &right)
+            {
+                (
+                    Some(Expression::BoolLiteral { 
+                        loc: *loc,
+                        value: left < right
+                    }),
+                    true,
+                )
+            } else {
+                (None, true)
+            }
+        }
+        Expression::Equal {
+            loc,
+            left,
+            right,
+        } => {
+            let left = eval_constants_in_expression(left, diagnostics).0;
+            let right = eval_constants_in_expression(right, diagnostics).0;
+
+            if let (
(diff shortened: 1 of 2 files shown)


Nirhar:
As a first time contributor to this project, I'd like to ask some questions:
1. How do I add tests for my patch?
2. Is there any formatting rules that I have to follow, or do we use any tools like clang-format?

I am new to rust, so if you find anything that goes against conventional programming style in rust, please let me know!

src/sema/eval.rs
```diff
@@ -729,6 +729,150 @@ pub(crate) fn eval_constants_in_expression(
...
+
+            if let (
+                Some(Expression::NumberLiteral { value: left, .. }),
+                Some(Expression::NumberLiteral { value: right, .. }),
+            ) = (&left, &right)
+            {
+                (
+                    Some(Expression::BoolLiteral { 
+                        loc: *loc,
+                        value: (left.cmp(right) == core::cmp::Ordering::Greater)
```
seanyoung, line 748: This is a very elaborate way of writing `left > right`

LucasSte, line 748: Why not `left > right`?

seanyoung (changes requested):
One way to write a test for this is to add test to `tests/codegen_tests/..` which does something like the following:

```
contract C {
     function foo() public returns (bool) {
           int a = 100;
           
           return a < 200;
     }
     
     function bar() public returns (bool r) {
           int a = 100;
           int b = 200;
           
           r = a < b;
     }
}
```
Make sure the test ensures that the code generation constant folds the expressions and generates `false` for both functions.

LucasSte:
> 2. Is there any formatting rules that I have to follow, or do we use any tools like clang-format?

Have you checked our contributing guide https://solang.readthedocs.io/en/latest/contributing.html?

src/sema/eval.rs
Nirhar: Sorry, for a moment I thought literals did support operator overloading for comparison operations. Have corrected it!

Nirhar: Have corrected it!

Nirhar:
> One way to write a test for this is to add test to `tests/codegen_tests/..` which does something like the following:
> 
> ```
> contract C {
>      function foo() public returns (bool) {
>            int a = 100;
>            
>            return a < 200;
>      }
>      
>      function bar() public returns (bool r) {
>            int a = 100;
>            int b = 200;
>            
>            r = a < b;
>      }
> }
> ```
> 
> Make sure the test ensures that the code generation constant folds the expressions and generates `false` for both functions.

I've added a similar test case as you've suggested, and when I try to execute it as:
```
<path-to-solang> compile --target polkadot --emit cfg
```
I find `ty:bool %r = (signed less int256 100 < int256 200)` which suggests to me that my optimization did not kick in. I'm not sure if cfg is the right place to investigate the results of optimizations. I also tried to emit llvm-ir to inspect (with `--emit llvm-ir`), however, nothing was printed to the console when I do so.

src/sema/eval.rs
```diff
@@ -729,6 +729,150 @@ pub(crate) fn eval_constants_in_expression(
...
+                Some(Expression::NumberLiteral { value: right, .. }),
+            ) = (&left, &right)
+            {
+                (
+                    Some(Expression::BoolLiteral { 
+                        loc: *loc,
+                        value: left > right
+                    }),
+                    true,
+                )
```
seanyoung, line 751: Please don't optimize the code in sema, this should be done in codegen. 

Also this does not take signed-ness into account. Unsigned/signed compare are different.

seanyoung:
PR has not been changed for as long time, closing.

PR #1659, opened 2024-07-14
Add print functionality to Soroban contracts

salaheldinsoliman:
This PR adds static string print functionality to Soroban contracts. This serves the following:

1. `print()` statements
2. Logging runtime errors.

However, the following findings might be interesting:
In both Solana and Polkadot, the VM execution capacity can grasp a call to `vector_new` in the `stdlib`: https://github.com/hyperledger/solang/blob/06798cdeac6fd62ee98f5ae7da38f3af4933dc0f/stdlib/stdlib.c#L167

However, Soroban doesn't. That's why Soroban would need Solang to implement a more efficient way of printing dynamic strings. 
@leighmcculloch

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,12 +95,11 @@ impl Target {
 
     /// Size of a pointer in bits
     pub fn ptr_size(&self) -> u16 {
-        if *self == Target::Solana {
+        match *self {
             // Solana is BPF, which is 64 bit
-            64
-        } else {
+            Target::Solana => 64,
             // All others are WebAssembly in 32 bit mode
-            32
+            _ => 32,
         }
     }
 
--- a/integration/soroban/.gitignore
+++ b/integration/soroban/.gitignore
@@ -6,3 +6,5 @@
 !package.json
 node_modules
 package-lock.json
+*.txt
+*.toml
--- a/integration/soroban/runtime_error.sol
+++ b/integration/soroban/runtime_error.sol
@@ -1,0 +1,9 @@
+contract Error {
+    uint64  count = 1;
+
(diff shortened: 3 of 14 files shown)


src/lib.rs
```diff
@@ -95,7 +95,7 @@ impl Target {
 
     /// Size of a pointer in bits
     pub fn ptr_size(&self) -> u16 {
-        if *self == Target::Solana {
+        if *self == Target::Solana || *self == Target::Soroban {
```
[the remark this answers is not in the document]
salaheldinsoliman: Thanks, that was a typo

salaheldinsoliman:
@seanyoung Can I get a second look on this one?

.github/workflows/test.yml
```diff
@@ -485,13 +485,16 @@ jobs:
...
     - run: npm install
       working-directory: ./vscode
     - run: npm run compile
       working-directory: ./vscode
     - name: Run headless test
-      run: xvfb-run npm test
+      run: |
+          export DBUS_SESSION_BUS_ADDRESS=/dev/null
+          export NO_AT_BRIDGE=1
+          xvfb-run --auto-servernum --server-args="-screen 0 1920x1080x24" npm test -- --disable-gpu
```
seanyoung, line 497: These changes don't fix the issue of the vscode test not working. For some reason the tap package is not installed, it's not `xvfb-run`

seanyoung (approved):

seanyoung:
LGTM, thank you!

.github/workflows/test.yml
salaheldinsoliman: Let me just remove that and address it in a further PR then

PR #1664, opened 2024-08-24
Support different Soroban storage types

salaheldinsoliman:
This PR addresses that Soroban has 3 different storage types: persistent, temporary and instance. 
It modifies the parser to accept `persistent, temporary and instance` keywords during contract variable instantiation. If no storage type is provided, the storage defaults to `persistent` and raises a warning telling the developer: `variable x has no storage type defined, defaulting to persistent`.


src/sema/variables.rs
```diff
@@ -233,9 +235,22 @@ pub fn variable_decl<'a>(
 
                 visibility = Some(v.clone());
             }
+            pt::VariableAttribute::StorageType(s) => {
+                storage_type = Some(s.clone());
```
seanyoung, line 239: Here you don't check for duplicate entries. For example:

```
contract c {
    int temporary persistent v;
}
```

```

File: src/sema/variables.rs [hunk 2]
```diff
@@ -233,9 +235,22 @@ pub fn variable_decl<'a>(
...
 
+    if ns.target == Target::Soroban && storage_type.is_none() {
+        ns.diagnostics.push(Diagnostic::warning(
+            def.loc,
+            format!(
+                "storage type not specified for `{}`, defaulting to `persistent`",
+                def.name.as_ref().unwrap().name
+            ),
+        ));
+    }
```
seanyoung, line 252: If the `storage_type` is specified and the target is not Soroban, we should give an error and not silently ignore.

src/emit/binary.rs
```diff
@@ -1015,6 +1015,7 @@ impl<'a> Binary<'a> {
                 Type::FunctionSelector => {
                     self.llvm_type(&Type::Bytes(ns.target.selector_length()), ns)
                 }
+                Type::Void => BasicTypeEnum::IntType(self.context.i64_type()),
```
seanyoung, line 1018: There is a llvm/inkwell `self.context.void_type().into()` for this.

src/codegen/expression.rs
```diff
@@ -58,9 +58,21 @@ pub fn expression(
...
+                loc: _,
+                ty: _,
+                var_no,
+                contract_no,
+            } = *expr.clone()
+            {
+                let var = ns.contracts[contract_no].variables.get(var_no).unwrap();
+
+                storage_type = var.storage_type.clone();
+            }
```
seanyoung, line 72: you could do
```rust
  let storage_type =  if let ast::Expression::StorageVariable {
                loc: _,
                ty: _,
                var_no,
                contract_no,
            } = *expr.clone()
            {
                let var = ns.contracts[contract_no].variables.get(var_no).unwrap();

               var.storage_type.clone()
            } else {
               None
          };
```

src/codegen/expression.rs
```diff
@@ -1232,13 +1244,35 @@ fn post_incdec(
...
+        loc: _,
+        ty: _,
+        var_no,
+        contract_no,
+    } = var
+    {
+        let var = ns.contracts[*contract_no].variables.get(*var_no).unwrap();
+
+        storage_type = var.storage_type.clone();
+    }
```
seanyoung, line 1260: same here

src/codegen/expression.rs
```diff
@@ -2676,6 +2734,19 @@ pub fn assign_single(
...
+                loc: _,
+                ty: _,
+                var_no,
+                contract_no,
+            } = left
+            {
+                let var = ns.contracts[*contract_no].variables.get(*var_no).unwrap();
+
+                storage_type = var.storage_type.clone();
+            }
```
seanyoung, line 2748: and here

solang-parser/src/lexer.rs
```diff
@@ -553,6 +561,9 @@ static KEYWORDS: phf::Map<&'static str, Token> = phf_map! {
     "unchecked" => Token::Unchecked,
     "assembly" => Token::Assembly,
     "let" => Token::Let,
+    "persistent" => Token::Persistent,
+    "temporary" => Token::Temporary,
+    "instance" => Token::Instance,
```
seanyoung, line 566: This does affect Polkadot/Solana, but never mind, I don't know what other solution there is (other than a better parser generator).

seanyoung (changes requested):

tests/soroban_testcases/storage.rs
```diff
@@ -39,3 +39,89 @@ fn counter() {
     let expected: Val = 10_u64.into_val(&src.env);
     assert!(expected.shallow_eq(&res));
 }
+
+#[test]
+fn different_storage_types() {
+    let src = build_solidity(
+        r#"contract sesa {
```
silence48, line 46: I'd suggest renaming the storage variables `sesa`, `sesa1`, `sesa2`, and `sesa3` to something more descriptive, unless they need to stay like this for some reason

For example, `temporaryCount`, `instanceCount`, and `persistentCount` may help clarify what the variables are intended to test and demonstrate.

I figured i'd mention this because i was trying to figure out what `sesa3` was used for,

src/codegen/cfg.rs
```diff
@@ -1027,7 +1029,7 @@ impl ControlFlowGraph {
                 true_block,
                 false_block,
             ),
-            Instr::LoadStorage { ty, res, storage } => format!(
+            Instr::LoadStorage { ty, res, storage, .. } => format!(
```
silence48, line 1032: a comment clarifying about how the system behaves when the storage type is omitted might be helpful to people looking at this in the future.

integration/soroban/storage_types.spec.js
```diff
@@ -0,0 +1,87 @@
...
+    // Load contract
+    contract = new StellarSdk.Contract(contractAddr);
+
+    // Initialize the contract if necessary (adapt according to your contract logic)
+    await call_contract_function("init", server, keypair, contract);
+  });
+
+  it('check initial values', async () => {
+    // Check initial values of all storage variables
+    let sesa = await call_contract_function("sesa", server, keypair, contract);
```
silence48, line 39: To expand testing coverage, consider adding a test case for scenarios where a variable **lacks** a specified storage type and defaults to `persistent`. This would ensure that the defaulting mechanism works as intended in Soroban contracts.


Also, what I said previously about `sesa` not being a descriptive variable.

src/emit/soroban/target.rs
```diff
@@ -43,7 +43,18 @@ impl<'a> TargetRuntime<'a> for SorobanTarget {
         slot: &mut IntValue<'a>,
         function: FunctionValue<'a>,
         ns: &ast::Namespace,
+        storage_type: &Option<StorageType>,
     ) -> BasicValueEnum<'a> {
+        let storage_type = if let Some(storage_type) = storage_type {
```
silence48, line 48: To reduce code duplication and improve maintainability, consider extracting the mapping of StorageType to integer values into a helper function. This will ensure consistency across different functions and make future updates easier.

Both storage_load and storage_store functions contain similar code for mapping StorageType.

```rust
fn map_storage_type(storage_type: Option<&StorageType>) -> u64 {
    if let Some(storage_type) = storage_type {
        match storage_type {
            StorageType::Persistent(_) => 0,
            StorageType::Temporary(_) => 1,
            StorageType::Instance(_) => 2,
        }
    } else {
        0 // Default to Persistent
    }
}
```

Then, in your functions, you can use:

```rust
let storage_type = map_storage_type(storage_type);
```

src/emit/polkadot/target.rs
```diff
@@ -17,7 +17,7 @@ use inkwell::values::{
     ArrayValue, BasicMetadataValueEnum, BasicValueEnum, FunctionValue, IntValue, PointerValue,
 };
 use inkwell::{AddressSpace, IntPredicate};
-use solang_parser::pt::Loc;
+use solang_parser::pt::{Loc, StorageType};
```
silence48, line 20: The `storage_type` parameter has been added to trait methods and their implementations for all targets, but it's unused in non-Soroban targets.

Ensure that adding the `storage_type` parameter does not negatively impact other targets. If the parameter is irrelevant for Polkadot and Solana, consider using default implementations or conditional compilation to exclude it from those targets.

src/sema/variables.rs
silence48: This confused me for a second, but I assume you're suggesting that we add a check to prevent declaring two storage type modifiers for the same variable in a contract. However, in Soroban, each storage 'map' (e.g., temporary, persistent, instance) operates independently of the others. It's possible to store different (or even the same) values for the same key across different storage types. For example, the key 123_u32 might hold the value 100_u32 in temporary storage, while in persistent storage, the same key could store abcd or 100_u32.

So, if both modifiers are handled correctly, Soroban would technically allow this. That said, while it's not typically recommended—since it consumes more storage—there could be edge cases where this pattern is useful, perhaps as a way to initialize data across namespaces. In those cases, such declarations might be valid.

silence48 (commented):
All in all the code looks good, and seems to implement the storage correctly.

for solidity dev's who might be looking at how this works, and are not familiar with the multiple types of storage, it would be helpful to explicitly document the best practices for choosing a storage type (persistent vs. instance vs. temporary). This could be inline in the code, or a link to an external guide (e.g., Stellar documentation) would suffice.

 Consider adding a function or hook to extend the TTL for instance storage (perhaps something like an extend_ttl method), as instance storage typically has a limited lifespan tied to the contract instance. This could be helpful for contracts that need to maintain instance data for longer periods.
 
Currently, storage types are limited to Soroban targets, but if these types are used in other contexts, you should ensure proper error handling is in place.
I'd suggest adding error handling for these cases on those targets if necessary.

fmt/src/formatter.rs
```diff
@@ -3333,6 +3333,11 @@ impl<'a, W: Write> Visitor for Formatter<'a, W> {
                 self.visit_list("", idents, Some(loc.start()), Some(loc.end()), false)?;
                 None
             }
+            VariableAttribute::StorageType(s) => match s {
```
silence48, line 3336: is there any reason the order of the matching should match the [order of the enum?](https://github.com/stellar/rs-soroban-env/blob/cf469d2d4be82f83c24b75aeb8a133cfb82a23cd/soroban-env-common/src/storage_type.rs#L12)

solang-parser/src/pt.rs
```diff
@@ -955,6 +955,24 @@ pub enum VariableAttribute {
...
+
+    /// Storage type.
+    StorageType(StorageType),
+}
+
+/// Soroban storage types.
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "pt-serde", derive(Serialize, Deserialize))]
+#[repr(u8)] // for cmp; order of variants is important
+pub enum StorageType {
```
silence48, line 967: This doesn't match the storatype enum order on soroban side

src/emit/binary.rs
salaheldinsoliman: Soroban functions always return a value encoded in 64 bits. If functions return type is void, that means it returns `2_u64`. However, I should have explained this in a comment

src/sema/variables.rs
salaheldinsoliman: I think @seanyoung meant this case:
`uint64 public temporary instance persistent sesa = 1;`

src/codegen/cfg.rs
salaheldinsoliman: This is the implementation of `instr_to_string`, which is used in Solang if a Solidity dev wants to debug the generated IR (using `--emit cfg).
To simplify the output of this, I omitted the storage type option as I think it is not relevant at the IR generation stage.
WDYT?

silence48 (commented):
Changes look good!

src/emit/binary.rs
seanyoung: That is Soroban specific, and that doesn't belong in the llvm type system.

What's the return value? Is it a success/failure code? That's the same as Solana and Polkadot, and that's implemented in the function emit code.

solang-parser/src/lexer.rs
seanyoung: Unfortunately the keywords can't be made context-sensitive with the current parser generator. Of course, this is something I would really love and this would be such a nicer feature (for lalrpop).

Having said that, it might be possible to replace the keywords `persistent`, `temporary`, and `instance` with `SolIdentifier` and then checking for the keywords in sema.

src/sema/variables.rs
seanyoung: Would be nice to have test cases for these

solang-parser/src/lexer.rs
seanyoung: A way around this is to make these keywords Soroban specific, so they do not apply for Solana/Polkadot. This could be implemented in the lexer.

src/emit/binary.rs
salaheldinsoliman: @seanyoung The return value is not a success/failure, it is just the returned data
If no data is returned, it returns void (which is a uint64 in Soroban)

PR #1673, opened 2024-11-12
Rust 1.75.0

salaheldinsoliman:
Update to Rust 1.75.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: recursive
-    - uses: dtolnay/rust-toolchain@1.74.0
+    - uses: dtolnay/rust-toolchain@1.82.0
     - name: Build
       run: cargo build --verbose --release
     - name: Run tests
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: recursive
-    - uses: dtolnay/rust-toolchain@1.74.0
+    - uses: dtolnay/rust-toolchain@1.82.0
```
seanyoung, line 18: 1.75 should suffice here and everywhere below (keep the 1.82.0 fixes though).

Problem is that you are forcing anyone who uses the solang or solang-parser crate to update to 1.82.0, when it should work with older versions.

salaheldinsoliman: Solang depends on a crate which requires `1.75.0`: https://github.com/hyperledger-solang/solang/actions/runs/11792189166/job/32845426551?pr=1664#step:6:505

Should I upgrade to `1.75.0` instead?

seanyoung: so it should be the lowest version that works

seanyoung (approved):

seanyoung:
Looks good, let's merge :)

Issue #1672, opened 2024-11-02
Add constructor support to Solidity contracts


salaheldinsoliman:
Soroban protocol 21 did not support constructors (functions exposed by the contract, which the Vm calls upon contract deployment).
That means that a developer would have to call `init` manually from the command line.

`init` is a function that Solang embeds in Solidity contracts, which loops over storage variables and initializes them.
https://github.com/hyperledger-solang/solang/blob/5176166545df6090131565a990b303d9cfabf37e/src/emit/soroban/mod.rs#L254

What needs to be done:
Upgrade protocol version in contract to be 22, and change the name `init` to `__constructor`, and add tests

tareknaser:
> change the name `init` to `__constructor`

Just to make sure I understand this correctly,
If we just rename `init` to `__constructor`, it seems we still wouldn’t fully support constructors as outlined in [CAP-0058](https://github.com/stellar/stellar-protocol/blob/5ef1b45cbab54f57c5ddfc98adb964f26b68bb3b/core/cap-0058.md). Developers would still need to manually call it, as they currently do with `init`.

There are a few things I’m unsure about:
- Since `__constructor` is optional, what happens if the developer doesn’t implement it?
	- Do we just expose an empty `__constructor` function that only calls `storage_initializer`, similar to how `init` works now, and expect the developer to call it manually?
- I noticed that the new `__constructor` method can also [include custom logic](https://github.com/stellar/rs-soroban-sdk/blob/main/tests/constructor/src/lib.rs) that the developer wants to run on initialization, which might mean it can take parameters too. Will Solang support that?
- Should `__constructor` still be callable after the contract is deployed and initialized?

salaheldinsoliman:
@tareknaser `__constructor` should call `storage_initializer` before executing custom initialization logic. (not only call `storage_initializer`)

tareknaser:
Alright, here’s what I understand about the desired behavior:
- If the developer doesn’t implement `__constructor`, we should export a function called `__constructor` that only initializes storage variables by calling `storage_initializer`. This part seems straightforward; we could just rename what we have from `init` to `__constructor`.
- If the developer defines their own `__constructor` function with custom code, then we need to ensure that `__constructor` first calls `storage_initializer` and then executes the custom code. This is the part that’s proving tricky.

I’ve tried a few approaches to handle the second point, but they haven’t worked out so far:
[...]

PR #1674, opened 2024-11-22
Add Constructor Support for Solang Solidity Contracts on Soroban

tareknaser:
This PR is the first step towards implementing #1672. It adds constructor support for Solidity contracts in Solang. The next step, not covered in this PR, is handling cases where the user defines a `__constructor` function with custom logic. More details can be found in the issue discussion.

### Changes Made
- Renamed the `init` function (used to call `storage_initializers`) to `__constructor` so that it is recognized and invoked by the Soroban host environment.
- Updated tests and CI to work with Soroban version 22.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: recursive
-    - uses: dtolnay/rust-toolchain@1.75.0
+    - uses: dtolnay/rust-toolchain@1.81.0
     - name: Build
       run: cargo build --verbose --release
     - name: Run tests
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: recursive
-    - uses: dtolnay/rust-toolchain@1.75.0
+    - uses: dtolnay/rust-toolchain@1.81.0
```
salaheldinsoliman, line 18: Why is that change related to the PR?

salaheldinsoliman (approved):

.github/workflows/release.yml
tareknaser: `soroban-sdk` 22.0.0-rc.3.2 requires rustc 1.79.0 or newer, and `soroban-cli` (which we use in the CI) now requires rustc 1.81.0 or newer.

seanyoung (approved):
LGTM

PR #1675, opened 2024-11-29
Soroban: Support for Constructors

tareknaser:
This PR adds support for user-defined constructors in Soroban, specifically for constructors with no arguments. Note that constructors with arguments are not yet supported (see comments for details).
Part of #1672

**Other changes include:**
- Refactored code in `tests/soroban.rs` to allow providing constructors when registering contracts on the chain.
- Added tests for constructors in `tests/soroban_testcases/constructor.rs`.

### DIFF
--- a/src/emit/soroban/mod.rs
+++ b/src/emit/soroban/mod.rs
@@ -63,7 +63,7 @@ impl SorobanTarget {
         );
         binary.internalize(export_list.as_slice());
 
-        Self::emit_initializer(&mut binary, ns);
+        Self::emit_initializer(&mut binary, ns, contract.constructors(&ns).first());
 
         Self::emit_env_meta_entries(context, &mut binary, opt);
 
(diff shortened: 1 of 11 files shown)


src/emit/soroban/mod.rs
```diff
@@ -294,8 +296,20 @@ impl SorobanTarget {
...
 
+        // call the user defined constructor (if any)
+        if let Some(cfg_no) = constructor_cfg_no {
+            let constructor = binary.functions[cfg_no];
+            let constructor_name = constructor.get_name().to_str().unwrap();
+            binary
+                .builder
+                .build_call(constructor, &[], constructor_name)
+                .unwrap();
+        }
```
tareknaser, line 307: We can set `cfg.params` to the parameters of the constructor `FunctionValue`'s params. This will likely include a mapping that looks something like:
```rust
let mut params = vec![];
for param in constructor.get_params() {
	let param = match param.get_type() {
		BasicTypeEnum::IntType(int_type) => {
			let ty = match int_type.get_bit_width() {
				1 => ast::Type::Bool,
				32 => ast::Type::Uint(32),
				64 => ast::Type::Uint(64),
				_ => unreachable!(),
			};
			ast::Parameter::new_default(ty)
		}
		_ => unreachable!(),
	};
	params.push(param);
}
cfg.params = sync::Arc::new(params);
```
This won’t work because we still need to pass the arguments to `build_call()` (on line 296), which, of course, we don’t know.

src/emit/soroban/mod.rs
```diff
@@ -63,7 +63,7 @@ impl SorobanTarget {
         );
         binary.internalize(export_list.as_slice());
 
-        Self::emit_initializer(&mut binary, ns);
+        Self::emit_initializer(&mut binary, ns, contract.constructors(&ns).first());
```
tareknaser, line 67: Another approach I have in mind is to get the constructor `ControlFlowGraph` itself and prepend it with a call to `storage_initializer`, but this won't work either because the contract is immutable.

We could clone it and re-export it, but I’m not sure if that’s the best approach.

tareknaser (commented):
So adding support for constructors with custom arguments wasn't straightforward. I have two approaches in mind that I thought of, which I’ll explain below. 

Let me know what you think, or if there’s a better, more straightforward approach I might be missing.

seanyoung (approved):
Looks good to me, really needs @salaheldinsoliman approval for merging (as he looks after the Soroban target)

salaheldinsoliman (approved):

PR #1679, opened 2024-12-04
Add support for Soroban external calls 🎉 

salaheldinsoliman:
This PR adds support for Soroban external calls. This is the first step to support value transfers from inside Soroban contracts.
What a follow up PR should do is support encoding/decoding of a bigger range of value types.


src/emit/expression.rs
```diff
@@ -2147,6 +2149,16 @@ pub(super) fn expression<'a, T: TargetRuntime<'a> + ?Sized>(
...
 
+        Expression::PointerPosition { pointer } => {
+            let ptr = expression(target, bin, pointer, vartab, function, ns);
+            let data = bin.vector_bytes(ptr);
+            let res = bin
+                .builder
+                .build_ptr_to_int(data, bin.context.i64_type(), "sesa");
+
+            res.unwrap().into()
+        }
```
seanyoung, line 2160: Please help me understand `Expression::PointerPosition` . It takes a vector (e.g. `Type::DynamicString`), gets the pointer and converts the pointer to `i64` (even though pointers are 32 bit in wasm).

How about renaming it to `Expression::VectorData` ?

seanyoung (approved):
Please rename `Expression::PointerPosition`

Looks great otherwise!

src/emit/expression.rs
salaheldinsoliman: The `i64` rationale is that Soroban host always takes `i64` values. And that expression is only used in Soroban at the moment

PR #1709, opened 2025-02-02
Soroban: Add `extendTtl` Builtin Method 

### LINKED ISSUE #1669: Add builtin to extend TTL for Soroban storage variables
Soroban has a mechanism that ensures the active ledger state doesn't grow indefinitely. Every stored contract key has a certain TTL (Time to Live) that can be extended by calling a Soroban host function.

This can be handled by adding a Solang builtin `extend_ttl(var)`.

Relative docs:  https://developers.stellar.org/docs/build/guides/storage/choosing-the-right-storage

tareknaser:
This PR adds support for `extendPersistentTtl()` method in Soroban, along with a test and documentation. Also, the Soroban testing infrastructure has been refactored to allow more flexible environment manipulation.

[Documentation for `extend_ttl`](https://developers.stellar.org/docs/build/smart-contracts/getting-started/storing-data#managing-contract-data-ttls-with-extend_ttl)
Fixes #1669

#### Changes 
- Added support for `extendPersistentTtl` as a method on `uint64` for the Soroban target.
    - In the Soroban SDK, `extend_ttl` is a generic function (`IntoVal<Env, Val>`)
```rust
pub fn extend_ttl<K>(&self, key: &K, threshold: u32, extend_to: u32)
where
    K: IntoVal<Env, Val>,
```
but Solidity does not support that, so it's implemented as a method instead.
- One assertion in the `different_storage_types` test is affected due to changes in diagnostic capture. A follow-up PR will address this.

### DIFF
--- a/docs/language/builtins.rst
+++ b/docs/language/builtins.rst
@@ -666,3 +666,31 @@ Assuming `arg1` is 512 and `arg2` is 196, the output to the log will be ``foo en
     When formatting integers in to decimals, types larger than 64 bits require expensive division.
     Be mindful this will increase the gas cost. Larger values will incur a higher gas cost.
     Alternatively, use a hexadecimal ``{:x}`` format specifier to reduce the cost.
+
+.. _extendPersistentTtl:
+
+extendPersistentTtl(uint32 threshold, uint32 extend_to) 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+The ``extendPersistentTtl()`` method allows extending the time-to-live (TTL) of a contract storage entry.
+
+If the entry's TTL is below threshold ledgers, this function updates ``live_until_ledger_seq`` such that TTL equals ``extend_to``. The TTL is defined as:
+
+.. math::
+
+TTL = live_until_ledger_seq - current_ledger
+
+
+.. note:: This method is only available on the Soroban target
+
+.. code-block:: solidity
+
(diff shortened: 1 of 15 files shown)


tareknaser:
Update: `extendPersistentTtl` is now renamed to `extendTtl` for both Persistent and Temporary storage types. For Instance storage type, a new builtin called `extendInstanceTtl` has been added as a standalone global function (not a method).  

Tests for all storage types have been added in `tests/soroban_testcases/ttl.rs`.

salaheldinsoliman (approved):
Great!
Looks good but could you please:
1- Sign your commits for the DCO to pass
2- Add a test under `integration/soroban`?

tareknaser:
> 1- Sign your commits for the DCO to pass

Sure, should be fixed now

> 2- Add a test under integration/soroban?

This is a bit tricky. Writing integration tests for the time-to-live functionality would require a way to access the current ledger state during tests (like we do now in `tests/soroban_testcases/ttl.rs` using `soroban_sdk::testutils`). We’d need to do something similar in JavaScript.

There’s a package `@stellar/stellar-sdk` that we currently use for Soroban integration tests, but I’m not sure if it has all the testing utilities we’d need. I haven’t looked into it deeply yet, but the general idea is that we’d have to update the testing setup to make it work (if possible).

For now, I’d like to get this PR through, and in a follow-up PR, I’ll add support for other variable types in `extendTtl()` method with tests for each type and explore adding integration tests.

PR #1732, opened 2025-02-08
abi.decode("", ()) has no returns

seanyoung:
This should be compiled as: function returning void

Fixes: https://github.com/hyperledger-solang/solang/issues/1727


seanyoung:
Polkadot integration tests are failing because they've bitrotten. Nothing to do with this change.

LucasSte (approved):

PR #1753, opened 2025-03-09
Fix Polkadot Integration Test in CI

tareknaser:
Tracing this back, I think this issue was introduced in #1673

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -424,6 +424,7 @@ jobs:
         node-version: '16'
     - run: npm install
       working-directory: ./integration/polkadot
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -424,6 +424,7 @@ jobs:
         node-version: '16'
     - run: npm install
       working-directory: ./integration/polkadot
+    - uses: dtolnay/rust-toolchain@1.82.0
```
seanyoung, line 427: I think the failure is in the solang-substrate-ci docker image, not in github actions itself

seanyoung (changes requested):

PR #1778, opened 2025-04-08
Check command statuses in build.rs

### LINKED ISSUE #1684: Build-script command statuses are not checked
The build scripts use `Command::output` to run `make` and `llvm-config`: https://github.com/hyperledger-solang/solang/blob/32a45ea1efbcd2b148df0d6111ff47c15ae7973a/build.rs#L8-L17

However, the [`status` fields](https://doc.rust-lang.org/std/process/struct.Output.html#structfield.status) of the returned `Output` structs are not checked. So if either command completes with a non-zero exit code, it will go unnoticed.

Moreover, `Command::output` has the unfortunate side effect of capturing the output, i.e., stdout and stderr are not forwarded to the terminal. So if an error occurs, it could be hidden.

Would you consider a PR to check the commands' statuses?

smoelius:
This PR fixes #1684 by checking the statuses of two commands in the build.rs file.

I never actually received a response to #1684. So I won't be insulted if the PR is rejected as unnecessary.


smoelius:
I'm assuming the CI failures are not due to this PR. Please tell met if that is incorrect.

seanyoung (approved):
Good addition - I've hit this problem too and it's not obvious

PR #1779, opened 2025-04-09
Fix two infinite loops in codegen

smoelius:
This PR fixes infinite lops in the reaching definitions and dead storage code.

The two added test files are based on ones from Solidity:
- https://github.com/ethereum/solidity/blob/develop/test/libsolidity/smtCheckerTests/loops/do_while_break_2.sol
- https://github.com/ethereum/solidity/blob/develop/test/libsolidity/smtCheckerTests/loops/do_while_break_2_fail.sol

If you add `dbg!(block_no)` between the next two lines and run Solang on either of the test files, you will find that a block is visited repeatedly: https://github.com/hyperledger-solang/solang/blob/279ee911260d74c78acace41d753c5376a7e942e/src/codegen/reaching_definitions.rs#L56-L57

If you fix that infinite loop and add `dbg!(block_no)` between the next two lines, you will again find that a block is visited repeatedly: https://github.com/hyperledger-solang/solang/blob/279ee911260d74c78acace41d753c5376a7e942e/src/codegen/dead_storage.rs#L109-L110

In both cases, the problem appears to be that edges are added to the `blocks_todo` queue unconditionally.

My understanding of these algorithms is that they try to iterate to a fixpoint. Hence, I changed each loop so that:
- a flag records whether a _state change_ has occurred
- at the end of each loop iteration, the edge is added to the queue only if the flag is set

Furthermore, I analyzed each loop to try to determine what constitutes a state change. On this point, I invite scrutiny.

Nits are welcome on anything, though.

### DIFF
--- a/src/codegen/dead_storage.rs
+++ b/src/codegen/dead_storage.rs
@@ -141,11 +141,11 @@ fn reaching_definitions(cfg: &mut ControlFlowGraph) -> (Vec<Vec<Vec<Transfer>>>,
         );
 
         for edge in cfg.blocks[block_no].successors() {
+            let mut changed = false;
+
             if !block_vars.contains_key(&edge) {
-                blocks_todo.insert(edge);
-                block_vars.insert(edge, vec![vars.clone()]);
+                changed |= block_vars.insert(edge, vec![vars.clone()]).is_none();
             } else if block_vars[&edge][0] != vars {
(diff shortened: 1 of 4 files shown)


seanyoung (approved):
Totally makes sense. Thank you very much @smoelius

smoelius:
Thank you, @seanyoung!

PR #1789, opened 2025-05-08
Baby steps

smoelius:
The current changes allow me to compile the Solidity program below, and
the resulting wasm file passes `cargo stylus --check`.

```sol
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

contract Counter {
    uint256 public number;

    function setNumber(uint256 newNumber) public {
        number = newNumber;
    }

    function increment() public {
        number++;
    }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ Cargo.lock
 .vscode/
 
 /test_snapshots
+/llvm16.0*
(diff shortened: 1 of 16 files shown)


salaheldinsoliman:
Why did you close that one?
Looks pretty interesting 👍

PR #1791, opened 2025-05-12
Fix typos in tests/polkadot.rs

smoelius:
One is in an assertion message. All others are in comments.


seanyoung (approved):
Looks good, thank you!

seanyoung:
I've been looking at those CI failures, they're totally unrelated

PR #1793, opened 2025-05-13
Fix a small typo affecting `Expression::GetRef`

smoelius:
From what I can tell, this change affects no tests.


seanyoung (approved):

seanyoung:
Thank you @smoelius

PR #1795, opened 2025-05-15
Fix: Match Solidity behavior for memory array deletion (issue #1785)

### LINKED ISSUE #1785: [Inconsistent with solc]: Deleting element in arrary causes inconsistent execution results.
Solang version: v0.3.3-70-g32a45ea1
Description: In solc + EVM, the value of y is 2, but in solang + SVM, the value of y is 0. The reason is that in Solang, **delete data[0];** deletes the entire data. However, in Solc, **delete data[0];** only deletes one element from data.

Reproduction: I set up a Solang compilation and execution [environment ](https://github.com/Subway2023/solang)to test the program.

```solidity
contract C {

    function len() public returns (uint ret) {
        uint[] memory data = new uint[](2);
        data[0] = 234;
        data[1] = 123;
        delete data[0];
        delete data[1];
        assembly {
            ret := mload(data)
        }
    }
}
```

guptapratykshh:
## Description
This PR fixes behavior of `delete `operator when used with memory arrays to match Solidity's behavior. Previously, the delete operator was not correctly handling memory array elements, which could lead to unexpected behavior.

## Related Issue
Closes #1785


src/sema/statements.rs
```diff
@@ -724,15 +727,33 @@ fn statement(
...
 
-                        Ok(true)
+                        return Ok(true);
                     } else {
+                        // For memory arrays and other types, we should only delete the element
+                        // by assigning the default value, not delete the entire array
+                        // Issue #1785 - match Solc behavior for delete array[index]
                         ns.diagnostics.push(Diagnostic::warning(
                             *loc,
                             "argument to 'delete' should be storage reference".to_string(),
```
seanyoung, line 737: Here are you still generating a warning. Should be removed for array elements.

src/sema/statements.rs
```diff
@@ -724,15 +727,33 @@ fn statement(
...
+                        
+                        if let Ok(default_expr) = get_default_value(*loc, element_ty, ns, diagnostics) {
+                            let assign = Expression::Assign {
+                                loc: *loc,
+                                ty: expr.ty(),
+                                left: Box::new(expr),
+                                right: Box::new(default_expr),
+                            };
+                            
+                            res.push(Statement::Expression(*loc, true, assign));
```
seanyoung, line 751: The problem here is that the AST should be an accurate representation of the source code, not of the code it will generate - that happens in codegen.

So in sema, add:
```rust
res.push(Statement::Delete(...);
```
Then in codegen, for array elements, set a default value, here:

https://github.com/hyperledger-solang/solang/blob/main/src/codegen/statements/mod.rs#L217

seanyoung (changes requested):
Thank you for your contribution. The main problem is that output from sema should be the same as the source tree, i.e. you can't replace a `delete` statement with something else. Many components, e.g. solang language server, depend on the AST being accurate.

PR #1801, opened 2025-05-30
Handle `abi.encode()` with empty args

smoelius:
This is a small bugfix for the case when `abi.encode()` is called with empty arguments.

Without this change, if you run the following command, you get a panic:
```sh
$ cargo run compile --target=polkadot testdata/solidity/test/libsolidity/semanticTests/abiEncoderV2/abi_encode_v2.sol
...
thread 'main' panicked at src/codegen/encoding/mod.rs:169:46:
index out of bounds: the len is 0 but the index is 0
```
The cause of the panic is this function:
```sol
    function f0() public pure returns (bytes memory) {
        return abi.encode();
    }
```
With the change, the command completes successfully.

### DIFF
--- a/tests/codegen_testcases/solidity/abi_encode_v2.sol
+++ b/tests/codegen_testcases/solidity/abi_encode_v2.sol
@@ -1,0 +1,58 @@
+// RUN: --target polkadot --emit cfg
+
+pragma abicoder               v2;
+
+
+contract C {
+    struct S {
+        uint256 a;
+        uint256[] b;
+    }
+
+    // CHECK: C::function::f0
+    function f0() public pure returns (bytes memory) {
+        return abi.encode();
+    }
+
+    function f1() public pure returns (bytes memory) {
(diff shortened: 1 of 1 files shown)


smoelius:
> A test case would be nice

I added `tests/codegen_testcases/solidity/abi_encode_v2.sol`, which is a slight modification of `testdata/solidity/test/libsolidity/semanticTests/abiEncoderV2/abi_encode_v2.sol`:
```sh
$ diff testdata/solidity/test/libsolidity/semanticTests/abiEncoderV2/abi_encode_v2.sol tests/codegen_testcas
es/solidity/abi_encode_v2.sol 
0a1,2
> // RUN: --target polkadot --emit cfg
> 
9a12
>     // CHECK: C::function::f0
```
Without the change in 4619c04c343d108c802296d0fb5497acd69ee17f, `cargo test --test codegen solidity_testcases` fails. With the change, the command succeeds.

PR #1802, opened 2025-06-02
Rename `binary` to `bin`

### LINKED ISSUE #1799: Use of both `bin` and `binary` to name `Binary` values
This is somewhat OCD, but the code uses both `bin` and `binary` to name `Binary` values:
```
$ grep -r 'bin: &Binary' src | wc -l
      70
$ grep -r 'binary: &Binary' src | wc -l
      92
```
Would you consider a PR to adopt one name consistently (I don't care which) and to configure Clippy's [`disallowed-names` lint](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_names) to disallow the other?

smoelius:
Fixes #1799 by renaming `binary` to `bin`.

This turned out to be a smaller change than going the other way (1205 replacements vs. 1496).

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,2 +1,3 @@
 too-many-arguments-threshold = 13
 cognitive-complexity-threshold = 37
+disallowed-names = ["binary"]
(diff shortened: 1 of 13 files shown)


seanyoung (approved):
LGTM

Thank you very much!

PR #1857, opened 2026-04-09
Replace -O none/less/default/aggressive with standard -O0 -O1 -O2 -O3 -Os -Oz flags (fixes #534)

### LINKED ISSUE #534: Replace `-O less` command line option with `-O0`, `-O3`, `-Oz`, and `-OZ`

Sukuna0007Abhi:
Fixes #534

Replaces the non-standard `-O none/less/default/aggressive` flags with standard GCC/Clang-style `-O0`, `-O1`, `-O2`, `-O3`, `-Os`, `-Oz`. Old values still work for backward compatibility.

Key change from the previous attempt (PR #850): the `Options` struct now tracks both `opt_level` (speed) and `opt_size_level` (size) separately, so `-Os` and `-Oz` can properly communicate size optimization intent to LLVM.

**Mapping:**
- `-O0` → no optimization
- `-O1` → basic optimization  
- `-O2` → default (same as before)
- `-O3` → aggressive optimization
- `-Os` → optimize for size
- `-Oz` → aggressively optimize for size

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -58,8 +58,11 @@ Options:
   Only compile the code for the specified contracts. If any those contracts cannot be found, produce an error.
 
 -O *optimization level*
-  This takes one argument, which can either be ``none``, ``less``, ``default``,
-  or ``aggressive``. These correspond to llvm optimization levels.
+  This takes one argument, which can be ``0``, ``1``, ``2``, ``3``, ``s``,
+  or ``z``. These correspond to standard GCC/Clang optimization levels.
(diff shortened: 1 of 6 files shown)


Sukuna0007Abhi:
Pls review @seanyoung @salaheldinsoliman

PR #1861, opened 2026-04-15
fix(parser): treat persistent/temporary/instance as identifiers for non-Soroban targets

### LINKED ISSUE #1847: Solang parser incompatibility from introducing Soroban storage types
This PR: #1664 introduces three new keywords to Solang's parser, causing incompatibilities with Solidity parser. Thanks @smoelius for raising this up 👍

veyron-kairo:
Fixes #1847.

PR #1664 added `persistent`, `temporary`, and `instance` as reserved keywords globally, which broke standard Solidity code using those words as identifiers on non-Soroban targets.

The fix makes the lexer target-aware via a `soroban` flag on the `Lexer` struct. When the flag is off (default), the three tokens fall back to plain identifiers. A new `parse_soroban()` entry point enables the flag for Soroban targets; `sema` dispatches to the right parser based on the compilation target.

Two regression tests cover both modes.

### DIFF
--- a/solang-parser/src/lexer.rs
+++ b/solang-parser/src/lexer.rs
@@ -361,6 +361,10 @@ pub struct Lexer<'input> {
     last_tokens: [Option<Token<'input>>; 2],
     /// The mutable reference to the error vector.
     pub errors: &'input mut Vec<LexicalError>,
+    /// When true, `persistent`, `temporary`, and `instance` are lexed as
+    /// keywords (Soroban storage types). When false, they are treated as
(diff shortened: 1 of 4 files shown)


sridhar-panigrahi:
@salaheldinsoliman , please let me know your thoughts on this !

salaheldinsoliman (approved):

veyron-kairo:
Thanks for the review and approval, @salaheldinsoliman! 🙏 The branch is up to date with `main` and ready to go.

PR #1893, opened 2026-04-30
feat(soroban): implement `emit_event` via `contract_event` host function

0xull:
### Description
any solidity contract using `emit` would panic at compile time on the soroban target. the codegen dispatch in `src/codegen/events/mod.rs` had a `todo!()` for soroban, and `emit_event` in `src/emit/soroban/target.rs` was stubbed with `unimplemented!()`, so the path from solidity event emission down to the host was never wired up.

### Key Changes

- added `SorobanEventEmitter` in `src/codegen/events/soroban.rs`; indexed fields map to topics, first non-indexed field becomes the data val, falls back to zero val when there is no data
- wired `SorobanEventEmitter` into `new_event_emitter()` target dispatch
- added `ContractEvent` (`x.1`) to `HostFunctions` enum with type signature and wasm import registered in `declare_externals`
- implemented `emit_event` in `SorobanTarget` using `VectorNew` + `VecPushBack` to build the topics vec then calling `contract_event(topics, data)`

### Tests
two tests in `tests/soroban_testcases/events.rs` covering zero-topic and indexed-topic emission, both verified against the embedded soroban host.

### DIFF
--- a/tests/soroban_testcases/events.rs
+++ b/tests/soroban_testcases/events.rs
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+use crate::build_solidity;
+use soroban_sdk::testutils::Events as _;
+use soroban_sdk::{IntoVal, Val};
+
+#[test]
+fn emit_event_no_topics() {
+    let src = build_solidity(
+        r#"contract EventEmitter {
+            event Transfer(uint64 value);
+            function doTransfer(uint64 value) public {
+                emit Transfer(value);
+            }
+        }"#,
+        |_| {},
+    );
+
+    let addr = src.contracts.last().unwrap();
+    src.invoke_contract(addr, "doTransfer", vec![42_u64.into_val(&src.env)]);
+
+    let events = src.env.events().all();
+    assert_eq!(events.len(), 1);
+    let (_, topics, data) = events.get(0).unwrap();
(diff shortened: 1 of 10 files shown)


tests/soroban_testcases/events.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let addr = src.contracts.last().unwrap();
+    src.invoke_contract(
+        addr,
+        "doTransfer",
+        vec![1_u64.into_val(&src.env), 42_u64.into_val(&src.env)],
+    );
+
+    let events = src.env.events().all();
+    assert_eq!(events.len(), 1);
```
salaheldinsoliman, line 49: We should add an assert that the emitted event contains the expected values,
This makes sure there are no encoding/decoding errors.

salaheldinsoliman (changes requested):
Thank's! Looks great overall.
Aside form the comment could you update Soroban docs and support matrix accordingly?

0xull:
> Thank's! Looks great overall. Aside form the comment could you update Soroban docs and support matrix accordingly?

done, updated the support matrix, language compatibility, and examples coverage docs.

tests/soroban_testcases/events.rs
0xull: done, both tests now assert the exact topic and data values against what was passed in.

PR #1903, opened 2026-05-12
fix(soroban): reject unsupported ABI types before codegen

### LINKED ISSUE #1897: Improve error reporting
# Improve Soroban error reporting for LLVM aborts and compiler panics

## Summary

Some unsupported or partially supported Soroban code paths currently terminate the compiler with internal panics or LLVM assertions. These should become normal Solang diagnostics with source spans where possible.

This matters because users currently see implementation details such as `not yet implemented`, Inkwell variant panics, or LLVM assertions instead of actionable compiler errors.

## Environment

```text
solang version v0.3.4-21-g75da25ce
```

All examples below were compiled with:

```bash
target/debug/solang compile --target soroban <file>
```

### Reproducer

```solidity
contract StringReturnLLVMError {
    function make() public returns (string memory) {
        return "hello";
    }
}
```

Command:

```bash
target/debug/solang compile --target soroban /tmp/solang_soroban_llvm_string_return.sol
```

### Actual behavior

```text
warning: function can be declared 'pure'
  ┌─ /tmp/solang_soroban_llvm_string_return.sol:2:5
  │
2 │     function make() public returns (string memory) {
  │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

solang: /home/runner/work/solang-llvm/solang-llvm/llvm-project/llvm/lib/IR/Value.cpp:505: void llvm::Value::doRAUW(llvm::Value*, llvm::Value::ReplaceMetadataUses): Assertion `New->getType() == getType() && "replaceAllUses of value with new value of different type!"' failed.
/bin/bash: line 12: 177657 Aborted                 (core dumped) target/debug/solang compile --target soroban "$tmpfile"

[exit 134]
```

### Expected behavior

Solang should not abort inside LLVM. It should either compile this Soroban string return correctly, or emit a normal diagnostic with a source span if returning `string memory` is unsupported for Soroban.

### Reproducer

```solidity
contract EventPanic {
    event Ping(uint64 value);

    function emit_ping(uint64 value) public {
        emit Ping(value);
    }
}
```

Command:

```bash
target/debug/solang compile --target soroban /tmp/event_panic.sol
```

### Actual behavior

```text
thread 'main' panicked at src/codegen/events/mod.rs:54:28:
not yet implemented
```

The compiler exits with code `101`.

### Expected behavior

Solang should not panic. If event emission is unsupported for Soroban, the compiler should emit a normal diagnostic with a source span pointing at:

```solidity
emit Ping(value);
```

### Reproducer

```solidity
contract BytesPanic {
    function len(bytes memory data) public returns (uint64) {
        return data.length;
    }
}
```

Command:

```bash
target/debug/solang compile --target soroban /tmp/solang_soroban_bytes_panic.sol
```

### Actual behavior

```text
thread 'main' panicked at src/codegen/encoding/soroban_encoding.rs:212:14:
not implemented: unimplemented ty DynamicBytes in soroban decoder
```

The compiler exits with code `101`.

### Expected behavior

Solang should not panic. If dynamic `bytes` parameters are unsupported for Soroban, the compiler should emit a normal diagnostic with a source span pointing at the `bytes memory data` parameter.

## Desired outcome

These classes of failures should be converted into user-facing diagnostics before reaching LLVM, Inkwell unwraps, `todo!()`, `unimplemented!()`, or `panic!()` paths.

At minimum, unsupported Soroban ABI types/features should be rejected during semantic analysis or an explicit target-support validation pass, with diagnostics that identify:

- the unsupported construct,
- the Soroban target limitation,
- the source span that triggered the failure.

mohamedbasuony:
## Summary

Fixes the compiler-side Soroban diagnostics covered by #1897.

Unsupported Soroban paths could previously reach codegen, Soroban encoder/decoder logic, spec emission, or LLVM and fail with internal panics, `todo!()`, `unimplemented!()`, Inkwell errors, or LLVM assertions. This PR adds Soroban codegen-side validation so those user-triggerable cases are reported as normal Solang diagnostics before the unsafe paths are reached.

## What changed

- Move unsupported Soroban ABI validation out of `sema` and into the Soroban codegen path.
- Validate unsupported public/external Soroban function parameters and return values before dispatch wrapper generation.
- Validate unsupported public accessor return values before generated accessor code can reach unsupported Soroban lowering.
- Validate unsupported Soroban event argument types before event encoding.
- Guard multiple external return values and unsupported array/struct/string/bytes cases that could otherwise reach Soroban spec, encoder/decoder, Inkwell, or LLVM failures.
- Add precise diagnostics for internal string-helper patterns that could otherwise trigger LLVM/Inkwell failures.
- Keep supported Soroban behavior unchanged, including supported event emission and public string accessors.
- Target-gate the Soroban validation and early returns so non-Soroban diagnostic accumulation is unchanged.

## Notes

- This does not catch panics globally.
- This does not wrap compiler execution.
- This does not suppress LLVM assertions.
- Unsupported Soroban cases are rejected before the relevant codegen/emit paths are reached.
- Non-Soroban targets are not affected by these Soroban diagnostics.

## Verification

Ran locally:

- `cargo fmt --all -- --check`
- `cargo build --workspace`
- `cargo test --test soroban`
- `cargo test --test contract solana_contracts -- --nocapture`
- `cargo clippy --workspace`
- `cargo test --workspace`

Manual checks:

- `dynamic_bytes_arg.sol`: emits a normal diagnostic; old Soroban decoder panic is gone.
- `string_return.sol`: emits a normal diagnostic; old LLVM abort is gone.
- `event_panic.sol`: still compiles on this checkout.
- `examples/soroban/token.sol`: still compiles with existing warnings.
- Unsupported event argument stress cases emit diagnostics instead of panicking.
- Internal string-helper stress cases emit diagnostics instead of reaching LLVM/Inkwell failures.
- Public string accessor, internal bytes helper, and non-Soroban bytes/string controls remain unaffected.

Fixes #1897.

### DIFF
--- a/src/codegen/encoding/soroban_encoding.rs
+++ b/src/codegen/encoding/soroban_encoding.rs
@@ -3,6 +3,7 @@
 use crate::codegen::cfg::InternalCallTy;
 use crate::codegen::cfg::{ControlFlowGraph, Instr};
 use crate::codegen::encoding::create_encoder;
+use crate::codegen::error::CodegenError;
 use crate::codegen::vartable::Vartable;
 use crate::codegen::HostFunctions;
 use crate::codegen::{Builtin, Expression};
@@ -209,7 +210,14 @@ pub fn soroban_decode_arg(
             }
         }
 
-        _ => unimplemented!("unimplemented ty {:#?} in soroban decoder", ty),
+        _ => panic!(
+            "{}",
+            CodegenError::unsupported_soroban_type(
+                arg.loc(),
+                "by the Soroban decoder",
+                ty.to_string(ns),
+            )
+        ),
     }
 }
 
@@ -722,7 +730,14 @@ pub fn soroban_encode_arg(
             expr: encode_vector(item.clone(), cfg, vartab),
         },
 
-        _ => todo!("Type not yet supported in soroban encoder: {:?}", item.ty()),
+        _ => panic!(
+            "{}",
+            CodegenError::unsupported_soroban_type(
+                item.loc(),
+                "by the Soroban encoder",
+                item.ty().to_string(ns),
+            )
+        ),
     };
 
     cfg.add(vartab, ret);
--- a/src/codegen/error.rs
+++ b/src/codegen/error.rs
@@ -1,0 +1,279 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::Target;
+use solang_parser::diagnostics::Diagnostic;
+use solang_parser::pt::Loc;
+use std::fmt;
+
+#[allow(dead_code)]
+#[derive(Debug, Clone)]
+pub enum CodegenError {
+    UnsupportedSorobanType {
+        loc: Loc,
+        operation: String,
+        ty: String,
+    },
+    UnsupportedSorobanOperation {
+        loc: Loc,
+        operation: String,
+    },
+    MissingRuntimeHelper {
+        helper: String,
+        operation: String,
+        target: Target,
+    },
+    LlvmBuilder {
+        operation: String,
+        error: String,
+    },
+    MissingLlvmEntity {
+        operation: String,
+        entity: String,
+    },
+    InvalidCfgInvariant {
+        operation: String,
+        reason: String,
+    },
+    NumericConversion {
+        operation: String,
+        value: String,
+        target_type: String,
+    },
+}
+
+#[allow(dead_code)]
+impl CodegenError {
+    pub fn unsupported_soroban_type(loc: Loc, operation: impl Into<String>, ty: String) -> Self {
+        Self::UnsupportedSorobanType {
+            loc,
+            operation: operation.into(),
+            ty,
+        }
+    }
+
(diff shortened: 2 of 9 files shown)


salaheldinsoliman:
This change shouldn't be in `sema`. Instead, it should be handled in `codegen`, where target specific encoding/decoding logic is implemented . 
Moreover, I expected to see:
1-  `panic!`s without error messages addressed
2-  Areas where LLVM panics could occur safely guarded by human readbale messages.

Changes should be mostly in `emit` and `codegen`

salaheldinsoliman (changes requested):
Left a comment

mohamedbasuony:
Thanks, I see your point. I put the checks in sema as target validation, but I agree that this is too early/general for Soroban-specific ABI handling. The PR will be updated so that it handles unsupported Soroban ABI cases in the emit/codegen path, closer to the encoding/decoding logic.

Issue #1901, opened 2026-05-10
Support all Stellar examples


salaheldinsoliman:
# Add Solang Soroban examples for missing upstream Soroban example coverage

## Summary

Solang has Soroban Solidity examples and tests for several upstream Soroban concepts, including token, timelock, TTL, auth, cross-contract calls, atomic swap, liquidity pool, logging via `print`, and storage/array basics.

However, a number of examples from `stellar/soroban-examples` do not have an obvious Solidity/Solang counterpart yet. This issue tracks those gaps so we can add Solidity examples, integration tests, or explicit blocker issues for unsupported Soroban features.

Upstream reference:

- Repository: https://github.com/stellar/soroban-examples
- Snapshot checked: https://github.com/stellar/soroban-examples/tree/7b168174ae1268dab91a0190d80a94ab7ff41b59

## Missing or incomplete example coverage

| Upstream example | Link | Notes for Solang |
| --- | --- | --- |
| `account` | https://github.com/stellar/soroban-examples/tree/7b168174ae1268dab91a0190d80a94ab7ff41b59/account | Custom account contract with multi-signer policy and `__check_auth`; likely needs explicit Solang support/design for account contracts and auth context types. |
| `atomic_multiswap` | https://github.com/stellar/soroban-examples/tree/7b168174ae1268dab91a0190d80a94ab7ff41b59/atomic_multiswap | Batch swaps using a `SwapSpec` struct/vector shape; useful follow-up to current atomic swap coverage. |
| `bls_signature` | https://github.com/stellar/soroban-examples/tree/7b168174ae1268dab91a0190d80a94ab7ff41b59/bls_signature | Custom account-style example using BLS signature verification; likely blocked on crypto host-function bindings and bytes/vector support. |
| `custom_types` | https://github.com/stellar/soroban-examples/tree/7b168174ae1268dab91a0190d80a94ab7ff41b59/custom_types | Struct-backed state returned from public APIs; this is valuable because struct return/type support is currently fragile. |
| `deployer` | https://github.com/stellar/soroban-examples/tree/7b168174ae1268dab91a0190d80a94ab7ff41b59/deployer | Deploys another contract Wasm and invokes initialization; needs Solang story for deploying Wasm/contracts from contracts. |
[...]

PR #1907, opened 2026-05-17
feat: add pause example (Refs #1901)

Amit5601:
This PR adds the Solidity equivalent of the pause contract to the Soroban examples directory. This directly addresses missing/incomplete example (pause) in the upstream example coverage tracking issue.

### DIFF
--- a/docs/examples/soroban/pause/pause.sol
+++ b/docs/examples/soroban/pause/pause.sol
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+contract Pause {
+    bool private _isPaused;
+
+    function paused() public view returns (bool) {
+        return _isPaused;
+    }
+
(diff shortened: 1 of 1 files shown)


mohamedbasuony:
Thanks for adding this. Two small things should be fixed first:

- The upstream `pause` example stores `PAUSED` in instance storage. In Solang Soroban, unqualified state variables default to persistent storage, so this should likely be `bool private instance _isPaused;` to match the upstream behavior.
- Please change `Fixes #1901` to `Refs #1901` or `Part of #1901`. #1901 tracks all missing Stellar examples, so this PR should not close it by itself.

Amit5601:
@mohamedbasuony i have changed the title and added the required code as well. The vs code extension ci check is failing in this PR. Since this PR only adds a Solidity file to the docs, this looks like an unrelated Node environment issue in the runner. Let me know if everything is good to go.

PR #1918, opened 2026-05-28
feat(soroban): implement get/set storage bytes subscript

aryanbaranwal001:
**Description**

Adds support for reading or writing a particular byte of a `bytes` state variable on soroban by implementing  previously stubbed functions of `get_storage_bytes_subscript` and `set_storage_bytes_subscript` with `unimplemented!` in `soroban/target.rs`.

The following contract now compiles to a valid WASM binary

```solidity
// SPDX-License-Identifier: Apache-2.0
contract BytesSubscript {
    bytes public data;

    function set_data(bytes memory d) public {
        data = d;
    }

    function get_byte(uint32 i) public view returns (bytes1) {
        return data[i];
    }

    function set_byte(uint32 i, bytes1 v) public {
        data[i] = v;
    }
}
```

**Key Changes**

- `codegen/mod.rs`: added `BytesGet` and `BytesPut` variants to `HostFunctions` & their `"b.6"` / `"b.5"` name mappings.

- `emit/soroban/mod.rs`: added `BytesGet` and `BytesPut` arms to `function_signature`, and added `BytesLen` / `BytesGet` / `BytesPut` to the `declare_externals` host_functions list.

- `emit/soroban/target.rs`: implemented `get_storage_bytes_subscript` and `set_storage_bytes_subscript`.

**Tests**

- Added 8 tests in `tests/soroban_testcases/bytes_subscript.rs` covering read at every index, boundary writes, write isolation against unrelated bytes, sequential writes at the same index, and three out-of-the-bounds trap cases.

PR #1908, opened 2026-05-17
feat(soroban): support bytesN as function parameters and return values

aryanbaranwal001:
**Description**

Adds support for `bytesN` (`bytes1`–`bytes32`) solidity function parameters or return types for the Soroban target. The following now compiles to valid WASM and round trips through the host.

```solidity
contract BytesNCodec {
    function echo_bytes5(bytes5 x) public pure returns (bytes5) {
        return x;
    }
}
```

**Key Changes**

- `codegen/encoding/soroban_encoding.rs`: added `Type::Bytes(n)` arms to `soroban_decode_arg` and `soroban_encode_arg`. All `1 ≤ n ≤ 32` route through `BytesObject` via `bytes_copy_to_linear_memory` (decode) and `bytes_new_from_linear_memory` (encode), matching `soroban-sdk`'s `BytesN<N>` wire format.

**Tests**

- Added 3 tests in `tests/soroban_testcases/bytes_n_codec.rs` covering narrow, non-natural, and wide LLVM integer widths.

PR #1919, opened 2026-06-01
feat(soroban): implement sha256 and keccak256 builtins

aryanbaranwal001:
**Description**

The Solidity `sha256` and `keccak256` builtins were unimplemented on the Soroban target. This PR adds support for both, the following code now compiles and works as expected.

```solidity
contract Hasher {
    function sha(bytes memory input) public pure returns (bytes32) {
        return sha256(input);
    }
}
```

**Key Changes**

- `codegen/mod.rs`: added `HostFunctions::ComputeSha256` (`"c._"`) and `ComputeKeccak256` (`"c.1"`).
- `emit/soroban/mod.rs`: registered both in `function_signature` and added them to `declare_externals`.
- `codegen/expression.rs`: lower `Builtin::Sha256` / `Builtin::Keccak256` on Soroban, pass the input bytes (`args[0]`) through `soroban_encode_arg` to get a BytesObject handle, call the host fn, then decode the returned digest with `soroban_decode_arg`.

**Tests**

- Added 9 tests in `tests/soroban_testcases/hash.rs`: digest fixture asserts for both builtins, determinism and distinguishability, large input handling, cross algorithm divergence (`sha256 != keccak256`), and Sha256 / Keccak256 padding boundary fixtures.

**Note**

Other targets (solana, polkadot) implement hashing in `TargetRuntime::hash`, which takes a (`vector_bytes`, `vector_len`) ptr/len pair. We could do the same in `SorobanTarget::hash`, but on Soroban `bytes memory` is an `i64` BytesObject handle, so we'd first need a wrapper that copies the BytesObject contents into linear memory just to feed `hash` the pointer it expects. Cleaner approach is to intercept the builtin one stage earlier in codegen and lower it directly, wiring it with the Soroban hashing host functions. This follows the pattern that the existing Soroban arms for `RequireAuth`, `ExtendTtl` follows.


aryanbaranwal001:
This PR is stacked on #1918 because it includes the changes from #1908 and #1904, which add support for `bytesN` & `bytes` types respectively which aren't supported by soroban main. Both are required by the canonical Solidity signature for `sha256` / `keccak256`, which is `bytes memory → bytes32`. I kept this signature to stay consistent with  Polkadot / Solana counterparts.

PR #1931, opened 2026-07-09
soroban: support struct contract storage

Islam-Imad:
- Struct is represented as Vec Obj in constract storage memory. 
- Its encoder/decoder is different from the ABI since struct is represented as Map Obj in the ABI.
- We use vec_put to update a struct member in contract storage.
- We use vec_get to read a struct member from contract storage.


salaheldinsoliman:
@Islam-Imad Great!
Could you update the soroban support matrix in the docs accordingly? and also update this issue #1901 for the newly supported examples

Islam-Imad:
> @Islam-Imad Great! Could you update the soroban support matrix in the docs accordingly? and also update this issue #1901 for the newly supported examples

i will merge this one and then when it comes to merge #1932 i will update the matrix with struct support.
since struct support is divided in these two PRs.

PR #1932, opened 2026-07-09
soroban: support ABI struct encode/decode via named-field MAP OBJ

Islam-Imad:
Since struct is represented as a map object we implement the encoder/decoder according to this information.
- when we try to encode a struct we utilize the map_put host function, it automatically solve the sorting problem of struct member lexicographical by name. also if we try to use the unpack_map_obj_from_linear_memory host function we should allocate heap memory to able to do that but map_put usage help us also to avoid this problem entirely.
- when we try to decode a struct we utilize the map_get host function. (why? the same reason for encoder)
- in this batch we have focused on correctness not optimization, later batches will do focus on optimizations in-sha allah


Islam-Imad:
we need to make sure we emit the right specs. 
we will need to read and parse the wasm output and check the custome section.
this world will be deffered.

src/emit/soroban/mod.rs
```diff
@@ -428,6 +385,84 @@ impl SorobanTarget {
...
+                    .expect("struct field count exceeds spec limit"),
+            })
+            .write_xdr(&mut spec)
+            .unwrap_or_else(|_| panic!("writing struct spec entry for {:?}", decl.id.name));
+            Self::add_custom_section(context, &bin.module, "contractspecv0", spec.inner);
+        }
+    }
+
+    fn collect_struct_deps(ty: &ast::Type, ns: &ast::Namespace, acc: &mut Vec<usize>) {
+        match ty.deref_any() {
```
Islam-Imad: **Bug: `collect_struct_deps` silently collects nothing — no UDT spec entries are ever emitted**

`deref_any()` only strips `Ref` and `StorageRef`; it does **not** unwrap `SorobanHandle`. `dispatch.rs` wraps every public CFG's params/returns in `SorobanHandle(...)` and flips the original CFG's `public` flag to `false`, so `emit_struct_spec_entries` only ever sees wrapper CFGs whose param types are `SorobanHandle(Struct(UserDefined(n)))`. When `collect_struct_deps` calls `ty.deref_any()`, the `SorobanHandle` is returned unchanged, matches `_ => {}`, and the struct index is never pushed. The `structs` accumulator stays empty for every public function, and `emit_struct_spec_entries` emits zero `UdtStructV0` entries. Meanwhile `type_to_spec` *does* handle `SorobanHandle` (line 198), so the function spec correctly references the UDT by name — but with no matching definition entry, the contract spec is incomplete and SDK tooling (e.g. `soroban contract bindings`) will fail to resolve the type.

**Fix:** mirror the `type_to_spec` approach — match `SorobanHandle(inner)` (and `Ref(inner)`) explicitly and recurse, rather than delegating to `deref_any()`:

```rust
fn collect_struct_deps(ty: &ast::Type, ns: &ast::Namespace, acc: &mut Vec<usize>) {
    match ty {
        ast::Type::Ref(inner)
        | ast::Type::SorobanHandle(inner)
        | ast::Type::StorageRef(_, inner) => Self::collect_struct_deps(inner, ns, acc),
        ast::Type::Struct(ast::StructType::UserDefined(n)) => {
            if !acc.contains(n) {
                acc.push(*n);
                for f in &ns.structs[*n].fields {
                    Self::collect_struct_deps(&f.ty, ns, acc);
                }
            }
        }
        ast::Type::Array(elem, _) => Self::collect_struct_deps(elem, ns, acc),
        _ => {}
    }
}
```

src/emit/soroban/mod.rs
```diff
@@ -187,31 +193,43 @@ impl HostFunctions {
...
             ast::Type::Uint(32) => ScSpecTypeDef::U32,
             ast::Type::Int(32) => ScSpecTypeDef::I32,
             ast::Type::Enum(_) => ScSpecTypeDef::U32,
             ast::Type::Uint(64) => ScSpecTypeDef::U64,
             ast::Type::Int(64) => ScSpecTypeDef::I64,
             ast::Type::Int(128) => ScSpecTypeDef::I128,
             ast::Type::Uint(128) => ScSpecTypeDef::U128,
+            ast::Type::Int(256) => ScSpecTypeDef::I256,
+            ast::Type::Uint(256) => ScSpecTypeDef::U256,
+            ast::Type::Int(_) => ScSpecTypeDef::I32,
```
Islam-Imad, line 213: **Bug: `type_to_spec` missing `Uint(_)` catch-all — compiler panics for structs with `uint8`/`uint16`/`uint24` fields**

The match covers `Uint(32)`, `Uint(64)`, `Uint(128)`, `Uint(256)` explicitly and has a catch-all `Int(_) => I32` for small signed integers, but there is **no** corresponding `Uint(_)` catch-all. A struct field of type `uint8`, `uint16`, or `uint24` falls through to `_ => panic!("unsupported type in spec: {ty:?}")`, crashing the compiler.

This is currently masked by the `collect_struct_deps` / `SorobanHandle` bug (see other comment) — because no structs are ever collected, `type_to_spec` is never called on field types. Once that bug is fixed, a `uint8` struct field will trigger this panic.

**Fix:** add a `Uint(_)` catch-all that maps small unsigned ints to `U32`, mirroring the existing `Int(_)` arm:

```rust
ast::Type::Uint(_) => ScSpecTypeDef::U32,
ast::Type::Int(_) => ScSpecTypeDef::I32,
```

PR #1941, opened 2026-07-16
Fix soroban i128/u128 encoding high-bit truncation and signed small-f…

### LINKED ISSUE #1938: Fix `int128`/`uint128` ABI small-form encode/decode (Soroban)
### Bug 1 — `encode_i128` drops the high 64 bits

`encode_i128` (`encoding.rs:806`) gates the small form on the low limb only
(`(lo >> 56) == 0`), ignoring `high`. So `2^64 + 5` (`high=1, lo=5`) is wrongly
encoded as small form and loses its high 64 bits.

**Fix:** like `encode_i256` in #1933, refactor `encode_i128` to take the whole
128-bit value and gate the small form on **`(value >> 8) == 0`** (shift the full
value, `signed` for `Int(128)`); split into `lo`/`high` only on the host-object
branch.

### Bug 2 — `decode_i128` mis-signs the small form

The small-form branch (`encoding.rs:1133`) always uses an unsigned shift +
`ZeroExt`, so a negative `int128` (e.g. `-1`) decodes to a large positive value.
Zero-extend is correct for `Uint(128)` but wrong for `Int(128)`.

**Fix:** make it type-dependent (as `decode_i256` in #1933):
`ShiftRight { signed: is_signed }` then `SignExt` for `Int(128)` / `ZeroExt` for
`Uint(128)`.

### Testing

Start from this reproduction test for Bug 1 (`tests/soroban_testcases/i128_u128.rs`).
It **fails today** (returns `5` instead of `2^64 + 5`) and must pass after the fix:

```rust
#[test]
fn uint128_high_limb_not_dropped_on_encode() {
    let runtime = build_solidity(
        r#"contract test {
            function id(uint128 a) public returns (uint128) { return a; }
        }"#,
        |_| {},
    );
    let addr = runtime.contracts.last().unwrap();
    let value: u128 = (1u128 << 64) + 5; // high limb set, low limb < 2**56
    let arg: Val = value.into_val(&runtime.env);
    let res: Val = runtime.invoke_contract(addr, "id", vec![arg]);
    let got: u128 = FromVal::from_val(&runtime.env, &res);
    assert_eq!(got, value, "uint128 high 64 bits were dropped on encode");
}
```

```
assertion `left == right` failed: uint128 high 64 bits were dropped on encode
  left: 5
 right: 18446744073709551621   // 2^64 + 5
```

Then extend coverage across the small-form and host-object ranges and, for signed,
negatives (Bug 2) — same style as the i256 tests in #1933
(`tests/soroban_testcases/i256_u256.rs`).

```bash
cargo test --test soroban
cargo fmt --all
```

Ahmed0427:
Fixes #1938

* **Fixes `i128`/`u128` encoding truncation:** Gated the small-form encoding on the full 128-bit value (`value >> 8 == 0`) rather than just the low 64-bit limb, preventing high-bit truncation.

* **Fixes signed small-form decoding:** Replaced the default zero-extension with sign-extension (`SignExt`) when decoding signed `i128` small-form values to prevent negative numbers from decoding as positive.

* **Tests:** Added the reproduction test and expanded coverage for both small-form/host-object boundaries and negative numbers.

### DIFF
--- a/src/codegen/targets/soroban/encoding.rs
+++ b/src/codegen/targets/soroban/encoding.rs
@@ -774,7 +774,7 @@ fn encode_i128(
         left: high.clone().into(),
         right: Expression::NumberLiteral {
             loc: pt::Loc::Codegen,
-            ty: Type::Uint(64),
+            ty: high.ty(),
             value: BigInt::from(0_u64),
         }
         .into(),
@@ -1140,7 +1140,7 @@ fn decode_i128(cfg: &mut ControlFlowGraph, vartab: &mut Vartable, arg: Expressio
         left: arg.clone().into(),
         right: Expression::NumberLiteral {
             loc: pt::Loc::Codegen,
-            ty: Type::Int(64),
+            ty: Type::Uint(64),
             value: BigInt::from(8_u64),
         }
         .into(),
--- a/src/codegen/targets/soroban/encoding.rs
+++ b/src/codegen/targets/soroban/encoding.rs
@@ -763,26 +763,15 @@ fn encode_i128(
     };
 
     vartab.new_dirty_tracker();
+
+    let check_lo = cfg.new_basic_block("check_lo".to_string());
     let fits_in_56_bits = cfg.new_basic_block("fits_in_56_bits".to_string());
     let should_be_in_host = cfg.new_basic_block("should_be_in_host".to_string());
     let return_block = cfg.new_basic_block("finish".to_string());
 
-    let high_8_bits = Expression::ShiftRight {
+    let high_is_zero = Expression::Equal {
         loc: pt::Loc::Codegen,
-        ty: Type::Uint(64),
-        left: lo.clone().into(),
-        right: Expression::NumberLiteral {
-            loc: pt::Loc::Codegen,
-            ty: Type::Uint(64),
-            value: BigInt::from(56_u64),
-        }
-        .into(),
-        signed: false,
-    };
-
-    let cond = Expression::Equal {
-        loc: pt::Loc::Codegen,
-        left: high_8_bits.clone().into(),
+        left: high.clone().into(),
         right: Expression::NumberLiteral {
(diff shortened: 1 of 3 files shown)


src/codegen/targets/soroban/encoding.rs
```diff
@@ -763,26 +763,59 @@ fn encode_i128(
...
+    cfg.set_basic_block(check_lo);
+
+    // check if the low limb fits within the small representation limit
+    // signed positive must stay under 55 bits to avoid sign-extension confusion.
+    // unsigned can use up to 56 bits
+    let shift_amount = match int128_ty {
+        Type::Uint(128) => 56_u64,
+        Type::Int(128) => 55_u64,
+        _ => unreachable!(),
+    };
```
Islam-Imad, line 801: yeah it make sense

src/codegen/targets/soroban/encoding.rs
```diff
@@ -763,26 +763,59 @@ fn encode_i128(
     };
 
     vartab.new_dirty_tracker();
+
+    let check_lo = cfg.new_basic_block("check_lo".to_string());
```
Islam-Imad, line 767: the current implementation is correct, but we can do all our checks in one basic block instead of two.
but according to opened issue its enough now and we don't need to update it.

tests/soroban_testcases/i128_u128.rs
```diff
@@ -0,0 +1,79 @@
...
+        (1 << 55) - 1,
+        1 << 55,
+        (1 << 63) - 1,
+        -(1 << 55),
+        -(1 << 55) - 1,
+        -1,
+        i64::MIN as i128,
+        i128::MIN,
+        i128::MAX,
+    ];
```
Islam-Imad, line 68: good coverage for edge cases.

Islam-Imad (commented):
LGTM

PR #1978, opened 2026-07-22
feat(soroban): implement sha256 host function builtin

### LINKED ISSUE #1973: Support `sha256` on Soroban
> **⚠️ Prior art — open PR [#1919](https://github.com/hyperledger-solang/solang/pull/1919)**
> ("feat(soroban): implement sha256 and keccak256 builtins") **already implements
> this** (and `keccak256`). Review / rebase / finish that PR rather than starting
> fresh. What it does: adds `HostFunctions::ComputeSha256` (`c._`) +
> `ComputeKeccak256` (`c.1`), lowers `Builtin::Sha256`/`Keccak256` by passing
> `args[0]` through `soroban_encode_arg` → host call → `soroban_decode_arg`, and adds
> 9 tests in `tests/soroban_testcases/hash.rs`. Two deltas from the sketch below: it
> lowers in `codegen/expression.rs` (intercepting the builtin one stage earlier)
> rather than `lower_builtin`, and it predates the tree reorg (touches
> `src/codegen/encoding/soroban_encoding.rs`, now `src/codegen/targets/soroban/encoding.rs`),
> so it **needs a rebase**. The spec below documents the intended end state.

### Background

`sha256(bytes)` currently **panics at compile time** on Soroban — it routes
through the generic hash path (`src/emit/expression.rs:2570`) into
`SorobanTarget::hash()` (`src/emit/soroban/target.rs:1125`), which calls
`unsupported_soroban(...)`.

Soroban's host provides the hash directly. In the `crypto` module of the host
interface (`env.json`), `compute_hash_sha256` (symbol **`c._`**) takes a
`BytesObject` and returns a 32-byte `BytesObject`. Host-function definition:
https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-common/env.json#L2248
(module `crypto`; verified against v23.0.1).

`sha256` takes `bytes` and returns `bytes32` (`src/sema/builtin.rs:133`).
Target-specific builtins are lowered in the **codegen** layer
(`SorobanTarget::lower_builtin`).

> This is the same shape as the `keccak256` issue; if that one is done first, this
> is a small follow-up (only the host function differs).

### Task

1. **Add the host function.** `ComputeHashSha256` in the `HostFunctions` enum:
   - `src/codegen/mod.rs` — enum variant + `"c._"` in `name()`.
   - `src/emit/soroban/mod.rs` — signature `(i64) -> i64` in `function_signature`
     and add to `declare_externals`.

2. **Lower the builtin** in `lower_builtin` (`src/codegen/targets/soroban/mod.rs`),
   reusing the existing bytes encode/decode helpers:
   - evaluate `args[0]` (the `bytes`) with `expression(...)`;
   - `soroban_encode_arg(input, cfg, vartab, ns)` → a `BytesObject` handle;
   - `Instr::Call { HostFunction: ComputeHashSha256, args: [bytes_obj] }` into a
     temp `Type::Uint(64)`;
   - `soroban_decode_arg(hash_obj, cfg, vartab, ns, Some(Type::Bytes(32)))` → the
     `bytes32` result; return `Some(...)`.

Leave the emit `hash` stub as-is (still used by the unsupported hashes).

### Testing

Add a test under `tests/soroban_testcases/` — a contract returning `sha256(input)`
— and assert the 32-byte result against a known SHA-256 test vector (e.g.
`sha256("abc")` =
`0xba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad`).

To generate / verify expected digests for your own inputs, use an online hash
calculator: https://xorbin.com/tools/sha256-hash-calculator

```bash
cargo test --test soroban
cargo fmt --all
```

Ahmed0427:
## Support sha256 on Soroban

Closes #1973

Implements sha256(bytes) for the Soroban target. Previously this panicked at compile time via SorobanTarget::hash() -> unsupported_soroban(...).

### Changes
- Add HostFunctions::ComputeHashSha256 (host symbol c._, Soroban's compute_hash_sha256) in src/codegen/mod.rs

- Add its signature ((i64) -> i64) and external declaration in src/emit/soroban/mod.rs

- Lower Builtin::Sha256 in SorobanTarget::lower_builtin (src/codegen/targets/soroban/mod.rs): encode input via soroban_encode_arg, call the host function, decode result via soroban_decode_arg as Bytes(32)

- Add a test in tests/soroban_testcases/hash.rs verifying the result against the known sha256("abc") vector

### Note
This is a fresh implementation against current main rather than a rebase of #1919, since that PR predates the codegen/targets tree reorg. Same end state described in the issue.

### DIFF
--- a/src/codegen/targets/soroban/mod.rs
+++ b/src/codegen/targets/soroban/mod.rs
@@ -302,6 +302,34 @@ impl TargetCodegen for SorobanTarget {
                 );
                 Some(address_var)
             }
+            ast::Builtin::Sha256 => {
+                let input = expression(&args[0], cfg, contract_no, func, ns, vartab, opt, self);
+                let bytes_obj = soroban_encode_arg(input, cfg, vartab, ns);
+                let hash_obj = vartab.temp_anonymous(&Type::Uint(64));
+                cfg.add(
+                    vartab,
+                    Instr::Call {
+                        res: vec![hash_obj],
+                        call: InternalCallTy::HostFunction {
+                            name: HostFunctions::ComputeHashSha256.name().to_string(),
+                        },
+                        args: vec![bytes_obj],
+                        return_tys: vec![Type::Uint(64)],
+                    },
+                );
+                let hash_expr = Expression::Variable {
+                    loc: pt::Loc::Codegen,
+                    ty: Type::Uint(64),
+                    var_no: hash_obj,
+                };
+                Some(soroban_decode_arg(
+                    hash_expr,
+                    cfg,
+                    vartab,
+                    ns,
+                    Some(Type::Bytes(32)),
+                ))
+            }
             ast::Builtin::BlockNumber => {
                 let block_var_no = vartab.temp_name("block_number", &Type::Uint(64));
                 let block_var = Expression::Variable {
--- a/tests/soroban_testcases/hash.rs
+++ b/tests/soroban_testcases/hash.rs
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0
+use crate::build_solidity;
+use soroban_sdk::{Bytes, BytesN, IntoVal, TryFromVal};
+
+#[test]
+fn sha256_basic() {
+    let runtime = build_solidity(
+        r#"contract HashTest {
+            function hash(bytes memory input) public pure returns (bytes32) {
+                return sha256(input);
+            }
+        }"#,
+        |_| {},
+    );
+
+    let addr = runtime.contracts.last().unwrap();
+
+    let input = Bytes::from_slice(&runtime.env, b"abc");
+    let result = runtime.invoke_contract(addr, "hash", vec![input.into_val(&runtime.env)]);
+
+    let result_bytes = BytesN::<32>::try_from_val(&runtime.env, &result).unwrap();
+
+    // let expected = BytesN::<32>::from_array(
+    //     &runtime.env,
+    //     &hex_literal::hex!("ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"),
+    // );
+
+    let expected = BytesN::<32>::from_array(
+        &runtime.env,
+        &[
+            0xba, 0x78, 0x16, 0xbf, 0x8f, 0x01, 0xcf, 0xea, 0x41, 0x41, 0x40, 0xde, 0x5d, 0xae,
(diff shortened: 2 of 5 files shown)


tests/soroban_testcases/hash.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let input = Bytes::from_slice(&runtime.env, b"abc");
+    let result = runtime.invoke_contract(addr, "hash", vec![input.into_val(&runtime.env)]);
+
+    let result_bytes = BytesN::<32>::try_from_val(&runtime.env, &result).unwrap();
+
+    // let expected = BytesN::<32>::from_array(
+    //     &runtime.env,
+    //     &hex_literal::hex!("ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"),
+    // );
```
Islam-Imad, line 26: if there is an issue here please report it as an issue. if its useless so remove them.

src/codegen/targets/soroban/mod.rs
```diff
@@ -302,6 +302,34 @@ impl TargetCodegen for SorobanTarget {
                 );
                 Some(address_var)
             }
+            ast::Builtin::Sha256 => {
+                let input = expression(&args[0], cfg, contract_no, func, ns, vartab, opt, self);
+                let bytes_obj = soroban_encode_arg(input, cfg, vartab, ns);
+                let hash_obj = vartab.temp_anonymous(&Type::Uint(64));
```
Islam-Imad, line 308: please use `vartab.temp_name` instead of  `anonymous` since it help use to debug the CFG.

tests/soroban_testcases/mod.rs
```diff
@@ -10,6 +10,7 @@ mod bytes_pass;
 mod constructor;
 mod cross_contract_calls;
 mod events;
+mod hash;
```
Islam-Imad, line 13: could u rename the file name from `hash.rs` to `sha256.rs`

src/codegen/targets/soroban/mod.rs
```diff
@@ -302,6 +302,34 @@ impl TargetCodegen for SorobanTarget {
                 );
                 Some(address_var)
             }
+            ast::Builtin::Sha256 => {
```
Islam-Imad, line 305: also could u add assert_eq to make sure we expect the right number of arguments.

Islam-Imad:
also try to change the name of the branch to only focus on sha256

PR #1984, opened 2026-07-26
codegen: route storage bytes/string length through TargetCodegen

salaheldinsoliman:
## Summary

Stacked on #1983 (final PR of the stack: #1982 → #1983 → this).

Since #1927, shared codegen lowered storage `bytes`/`string` `.length` through Soroban host functions **unconditionally, for every target**. On Polkadot the emitted CFG then contained host-function calls that don't exist in its module, and `polkadot_tests::strings::bytes_storage` panicked in emit (`module.get_function(...)` returned `None` for the Soroban bytes-length host function).

## Fix

Dispatch the `DynamicBytes`/`String` storage-length lowering through `TargetCodegen::lower_storage_array_length`, so shared codegen no longer inspects `ns.target`:

- **Soroban** impl does the host-call lowering (`soroban_bytes_length`/`soroban_strings_length`).
- **Polkadot** impl restores its pre-#1927 behaviour: `StorageArrayLength` for `bytes`/`string` (lowered in emit), slot load for dynamic arrays.
- **Solana**'s passthrough impl is unaffected; EVM delegates to Polkadot's.

## Testing

Covered by the existing `polkadot_tests::strings::bytes_storage`/`bytes_storage_subscript` (panicked before, pass after) — note these run in the `tests/polkadot.rs` suite, which CI does not currently execute. The soroban `bytes_pass`/`strings` suites cover the Soroban side and stay green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/codegen/expression.rs
+++ b/src/codegen/expression.rs
@@ -15,8 +15,7 @@ use crate::codegen::constructor::call_constructor;
 use crate::codegen::interface::TargetCodegen;
 use crate::codegen::targets::polkadot::return_code as polkadot;
 use crate::codegen::targets::soroban::{
-    soroban_bytes_length, soroban_strings_length, soroban_struct_load, soroban_struct_member_load,
-    soroban_struct_member_store,
+    soroban_struct_load, soroban_struct_member_load, soroban_struct_member_store,
 };
 use crate::codegen::unused_variable::should_remove_assignment;
 use crate::codegen::{Builtin, Expression};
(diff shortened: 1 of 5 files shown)


salaheldinsoliman:
Pushed b2ece320 onto this branch: the Soroban impl derived the array's type from the *lowered* expression, but for storage variables that is the storage slot (uint64), so bytes/string lengths fell into the StorageArrayLength passthrough with a 64-bit type — comparing such a length against a uint32 built an ICmp with mismatched operands and aborted LLVM (caught by `bytes_storage_queue`). The declared array type is now threaded through the trait and dispatched on directly. Full soroban suite green (227), polkadot bytes_storage tests green.

PR #1988, opened 2026-07-27
refactor(soroban): lower extendTtl and extendInstanceTtl in codegen

### LINKED ISSUE #1971: Fully lower `extendTtl` / `extendInstanceTtl` from emit to codegen (Soroban)
The Soroban backend is being refactored to lower target-specific builtins in the
**codegen/CFG layer** (`SorobanTarget::lower_builtin`) instead of hand-writing LLVM
IR in the **emit layer**. `block.timestamp` was migrated this way in #1970 (issue
#1936); builtins already fully migrated include `GetAddress`, `BlockNumber`,
`RequireAuth`, and `AuthAsCurrContract` (see `src/codegen/targets/soroban/mod.rs`).

Dispatch: `src/codegen/expression.rs:1398` calls `lower_builtin`; if it returns
`Some(e)` the builtin is fully handled in codegen, otherwise it falls through to
emit's `TargetRuntime::builtin()`.

`extendTtl` / `extendInstanceTtl` are still (partly) done the old way and carry a
standing `// TODO` to move them (`src/emit/soroban/target.rs:963`):

- **`ExtendTtl`** — already has a `lower_builtin` arm
  (`src/codegen/targets/soroban/mod.rs:772`), but it only resolves the storage type
  from the variable declaration and then **re-emits** a codegen
  `Expression::Builtin { kind: Builtin::ExtendTtl }`. The actual host call is still
  hand-written in LLVM IR at `src/emit/soroban/target.rs:958‑1043`. This is a
  *partial* move — that emit arm is still live, not dead code.
- **`ExtendInstanceTtl`** — has **no** `lower_builtin` arm; it takes the default
  `expr_builtin` path and is lowered entirely in emit at
  `src/emit/soroban/target.rs:1044‑1110`.

Both host functions are already wired — no changes to the `HostFunctions` enum or
`declare_externals` are needed:

- `HostFunctions::ExtendContractDataTtl` (`l.7`):
  `extend_contract_data_ttl(k_val, t_storage_type, threshold_u32_val, extend_to_u32_val) -> i64`
- `HostFunctions::ExtendCurrentContractInstanceAndCodeTtl` (`l.8`):
  `extend_current_contract_instance_and_code_ttl(threshold_u32_val, extend_to_u32_val) -> i64`

The `threshold` / `extend_to` arguments (both `uint32`) must be passed as encoded
`U32Val`s. The emit code builds these by hand with `(x << 32) + 4`. The codegen
equivalent already exists — reuse it, don't reimplement:
`encode_object(loc, value, 32, tags::U32)` (`src/codegen/targets/soroban/encoding.rs:2371`,
`tags::U32 = 4`). `k_val` (the storage slot) and `t_storage_type` are passed as raw
`i64` vals, not encoded.

### Task

1. **Finish the `ExtendTtl` arm** in `lower_builtin`
   (`src/codegen/targets/soroban/mod.rs:772`):
   - Keep the existing storage-type resolution
     (`Temporary => 0`, `Persistent => 1`, `Instance => panic` with the current message).
   - Instead of returning `Some(Expression::Builtin { kind: Builtin::ExtendTtl, .. })`:
     - Encode `threshold` and `extend_to` as `U32Val`s with
       `encode_object(*loc, <arg>, 32, tags::U32)`.
     - Emit `Instr::Call { HostFunction: ExtendContractDataTtl,
       args: [slot_no, storage_type, threshold_u32val, extend_to_u32val] }` into a
       temp `Type::Int(64)` (mirror the `GetAddress` arm at `mod.rs:277`).
     - Return `Some(...)` of the result variable expression.

2. **Add an `ast::Builtin::ExtendInstanceTtl` arm** to `lower_builtin`:
   - Evaluate the two args (`threshold`, `extend_to`) and encode each as a `U32Val`
     with `encode_object(..., 32, tags::U32)`.
   - Emit `Instr::Call { HostFunction: ExtendCurrentContractInstanceAndCodeTtl,
     args: [threshold_u32val, extend_to_u32val] }` into a temp `Type::Int(64)`.
   - Return `Some(...)` of the result variable expression.

3. **Delete both emit arms** from `src/emit/soroban/target.rs`:
   - `Builtin::ExtendTtl` (~lines 958‑1043, including the `// TODO`).
   - `Builtin::ExtendInstanceTtl` (~lines 1044‑1110).

No changes needed to the `HostFunctions` enum / `declare_externals` — both host
functions are already declared and imported.

> Note (optional improvement, not required): the current emit path requires
> `threshold` / `extend_to` to be compile-time `NumberLiteral`s and shifts them at
> compile time. Using `encode_object` produces a runtime `ShiftLeft`, so the
> literal-only restriction can be relaxed if desired — keeping scope minimal is fine.

### Testing

Existing coverage lives in `tests/soroban_testcases/ttl.rs` (persistent/temporary
`extendTtl`, `extendInstanceTtl`, and the instance-storage guard).

```bash
cargo test --test soroban ttl
cargo test --test soroban
cargo fmt --all
```

Ahmad-Faraj:
Fixes #1971

extendTtl was half migrated. Its codegen arm resolved the storage type and then
re-emitted the builtin, so the host call was still hand written LLVM IR in emit.
extendInstanceTtl had no codegen arm at all.

Both now emit the host call from lower_builtin, and the two emit arms are
deleted. The threshold and extend_to arguments go through the existing
encode_object helper instead of a compile time shift, so they no longer have to
be number literals.

Existing ttl tests cover both paths.

### DIFF
--- a/src/codegen/targets/soroban/mod.rs
+++ b/src/codegen/targets/soroban/mod.rs
@@ -828,7 +828,8 @@ impl TargetCodegen for SorobanTarget {
                 Some(call_res_var)
             }
             ast::Builtin::ExtendTtl => {
-                let mut arguments: Vec<Expression> = args
+                assert_eq!(args.len(), 3, "extendTtl expects 3 arguments");
+                let arguments: Vec<Expression> = args
                     .iter()
                     .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt, self))
                     .collect();
@@ -852,17 +853,71 @@ impl TargetCodegen for SorobanTarget {
                     ),
                 };
 
-                arguments.push(Expression::NumberLiteral {
+                let slot = Expression::NumberLiteral {
                     loc: *loc,
-                    ty: Type::Uint(32),
+                    ty: Type::Uint(64),
+                    value: BigInt::from(var_no),
+                };
+                let storage_type = Expression::NumberLiteral {
+                    loc: *loc,
+                    ty: Type::Uint(64),
                     value: BigInt::from(storage_type_usize),
-                });
+                };
+                let threshold =
+                    encoding::encode_object(*loc, arguments[1].clone(), 32, encoding::tags::U32);
+                let extend_to =
+                    encoding::encode_object(*loc, arguments[2].clone(), 32, encoding::tags::U32);
 
-                Some(Expression::Builtin {
+                // The host functions return Void, so the result is not decoded.
+                let res = vartab.temp_name("extend_ttl", &Type::Int(64));
+                cfg.add(
+                    vartab,
+                    Instr::Call {
+                        res: vec![res],
+                        return_tys: vec![Type::Int(64)],
+                        call: InternalCallTy::HostFunction {
+                            name: HostFunctions::ExtendContractDataTtl.name().to_string(),
+                        },
+                        args: vec![slot, storage_type, threshold, extend_to],
+                    },
+                );
+                Some(Expression::Variable {
                     loc: *loc,
-                    tys: vec![Type::Int(64)],
-                    kind: (&builtin).into(),
-                    args: arguments,
+                    ty: Type::Int(64),
+                    var_no: res,
+                })
+            }
+            ast::Builtin::ExtendInstanceTtl => {
+                assert_eq!(args.len(), 2, "extendInstanceTtl expects 2 arguments");
+                let arguments: Vec<Expression> = args
+                    .iter()
+                    .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt, self))
+                    .collect();
+
+                let threshold =
+                    encoding::encode_object(*loc, arguments[0].clone(), 32, encoding::tags::U32);
+                let extend_to =
+                    encoding::encode_object(*loc, arguments[1].clone(), 32, encoding::tags::U32);
+
+                // The host functions return Void, so the result is not decoded.
+                let res = vartab.temp_name("extend_instance_ttl", &Type::Int(64));
+                cfg.add(
+                    vartab,
+                    Instr::Call {
+                        res: vec![res],
+                        return_tys: vec![Type::Int(64)],
+                        call: InternalCallTy::HostFunction {
+                            name: HostFunctions::ExtendCurrentContractInstanceAndCodeTtl
+                                .name()
+                                .to_string(),
+                        },
+                        args: vec![threshold, extend_to],
+                    },
+                );
+                Some(Expression::Variable {
+                    loc: *loc,
+                    ty: Type::Int(64),
+                    var_no: res,
                 })
             }
             _ => None,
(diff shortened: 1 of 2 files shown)


Islam-Imad:
@Ahmad-Faraj could u please rebase on main and solve the conflicts

src/codegen/targets/soroban/mod.rs
```diff
@@ -852,17 +794,66 @@ impl TargetCodegen for SorobanTarget {
...
+                let storage_type = Expression::NumberLiteral {
+                    loc: *loc,
+                    ty: Type::Uint(64),
                     value: BigInt::from(storage_type_usize),
-                });
+                };
+                let threshold = encode_object(*loc, arguments[1].clone(), 32, tags::U32);
+                let extend_to = encode_object(*loc, arguments[2].clone(), 32, tags::U32);
+
+                let res = vartab.temp_anonymous(&Type::Int(64));
```
Islam-Imad: please avoid using `temp_anonymous` use `temp_name` instead since the cfg looks cleaner.

src/codegen/targets/soroban/mod.rs
```diff
@@ -852,17 +794,66 @@ impl TargetCodegen for SorobanTarget {
...
+            ast::Builtin::ExtendInstanceTtl => {
+                let arguments: Vec<Expression> = args
+                    .iter()
+                    .map(|v| expression(v, cfg, contract_no, func, ns, vartab, opt, self))
+                    .collect();
+
+                let threshold = encode_object(*loc, arguments[0].clone(), 32, tags::U32);
+                let extend_to = encode_object(*loc, arguments[1].clone(), 32, tags::U32);
+
+                let res = vartab.temp_anonymous(&Type::Int(64));
```
Islam-Imad: Likewise

src/codegen/targets/soroban/mod.rs
```diff
@@ -852,17 +794,66 @@ impl TargetCodegen for SorobanTarget {
...
+                cfg.add(
+                    vartab,
+                    Instr::Call {
+                        res: vec![res],
+                        return_tys: vec![Type::Int(64)],
+                        call: InternalCallTy::HostFunction {
+                            name: HostFunctions::ExtendCurrentContractInstanceAndCodeTtl
+                                .name()
+                                .to_string(),
+                        },
```
Islam-Imad: could u utlize the host_call function defined in `encoding.rs` file at line 2240
```rust
fn host_call(res: Vec<usize>, host_fn: HostFunctions, args: Vec<Expression>) -> Instr 
```

Islam-Imad:
@Ahmad-Faraj run the soroban testscases locally and check what is wrong
:face_in_clouds:

Islam-Imad:
run `cargo test --test soroban`

Islam-Imad:
LGTM

PR #1992, opened 2026-08-03
fix(resolver): prefer the longest matching import remapping

grandpig:
## Summary

- Prefer the longest matching import remapping prefix.
- Make remapping selection independent of rule order when prefixes have different lengths.
- Preserve the existing overwrite behavior for duplicate prefixes.
- Add regression tests for both remapping orders.

## Problem

`FileResolver::resolve_file()` currently overwrites the remapped path every time a rule matches. As a result, the last matching rule wins, even when another rule has a longer matching prefix.

For example:

```text
@lib/=vendor/base/
@lib/math/=vendor/math/
```

The import:

`import "@lib/math/Calc.sol";`

must resolve to:

`vendor/math/Calc.sol`

However, when the shorter rule appears after the longer rule, the current implementation incorrectly resolves it to:

`vendor/base/math/Calc.sol`

## Changes

The resolver now selects the matching remapping with the longest prefix and applies only that remapping.

Solidity's complete remapping algorithm also considers remapping context. Solang currently does not expose context in its import map representation, so this change applies the longest-prefix rule within Solang's existing prefix-to-target model.

Duplicate prefixes retain the existing behavior: add_import_map() replaces the earlier target, and the CLI emits an overwrite warning.

### DIFF
--- a/src/file_resolver.rs
+++ b/src/file_resolver.rs
@@ -220,17 +220,28 @@ impl FileResolver {
         }
 
         // first check maps
-        let mut remapped = path_filename.clone();
+        let mut best_match: Option<(&OsString, &PathBuf, &Path)> = None;
 
-        for import_map_no in 0..self.import_paths.len() {
-            if let (Some(mapping), target) = &self.import_paths[import_map_no].clone() {
-                if let Ok(relpath) = path_filename.strip_prefix(mapping) {
-                    remapped = target.join(relpath);
-                }
+        for (mapping, target) in &self.import_paths {
+            let Some(mapping) = mapping else {
+                continue;
+            };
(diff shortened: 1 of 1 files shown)


grandpig:
@seanyoung Just a gentle follow-up in case this got buried. This is focused on FileResolver remapping selection, making the longest matching prefix win regardless of rule order, with regression coverage for both orders and duplicate prefixes. Would appreciate a look when you have time. Thanks!

grandpig:
@seanyoung Could you please take a look when you have a chance? Thanks!

seanyoung (approved):
Looks great, thank you very much.

